PR #15, opened 2020-08-15
pre-commit-config

mLewisLogic:
@cjdsellers 

So I attempted to build the Cython files and got a failure at the `flake8` step.
The actual issue was that `nautilus_trader` was a submodule within a larger project that already had its own `.flake8` file that was getting picked up (and had more/stricter checks). The way the invocation was being made (within [Poetry](https://python-poetry.org)), it was picking up the broader `.flake8` project file, rather than `nautilus_trader`s.

So, rather than doing the sensible thing and just fixing that issue, I reworked the linting system and moved it over to [pre-commit](https://pre-commit.com) hooks.
No need to run `flake8` before every install (that's actually a kinda awkward place to do lint checks). The current repo now passes the current `.flake8` tests. Any file changes will be run through `flake8` before they can be committed (generally nice, though occassionally annoying if you want to sneak in a messy checkpoint commit that's in a semi-garbage state).

I added some new pre-commit hooks as well. They're the reason for the chunderous PR. Little things like removing extraneous whitespace and ensuring a consistent newline at the end of files.

Additional pre-commit hooks of note:
- https://github.com/pre-commit/pre-commit-hooks
  - Lots of minor cleanups
- https://github.com/prettier/prettier
  - Mostly to auto-format *.md files
- https://github.com/timothycrosley/isort
  - Will auto-format and sort your imports. This one is quite useful. I also wrote and tested a config file for it (`.isort.cfg`). I left this commented out in the PR and un-run because it will cause a huge diff that will be hard to review. I'll leave that up to you to uncomment it and run. Once you get through the pain of doing it once it will keep your code in a much nicer steady-state and make you worry about code format far less.
- https://github.com/psf/black
  - This is a gamechanger. It's a widely used, opinionated Python formatter. It's right in its style choices like 99% of the time. It's useful enough to bear with it the 1% of times that it produces an ugly line or two (and I've never seen it "break" code). It makes code formatting something that fades into the background. I left this commented because the diff for the first PR is going to be stupid big. Again, once you endure the one-time pain, it is quite nice on the other side.
- https://gitlab.com/pycqa/flake8
  - Runs `flake8` on commit. Also includes some `flake8` plugins that pass. Commented out the ones that have issues so as to minimize the functional changes in this PR.

The pre-commit hooks will run on whatever files are in the changeset when you try to write a commit.
If you want to run the hooks against all files, run `pre-commit run --all-files`.

With all of this set up, I dropped the `flake8` run in the `setup.py` file. Shouldn't be necessary anymore. Cython files built fine. Tests mostly ran (though integrations failed because I didn't have a Redis server running).

-Mike


cjdsellers:
I like the introduction of a pre-commit hook, I agree this is a more sensible place to do any linting checks. 

Prettier looks handy as the .md files did need some cleaning up.

I'll do some testing with isort, I had my own style of importing from lowest dependency first which is going to be hard to maintain for other developers.

I watched the Lukasz Langa PyCon2019 talk. I definitely like the premise behind it, I agree with a lot of the styling from what I saw re parameter hanging and white space. It'll be interesting to see just how different the code formatting will be. There's a potential issue with it recognizing Cython code, but we'll give it a try.

Thanks for your efforts here! I'll merge.

PR #28, opened 2020-09-18
MaxBidMinAsk

mLewisLogic:
Indicator that tracks the max bid and min ask within a recent time window.
Uses deques to prune the collection as aggressively as possible.

I tried using `libcpp.deque` but ran afoul of trying to put cdef Python classes into it. In retrospect, of course that wouldn't work. Will keep it in my back pocket if there's a time where I need a deque that only has C primitives in it.


mLewisLogic:
Needs docstring cleanup, but curious for initial feedback.

PR #32, opened 2020-09-19
Commission open/increase position changes

scoriiu:


cliffsch:
## Implementation complete (2026-04-22)

Gate confirmed: OCO was never implemented in the Rust adapter; we now own adapter development.

### Changes made

**Python runner** (`scripts/nautilus_live_runner.py`):
- `_BracketState`: added `pending_exit: bool = False` flag
- `_submit_entry`: pre-generates stop/target COIDs and pre-registers bracket legs at entry submission time; computes tick offsets from bar close price and injects `px_stop_ticks`, `px_target_ticks`, `px_stop_coid`, `px_target_coid` tags into the entry MarketOrder
- `on_order_filled` (entry path): no longer calls `_submit_bracket`; arms watchdog directly after recording fill price
- `on_order_filled` (exit path): sets `bracket.pending_exit = True` immediately before cleanup
- `on_order_canceled`: if `bracket.pending_exit` is True, treats cancel as expected OCO sibling cleanup (deregister + return); otherwise escalates to `_handle_bracket_exit_failure`

**Rust adapter** (`external/nautilus-projectx/crates/adapters/projectx/src/execution.rs`):
- Added `PendingOcoBracket` struct tracking pre-assigned stop/target COIDs for server-side OCO children
- Added `pending_oco_brackets: Arc<DashMap<(i64, String), PendingOcoBracket>>` field
- Added `extract_bracket_tags()` helper: reads `px_stop_ticks` / `px_target_ticks` / `px_stop_coid` / `px_target_coid` from entry order tags
- `submit_order`: populates `stop_loss_bracket` (type=4 Stop) and `take_profit_bracket` (type=1 Limit) in `PlaceOrderReq`; after entry ACK, inserts into `pending_oco_brackets` and pre-registers child COIDs in `order_meta_by_client`
- `handle_user_orders_event`: when a `GatewayUserOrder` arrives with no client mapping, checks `pending_oco_brackets`; if matched, registers the venue→client mapping and emits a synthetic `OrderAccepted` event for the child

All 31 Rust unit tests pass. Python syntax clean.

### Acceptance still needed
- 20 consecutive clean PRAC round-trips with no `ghost_position_detected` events
- `events.jsonl` inspection: no `cancel_order` calls for bracket siblings; all sibling `OrderCanceled` arrive broker-initiated
- Nautilus parity tests: `run_nautilus_parity.py` (backtest path unaffected — bracket fields are live-only)

PR #41, opened 2020-09-22
BidAskMinMax

mLewisLogic:
Cleaned up the indicator. Now it tracks both min/max of both bid/ask. It's a lot cleaner and includes better docstrings.


mLewisLogic:
Also, this adds a `indicators/utils` module as a place for `WindowedMinMaxPrices`. Made sense to extract it from the main indicator and wanted a neutral place to put it in case it was useful for reuse.

mLewisLogic:
Lemme know if you want it in a different place.

cjdsellers:
The indicator itself looks good. I'm hesitant to introduce a general `utils` module into indicators. I've found in the past utils start attracting all kinds of helpers and random code. So two other options could be;
- Extract the more general `windowed` function if and when necessary.
- If you have something on the cards already which is going to need the function, lets call the new module functions instead of utils, which will specifically be for generic indicator functions.

Also for the unit test, could be split the test with the act-assert-act-assert type flow into some other tests or change the flow just so there's a cleaner arrange-act-assert flow?

mLewisLogic:
K, I'm happy to fold the `WindowedMinMaxPrices` class back into `BidAskMinMax` if needed. My preference is that module/file == 1 class (makes finding things much easier), but that's probably more of a trait picked up in other languages like Scala than a hard Pythonism. Def recognize that in this repo modules are more of functionality collections rather than the container for a class.

mLewisLogic:
K, collapsed the files together and tightened up the tests to match your AAA style.

cjdsellers:
Good points, I'm also partial to breaking out classes/functions into separate files for ease of navigation. The 2-3 thousand line Python files can make things more difficult.

PR #50, opened 2020-09-23
Extend commission model to maker and taker

scoriiu:


cjdsellers:
Thank you for that. Do you think its worth squashing commits of this size next time?

scoriiu:
For sure. You can also give this kind of feedback as comments on the PR and I will address them prior to merging

PR #59, opened 2020-09-26
Test position change

scoriiu:


cjdsellers:
For this one you were calling .as_double() on a Python float. The API for value objects is actually the next thing to get refactored to avoid the need for these type changes.

Also, for assertEqual the expected value comes first, and then the actual.

I'm going to add base_currency into fill events and on the position. Then we can work with it in calculations.

PR #67, opened 2020-10-05
Poetry

mLewisLogic:
@cjdsellers 

Intent is to enable a full Poetry flow (without using setup.py at all), while still supporting the existing setup.py flow until you are comfortable enough to turn it off.

It produces sdist as well as a wheel.

Version is bumped (and is entirely controlled within pyproject.toml).
Some dependency upgrades, but I tried to minimize them to keep this focused. Pytest was upgraded to support xdist and parallel testing, and a few others to resolve dependency conflicts (which pip will just paper order based upon order it sees the dependencies...).

This also enable parallel tests. They're running pretty fast. 17 sec on my machine with coverage, without performance tests.


cjdsellers:
This is looking good!
Having the wheel is necessary and appreciated, it's really going to enhance the installation experience.

That's fine regarding the dependency upgrades. I'm generally keeping pace with the latest versions in most other libraries.

From a short read using poetry.lock sounds like the way to go. I think most users these days are comfortable with maintaining a separate env for large projects like this one.

Regarding the errors, I actually haven't seen anything throw from those locations before.

mLewisLogic:
The `tests/unit_tests/core/test_core_fractions.py` is legit because it is trying to `import quicktions`.
The `tests/unit_tests/core/test_core_uuid.py::UUIDTests::test_create_uuid1` error might have to do with the packaging setup.

mLewisLogic:
Probably something related to `cdef enum` in the pxd file.

cjdsellers:
Ok, I see what was happening in the fractions tests, now correctly importing the nautilus version.

mLewisLogic:
Sorry I haven't kept up with communication this week. Life's been busy.
Pushed my current WIP for this branch.
Having some coverage issues, but the overall thrust is good.

Getting coverage working is currently blocked on this issue: https://github.com/cjolowicz/nox-poetry/issues/127
Basically, we need to build a "debug" build with all of the tracing settings available for coverage to work. Poetry doesn't natively support variations lke that. It will have a single wheel build output path (the normal optimized version). If we want debug variations like that, we'll have flag it via env params that are active as the package is being installed locally via an sdist .tar.gz package) (within a nox session).

Having the above working like that would allow us to have a nox session for normal testing with the optimized code, and another session for profiling/coverage. They would build within separate venvs, so wouldn't trip over one another.

cjdsellers:
This sounds like a good approach. The separate build for coverage was always going to be necessary as you do need the `profile` and `linetrace` compiler directives set to True.

cjdsellers:
Apologies, I keep bumping the version which isn't really relevant in this beta phase anyway.

cjdsellers:
I see you're almost there with this one, great work! I'll await your final go ahead before I merge.

cjdsellers:
Am I correct in my understanding that with the wheels setup we won't need to keep Cython as an installation dependency anymore?

bohblue2:
> Am I correct in my understanding that with the wheels setup we won't need to keep Cython as an installation dependency anymore?

Yes We can deprecate cython installation dependency with wheel.

SEE: https://stackoverflow.com/questions/4505747/how-should-i-structure-a-python-package-that-contains-cython-code

bohblue2:
By the way, nice work! @mLewisLogic

mLewisLogic:
Cython is required to build and package the wheels. It packages up the .pyx and the .so files.
So the end-user will be able to quickly install without needing to perform either the .pyx -> .c or the .c -> .so steps. The second one is the most computationally heavy anyways, so including the .so files makes sense.

cjdsellers:
Understood,

There's some guidance here http://docs.cython.org/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules recommending to distribute the generated `.c` files.

The approach of distributing the `.so` files will achieve the same outcome for users installing via PyPI. If a user wanted to specifically compile with their own version of GCC say (for whatever reason), then they'd simply do that directly from source anyway.

So we'll distribute the package wheels compatible with 64-bit versions of Windows, OSX and Linux then?

mLewisLogic:
@cjdsellers ya, that's the thinking.
64-bit cross-platform wheels are available for easy install.
sdist is available for anybody who wants to compile/build in their own particular way.
If you're particular about GCC flags, then wheels aren't for you in this case. Compiling the parallel sdist would allow that flexibility.
In my mind, for a huge Cython codebase like this, the primary goal of wheels should be to take care off all pre-compilation for the end-user so that installing from pypi can be done in seconds rather than minutes.

cjdsellers:
Absolutely agree

mLewisLogic:
The poetry work has been going overall quite well.At this point it nicely builds and tests across 3.6-3.8 for Ubuntu and MacOS. Lots of caching, much fast.I'm blocked on getting it working for Windows at the moment. Some weird linker error in the Cython build with no meaningful error message: https://github.com/nautechsystems/nautilus_trader/runs/1303943191?check_suite_focus=true
I'm punting on Window support in the CI for now. The structure is there to allow it, if the C compilation/linking error can be solved. Would want to see community demand for Windows wheels before I was motivated to crack it open again.
Strangely, I'm getting consistent test failures from `tests/unit_tests/common/test_common_clock.py` when running on MacOS. Actually runs fine on my local mac, only fails in CI. Not sure how Timer is implemented, but I'm wondering if it is sensitive to CPU load or something like that?https://github.com/nautechsystems/nautilus_trader/runs/1304090079?check_suite_focus=true

Integration tests run on MacOS and Ubuntu. Redis is installed via brew for mac and apt-get for Ubuntu.
No performance tests in the CI. The failure threshold is clearly pegged to your machine, so doesn't make sense to run that pass/fail in a different environment.
I moved the sphinx/readthedocs to be on poetry/nox. Dropped the build/make/requirements.txt files. `nox -s build_docs` gets you your local build now.The new `.readthedocs.yml` file is untested. Unsure of the best way to do that.

cjdsellers:
Lets press on without specific Windows builds in the CI then. I think VMs are becoming quite easy with Windows 10 these days, I also have a Windows 10 VM on my local system I can do some testing with to see if its running there anyway.

I'll look into the `Clock`, I think both the tests and the clock itself could do with some refactoring for starters. Probably that interval just isn't long enough for the test on a lower spec server.

Agreed on the performance testing, really only for a developer to do relative profiling on their local system - the asserts are unnecessary really and have triggered for me intermittently also.

Good work on Redis, its great we can run those integration tests across Linux and MacOS.

Readthedocs is an ongoing project for me. I need to be able to compile the `.so`s on their build for all of the docs to be generated by sphinx (works on my local of course). Also some further templating to tidy up - not to mention the docs themselves need attention. I'm not firm on any specific config setups, as long as it works and is simple - your approach there sounds good

cjdsellers:
It turns out the clock/timer tests were too fragile, occasionally relying on things happening within a 50ms window. I've refactored to preserve the intent of the tests, with far less fragility.

cjdsellers:
This is looking really solid. I'll await your go ahead before merging.

mLewisLogic:
@cjdsellers I think the main body of work is ready here. Wouldn't mind getting past rebase hell.

Only loose thread is that the commit tagging and publishing isn't tested (it only triggers on master).
My inclination is that we go ahead and merge. There's a chance that the first pass of the package publishing will need a bugfix, but that seems like a small issue all things considered (and would be pretty easy to fix).

PR #105, opened 2020-11-02
Develop

cjdsellers:
New improved CI pipeline.


cjdsellers:
Pausing for a quick review on this one. Some immaterial minor tweaks and mainly checking the docs still build with the trimmed `docs/source/requirements.txt`, and also giving the tag-release flow a test run.

docs/source/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
mLewisLogic: I don't believe that this file is necessary anymore.

docs/source/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
mLewisLogic: Same here.

This should now be handled by `nox -s build_docs`.

docs/source/requirements.txt
```diff
@@ -0,0 +1,4 @@
+cython>=0.29.21
```
mLewisLogic, line 1: This file can get dropped.

`nox -s build_docs` handles things locally.
The RTD config file uses pip to install necessary packages for documentation.

docs/source/Makefile
mLewisLogic: Is this something that RTD specifically needs?

mLewisLogic:
Intent with the poetry branch was to enable doc building locally via nox, and for RTD via the `pip .` with extras=docs.
I've tested the local aspect successfully. The RTD aspect needs testing though.

cjdsellers:
Understood. I'll remove those redundancies then and do a test run.

cjdsellers:
> Intent with the poetry branch was to enable doc building locally via nox, and for RTD via the `pip .` with extras=docs.
> I've tested the local aspect successfully. The RTD aspect needs testing though.

```
Running Sphinx v1.8.5
loading translations [en]... done

  File "/home/docs/checkouts/readthedocs.org/user_builds/nautilus-trader/envs/latest/lib/python3.8/site-packages/sphinx/registry.py", line 472, in load_extension
    mod = __import__(extname, None, None, ['setup'])
ModuleNotFoundError: No module named 'numpydoc'
```

For some reason didn't install `numpydoc`. I'm surprised they don't have that in their standard environment actually.

cjdsellers:
I can handle getting RTD working.

mLewisLogic:
K, that is strange.
`.readthedocs.yaml` should include the `extras=docs` requirement.

That maps to 
```
[tool.poetry.extras]
docs = ["numpydoc", "sphinx", "sphinx-rtd-theme"]
```
in `pyproject.yaml`.

So if everything was going to plan, it should install normal prod dependencies (not dev) + the extras from the `docs` group.

cjdsellers:
Yep, it's there in `.readthedocs.yaml`

codecov is generating some cool reports now

cjdsellers:
I can also flip the docs build to trigger on any branch too if you want to trouble shoot

mLewisLogic:
Not going to be able to troubleshoot tonight.
I did a quick test on a fresh repo/venv and `pip install ".[docs]"` installs the whole package with the extras, no Poetry required.

cjdsellers:
Possibly we could hack this with adding `path: .[docs]` into the config yaml?

mLewisLogic:
Worth a shot?
Seems strange since `extra_requirements` seems pretty clearly supported: https://docs.readthedocs.io/en/stable/config-file/v2.html#packages

mLewisLogic:
Hrmm, the `version: 2` portion from the docs got dropped during a rebase. Wonder if it is related?

cjdsellers:
I think I've found the fix for the fetch depth thing for `test-tag-publish`. Added it in for the next push

cjdsellers:
> Hrmm, the `version: 2` portion from the docs got dropped during a rebase. Wonder if it is related?

That actually wasn't you. I changed the formatting of the config to match one which was working with another Cython project.

cjdsellers:
I'll change the RTD trigger branch to develop and trouble shoot a little later.

cjdsellers:
> Worth a shot?
> Seems strange since `extra_requirements` seems pretty clearly supported: https://docs.readthedocs.io/en/stable/config-file/v2.html#packages

I agree, the current config looks correct according to this...

cjdsellers:
Its fixed. RTD config upgraded to V2 using Python 3.8

requirements.txt not required its doing the right thing calling `pip install .[docs]` as per the config.

The actual problem was hiding in the `conf.py`.
This line is only required for calling `make` locally
```
sys.path.insert(0, os.path.abspath("../.."))
```

It sends the RTD build server off in the wrong direction. Just had to remove that one line!

Pushed again to `master`. Almost there, just failing on `Create git tag`

PR #123, opened 2020-11-12
Build cleanup

mLewisLogic:
1. Attempts to fix the test-tag-publish flow for master.
2. Enables `test` workflows for non-develop branches
3. Trims the size of the published sdist from 15mb to 200kb (by dropping docs/scripts/tests and nautilus_trader/_internal from the package).
4. Trims the size of the published MacOS wheel by 800kb (7.5mb -> 6.7mb).

I noticed that the linux wheels are much larger. Guessing it's just a byproduct of how manylinux2014 works.

I'm up for debating whether _internal should stay as part of the core shipped artifact. I'd argue that it's really big, and almost alway stale. It's really just one sliver of data that should belong in a user database. Getting it into that database probably shouldn't involve storing it in git.

### DIFF
--- a/.github/workflows/test-tag-publish.yml
+++ b/.github/workflows/test-tag-publish.yml
@@ -7,9 +7,64 @@ on:
     branches: [ master ]
 
 jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
+        python-version: [ 3.6, 3.7, 3.8 ]
+    name: test - Python ${{ matrix.python-version }} (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
 
-  test-tag-publish:
+    steps:
+      - uses: actions/checkout@v2
 
+      # Python setup
+      - name: Set up Python environment
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # Install build dependencies
+      - name: Install build dependencies
+        run: python -m pip install --upgrade pip setuptools wheel poetry nox --use-feature=2020-resolver
+
+      # Setup Poetry caching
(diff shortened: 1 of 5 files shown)


cjdsellers:
Great work,

I forgot how large those files were, `_internal` was always meant to be a temporary solution so I've moved the data out of there into `tests/test_kit/data` for starters. The classes which had csv paths for that data in the constructors have been changed to take a `pd.DataFrame` instead (way more flexible), the unit tests now just `pd.read_csv` from the test kit data path to keep those tests passing.

As for the `RolloverInterestCalculator` this is a little more tricky to extract as it's hardwired into the `SimulatedExchange`. So as a fix for this I'm creating the concept of `SimulationModule` which will be pluggable components into a `SimulatedExchange` following a microkernel type pattern https://www.oreilly.com/library/view/software-architecture-patterns/9781491971437/ch03.html

So as the backtest is iterating it'll pass through those modules to apply additional business logic/simulations. That'll fix this up for now and keep the dists smaller.

The rollover interest is really FX specific anyway for funding "overnight" positions, other asset classes do it differently or not at all. The backtest example is currently `GBP/USD` so I'll make sure the module is included there to keep everything consistent.

Just coding it up now.

cjdsellers:
Ok, that's done, about to push it up. It's a little rough at the moment but working.

```
    # Option plug in module to simulate rollover interest
    # The data is coming from packaged test data
    interest_rate_data = pd.read_csv(os.path.join(PACKAGE_ROOT + "/data/", "short-term-interest.csv"))
    fx_rollover_interest = FXRolloverInterestModule(rate_data=interest_rate_data)

    engine.plug_simulation_module(Venue('SIM'), fx_rollover_interest)
```

cjdsellers:
No effect on performance.

PR #125, opened 2020-11-13
Develop

cjdsellers:


cjdsellers:
For the math ops involving `BaseDecimal`, the `decimal.Decimal` needs to be extracted out. If the object in question is on the right side of the expression then you can no longer call `self ` as that param name overrides `self` with the incoming left side arg. I'll see if I can clean this up going with `a, b` and getting ride of the `@staticmethod`s

PR #169, opened 2021-01-09
Test0

jpmediadev:


cjdsellers:
It looks like the pre-commit checks just failed because `nautilus_trader/backtest/loaders.pxd` didn't end with a new line. Pedantic I know but it helps keep everything clean. I'd recommend checking out https://pre-commit.com/ 

The installation is very easy and it'll check all pre-commit checks automatically for you on a push.

There were 5 failing tests. Looks like it might be related to gzip, possibly when decompressing data you've added? Give it a try with uncompressed data although you may have to trim the amount right back as we'd like to avoid accumulating large amounts of tests data in the repo.

```
 >           uncompress = self._decompressor.decompress(buf, size)
E           zlib.error: Error -3 while decompressing data: invalid distance too far back

../../../hostedtoolcache/Python/3.7.9/x64/lib/python3.7/gzip.py:482: error
```

Otherwise have a closer look by clicking on `Details` besides one of the failing checks.

What you're working on is looking really good indeed, your tests are great.

PR #170, opened 2021-01-10
Develop

cjdsellers:


netomenoci (commented):
## 🔍 Deep Code Review — PR #170

Four parallel review tracks were run: **Bug Hunting**, **Concurrency Analysis**, **Breaking Changes**, and **Test Coverage**. Below is the consolidated report.

---

### CRITICAL / HIGH Severity Issues

#### 1. `_on_account_state()` crashes on unknown currency — CRITICAL
**File:** `nautilus_trader/adapters/ccxt/execution.pyx`

All three balance loops (total, free, used) have the same bug — after logging that a currency is `None`, there is no `continue`. Execution falls through to `Money(amount, currency)` with `currency=None`. Since `Money.__init__` expects `Currency not None`, this will **TypeError and crash**, killing the entire balance update path.

**Bonus:** The error message is copy-pasted — all three loops say "total balance" even for free and locked.

---

#### 2. Race condition: `_watch_balances` starts before currencies are loaded — HIGH
**File:** `nautilus_trader/adapters/ccxt/execution.pyx` — `connect()`

```python
self._loop.create_task(self._connect())              # loads currencies async
self._watch_balances_task = self._loop.create_task(self._watch_balances())  # starts immediately
```

Both tasks launch concurrently. `_watch_balances` can receive a balance update **before** `_load_currencies()` finishes inside `_connect()`. With `self._currencies` still empty, every lookup returns `None` → triggers Bug #1 → crash.

---

#### 3. BinanceDataClient never sets `initialized = True` — HIGH
**File:** `nautilus_trader/adapters/binance/data.pyx` — `_connect()`

`_connect()` sets `self._is_connected = True` but **never** sets `self.initialized = True`. The old path through `DataEngine.update_instruments_all()` → `_request_instruments()` used to set it, but that was removed. `CCXTDataClient` correctly sets both flags — Binance does not.

Result: `TradingNode._wait_for_engines()` spins forever → **permanent startup hang**.

---

#### 4. OandaDataClient never sets `initialized = True` — HIGH
**File:** `nautilus_trader/adapters/oanda/data.pyx`

Same issue as #3. The `self.initialized = True` line was removed from `_request_instruments()` and never added to `_connect()`. Same permanent startup hang result.

---

### MEDIUM Severity Issues

#### 5. `_update_balances()` missing `return` after capability check
After logging `"fetch_balance not available"`, code falls through to call `fetch_balance()` anyway.

#### 6. `_load_currencies` TypeError workaround is broken
`except TypeError: response = self._client.fetch_currencies` assigns the **bound method object** (not its return value). The subsequent `response.items()` will `AttributeError`. Same pattern in `_update_balances` and `_watch_balances`.

#### 7. Double `self._client.close()` on shutdown
`_watch_balances()` `finally` block calls `close()`, then `_disconnect()` also calls `close()`. Double-close can error depending on the CCXT exchange implementation.

#### 8. Cancelled tasks never awaited
Both `CCXTDataClient` and `CCXTExecutionClient` cancel tasks in `_disconnect()` but comment out appending them to `stop_tasks` (TODO notes). `asyncio.gather(*stop_tasks)` is always a no-op. Cleanup code in cancelled tasks' `finally` blocks may race with the subsequent `self._client.close()`.

#### 9. Oanda `call_later()` from thread pool — not thread-safe
`_connect()` runs via `run_in_executor`, then calls `self._loop.call_later()` from the worker thread. `call_later` is not thread-safe per asyncio docs. Should use `call_soon_threadsafe()` to schedule a function that itself calls `call_later()`.

#### 10. `_connect()` failure = permanent hang, no retry
If `_connect()` throws (e.g., network error during `load_all_async()`), `_is_connected` and `initialized` stay `False` forever. No retry mechanism. `_wait_for_engines()` has no timeout.

---

### Test Coverage Gaps

| Gap | Severity |
|-----|----------|
| `_on_account_state` error path (Bug #1) — no test | **HIGH** |
| `_update_balances` missing return (Bug #5) — no test | **MEDIUM** |
| `_load_currencies` happy path — no assertions on parsed currencies | **MEDIUM** |
| `_on_account_state` happy path — no assertions on `AccountState` events | **MEDIUM** |
| `_watch_balances` streaming loop — no direct test | **LOW** |

Test removals (91 lines from `test_data_engine.py`, 22 from perf tests) are **justified** — they tested deleted APIs. CCXT test reductions were setUp de-duplication, not test removal.

---

### Breaking Changes

**Zero breaking changes to alpha_punting code.** All removed APIs (`Connect`/`Disconnect` classes, `update_instruments_all()`, CCXT adapter) are old upstream nautilus_trader code already superseded in the current codebase.

---

netomenoci:
## Bug #1 — CRITICAL: `_on_account_state()` crashes on unknown currency

**File:** `nautilus_trader/adapters/ccxt/execution.pyx` — `_on_account_state()`

All three balance loops (total, free, used) have the same bug:

```python
currency = self._currencies.get(code)
if currency is None:
    self._log.error(f"Cannot update total balance for {code} (no currency loaded).")
balances.append(Money(amount, currency))  # ← NO continue! Money(amount, None) → crash
```

After logging the error, there is no `continue`. Execution falls through to `Money(amount, currency)` with `currency=None`. Since `Money.__init__` declares `Currency currency not None`, this will **TypeError and crash**, killing the entire balance update path.

**Bonus:** The error messages are copy-pasted — all three loops say "total balance" even for free and locked.

**Fix:** Add `continue` after each error log.

cjdsellers:
Hi @netomenoci,

It looks like an agent may have posted these review comments here by mistake - they reference a private project name. If you're using Claude or similar, you could consider adding this to your project's `CLAUDE.md` to prevent it from interacting with the upstream:

```
Do not post comments, reviews, create PRs, or push to upstream repositories.
```

And to prevent accidental git pushes you could consider running:

```
git remote set-url --push upstream no_push
```

This is the second time something has ended up on the upstream repo accidentally. Please be more careful going forward.

PR #176, opened 2021-01-10
fix two streams (quotes & trades) in one time

jpmediadev:
check in the morning if  the error is same )


ian-wazowski:
It worked for me

jpmediadev:
i am new in co-development on github, tell me please which commands to rebase to make everything look nicer?

cjdsellers:
You could try creating a fresh branch for starters. It would be good to cleanup your `Stanislav` non-GitHub user.

I think rebasing should be enough, I'm not sure why its picking up old commit history...
I also did some slight formatting changes in `wrangling.pyx` so please update that.

I'm squashing your PR into one commit anyway so I think they're ignored in the source control.
It would be good to clean up the PRs though but not sure on tooling at your end or the best commands.

Just implementing some live execution stuff so will attend to this a little later.

cjdsellers:
I'm happy to merge this branch though, keep this open however have a go at fetching upstream.

jpmediadev:
done.

cjdsellers:
Great, I suppose if the user knows there's a misalignment of quotes and trades that's up to them.

Thanks for providing that test too!
We should probably have a test to check the logic of that min/max timestamping too?

jpmediadev:
I thought so too, but then I found  several checks in the code, maybe that's enough?

nautilus_trader/backtest/engine.pyx
 
```
Condition.equal(start.tz, pytz.utc, "start.tz", "UTC")
 Condition.equal(stop.tz, pytz.utc, "stop.tz", "UTC")
 Condition.true(start >= self._data_producer.min_timestamp, "start >= data_client.min_timestamp")
 Condition.true(start <= self._data_producer.max_timestamp, "stop <= data_client.max_timestamp")
 Condition.true(start < stop, "start < stop")
```

PR #263, opened 2021-04-08
Status and Close price

limx0:
This PR add a couple of events and enums that allow the reporting of statuses and closing prices for betting/equity markets. I've made a first pass at the statuses I think are applicable (see below).  These statuses could be used to reject orders in the exec engine or other things down the track, but at the moment will at least allow strategies to subscribe to events that may be of interest to them.  

Events:
- `VenueStatusEvent` - The entire venue has changed status (if applicable).
- `InstrumentStatusEvent` - A particular instrument has changed status. 
- `InstrumentClosePrice` - The official closing price for an instrument (with a few different types), which is typically published at the end of a session or when an instrument expires (if applicable). 

Enums
- `VenueStatus` - a Venue wide status (if applicable). Variations described below
- `InstrumentStatus` - An instrument specific status. Variations described below
- `InsturmentCloseType`  - The type or reason for the close price being published. Currency `end_of_session / expired`

Description of statuses:
- `CLOSED` - Venue/Instrument not open for trading. 
- `PRE_OPEN` - Some markets allow orders to be send before the market open, but trading does not occur in this status. Also know as pre-market auction.
- `OPEN` - Market is open for orders and trading.
- `PAUSE` - Also known as suspend / halt. Market is not available for matching, but some markets may allow the cancelling of orders.
- `PRE_CLOSE` - Closing auction / special status for between OPEN and CLOSE.

Open to any suggestions on naming also.


cjdsellers:
Excellent work once again!

PR #266, opened 2021-04-10
Betfair fixes & Add OrderUpdated

limx0:
This PR makes several small changes:
- Fixes some issues with Betfair order insert/replace/cancels.
- Adds `subscribe_order_book_deltas` to the DataClient and `_handle_subscribe_order_book_deltas` to the DataEngine. This allows a user to subscribe to the raw orderbook snapshot and update/diff messages.
- Adds `_handle_order_updated` to the ExecutionEngine
- Adds `calculate_data_size` to backtest container, allowing skipping of this step (it has caused some issues with betfair data for reasons unknown
- Fixes some issues with the shutdown in the `sockets` module

Some more testing required for the `OrderUpdated` changes


cjdsellers:
Many thanks!

PR #280, opened 2021-04-17
Add poetry markers for windows 32bit platform

ian-wazowski:
There is build issue when installing pyarrow for windows 32bit.

pyarrow does not support windows 32bit.


cjdsellers:
Thanks. 

After a change to `pyproject.toml` we need to run `poetry update` to generate a new `poetry.lock` file.

PR #285, opened 2021-04-21
[WIP] - Sim exchange order book

limx0:
This PR adds a full orderbook to the backtest SimulatedExchange with the following functionality:
- Orders will now be filled at a realistic price (and volume) based on the state of an instruments order book
- Limit orders can be placed into the actual book
- Partial fills are now supported

It also adds some more functionality and testing for the OrderBook class and removes `orderbook/util.py` that was hanging around but not being used

Tasks before this can be merged
- [ ] There are a handful (13) of failing tests in `test_backtest_exchange` that aren't immediately obvious how to fix, @cjdsellers could you please take a look?
- [ ] Performance of using the orderbook rather than just top bid/ask needs to be assessed (and likely optimised).
- [ ] A few more tests on the orderbook / partial fill functionality would be nice


limx0:
Closing in favour of #287

PR #290, opened 2021-04-22
Add new AssetTypes

ian-wazowski:
Add new AssetTypes: CALL, PUT, SPREAD

### DIFF
--- a/nautilus_trader/model/c_enums/asset_type.pxd
+++ b/nautilus_trader/model/c_enums/asset_type.pxd
@@ -22,6 +22,9 @@ cpdef enum AssetType:
     CFD = 5,
     OPTION = 6,
     WARRANT = 7,
+    CALL = 8,
+    PUT = 9,
+    SPREAD = 10,
 
 
 cdef class AssetTypeParser:
--- a/nautilus_trader/model/c_enums/asset_type.pyx
+++ b/nautilus_trader/model/c_enums/asset_type.pyx
@@ -31,6 +31,12 @@ cdef class AssetTypeParser:
             return "OPTION"
         elif value == 7:
             return "WARRANT"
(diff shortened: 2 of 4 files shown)


limx0:
We already have `OPTION` in the `AssetType` - I think for PUT/CALL we should add an attribute `KIND`, and probably lump SPREAD in there also

cjdsellers:
Thanks for the PR Ian,

I agree with @limx0 they're option types. Also for new enums the full pattern needs to be implemented including adding to the Parser and the unit tests. Its extra work but then gives us the magic of Python and C enums rolled into one!

ian-wazowski:
@limx0 @cjdsellers 
Thanks for feedbacks!, I will develop more and than send PR again.

cjdsellers:
Thanks for the PR Ian,

I believe this is following on from the discussion about the `network` subpackage?

There are still some things to figure out regarding the architecture of the components we'll be building on top of - however I would like to merge your code into `develop` following a couple of tweaks to the PR:

Would it be possible for you to remove the code relating to the option type enums, as there is an `Option` instrument in the works which will include the put and call kinds, possibly in another enum.

Also if you could remove the specific references to the exchanges as we're keeping `network` generic, and grouping implementation specific code in the appropriate adapters. 

You're welcome to start adapters for those exchanges if you like in a separate PR! 🚀

Following this it would make sense to retitle the PR to be specific to the network code.

We'll then be in a position to merge this 👍🏻

ian-wazowski:
@cjdsellers 

I forgot and kept the PR open without closing it. My mistakes, sorry :)

I've been working on a few things over the past few weeks. I hope I can do PR soon.

I understood what you said, Thanks for your kind feedback :)

Have a nice day!

PR #329, opened 2021-05-25
[WIP] Account balances

limx0:
WIP PR for updating the `AccountState` event and adding an `AccountBalance` type. This is a little more explicit than the separate `list[Money]` that we currently use.


ian-wazowski:
Good!

PR #336, opened 2021-06-03
Modified Relative Import on examples and added a trailing stop check for the ema_cross_trailing strategy.

thematz1:
When I ran a backtest with the ema_cross_trailing stop I was met with a value error. I added a None check for the self.trailing_stop in before the line with the client_order_id check. In addition relative imports were not working with pathlib for me so i changed it to use the os library only.


cjdsellers:
Ok great suggestions.

There's a small conflict as `self.data` has changed to `self.cache`. Could you please rebase on the latest which should resolve the conflict, then we can merge from there.

cjdsellers:
I just noticed as well you are diffing from the `master` branch. I think this needs to be from the `develop` branch.

cjdsellers:
I went ahead and applied these fixes myself, I hope that was ok with you.

I look forward to any further contributions!

PR #345, opened 2021-06-14
BUG: Check os env variable before checking the config

crazy25000:
Was testing out the live example and the account ID was never being passed. Tracked it down to here where it always returns the default value from the OS environment instead of checking the config. Save someone an hour or two :p 

https://github.com/nautechsystems/nautilus_trader/blob/f713b55f106871020d533d272aca61b80d2fd47c/nautilus_trader/adapters/oanda/factories.pyx#L64-L65


cjdsellers:
Great find!

Are you able to PR to the `develop` branch though? Then I'll run CI on it and we can marge.

crazy25000:
> Great find!
> 
> Are you able to PR to the `develop` branch though? Then I'll run CI on it and we can marge.

Done. Sorry about that. I thought selected dev.

crazy25000:
I also recommend that instead of returning a default to  raise an exception. Because without a proper ID and token, this functionality will never work.

cjdsellers:
Looking at this more closely. The intention for the config dict is to provide the names of environment variables which will be read at runtime. The value of the variable will be what we're after in this case. This PR will mean we are looking for an env var named "api_token" for example.

So if you set your env vars for `OANDA_API_TOKEN` and `OANDA_ACCOUNT_ID` you should get the expected behaviour.

Admittedly configuration could do with some improvements to make things clearer, and this is on the near term roadmap!

cjdsellers:
I've taken your advice and now raising `RuntimeError` if an expected environment variable is `None`.

e.g (where the env var name is changed by appending a 1 in each case).
```
RuntimeError: No value for environment variable OANDA_API_TOKEN1
RuntimeError: No value for environment variable OANDA_ACCOUNT_ID1
```

Issue #325, opened 2021-05-15
Convert tests to pytest


cjdsellers:
The intention is to move the testing framework completely over to `pytest`.
Currently all new tests are written in `pytest` however there remains a mixture of legacy `unittest` and `pytest` tests.

What this involves for each file which utilizes `unittest` is mainly the following:
- Remove `import unittest`.
- Change test class name to 'TestSomeName' (so starting with Test) to be 'discoverable' by pytest.
- Remove `unittest.TestCase` base class.
- Change `setUp` and `tearDown` method names to `setup` and `teardown` respectively.
- Change all usage of `self.assertEqual()` to an equivalent `assert expression`.
- Change all usage of `parameterized` to `@pytest.mark.parametrize`.
- Change all usage of `self.assertRaises()` to the `with pytest.raises` pattern.

This will enhance standardization of the test suite and allow the removal of `parameterized` as a dependency.

Any help on this somewhat large task will be greatly appreciated! It's also a great way to learn more about the codebase.

cjdsellers:
Closing thanks to @Shutchinson and @pkkm

PR #368, opened 2021-07-07
Convert unittest tests to pytest (tests/unit_tests/trading)

Shutchinson:
Issue Ticket: #325

For right now, only `tests/unit_tests/trading/test_trading_strategy.py` has been converted to use pytest. If the conversion is acceptable to the maintainers, I'll convert the rest on this PR.

In order to get the unit tests to run/pass, I did have to modify the `noxfile.py` `tests` session to include the `--extras ALL_EXTRAS` flag, otherwise it complained about a missing `betfairlightweight` import and probably would have complained about the other extras as well. I didn't include the `noxfile.py` changes in this PR as I wasn't sure if it was just me or actually a bug.

Pre-commit hooks were run, unit and integration tests passed. Let me know if I missed anything.

### DIFF
--- a/tests/unit_tests/model/test_model_position.py
+++ b/tests/unit_tests/model/test_model_position.py
@@ -14,9 +14,9 @@
 # -------------------------------------------------------------------------------------------------
 
 from decimal import Decimal
-import unittest
 
 from parameterized import parameterized
+import pytest
 
 from nautilus_trader.common.clock import TestClock
 from nautilus_trader.common.factories import OrderFactory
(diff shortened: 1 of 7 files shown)


cjdsellers:
Hi @Shutchinson,

Great work on this! The only point I would make is that for some of the assertions checking result == expected, the ordering is opposite to `unititest` (frustratingly). So for instance line 184 would be:
```
assert str(strategy) == "TradingStrategy-GBP/USD-MM"
```

Other than this happy for you to go right ahead! :chart_with_upwards_trend:

Shutchinson:
@cjdsellers I've made the changes you requested and expanded the number of files converted. Please review them when you get a chance.

The following files have been converted (7 total):

- `tests/unit_tests/trading/test_trading_*.py`
- `tests/unit_tests/model/test_model_position.py`

PR #380, opened 2021-07-09
Convert backtest and cache tests to pytest

pkkm:
Issue ref: https://github.com/nautechsystems/nautilus_trader/issues/325.

While doing the conversion, I've noticed a `print` in `test_backtest_exchange` on line 1705/1704. Is it supposed to be there?

In addition, there were two `assertTrue`s in `test_data_cache.py` that I'm quite sure were supposed to be `assertEqual`s instead. I've changed them to equality assertions, but this resulted in **two new failing tests**:

- `TestCache.test_bars_when_one_bar_returns_expected_list`
- `TestCache.test_add_bars_when_already_bars_does_not_add`

Could someone take a look at these?

### DIFF
--- a/tests/unit_tests/backtest/test_backtest_exchange.py
+++ b/tests/unit_tests/backtest/test_backtest_exchange.py
@@ -15,7 +15,6 @@
 
 from datetime import timedelta
 from decimal import Decimal
-import unittest
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


cjdsellers:
Yes, unfortunately @Shutchinson managed to complete all of the pytest conversion. This is a great find though and I'll look into this potential bug and make another comment here.

If there's something to fix did you want to do a separate PR for the fix yourself, or are you happy if I apply it?

pkkm:
I'm happy if you apply it. You'll probably be able to do it much quicker if you're the one investigating the bug.

cjdsellers:
For that print statement, looks like the test is WIP before fixing bracket orders which are currently a bit broken. Good spotting though.

cjdsellers:
For those two tests, they were actually errors in the tests themselves. They needed to pass the `bar_type` from the `TestStubs` to the bar (the bar types were different). So thanks for spotting those and I've now fixed :+1:

PR #390, opened 2021-07-14
Betfair fixes

limx0:
Refactor and fix a handful of things in the Betfair adapter
- Fix execution tests that were skipped
- Pull order and instrument from cache rather than computing
- Fix margins and notional value in Betting Instrument & add simple tests
- Add some more test fixtures and split BetfairTestStubs into BetfairTestStubs/BetfairDataProvider
- Add portfolio initial margin test


cjdsellers:
Solid work yet again :+1:

PR #391, opened 2021-07-18
ENH: Added linear regression indicator & crossover helpers

crazy25000:
### Info:
Added a simple linear regression indicator and cross over helpers. There are linear regression methods from external libraries too, this is the fastest vanilla Python version that I know of.

### TODO:
- [x] New unit tests
- [ ] ~Maybe add different regression types (could do it in another PR as an improvement)~


nautilus_trader/indicators/helpers.pyx
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns
+    -------
+    bint
+
+    """
+    if direction == "UP":
+        return cross_up(values1, values2)
+    elif direction == "DOWN":
+        return cross_down(values1, values2)
```
crazy25000, line 22: We could make these an Enum.

nautilus_trader/indicators/linear_regression.pyx
```diff
@@ -0,0 +1,99 @@
...
+#  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+# -------------------------------------------------------------------------------------------------
+
+from statistics import mean
```
crazy25000, line 16: Is there a C version/import that we could use here?

nautilus_trader/indicators/helpers.pyx
cjdsellers: Yes potentially, or if there are ever only two possible values a bool (`bint`) would work also.

nautilus_trader/indicators/linear_regression.pyx
cjdsellers: Numpy does expose Cython 'headers' (`.pxd`), so it is possible to `cimport numpy`.

Definitely something to play around with for performance optimization.

nautilus_trader/indicators/helpers.pyx
cjdsellers: Were you planning on adding further indicators which use this `cross_over` function? I'd suggest until then we keep this function with the `LinearRegression` code.

cjdsellers (commented):
Thanks for your contribution! I'll kick off CI and also left a few review comments.

cjdsellers:
Looking good. We also need to add some tests for this indicator.

cjdsellers:
You'll also need to setup pre-commit which will format your code and run linting checks (otherwise CI will consistently fail).

Run:
```
pre-commit install
```

Which should install the necessary hook which will run it automatically on commit in future.

You can also manually run pre-commit with:
```
pre-commit run --all-files
```

nautilus_trader/indicators/helpers.pyx
crazy25000: I'm not entirely sure where to move these to. They're helpers that can be used directly in indicators or in strategies. I do plan on using them more, but if everyone doesn't find them useful, can remove it from PR and keep it local.

cjdsellers:
Thanks for adding the test, looks good.

Once you manage to setup pre-commit that should clear the check. I'm looking into the `performance` flows cred issue.

nautilus_trader/indicators/helpers.pyx
cjdsellers: You could move the function into the same file, that's where I would move it for now.

cjdsellers:
I just learned that GitHub tokens can only be used on push actions, so I've removed the performance workflow from PRs. If you rebase/merge on the latest `develop` your next commit should clear that check also.

crazy25000:
> Thanks for adding the test, looks good.
> 
> Once you manage to setup pre-commit that should clear the check. I'm looking into the `performance` checks cred issue.

I've always had pre-commit setup, wonder if it's not automatically applying. I've had to apply it manually twice now.

crazy25000:
@cjdsellers I'd also recommend updating checks for PRs to automatically rerun after new commits so you don't have to manually kick them off every time.

nautilus_trader/indicators/helpers.pyx
crazy25000: I will remove the helpers from this PR, that way we can keep it independent of the indicator logic. I will just use it in a local file.

cjdsellers:
> @cjdsellers I'd also recommend updating checks for PRs to automatically rerun after new commits so you don't have to manually kick them off every time.

That's actually a GitHub security feature, its just because you haven't yet landed a commit into the code base, once that occurs CI won't need to be kicked off manually.

cjdsellers:
> > Thanks for adding the test, looks good.
> > Once you manage to setup pre-commit that should clear the check. I'm looking into the `performance` checks cred issue.
> 
> I've always had pre-commit setup, wonder if it's not automatically applying. I've had to apply it manually twice now.

Did you run both
```
pip install pre-commit
```

as well as
```
pre-commit install
```

cjdsellers:
Looking all good. Will merge this a little later.

crazy25000:
> > > Thanks for adding the test, looks good.
> > > Once you manage to setup pre-commit that should clear the check. I'm looking into the `performance` checks cred issue.
> > 
> > 
> > I've always had pre-commit setup, wonder if it's not automatically applying. I've had to apply it manually twice now.
> 
> Did you run both
> 
> ```
> pip install pre-commit
> ```
> 
> as well as
> 
> ```
> pre-commit install
> ```

Yup, it was installed. I ended up deleting the hook and recreating it. Works now.

PR #393, opened 2021-07-20
First pass on adapter template

limx0:

PR #395, opened 2021-07-21
TST: Initial example using data template + interactive brokers connect

crazy25000:
Based on #393 , attempt at creating a base adapter for IB. Managed to get the the connect and disconnect methods with the API so far.

Would be nice if we could have a generic class that would automatically instantiate components like clock, logger, portfolio..etc by passing a simple dict config. Maybe I didn't implement this correctly, but I wasn't able to inherit `TemplateLiveMarketDataClient` and get it to run successfully without constructing these components.

### DIFF
--- a/examples/live/data_template_test.py
+++ b/examples/live/data_template_test.py
@@ -1,0 +1,117 @@
+from abc import ABC
+import asyncio
+
+import ib_insync as ibi
+
+from nautilus_trader.adapters._template.data import TemplateLiveMarketDataClient
+from nautilus_trader.cache.cache import Cache
+from nautilus_trader.common.clock import LiveClock
+from nautilus_trader.common.logging import LiveLogger
+from nautilus_trader.common.uuid import UUIDFactory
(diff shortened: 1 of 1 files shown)


cjdsellers:
Ok great progress!

So the intention of the templates is to act as examples, rather than base classes. So you could inherit directly from `LiveMarketDataClient` and `LiveExecutionClient` (ABC not necessary).

So the `TradingNode` allows many different clients to be built using the builder pattern through abstract factories. Checkout `node_builder.pyx` and how clients are added to a node in the examples. You'll see that's where the clock and logger components will come from.

Config will come from the dictionary which is passed into `TradingNode`.

crazy25000:
> So the intention of the templates is to act as examples, rather than base classes.

If these templates are the minimum requirements for every adapter, should we create base classes?

>So the TradingNode allows many different clients to be built using the builder pattern through abstract factories. Checkout node_builder.pyx and how clients are added to a node in the examples. You'll see that's where the clock and logger components will come from.

I've been playing with `TradingNode` and do like the builder pattern.

PR #443, opened 2021-09-22
Streaming backtest

limx0:
# WIP

This is a very WIP first pass on adding streaming to backtests. 

This PR adds a `batch_size_bytes` to the `BacktestNode._run` method and a new `backtest_runner` method that allows batching loading data and passing to the engine. 

Still to do:
- [x] This will almost certainly break dask tests
- [ ] We probably want to spin up a separate `Process` to handle loading the data and passing through to the main thread via a Queue (so the engine just pops data from the queue and runs)


cjdsellers:
I've upgraded the logic of the `BacktestEngine` runs to help support this.
There's also been a change to the method signature of `run()` which no longer takes the strategies there.

Added `engine.run_streaming()`.
Added `engine.end_streaming()`.

PR #492, opened 2021-11-08
Add custom_summaries backtests

limx0:
This PR adds a `custom_summaries` optional config to `BacktestEngineConfig`. This allows adding custom callbacks to run after a backtest and be gathered with the `BacktestResult`.

A simple example: 

```python

def buy_count(engine):
    return {"buys": len([o for o in engine.cache.orders() if o.side == OrderSide.BUY])}

# Setup backtest
backtest_run_config = BacktestRunConfig(
    ... # Other inputs
    engine=BacktestEngineConfig(custom_summaries={"buy_count": buy_count})
)
node = BacktestNode()

# Run backtest
results = node.run_sync([backtest_run_config])
assert results[0].custom_summaries["buy_count"] == {"buys": 48}



```

### DIFF
--- a/nautilus_trader/backtest/config.py
+++ b/nautilus_trader/backtest/config.py
@@ -15,11 +15,13 @@
 
 import dataclasses
 import importlib
+import inspect
 from datetime import datetime
-from typing import Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union
 
 import pydantic
 from dask.base import tokenize
+from pydantic import validator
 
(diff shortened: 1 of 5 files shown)


limx0:
Going to close this one until I have a more clear use case

cjdsellers:
We could probably add a few of the extra summaries you wanted by default too, as I think they're useful.

What do you think?

limx0:
Possibly - but I think I prefer to use of some custom hook because it will allow users to retrieve the exact data they want/need for their use case.

PR #498, opened 2021-11-21
Fix typeo in docs and test codes

ian-wazowski:
Fix typeo in docs and test codes.


cjdsellers:
Great finds, many thanks! :pray:

PR #528, opened 2022-01-03
Docs quick start

limx0:
### Docs

This PR adds a couple of new doc pages
- `getting_started/quick_start` - a "get up and running as fast as possible" backtest quick start example
- `user_guide/loading_external_data` - how to load external data into the data catalog

### Code
It also makes a handful of changes to ensure these docs work:
- Logs the time taken to load data and count of events from parquet and subsequently into the engine when using `BacktestNode`
- Exposes the `LoggerAdapter` in the `BacktestEngine` so the `BacktestNode` can log some details about data loading.
- Allows returning the raw `BacktestEngine` from a `BacktestNode.run`, so users can access more detailed information about the backtest run
- Allows `BacktestRunConfig` to accept a `TradingStrategy` instance as well as an `ImportableStrategyConfig` for instances where the BacktestRunConfig is used in a script or notebook, with a TradingStrategy defined in the same file. 
- Remove the `source` kwarg from `ImportableActorConfig`. This is untested and unused.


.gitignore
```diff
@@ -45,4 +45,5 @@ PERF.JSON
 output.json
 .ipynb_checkpoints
 examples/backtest/notebooks/catalog
-nautilus_trader/**/.gitignore
\ No newline at end of file
+nautilus_trader/**/.gitignore
+docs/**/*.ipynb
```
cjdsellers, line 49: Are we not committing the raw notebooks anymore?

PR #626, opened 2022-04-30
Add benchmark setup

twitu:
# Pull Request

Benchmark f64_to_fixed_i64
Add tests for Buffer

## Type of change

- [ x ] Add tests
- [ x ] Add benchmarks

tested and benchmarked


twitu:
Running `cargo bench` generates detailed reports for the benchmarks. Currently however there is one issue that the iai bench is running but not returning formatted results. It's something to do with `Cargo.toml` configuration which only allows adding one name field.

I've raised an issue regarding this - bheisler/criterion.rs/issues/575

PR #638, opened 2022-05-07
Couple of doc fixes

limx0:
- Fix quick_start / backtest example docs
- Add spiel about running docker jupyterlab


cjdsellers:
Many thanks!

PR #642, opened 2022-05-14
Implement logging in Rust

twitu:
# Pull Request

Implement logger class and helper enums in Rust. Export as a python module under `common.logging`.

## Type of change

Manually running the python interpreter and calling the methods


twitu:
[ ] - figure out how to handle annotations
[ ] - pass msg and component from python more efficiently probably using PyUnicode or some Python native object type
[ ] - figure out if what other methods are needed to start using it in TestEngine
[ ] - try to create the module system in a module a way. as in `logging.rs` creates the logging module not at the top level `lib.rs`.

PR #647, opened 2022-05-17
[WIP] Trait and Macro for exporting C API for identifiers

twitu:
# Pull Request

Macro to implement C API for identifiers

## Type of change
- [ x ] Refactor

This is what the macro generates. Requesting feedback before applying this across other identifiers.

```
    impl InterconvertPyString for Venue {
        #[export_name = "Venue_from_pystr"]
        unsafe extern "C" fn from_pystr(ptr: *mut ffi::PyObject) -> Self {
            Venue {
                value: Box::new(pystr_to_string(ptr)),
            }
        }
        #[export_name = "Venue_to_pystr"]
        unsafe extern "C" fn to_pystr(&self) -> *mut ffi::PyObject {
            string_to_pystr(self.value.as_str())
        }
    }
```

Not tested currently


twitu:
I've also added a `DeRef` trait which allows Venue to be referenced as if if were it's inner string content itself. It's a useful idiom when operating on the Rust side of things.

twitu:
A couple of factors combine to make this impossible.
1. cbindgen does not generate bindings for functions defined inside traits.
2. this means that we need unique names for each ffi functions for e.g. `uuid4_to_pystr`
3. Currently it is not possible to generate identifier names from passed tokens inside a macro

This pretty much kills any chance of generating extern functions generically from macros.

[SO answer](https://stackoverflow.com/questions/24905160/is-it-possible-to-write-a-rust-macro-that-will-expand-into-a-function-method-sig)
https://github.com/rust-lang/rust/issues/12249
https://github.com/rust-lang/rust/issues/29599

PR #648, opened 2022-05-19
Fix PositionStatusReport net_qty sign

sidnvy:
# Pull Request

Quantity object can't be negative, `PositionStatusReport.net_qty` will always be positive here.
https://github.com/nautechsystems/nautilus_trader/blob/5446cb7fe12a8bc9477997f2c71bc20eaf3bfc33/nautilus_trader/execution/reports.pyx#L416
Please include a summary of the changes.

## Type of change

Tests added. (pytests can't run on my local env, it reports segmentation fault, will check later


cjdsellers:
Good catch!

PR #661, opened 2022-06-01
Rust ffi methods for monotonic time

twitu:
# Pull Request

Add Rust ffi methods for monotonic time for timestamp, ms, us, ns

## Type of change

Test cases for all

### DIFF
--- a/nautilus_core/Cargo.lock
+++ b/nautilus_core/Cargo.lock
@@ -396,6 +396,7 @@ name = "nautilus_core"
 version = "0.1.0"
 dependencies = [
  "cbindgen",
+ "lazy_static",
  "pyo3",
  "uuid",
 ]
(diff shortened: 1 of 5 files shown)


twitu:
Some key ideas behind the changes.

* Monotonic time has an arbitrary starting point and all offset is calculated from that starting `instant`. So any function that returns elapsed time must have access to the "instant". Since instant is readonly, and shareable across threads it's better to have to it available as a static variable. This removes the need from passing "instant" to the function and also passing across the C boundary.
* I evaluated `lazy_static` and `once_cell` which are the defacto heap allocated static variable libraries. While `once_cell` has a more flexible API. The flexibility comes with at a cost, `once_cell` apis are `pub fn get(&self) -> Option<&T>`. However this option really is unnecessary when we know "instant" will always be initialized in static. This is not a problem with `lazy_static` (it uses a tiny bit of unsafe :grin:)

Final thoughts
* Renaming the methods to something that indicates that it returns elapsed time/duration might be better. From `mono_unix_timestamp` to
    * `time_elapsed`
    * `mono_time_duration`
    * ... :question: 
* We can benchmark between `once_cell` and `lazy_static` before finalizing on them. However an integrated bench like running the whole engine might be a better choice than a micro bench.

PR #671, opened 2022-06-10
C ABI compatible test clock

twitu:
# Pull Request

C compatible test clock with appropriate types for handlers and event that cross Python boundary

## Type of change

Few tests added. More tests to be added.


cjdsellers:
As discussed, will merge post talk. Many thanks! 🙏

PR #674, opened 2022-06-12
Feature: ABC for DataCatalog

jordanparker6:
# Pull Request

An ABC for the DataCatalog has been defined ('BaseDataCatalog').

The current DataCatalog implementation has been refactored to use the BaseDataCatalog and renamed to an implementation name ('ParquetDataCatalog'). Future work may define an implementation of the DataCatalog for SQL-like interfaces.

Please note that the renaming of the DataCatalog to ParquetDataCatalog will introduce some breaking changes for end users. This could be avoided by creating a generic DataCatalog class that operates as a factory for each implementation where the default parameters initialises the ParquetDataCatalog. For discussion if appropriate.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

The test cases have been updated to reflect import path changes and renaming. No additional test cases have been added because no new functionality has been introduced.


cjdsellers:
Really excellent work on this.

Looks like there's some CI failures for pre-commit. Please run:

```
pre-commit install
pre-commit run --all-files
```

And then diff again 👍

jordanparker6:
Thanks. I have pushed changes post observing the CI issue. 

Please note that a fair few other functions outside of the DataCatalog implementations are coupled to the ParquetDataCatalog's methods / attributes. This may be something I can explore / de-couple post a go at implementing a SQL-like DataCatalog, pushing more common functionality to the BaseDataCatalog. 

The SQL-like DataCatalog may be more involved and will require some decisions around ORM to fit with the current entities. I also don't want to introduce a I/O bottleneck. 

Keen to get my hands dirty with some rust if there are some bite size tasks you need help during the migration of the Cython to Rust. Feel free to hit me up in Discord.

cjdsellers:
Great stuff!

We want to keep everything stable until after the talk this week, so expect the merge to happen after that.

Issue #662, opened 2022-06-02
Add more indicators


graceyangfan:
As disscussed in #515 I decided to moving more indicators to nautilus_trader with the inside functions, the main reference libaries are:
(1) [pandas-ta](https://github.com/twopirllc/pandas-ta)
(2) [finta](https://github.com/peerchemist/finta)
(3) https://github.com/mrjbq7/ta-lib
now,I am working on pandas-ta.In fact,I have rewrite some pandas-ta indicators with nautilus_trader's inside functions,I will gradually pull them.

cjdsellers:
Many thanks for your efforts!

I haven't looked too closely at libraries 1 or 2 yet, however we do want to be mindful of bloating the project dependencies. I think more indicators are valuable, at worse we can make them optional.

graceyangfan:
@cjdsellers Yes,using those libraries as a reference.So it is not always to repetition these indicators 100%.Now,the two added indicators give the same results with pandas-ta during my test,but I am not sure if I should include this test in unit-test code, because this introduce a new library.

cjdsellers:
I think your current approach is good. Just using the other library is a reference to aid development, and inform the test cases - without having to take a dependency.

PR #682, opened 2022-06-19
Add VerticalHorizontalFilter Indicator 

graceyangfan:
# Pull Request

Add VerticalHorizontalFilter Indicator  #662 

## Type of change

Add unit test for indicator,personal test provide same values as pandas-ta


cjdsellers:
Thanks so much for another one! If I may offer a few small nitpicks:

- When a method only had a single parameter, it's probably not necessary to take a newline, e.g. this is probably fine:
```python
cpdef void update_raw(self, double close) except *
```

- We generally don't add a docstring to private methods (those named with a leading underscore), and so for example I've been removing the docstring from this method (the name is also nice and clear anyway):
```python
cdef void _check_initialized(self) except *
```

I think its much neater code grouping that initialization logic into a method with a clear name like that as you've been doing. However consider the performance implications of an extra method call on every execution of `update_raw`, it depends on your use case, small overhead but it would be there. I'm happy to leave as is though.

Many thanks again! :pray:

cjdsellers:
One small performance optimization I just added, importing `fabs`  from `libc.math`. Cython exposes some libc for convenience which can be very handy.

graceyangfan:
@cjdsellers  tanks a lot，performance is very import！the fact is that I copy some code from exist indicator examples，all your suggestions is very good，I will take care about it

PR #696, opened 2022-07-13
Rust catalog C API

ghill2:
Additions:

1. Added catalog package to nautilus_core.

2. C API: Bar & QuoteTick:

* Return vector of rust structs to Cython.
* Initialize vector of rust structs to related class in Cython.

3. C API: Filter expressions -> Rust.

* Filter expressions are passed to Rust as strings str(ds.field).
* In Rust data is extracted from the filter_expr strings to perform the filter.

4. Data utility methods to benchmark/test the catalog.


twitu:
Catalog will behave like a iterator that will return a vector/chunk of structs.

1. It's inputs will be file path and filter expression, which you've already added. Although I'm not sure how best to represent filter information, keeping it as a string sounds good for now.
2. Catalog will return a vector of structs so your conversion from to C compatible type for the vector is pretty useful. we'll need it while we're interfacing with python
3. One thing you can add to the catalog struct is a next method, which take a catalog and return the next chunk of structs.

You can take a look at `schema.rs` in the `rs-schema` branch. It has a working example of how a struct specific catalog will be consumed. It's not well documented right now, but take a look at the `load_data_from_csv` for how a catalog is written `read_quote_tick_from_parquet` and how it is read.

cjdsellers:
Great progress @ghill2 and @twitu

My only comment there is we should follow the package structure of `nautilus_trader` and call the library `persistence`, as it may make sense for it to contain code other than the catalog.

cjdsellers:
@twitu gives his blessing to merge to his branch firstly :+1:

Issue #693, opened 2022-07-10
Improve for request_bars function 


graceyangfan:
# Feature Request

I am trying to using strategy with 10-SECOND-LAST-INTERNAL with binance,but the strategy need several  bars to initialize, with EXTERNAL bars, this can be realized by request_bars function,however for INTERNAL bars,I have to wait for live data to aggregate.So we need to improve request_bars functions:
(1)request EXTERNAL history bars several times to exceed the 1000 limit bars
(2)request INTERNAL history bars. first request  histrory trade_tick or quote_tick  and then auto aggregate them to get  INTERNAL history bars,of course there still need to deal with  several requests to exceed 1000 limit ticks

cjdsellers:
I believe @jordanparker6 has a PR in the works to address pagination of historical bar requests?

jordanparker6:
@cjdsellers yeah a while loop is resolving the pagination issues

graceyangfan:
@jordanparker6  any process on this,I have get binance future to request EXTERNAL history bars several times to exceed the 1000 limit bars,still need some time to test it

graceyangfan:
@cjdsellers Accurately,I have made improve to request_bars with several request to get external bars more than 1000,but binance did not provide more data for tick data,so it's difficult to make request_ticks work as I expect,maybe in the future we need to reuqest from a database or stream from other data provider,but that's a another topic,maybe close this?

cjdsellers:
Hi @graceyangfan 

We could, although we haven't properly handled pagination yet to request > 1000 bars from Binance. We never managed to merge the work by @jordanparker6 and I've noticed a couple of users have mentioned this issue, and implemented their own workarounds.

You're right that Binance doesn't make historical ticks available through the public API.

PR #736, opened 2022-08-20
Improve request_bars function for Binance

graceyangfan:
# Pull Request
#693 
1.Allow users to call request_bars to get more than 1000 EXTERNAL bars limit.
2.The `on_bar `function will perform for historical bars and living bars. A symbol `is_historical_bar` is adding to Strategy class to distinguish if current bar is a historical bar.

## Type of change

A new stratgy in request_history_bars.py and a live example binance_history_bars_testnet.py is used to test.More test will be added.


cjdsellers:
Hi @graceyangfan 

Thanks for your work here. I reviewed the diff several times and just have two questions:
- I can't see anywhere the data engine refers to the `on_preprocess_bars` method? Did you intend on doing another diff to wire this up?
- I see for the partial bar you changed the code from popping from the list of bars to just indexing at `[-1]`. I'm not sure this is a good change because in this case a partially formed bar would be added to the cache, and be passed through logic which expects a closed bar as per its `BarType` spec. Thoughts here?

One slight preference I would have also is changing the method name to `on_bar_preprocess` just to highlight it relates to bars and keeps it nicely grouped visually with `on_bar`.

graceyangfan:
@cjdsellers The logical of the data processing is:(1)request history bars,history bars  are sending into `on_preprocess_bars` to make some preprocess on bars , and process some registered nautilus indicators (2)after history bars,living bars are sending into `on_preprocess_bars`  to make some preprocess on bars,then nautilus_trader inside indicators compute,then on_bar where to create signals  for buying or selling.So `on_preprocess_bars` is just the same as nautilus_trader inside indicators,it need history bars to initialization and it need to perform both for history bars and living bars, nautilus_trader inside indicators is automatically, but it is only used for one bar_type. on_bar is only used for living bars.This is the difference.Of course,we could change on_bar to deal history bars and living bars and drop `on_preprocess_bars`  function.The rease I include the partial bar  is during my test I find that if I drop it,there will lose one bar between history bars and living bars in the end of history bars,maybe I need more test to confirm it.

cjdsellers:
I think rather than adding another method into the project, you could handle this by writing some logic in the strategy as per the suggestion in Discord https://discord.com/channels/924497682343550976/924499913386102834/1010423470443724820

On the arrival of each bar, check if you're initialized, if not, check for the existence of cached bars, if they are there then you can perform any preprocessing or other computation you like. The only difference here is that this occurs in the existing `on_bar` method, rather than adding another one into the platform (which then has some asymmetry as wouldn't we need to add a preprocess to every other data type which can include historical objects?)

graceyangfan:
@cjdsellers  As your suggestion,I will move on_bar to perform history data  like this: 
`
    def on_bar(self, bar: Bar, is_historical = False):
          if is_historical:
              do something 
          else:
              do something 
`

graceyangfan:
@cjdsellers As your suggestion,I will move on_bar to perform history data like this:
```
    def on_bar(self, bar: Bar, is_historical: bool = False):
          do something (for history bars and living bars preprocessing) 
          
          if is_historical:
             return 
          # just for living trading 
          do something to generate trading siginals 
```
This looks much better, but lots of examples and codes need to correct where has on_bar function.
Or something like this:
```
    def on_bar(self, bar: Bar):
          do something (for history bars and living bars preprocessing) 
          
          if self.is_historical_bar:
             return 
          # just for living trading 
          do something to generate trading siginals 
```
`is_historical_bar` adding to Strategy class

cjdsellers:
Were you able to write the logic you needed into the new `on_historical_data` method?

graceyangfan:
> Were you able to write the logic you needed into the new `on_historical_data` method?

@cjdsellers It is enough!I will do this at this  weekend!

cjdsellers:
Apologies, I could have phrased the question more clearly. Is the new functionality sufficient that we can now close this PR?

graceyangfan:
@cjdsellers Accurately,I have made improve to request_bars with several request to get external bars  more than 1000,but binance did not provide more data for tick data,so it's difficult to make request_ticks work as I expect,maybe in the future we need to reuqest from a database or stream from other data provider,but that's a another topic.

PR #792, opened 2022-09-27
Fix strategy clock not advancing

limx0:
Fixes strategy clock not advancing with backtest


cjdsellers:
Thanks Brad!

This bug must have been introduced  in the last few days when the kernel was decythonized.

Issue #836, opened 2022-10-31
Raise exception if `.` in Symbol or fix catalog bar_type suffix


rsmb7zhx:
# Feature Request

Not sure if this should come under bug however I would suggest to raise exception if '.' is present when creating Symbol and Venue.
If there is dot and you write parquet catalog using 'write_objects' it will and then write again with different bar_type, it will eventually overwrite the file on disk because suffix will never return proper bar_type in that case `suffix = df.iloc[0]["bar_type"].split(".")[1]`.

limx0:
Nice find @rsmb7zhx - this is definitely a bug.

Could you give us an example of a symbol/venue you're trying to write? This will help with us debugging/testing.

rsmb7zhx:
Because IB adapter is not complete yet and EUR/USD was causing issues, so I though quick fix is using IB localSymbol which is EUR.USD format for currency pairs (because the benefit is you don't need reverse conversion when dealing with IB objects). So I was download BID and ASK historical data and to my surprise files were being written on data receipt but file numbers were not increasing in folder :-D
I just noticed that because `InstrumentId.from_str('EUR.USD.IDEALPRO')` works without any issue and still allowing localSymbol fix could be `df.iloc[0]["bar_type"].split(".")[-1]`.

cjdsellers:
Now fixed from #840, many thanks!

PR #840, opened 2022-11-02
Fix to return Venue from Instrument ID

rsmb7zhx:
# Pull Request

Modified to make sure Venue is returned from Instrument ID similar to InstrumentId.from_str even in case Symbol have dot. This will fix the actual issue reported in #836.

## Type of change

Improvement in Instrument ID string split for the Venue.

Writing catalog will use proper Venue name which is messed up otherwise when Instrument ID have dot i.e. 'EUR.USD.IDEALPRO'


cjdsellers:
Thanks once again!

I just changed the base branch to `develop`. Please update the PR and we'll see if that kicks off the `build_pr` workflow.

rsmb7zhx:
Thanks. Yes I will choose 'develop' branch for any future changes, I think thats the correct way.

PR #850, opened 2022-11-09
Bar prices in get_xrate

ghill2:
# Changes

Changes to the get_xrate method. If no quotes are found for an instrument, the bar close prices are used instead.

Only bars with PriceType.BID, PriceType.ASK are supported.
Only uses the Bar prices if both bid and ask bars exist for the instrument.

## Type of change

[x ] New feature (non-breaking change which adds functionality)

How has this change been tested?

test_get_xrate_fallbacks_to_bars_if_no_quotes_returns_correct_rate
test_get_xrate_fallbacks_to_bars_if_no_quotes_returns_correct_rate_with_add_bars

### DIFF
--- a/nautilus_trader/cache/cache.pyx
+++ b/nautilus_trader/cache/cache.pyx
@@ -104,6 +104,7 @@ cdef class Cache(CacheFacade):
         self._trade_ticks = {}                 # type: dict[InstrumentId, deque[TradeTick]]
         self._order_books = {}                 # type: dict[InstrumentId, OrderBook]
         self._bars = {}                        # type: dict[BarType, deque[Bar]]
+        self._bars_for_instrument_id = {}      # type: dict[PriceType, dict[InstrumentId, Bar]]
         self._currencies = {}                  # type: dict[str, Currency]
         self._instruments = {}                 # type: dict[InstrumentId, Instrument]
         self._accounts = {}                    # type: dict[AccountId, Account]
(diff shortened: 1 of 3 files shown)


nautilus_trader/cache/cache.pyx
```diff
@@ -2072,17 +2087,25 @@ cdef class Cache(CacheFacade):
...
-            if not ticks:
+            if ticks:
+                bid = ticks[0].bid
+                ask = ticks[0].ask
+            else:
                 # No quotes for instrument_id
-                continue
+                bid_bar = self._bars_for_instrument_id[PriceType.BID].get(instrument_id)
+                ask_bar = self._bars_for_instrument_id[PriceType.ASK].get(instrument_id)
+                if not bid_bar or not ask_bar:
```
cjdsellers, line 2104: Cython will work more efficiently with the full `is None` rather than Python object truthiness

I can tidy this up though

nautilus_trader/cache/cache.pyx
```diff
@@ -1115,6 +1125,11 @@ cdef class Cache(CacheFacade):
         for bar in bars:
             cached_bars.appendleft(bar)
 
+        bar = bars[-1]
+        cdef PriceType price_type = bar.bar_type.spec.price_type
+        if price_type in (PriceType.BID, PriceType.ASK):
```
cjdsellers, line 1130: Will be more efficient breaking this out into an `if else` block, rather than the Python tuple thing. 

I can tidy this up though

cjdsellers (approved):
Great work, I think this is useful! :pray:

PR #884, opened 2022-12-02
Fix InstrumentId functions

twitu:
# Pull Request

`instrument_id_new` was used in two different ways in the code base. Most commonly to pass python strings as arguments but sometimes also to pass Symbol and Venue objects. However the implementation was assuming the inputs are only strings. This PR adds two functions for these two types of calls.

## Type of change

Test with large allocations. End to end test is remaining


twitu:
Implemented in https://github.com/nautechsystems/nautilus_trader/commit/e2a2882212fe053a2802070ebd835d1de71d473b

PR #888, opened 2022-12-03
Change docs for bypassing risk engine

DownBadCapital:
# Pull Request

Docs suggest that bypassing the risk engine turns off all risk checks. But, it still does rate limit checks.


cjdsellers:
I traced the code paths and you're correct, at the end of the `execution_gateway`, all submit order commands are then sent to the order throttler.

PR #892, opened 2022-12-04
IB balances filter and placeOrder for intended account_id

rsmb7z:
# Pull Request

This PR will make sure balances are returned for the intended account in case of multiple accounts on the TWS gateway. Also it will make sure the orders are placed to that intended account only. Furthermore this will add verification check that logged in account of the TWS/Gateway is same as expected, being logged into wrong account (account_id not matching with the Nautilus Config account_id) will Engine will not receive any account balances and orders will not be accepted by the Gateway.

## Type of change

Tested with TWS connectivity.


limx0:
Thanks for the PR @rsmb7z! We will look at this today. 

Just tagging for @cjdsellers  - it looks like this depends on #891

PR #898, opened 2022-12-06
Reuseability of IB Contract/Dict to InstrumentId

rsmb7z:
# Pull Request

This PR mainly provides re-usability to convert IB Contract to InstrumentId i.e. using the same dict in instrument_filters to provide InstrumentId in strategy etc (user needs to maintain one type of symbols to initiate Nautilus). So this way when providing the IB symbols in Instrument Provider Config, user can use this function to construct InstrumentId for strategy etc instead of manually defining or using duplicate code for helper function.

Few more changes are as follows:

- Changed `details.contract.primaryExchange or details.contract.exchange` to explicitly check if its 'SMART' then use primaryExchange because there are situations where contract may have primaryExchange set to '' (null string) instead of None which leads to error in InstrumentId creation (rust error with unexpected Venue). This doesn't apply to Options as they use SMART as the only option for exchange.
- STK: Modified to replace empty space in Equity localSymbol with hyphon. Reference equitites are 'BRK B' and 'BF B', they have space in exchange local symbol but on Yahoo refered as BRK-B and BF-B, some providers like finviz use dot instead of hyphon.

## Type of change

Converted InstrumentId as expected and all existing tests for IB are passed.


limx0:
LGTM!

PR #899, opened 2022-12-07
IB contract account cleanup

limx0:
Some minor cleanup in IB adapter:

- Add `account_id` config options (and TWS_ACCOUNT env var)
- Cleanup contract_details test stubs (moving from pkl to code definitions)


limx0:
@rsmb7z just FYI

PR #903, opened 2022-12-07
Instrument filter

limx0:
# Pull Request

Possibly clean up to instrument filters for IB adapter


rsmb7z:
Hi @limx0 
This is great, instrucing config class for IB Contracts and converstions utilities embedded into. Maybe I am excited and too early to comment but following points may help:
- We need to assign `localSymbol` (`symbol` field is available in IB Contract but that may have different value for example for Forex if we provide symbol='EUR' then we provide currency='USD' - I believe that make it overly complicated) when converting from InstrumentId and take care of string reverse parsing i.e. char changed in ib_contract_to_instrument_id
- `secType` will always be required to set (IB's requirement).
- Please add `localSymbol` as input parameters. `symbol` can be removed to avoid unncessary ambiguity.
- Whenever `primaryExchange` has value `exchange` needs to be set *SMART* (this is applicable to stocks, as other securities trade on their particular exchange so only `exchange` will exist whereas Options will have `exchange = 'SMART'` without the `primaryExchange'

limx0:
Thanks for the feedback @rsmb7z  - I'll fix those issues (I was just playing around with fields without testing).

This is a WIP at this stage, and waiting some feedback from @cjdsellers on https://github.com/nautechsystems/nautilus_trader/issues/904

limx0:
Closing this - implmented in #1036

PR #913, opened 2022-12-15
Fix memory leak

ghill2:

**Changes summary.**

[x] Add to_cstr rust C-API methods for classes/methods:
    [x] QuoteTick
    [x] TradeTick
    [x] BarSpecification
    [x] BarType
    [x] Bar
    [x] Currency
    [x] CurrencyCode
    [x] CurrencyName
    [x] UUID
    [x] AccountId
    [x] ClientId
    [x] ClientOrderId
    [x] ComponentIdComponentId
    [x] ExecAlgorithmId
    [x] InstrumentId
    [x] OrderListId
    [x] PositionId
    [x] Symbol
    [x] TradeId
    [x] VenueOrderId
    [x] Venue
    [x] logger_get_trader_id()
    [x] logger_get_machine_id()
    [x] time_event_name()

[x] Add cython cstring conversion method: cstr_to_pystr
    
[x] Update cython classes to use c_str methods
        [x] QuoteTick
        [x] TradeTick
        [x] BarSpecification
        [x] BarType
        [x] Bar
        [x] Currency
        [x] CurrencyCode
        [x] CurrencyName
        [x] UUID
        [x] AccountId
        [x] ComponentId
        [x] ClientId
        [x] InstrumentId
        [x] ClientOrderId
        [x] OrderListId
        [x] PositionId
        [x] Symbol
        [x] TradeId
        [x] Venue
        [x] VenueOrderId
        [x] ExecAlgorithmId
        [x] logger_get_trader_id()
        [x] logger_get_machine_id()
        [x] time_event_name()

### DIFF
--- a/nautilus_core/common/src/logging.rs
+++ b/nautilus_core/common/src/logging.rs
@@ -12,7 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
-
+use std::os::raw::c_char;
 use std::{
     fmt::Display,
     io::{self, BufWriter, Stderr, Stdout, Write},
@@ -22,7 +22,7 @@ use std::{
 use pyo3::ffi;
 
 use nautilus_core::datetime::unix_nanos_to_iso8601;
-use nautilus_core::string::{pystr_to_string, string_to_pystr};
+use nautilus_core::string::{pystr_to_string, string_to_pystr, string_to_cstr};
 use nautilus_core::uuid::UUID4;
 use nautilus_model::identifiers::trader_id::TraderId;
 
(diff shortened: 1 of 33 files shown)


cjdsellers:
I really appreciate the effort you've put in here, and looks like we only just beat you to a fix :pray: 

I see you went with a similar approach and extracted `string_to_cstr`, using C strings. We actually started with C strings on the first implementation, then transitioned to fixed buffers, then finally to pointers - which yielded the best performance on backtest benchmarks, but of course is so low-level we're responsible for decrementing our own reference counts so that memory can be deallocated (seems so obvious now).

So closing this for now, but we know where it is if we need to refer to an alternative way of handling strings! if you were able to run some side by side benchmarks with the latest fix this could be really interesting too (to confirm pointers are the right way to go while Cython is still in the codebase). I know these things take time to setup too, so you may not have the bandwidth for it.

Cheers!

PR #921, opened 2022-12-22
fix windows build: threads > 60

ghill2:
Fixes Cython hanging during cythonizing on Windows with thread count > 60


cjdsellers:
Thanks for the fix! looks like a pre-commit fail again. This should fix it:

```
pre-commit install
pre-commit run --all-files
```

cjdsellers:
Now implemented from 74af270888e21658ceff209f3f67295a9afab0e3

Many thanks! :pray:

PR #925, opened 2022-12-23
implement to_str cstring

ghill2:
Implement string conversion from Rust to Python using c strings.


cjdsellers:
Amazing work, and I think this is a huge improvement. Many thanks for your efforts here :pray:

PR #952, opened 2023-01-10
Improvements in IB Data Subscription

rsmb7z:
# Pull Request

- Used NT create_task for activating subscription instead of relying on ib_insync which requires nest_loop and is incompatible with uvloop
- Improvements in on_historical_bar_update event
- Added error handling hook to IB
- Resubscribe the Bar subscriptions upon connection lost between Gateway and IB i.e error codes 1101 and 1102 which is lost otherwise

## Type of change

Live subscription with TWS/Gateway and disconnecting internet connectivity to reproduce connection lost events.


limx0:
LGTM, thanks @rsmb7z !

Issue #897, opened 2022-12-06
Update or Ignore new Bar with same ts_event


rsmb7z:
# Feature Request

I noticed that if duplicates Bars (with same ts_event) are sent to Nautilus core it will simply process that as new Bar. If we can have some sort of flag to update or ignore, so in the adapter can be set accordingly to situation. But even without flag default behavior should be either update or ignore instead of adding as new bar.

cjdsellers:
Hi @rsmb7z

I just need a little more context on this one, under what scenario did duplicate bars occur?

When you say sent to Nautilus core, did you mean sent to the `DataEngine`? if so, I'm guessing these are externally aggregated bars?

rsmb7z:
Hi @cjdsellers 

> I just need a little more context on this one, under what scenario did duplicate bars occur?

**Live DataClient:**
So this requirement makes more sense for Swing Trading or Higher timeframes than HFT, while the system is as good as for all of these trading styles. Here is a scenario:
- I subscribe to one-hour external Bar with flag keepUpToDate set to True (referring IB as I am working)
- Now IB will keep sending this one-hour updated Bar every 5 seconds until hour is complete. Please note the Bar's unique timestamp (`ts_event`) identify in IB for the Bar is bar's starting time.
- Currently if the adapter's DataClient send the Bar to DataEngine it will be sent to subscribers as-is i.e. Strategy and Indicators will simply append in inputs deque as new Bar. On the other hand DataClient is aware it is updated Bar and can simply ignore it but then there is no purpose of keepUpToDate flag and Strategy is blind for one-hour until bar is complete. What if its even higher timeframe, so either you get the indicators values messed or you get nothing.

**Backtesting:**
Although data is assumed to be unique when loading to Backtest DataEngine but if for some reason you send duplicates, data will continue to process the bars and send to strategy/indicators (applies to ticks and bars both).


> When you say sent to Nautilus core, did you mean sent to the `DataEngine`? if so, I'm guessing these are externally aggregated bars?

Yes, by sent to core I meant DataEngine and is about externally aggregated.
For tick data I don't see the problem as of now but for Bars, before the on_bar in Strategy is called where possibly the user can control these checks but that is pretty much useless because before this Strategy.handle_bar will send the Bar to Indicators, apart from writing this common logic in each strategy, it will make strategy trade logic complicated.

So following could be possible solution:
- We need to possibly hande in DataEngine so data subscribers receives clean data
[...]

PR #955, opened 2023-01-11
Update or Ignore new Bar with same ts_event

rsmb7z:
# Pull Request

This fullfills the requirements described in #897.

- Fixes DataEngine to publish Bars with the latest ts_init [adds saftey specially in Backtesting where duplicate data may be there as error]
- Adds optional `is_revised` flag for Bar. Default is false and doesn't impact the existing implementation. Will open the way for requeting UpToDate bar from brokers and set this to True, which will allow actor/indicator to perform actions accordingly, specially indicators in this case i.e. append if is_revised=False else replace. ***We will need to make Indicators aware of `Bar.is_revised` though to get new functionality.***

## Type of change

- Intentionally adding duplicate External Bars, DataEngine will no longer publish duplicates.
- Requesting UpToDate bar works as expected.

### DIFF
--- a/nautilus_trader/cache/cache.pyx
+++ b/nautilus_trader/cache/cache.pyx
@@ -1046,6 +1046,8 @@ cdef class Cache(CacheFacade):
             bars = deque(maxlen=self.bar_capacity)
             self._bars[bar.bar_type] = bars
 
+        if bar._is_revised:
+            bars.popleft()
         bars.appendleft(bar)
 
         cdef PriceType price_type = <PriceType>bar._mem.bar_type.spec.price_type
--- a/nautilus_trader/data/engine.pxd
+++ b/nautilus_trader/data/engine.pxd
@@ -47,6 +47,7 @@ cdef class DataEngine(Component):
     cdef dict _routing_map
     cdef dict _order_book_intervals
     cdef dict _bar_aggregators
+    cdef dict _topic_last_ts
     cdef bint _build_time_bars_with_no_updates
 
     cdef readonly bint debug
--- a/nautilus_trader/data/engine.pyx
+++ b/nautilus_trader/data/engine.pyx
@@ -29,6 +29,7 @@ Alternative implementations can be written on top of the generic engine - which
 just need to override the `execute`, `process`, `send` and `receive` methods.
 """
 
+from libc.stdint cimport uint64_t
 from typing import Callable, Optional
 
 from nautilus_trader.common.enums import LogColor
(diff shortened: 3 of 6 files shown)


cjdsellers:
Thanks for the PR @rsmb7z 

I think I understand the requirements, what I'll do is implement something a little more efficient, and make this a config option to retain the existing behaviour.

For the `is_revised` flag on bars, how do you know the revision is for the bar at index 0 which is getting popped?

rsmb7z:
Hi @cjdsellers 
Thanks that will be great.

> For the `is_revised` flag on bars, how do you know the revision is for the bar at index 0 which is getting popped?

Revision will be only for the last Bar of bar_type. As in Cache last Bar is added at index 0, so whenever we receive revision we have to replace at index 0.

As you see DataEngine with condition `if bar.ts_event < self._topic_last_ts[topic][0] or bar.ts_init <= self._topic_last_ts[topic][1]:`, so only the Bar having same or higher ts_event and have higher ts_init will be processed. 

1. `ts_event` being unique id, and timeseries specially for Bars, I believe we need to make sure they are passed on to sequence because things onward are stored in list fashion, if mixed will lead to unexpected results in indicators and strategy where values depend on previous bar.
2. `ts_event` with last ts_event is allowed to allow revisions.
3. `ts_init` value is when the object is created/initialized, we don't need to process same object if duplicate or with older ts received for some reason, so only new is allowed

Just to give you more idea, currently InteractiveBrokersDataClient._on_historical_bar_update will simply return if when has_new_bar is False because that means the trigger was update of the last bar in list (always last in list). As soon new Bar starts, previous becomes final Bar. So this is the place where user's need wheather to receive the partial updates or not will apply.
```
    def _on_historical_bar_update(
        self,
        bars: list[BarDataList],
        has_new_bar: bool,
        bar_type: BarType,
    ):
        if not has_new_bar:
            instrument = self.instrument_provider.find(bar_type.instrument_id)
            bar = bars[-1]
            ts_event = dt_to_unix_nanos(bar.date)
            ts_init = self._clock.timestamp_ns()
            data = Bar(
                bar_type=bar_type,
                open=instrument.make_price(bar.open),
                high=instrument.make_price(bar.high),
                low=instrument.make_price(bar.low),
                close=instrument.make_price(bar.close),
                volume=instrument.make_qty(max(bar.volume, 0)),
                ts_event=ts_event,
                ts_init=ts_init,
                is_revised=True,
            )
            self._handle_data(data)
            return
```

cjdsellers:
Going to close this as we have most of the functionality implemented now.

Out of interest, why don't we just hold the partially built bar back until its closed - then send to the `DataEngine`?

rsmb7z:
> Out of interest, why don't we just hold the partially built bar back until its closed - then send to the `DataEngine`?

Responded in #897 as related to that as well.

PR #958, opened 2023-01-14
Implement catalog rust data read and write

ghill2:
# Pull Request

Load rust tick data within a start and end range as nautilus objects or dataframe using the ParquetDataCatalog. When passing multiple instrument_ids, objects are merged. 

Write rust tick data using the process_files method

## Type of change


nautilus_trader/model/data/tick.pyx
```diff
@@ -300,10 +301,26 @@ cdef class QuoteTick(Data):
...
+    cdef inline quote_tick_list_to_capsule(list items):
+        cdef QuoteTick_t * data = <QuoteTick_t *> PyMem_Malloc(len(items) * sizeof(QuoteTick_t))
+        if not data:
+            raise MemoryError()
+
+        cdef int i
+        for i in range(len(items)):
+            data[i] = (<QuoteTick> items[i])._mem
+
+        return PyCapsule_New(data, NULL, NULL)
```
twitu, line 314: The updated implementation for the writer takes a `CVec` instead of a just a pointer to a buffer of structs. We made this change to keep it consistent with parquet reader which returns a CVec, as you can see `capsule_to_quote_tick_list` method. So here you can put the pointer in a CVec, the appropriate len, and ignore `cap` field because it's not relevant here.

Also this'll currently leak memory because the buffer isn't deallocated after being written by the writer.

twitu (changes requested):

nautilus_trader/model/data/tick.pyx
ghill2: Thanks forthe help on this. I modified the method to store the buffer of structs inside of the CVec (and also set the len). Writing using the PyCapsule still is causing a segmentation fault.

PR #984, opened 2023-01-29
Add pre-commit hooks for Rust

twitu:
# Pull Request

Add pre-commit checks for Rust and fix failing pre-commit lints

* remove explicit free function for structs that don't have any heap allocate fields
* potentially memory leaking implementation in clocks, skipped test for now as it might need larger refactor

## Type of change

- [x] - Tooling improvement

pre-commit checks are passing. Rust tests are passing.


cjdsellers:
Many thanks!

Issue #970, opened 2023-01-17
Bar with OHLC having Decimal


rsmb7z:
# Feature Request

Hi @cjdsellers 
There is a scenario for BarType where all Bar charactristics are same except:

- Instead of Price_Type, its Data_Type such as `HISTORICAL_VOLATILITY`, or `OPTION_IMPLIED_VOLATILITY`
- OHLC hold Decimal instead of Price

For example `MSFT.NASDAQ-1-HOUR-HISTORICAL_VOLATILITY-EXTERNAL` and `MSFT.NASDAQ-1-HOUR-OPTION_IMPLIED_VOLATILITY-EXTERNAL`.

Do you think this is something can be integrated in existing Bar, or will make it complicated and seperate/clone object should be used?

cjdsellers:
So there's two parts to this:
- 1: How do we represent these aggregations as a `BarSpecification`?
- 2: What should the type of `Bar` be? what should the field types be?

For 1, we need to keep the parsed bar specification to three hyphen separated tokens:
```
{step}-{aggregation}-{price_type}
```

Unless we change that, I don't think a volatility really qualified as a price type. Because of that, they probably need to look more like this `MSFT.NASDAQ-1-HOUR_HISTORICAL_VOLATILITY-EXTERNAL`. Which is more or less what you have above, just one additional underscore.

For 2, we have three options:
- Just use the existing `Bar` and make the OHLC of type `Price`
- Create a new data type like `VolatilityBar`, which probably can't subclass `Bar` because its using the `Price` type
- Generalize `Bar` so that the `open`, `high`, `low`, `close` fields are any type of Python object, and/or create `PriceBar`

I think right now options 1 and 2 are the least invasive and don't involve breaking changes to the API. So we could experiment with those. I'm open to reviewing any PR along those lines.

rsmb7z:
> For 1, we need to keep the parsed bar specification to three hyphen separated tokens:

For this I have just added the `HistoricalVolatility` and `ImpliedVolatility` in the rust enum. Surprisingly it took care of underscore as in python and shows like `HISTORICAL_VOLATILITY` and `IMPLIED_VOLATILITY`. Although its weird to be in PriceType but this will keep the usage simpler and without created another BarSpec and BarType classes.

> For 2, we have three options:

[...]

PR #988, opened 2023-02-01
Added GenericBar and options for HV/IV

rsmb7z:
Fulfills requirement for #970.

### DIFF
--- a/nautilus_core/model/src/enums.rs
+++ b/nautilus_core/model/src/enums.rs
@@ -275,6 +275,8 @@ pub enum PriceType {
     Ask = 2,
     Mid = 3,
     Last = 4,
+    HistoricalVolatility = 11,
+    ImpliedVolatility = 12,
 }
 
 #[repr(C)]
--- a/nautilus_trader/core/includes/model.h
+++ b/nautilus_trader/core/includes/model.h
@@ -174,6 +174,8 @@ typedef enum PriceType {
     ASK = 2,
     MID = 3,
     LAST = 4,
+    HISTORICAL_VOLATILITY = 11,
+    IMPLIED_VOLATILITY = 12,
 } PriceType;
 
 typedef enum TimeInForce {
--- a/nautilus_trader/core/rust/model.pxd
+++ b/nautilus_trader/core/rust/model.pxd
@@ -148,6 +148,8 @@ cdef extern from "../includes/model.h":
         ASK # = 2,
         MID # = 3,
         LAST # = 4,
+        HISTORICAL_VOLATILITY # = 11,
+        IMPLIED_VOLATILITY # = 12,
 
     cpdef enum TimeInForce:
         GTC # = 1,
--- a/nautilus_trader/model/data/bar.pyx
+++ b/nautilus_trader/model/data/bar.pyx
@@ -14,7 +14,9 @@
 # -------------------------------------------------------------------------------------------------
 
 from cpython.datetime cimport timedelta
+from libc.stdint cimport uint8_t
 from libc.stdint cimport uint64_t
+from decimal import Decimal
 
 from nautilus_trader.core.correctness cimport Condition
 from nautilus_trader.core.data cimport Data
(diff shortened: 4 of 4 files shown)


cjdsellers:
Hi @rsmb7z 

For this one, I'm feeling like `HistoricalVolatility` and `ImpliedVolatility` are more like `BarAggregation` types than price types?

I think we can consider the current `Bar` object as a "price bar", as the OHLC are of type `Price`. We could introduce this `GenericBar` and for the moment keep the OHLC as `Decimal` as you've done.

I don't think its going to work declaring those types as a C `double` though, I would just keep this in Python similar to the `BinanceBar`, however inherit from `Data`:
```python
GenericBar(Data):
    ...
```

Which avoids us having to define a new type of bar every time we want to represent some OHLCV bar aggregation with decimal as the OHLC types.

Then, it depends where we intend on forming these bars as to whether we want to append these members to `BarAggregation` or potentially create a `GenericBarAggregation` to separate the price based bars, from generic bars?

If we don't intend on sending volatility updates as a separate data type into the system and having the `DataEngine` build these bars, then it may be best just defining a `GenericBarAggregation` enum? although having said that, things will only work with the current `BarType` if we add these to `BarAggregation`.

My guess is you intend to write some custom code to form these bars, and _then_ send those back to the `DataEngine` for publishing on the message bus?

rsmb7z:
Hi @cjdsellers 

> For this one, I'm feeling like `HistoricalVolatility` and `ImpliedVolatility` are more like `BarAggregation` types than price types?

These two are indeed replacement of `PriceType` in the `BarSpecification` whereas `BarAggregation` remains as is, which means these two types can have OHLC bar of any timeframe. In terms of OHLC values only difference is the Price and Percentage. That's why I have included these in the `PriceType` enum.

Following is the example snap:

image

*One thing important though, we should keep single `BarType` to keep things simpler and we can easily identify which class i.e Bar, GenericBar etc to use based on type and vice versa. However creating different `BarType` to accomodate the `BarSpecification` (`HistoricalVolatility` and `ImpliedVolatility`) may overly complicate the usage. In worst case there is no good alternative to accomodate for the `HistoricalVolatility` and `ImpliedVolatility` enum, we can keep in the `PriceType` as-is.*

> I don't think its going to work declaring those types as a C `double` though, I would just keep this in Python similar to the `BinanceBar`, however inherit from `Data`:
> Which avoids us having to define a new type of bar every time we want to represent some OHLCV bar aggregation with decimal as the OHLC types.

I tried using `Decimal` originally but that didn't worked because not supported in cython, yes we can use Python instead. You are right having `GenericBar(Data)` with Decimal will provide this to be used directly as well base class for any other sort of OHLC.

> If we don't intend on sending volatility updates as a separate data type into the system and having the `DataEngine` build these bars, then it may be best just defining a `GenericBarAggregation` enum? although having said that, things will only work with the current `BarType` if we add these to `BarAggregation`.

I think this is related to first question. `BarAggregation` of this bar and the original `Bar` have no difference. If we add these to `BarAggregation` enum we will loose the aggregation. Please have a look at following links:
https://interactivebrokers.github.io/tws-api/historical_bars.html
https://interactivebrokers.github.io/tws-api/tick_types.html

> My guess is you intend to write some custom code to form these bars, and _then_ send those back to the `DataEngine` for publishing on the message bus?

Yes correct. Will just flow like standard `Bar` where it will be consumed by Indicators and Strategy for decision making, and actions. Either loaded to the `DataEngine` for Backtesting or subscribed from the Live Broker in case of Live trading.

cjdsellers:
Thanks for providing the examples and references from the IB docs.

I can see that with the way you're using `PriceType`, it does fit quite conveniently into the `BarType`.

Is it logical that we might be aggregating the volatility of the `BID`, `ASK,`, `MID`, `LAST` prices though? in which case the logical bar type becomes something like.

`ABC-IMPLIED-VOLATILITY.NYSE-1-MINUTE-LAST-EXTERNAL` (assuming these are actually being aggregated externally by IB)?

If we extended the `PriceType` then it opens up many code paths to error with the way its used throughout the core of the codebase. Or at least every 'switch' statement would have to be checked that its handling the invalid type cases (which would nearly always include these volatility types as being invalid).

rsmb7z:
> Is it logical that we might be aggregating the volatility of the `BID`, `ASK,`, `MID`, `LAST` prices though? in which case the logical bar type becomes something like.

The volatility values are of the instrument so are complete in itself `HistoricalVolatility` and `ImpliedVolatility` whereas within BarType `BID`, `ASK,`, `MID`, `LAST` would always mean that it is `Price`. The scenario you are referring will be for the case of `Yield` which applies for Bonds. In that case enums for that will be added seperately such as `YieldBid`, `YieldAsk,`, `YieldLast`.


> `ABC-IMPLIED-VOLATILITY.NYSE-1-MINUTE-LAST-EXTERNAL` (assuming these are actually being aggregated externally by IB)?

This will be considered malformed. Now assuming these both are part of `PriceType` enums, it will be always follow the existing `BarType` string convension i.e `ABC.NYSE-1-MINUTE-HISTORICAL_VOLATILITY-EXTERNAL` or `ABC.NYSE-1-MINUTE-IMPLIED_VOLATILITY-EXTERNAL` - in case of external aggregation. _They can have `INTERNAL` aggregation when provided Historical Volatility Tick and Implied Volatility Tick data (as in the 2nd link Tick Id 23 and 24 relates to this). Just for information, that's not part of this PR._

> If we extended the `PriceType` then it opens up many code paths to error with the way its used throughout the core of the codebase. Or at least every 'switch' statement would have to be checked that its handling the invalid type cases (which would nearly always include these volatility types as being invalid).

True, it will be almost everywhere. It looks bit odd because of Enum's name `PriceType`, apart from this everything is perfect. Maybe this can have more suitable name to suit.

cjdsellers:
Hi @rsmb7z 

To move forward on this, all things considered I don't think a volatility is a price type, but can be based on a price type (so should be separate).

How many different time frames for volatility would be appropriate? we could just make them different bar aggregations, e.g.
- `MINUTE_HISTORICAL_VOLATILITY`
- `HOUR_IMPLIED_VOLATILITY`

Having said that, the combination of the step and the `BarAggregation` becomes the _rule_ for bar formation. So in the case of the current time based bars, the bar is closed at the end of the time interval and represents a data event at `ts_event`. 

In this way, aren't implied volatility bars also an aggregation sampled across time? This leads me back to a previous suggestion`ABC-IMPLIED-VOLATILITY.NYSE-1-MINUTE-LAST-EXTERNAL`

So a bar type is `{INPUT}-{STEP}-{AGGREGATION/RULE}-{EXTERNAL/INTERNAL}`. The input in the above cases is the bid, ask, mid, last implied volatility across `step * time` intervals?

cjdsellers:
Continuing discussion in https://github.com/nautechsystems/nautilus_trader/issues/970.

cjdsellers:
Closing this for now was we decided in #970 that it's currently only applicable to IB.

PR #989, opened 2023-02-05
Separate struct for log send and rcv

twitu:
# Pull Request

Separate logger sender and receiver structs across channel. This is so that each log event does not create it's own buffer writer.

## Type of change
 
- [x] Refactor/Redesign

By running backtests.


twitu:
The actual logging action is happening in the spawned thread, so the change does not show any substantial improvement in performance on the main thread. The decoupled design can be implemented later when needed.

PR #992, opened 2023-02-08
Add support for Binance aggregate trades

### LINKED ISSUE #918: Add Support for Aggregate Trades
# Feature Request

Binance provides a market data endpoint ([`GET /api/v3/aggTrades`](https://binance-docs.github.io/apidocs/spot/en/#compressed-aggregate-trades-list)) and a websocket stream ([`<symbol>@aggTrade`](https://binance-docs.github.io/apidocs/spot/en/#aggregate-trade-streams)) for trades in compressed/aggregate form, for both spot and futures.

The data endpoint can return historical trades between a `startTime` and `endTime` (limit 1000). Repeated requests could be used to build historical trade data, which cannot be done with only the currently supported/exposed [`GET /api/v3/trades`](https://binance-docs.github.io/apidocs/spot/en/#recent-trades-list) endpoint via `request_trade_ticks`. Hence the current limitation preventing historical time ranges in [`adapters/binance/spot/data.py#L429`](https://github.com/nautechsystems/nautilus_trader/blob/master/nautilus_trader/adapters/binance/spot/data.py#L429)

At the moment the extent of aggregate trades support are only on the HTTP & websocket layers within the Binance adapter. I propose these aggregate trade data sources be implemented completely and exposed to strategies.

To support aggregate trades, either a new data type could be created (`AggTradeTick`), or the existing `TradeTick` might be extended, replacing the `trade_id` field with `first_trade_id` and `last_trade_id`.

*I would be happy to work on this feature, I need it for my application and would prefer to contribute than make a workaround.*

poshcoe:
# Pull Request

Closes #918

Implements Binance aggregated trade ticks and allows users to access these endpoints instead of pure trade ticks through a new configuration variable (`use_agg_trade_ticks`)

Also includes further fixes to the `_subscribe_order` method logic in `data.py`

## Type of change

Subscription and HTTP requests for aggregated trade ticks have been tested on both Spot and Futures testnets.

## Required:

- [x] Binance Websocket aggTrade subscriptions & unsubscriptions
- [x] Binance HTTP aggTrade requests & handlers
- [x] Test coverage and pass
- [x] Documentation pass


poshcoe:
Sneaked another fix in here ([a25e170](https://github.com/nautechsystems/nautilus_trader/pull/992/commits/a25e1702ec86f0d2e16f3e086483abd800fff9fe)), where `BinanceCommonExecutionClient` was not passing the correct `OmsType` or `AccountType` down to the `LiveExecutionClient` for spot or margin accounts.

cjdsellers:
Excellent work here, a very useful feature!

OnlyC:
Trying this with 7 days history, got IP banned because of Too many requests. 
Looks like we doesn't have rate limiter, got the same error if I set ```instrument_provider=InstrumentProviderConfig(load_all=True)```
(This will run query_futures_commission_rate() for every instrument, one by one)

graceyangfan:
@poshcoe I use the agg-ticks in binance future live trading，but it seems that the history did not send to on_history_data，have you test this function？

poshcoe:
@OnlyC Until we have rate limiting I would recommend using the endpoint only for historical data that is not yet published on https://data.binance.vision/. Looking ahead, I expect even with rate limiting the endpoint will eventually slow to a crawl for very large time ranges...

@graceyangfan Interesting, I have tested this function and it works on my end. Is `on_historical_data` being called at all? Are you using a large time range? It might take a while to build the list of ticks in that case. The method only returns data once all ticks have been retrieved.

Note that the `on_historical_data` method is called for each `TradeTick` in the list returned by the endpoint. If you want to access the entire returned list directly, consider overriding the [`handle_trade_ticks`](https://github.com/nautechsystems/nautilus_trader/blob/master/nautilus_trader/common/actor.pyx#L2011) method in your strategy.

PR #1029, opened 2023-02-28
Add initial pass for IB tick scheme

limx0:
# Pull Request

First pass adding tick schemes to IB

## Type of change


nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
```diff
@@ -87,9 +105,10 @@ def parse_equity_contract(details: ContractDetails) -> Equity:
         price_increment=Price(details.minTick, price_precision),
         multiplier=Quantity.from_int(
             int(details.contract.multiplier or details.mdSizeMultiplier),
-        ),  # is this right?
+        ),  # TODO - is this right?
```
rsmb7z, line 108: Applies only to Futures and Options. We can either set to `Quantity.from_int(1)` or leave None.

nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
```diff
@@ -59,6 +60,22 @@ def sec_type_to_asset_class(sec_type: str):
     return asset_class_from_str(mapping.get(sec_type, sec_type))
 
 
+def get_contract_details_tick_scheme_name(details: ContractDetails) -> str:
+    exchange_to_market_rules = dict(
+        zip(details.validExchanges.split(","), details.marketRuleIds.split(",")),
+    )
+    unique_market_rules = sorted(set(exchange_to_market_rules.values()))
```
rsmb7z, line 67: If `SMART` present it will be at index so we don't need to sort it. In either case we can assign Index 0 as default to the Instrument.
I think what we can do is have an optional parameters (exchange: str = None). If that's none it returns default (Index 0). If the value is present for `exchange` it will return the schema for that exchange or raise error if not found. So then the same function `get_contract_details_tick_scheme_name` can be used within Strategy as well in case placing order to any other exchange.

rsmb7z:
Hi @limx0 
I came across different proposal than the inital one to name the schema like IB etc. I think the better approach with less maintenance would be:

1. Register TickScheme at glocal level instead of adapter, using Nautilus' own naming convension. As these `price_precision, min_tick, max_tick, increment` fields would make it unique so only one registeration would be required irrespective of which adapter need to use it.
2. Adapter will have a mapping dict for the Venue tick requirement to Nautilus TickScheme name.

limx0:
@rsmb7z - have you done something similar in #1036 and should we close this? 

I think either way best to create a new PR rather than try and rebase from here, so closing.

rsmb7z:
> @rsmb7z - have you done something similar in #1036 and should we close this?

No this is not included and I have not found this situation for other than Forex, will revisit the topic and raise new ticket if need be.

PR #1050, opened 2023-03-18
Halt Engine if Exception raised for generate_mass_status

rsmb7z:
# Pull Request

When account reconciliation is enabled and TradingNode position doesn't match the actual positions with the Broker, TradingNode would halt and doesn't continue. This is the correct and expecte behavior.
However, if for some reason there is exception raised within any of the three reports, which is quit possible, TradingNode simply raises the exception but continues on treating `reconciliation=True`.

## Type of change


cjdsellers:
This logic is correct, thanks for uncovering it!

PR #1062, opened 2023-03-23
Minor fix in Clock

rsmb7z:
# Pull Request

Minor fix in clock timer.

## Type of change


cjdsellers:
This is probably why the loop timer tests started failing.

PR #1064, opened 2023-03-24
Optional Venue Order Id on generate_order_submitted

rsmb7z:
# Pull Request

Allows optional Venue Order Id on `generate_order_submitted` which is useful where `VenueOrderId` is available before at the time of Sumit event i.e IB.

## Type of change

### DIFF
--- a/nautilus_trader/execution/client.pxd
+++ b/nautilus_trader/execution/client.pxd
@@ -94,6 +94,7 @@ cdef class ExecutionClient(Component):
         InstrumentId instrument_id,
         ClientOrderId client_order_id,
         uint64_t ts_event,
+        VenueOrderId venue_order_id=*,
     )
     cpdef void generate_order_rejected(
         self,
--- a/nautilus_trader/execution/client.pyx
+++ b/nautilus_trader/execution/client.pyx
@@ -310,6 +310,7 @@ cdef class ExecutionClient(Component):
         InstrumentId instrument_id,
         ClientOrderId client_order_id,
         uint64_t ts_event,
+        VenueOrderId venue_order_id = None,
     ):
         """
         Generate an `OrderSubmitted` event and send it to the `ExecutionEngine`.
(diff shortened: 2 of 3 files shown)


cjdsellers:
So we're not so sure on this one, what is the use case this will help with?

The concern is that the `VenueOrderId` is the identifier assigned by the venue, which you _normally_ don't know ahead of time (even if IB is the exception here).

There's a risk loosening this and allowing a user assigned `VenueOrderId` at submit might open up the potential for mismatched IDs and other issues for those implementing adapters.

Thoughts?

rsmb7z:
> The concern is that the `VenueOrderId` is the identifier assigned by the venue, which you _normally_ don't know ahead of time (even if IB is the exception here).

As for the IB, we first get the [NextValidID](https://interactivebrokers.github.io/tws-api/interfaceIBApi_1_1EWrapper.html#a09c07727efd297e438690ab42838d332) and assign to Order object when placing order which is then used by IB for any updates back. This ID wouldn't change during entire life of Order even if the session is re-connected. There is `PermId` (not to be confused from the v1) which is also assigned by IB however that is different unique order identity which doesn't help manage the Order through API.

I am assigning the Nautilus_Client_Id as reference which is available to us for updates once the submission is received. However during the time between `OrderSubmitted` and `OrderAccepted` if we have to cancel we don't have the `VenueOrderId` needed to cancel. Currently I am using hash but I thought this would be better approach.

> There's a risk loosening this and allowing a user assigned `VenueOrderId` at submit might open up the potential for mismatched IDs and other issues for those implementing adapters.

I believe `OrderSubmitted` will be by adapters only which wouldn't be exposed to end-user like Actor or Strategy where user could make such errors. Adapter side is more controlled and can have tests. In case this `OrderSubmitted` (could have been used by Actor/Strategy - I couldn't think of scenario) then, we may have some other solution.

cjdsellers:
So most venues allow you to send some kind of `ClientOrderId` which allows you to identify the order within your own system. This is so you can achieve exactly as you say, refer to the order and potentially cancel/modify prior to receiving an acknowledgement of order receipt/acceptance.

I think IB is the odd one out here, and in these cases its better for the adapter to handle this with something like a hashmap approach as you currently are, rather than loosening the domain model to accommodate it.

rsmb7z:
> I think IB is the odd one out here, and in these cases its better for the adapter to handle this with something like a hashmap approach as you currently are, rather than loosening the domain model to accommodate it.

In that case I am on track then. I will close this.

PR #1073, opened 2023-04-10
fix windows datafusion build

ghill2:
Disables the default feature causing the build issue.
https://github.com/BLAKE3-team/BLAKE3/issues/298


twitu:
Awesome workaround 🙌

Issue #705, opened 2022-07-21
Porting persistence layer to Rust


twitu:
This issue is to track the discussion and PRs for this large-ish port of the persistence layer into Rust.

## Parquet layer

In the parquet layer, the key component here is the parquet reader/writer that reads/writes to arrow format which gets converted to/from chunks of a Rust structs. The current working example in #699 introduces the following traits and structs. They are parameterized over a lifetime (explained below) and the struct which they are decoding encoding.

```rust
/// write vectors of struct A to a parquet file
pub struct ParquetWriter<A> {
    pub writer: FileWriter<File>,
    pub encodings: Vec<Vec<Encoding>>,
    pub options: WriteOptions,
    pub writer_type: PhantomData<*const A>,
}

/// read vectors of struct A from a parquet file
struct ParquetReader<'a, A> {
    file_reader: FileReader<&'a File>,
    reader_type: PhantomData<*const A>,
}

/// Any struct that wants to be decoded from a parquet file
/// must implement this trait.
trait DecodeFromChunk
where
    Self: Sized,
{
    fn decode(schema: &Schema, cols: Chunk<Arc<dyn Array>>) -> Vec<Self>;
}

/// Any struct that wants to be encoded and written to a parquet file
/// must implement this struct.
trait EncodeToChunk
where
    Self: Sized,
{
    fn encode_schema() -> Schema;
    fn encode(data: Vec<Self>) -> Chunk<Box<dyn Array>>;
}

/// parquet reader implements an iterator i.e. calling next on it reads
/// the next batch of chunks from disk converts them into vector of struct A
/// and returns the vector
impl<'a, A> Iterator for ParquetReader<'a, A>
where
    A: DecodeFromChunk,
{...}
```

## Interface layer

We need another layer on top of this to interface with Python/Cython bindings. Current persistence layer allows passing queries that project certain columns and filter certain rows. Since the parquet file reader is now provided by the rust arrow2 library, the python project filter expressions will have to be converted in this interfacing layer so that parquet layer (above) can understand it.

```rust
    /// Creates a new [`FileReader`] by reading the metadata from `reader` and constructing
    /// Arrow's schema from it.
    ///
    /// # Error
    /// This function errors iff:
    /// * reading the metadata from the reader fails
    /// * it is not possible to derive an arrow schema from the parquet file
    /// * the projection contains columns that do not exist
    pub fn try_new(
        mut reader: R,
        projection: Option<&[usize]>,
[...]

PR #1085, opened 2023-04-25
Benchmark catalog v2 for single and multi stream

twitu:
# Pull Request

Adds benchmarks for single and multi stream. Benchmarks measure how long it takes for catalog to load 10 M ticks from a single file and 72 M ticks from 61 files. This is a good stress test for the stream merging datafusion based catalog.

The python tests are expected to take more time because of the overhead of initializing python objects and acquiring GIL.

The benchmarks were run on an AMD 5600U machine and gave the following.

| Test | Rust | Python |
| -- | -- | -- |
| single stream (10M) | 2.4s | 3.2s |
| multi stream (72M, 61 files) | 23s | NA |

The multi-stream python bench requires 8+ gb of ram and could not be run.

An appropriate measure for the catalog is it's throughput which can be measured in terms of ticks per second. Here we can compare with previous iterations described in #705 

1. Rust catalog v2 - datafusion and kmerge ~ 3M/s
2. Rust catalog v1 - arrow2 and Python merge logic ~ (TODO)
3. Catalog - Pure Python - 100K/s

## Type of change

- [x] Performance tests

The benchmarks were run and passed.

### DIFF
--- a/nautilus_core/Cargo.lock
+++ b/nautilus_core/Cargo.lock
@@ -1703,6 +1703,7 @@ version = "0.4.0"
 dependencies = [
  "binary-heap-plus",
  "compare",
+ "criterion",
  "datafusion",
  "futures",
  "nautilus_core",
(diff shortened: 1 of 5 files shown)


twitu:
Closes #705

PR #1116, opened 2023-05-28
Improve Binance request_bars to get more bars

graceyangfan:
# Pull Request

Improve Binance request_bars to get more bars #736 

## Type of change

Test with binance live client request bars 1-min for 10 days


cjdsellers:
Thanks for the PR! Looks like it's mypy complaining:

```
nautilus_trader/adapters/binance/http/market.py:875: error: Incompatible types in assignment (expression has type "int", variable has type "Optional[str]")  [assignment]
```

nautilus_trader/adapters/binance/http/market.py
```diff
@@ -853,15 +853,28 @@ async def request_binance_bars(
...
+                limit=limit,
+                start_time=start_time,
+                end_time=end_time,
+            )
+            bars: list[BinanceBar] = [
+                kline.parse_to_binance_bar(bar_type, ts_init) for kline in klines
+            ]
+            all_bars.extend(bars)
+
+            if len(klines) < limit:
```
cjdsellers, line 870: We want to keep track of how many bars we have received, and break out of the loop when either we reach the `limit` (if any), or there are no more klines to parse into bars.

cjdsellers (commented):
Thanks for the PR! :pray:  @graceyangfan 

Just one small comment to double check on (see review).

cjdsellers:
So the way I'd expect the `limit` to work would be something like "get all bars in time range `start` to `end` up to a maximum of `limit`, and begin the request from the latest data and back towards the start". This would ensure the latest data is always included.

Does this sound like the behavior for this change?

I'm just finalizing quite a large diff which will integrate the core Rust order book into the platform, I'll all be able to merge once that in (the next day or so).

Many thanks again!

PR #1136, opened 2023-06-13
Support Pandas v2

rterbush:
# Pull Request
In Pandas v2, timezone aware datetimeindexes return 'datetime.timezone.utc' for tzinfo. This fails comparison when checking for timezone aware indexes. Explicitly setting these datetimeindexes to pytz.utc allows support for both Pandas 1.5.x and Pandas 2.x

## Type of change

Non-breaking change that allows support for Pandas ^1.5.3 as set in pyproject.toml dependencies.
This appears it will also apply cleanly to master
Slightly unsure about the change to cimport in `nautilus_trader/core/datetime.pyx`

- [ x] Bug fix (non-breaking change which fixes an issue)

unit_tests and acceptance_tests have passed

### DIFF
--- a/nautilus_trader/common/clock.pyx
+++ b/nautilus_trader/common/clock.pyx
@@ -14,12 +14,12 @@
 # -------------------------------------------------------------------------------------------------
 
 import asyncio
-import pytz
 from typing import Callable, Optional
 
 import cython
 import numpy as np
 import pandas as pd
+import pytz
 
 from cpython.datetime cimport datetime
 from cpython.datetime cimport timedelta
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @rterbush 

Great work on this and I'm happy to get this in and support pandas 2+. I could handle the actual dependency upgrade once this is merged.

Looks like the CI fail is just for the `black` formatting pre-commit step. 
If you could please run the following it should fix it up:

```
pre-commit install
pre-commit run --all-files
```

Once the git hook is installed with that first command (then you can never forget as it will automatically run on commit), another easier shortcut to run ad-hoc is the Make target from the top level:
```
make pre-commit
```

Cheers! :pray:

rterbush:
pre-commit has been applied @cjdsellers 
Thought I would like you be the one to make the commitment on dependencies. :-)

PR #1146, opened 2023-06-20
Use v2 design for raw sockets

twitu:
# Pull Request

Previous design was starving the writer when the reader was acquiring the lock but not receiving any messages. This uses the design from websockets and splits the connection into reader and writer ends that can be used indepdendently.

## Type of change

The rust test passes but it passed before as well. Perhaps a Python test is needed.


cjdsellers:
Awesome work. Agreed we need more tests on the Python side too, will do this once we clear out the old Cython and reassess the remaining tests.

PR #1153, opened 2023-06-26
Quick start updates

artnaz:
Fixed a few things while trying to go through `quick_start.ipynb` and stumbled upon some by me unresolved issues.

- Added some info to install Rust prior to developing
- added `jupyterlab` and `ipykernel` in development group of pyproject.toml to be able to launch the notebooks
- Cleaned up the notebook by fixing a hyperlink, added some additional descriptions and removed unnecessary imports
- **Important**: I copied the `./catalog` data (same as in the Docker image) to be able to load the data. Using the incumbent `ParquetDataCatalog.from_env()` gives environment error and I think it's for the user not straightforward how to load the data differently. Probably a different solution than mine would be better.
- **Important**: `node.run()` gives "RuntimeError: asyncio.run() cannot be called from a running event loop" (the commited file shows this as well). I found that this does not occur when running the same steps in a regular `.py` file. Not sure how to solve.


cjdsellers:
Hi @artnaz 

Thanks so much for providing this!

I agree on adding these dependencies to the development group. 

Do you think we could reduce these 4 MB+ parquet files down, or consider not including them? it was probably a mistake ever adding data files to the main repo, so we've been avoiding adding any more. 

We actually have a separate `nautilus_data` repo which contains some stub data https://github.com/nautechsystems/nautilus_data

artnaz:
Hi @cjdsellers, good point about jupyterlab, I agree that it's probably an unnecessary convenience with too much dependencies.

Using that `nautilus_data` repo would be much cleaner indeed. I'll see if I can pull that in within the notebook.

However, regarding the last point (`node.run()`), can somebody help with that? I noticed in the Discord chat that there were more people with the same problem. I'm not sure how to solve this as it seems to be a notebook-only problem.

benjaminsingleton:
I was able to solve `RuntimeError: asyncio.run() cannot be called from a running event loop` in jupyterlab by installing `nest-asyncio` (https://github.com/erdewit/nest_asyncio). In my first notebook cell, I execute the below code and everything seems to then work as expected.

```
import nest_asyncio
nest_asyncio.apply()
```

cjdsellers:
> I was able to solve `RuntimeError: asyncio.run() cannot be called from a running event loop` in jupyterlab by installing `nest-asyncio` (https://github.com/erdewit/nest_asyncio). In my first notebook cell, I execute the below code and everything seems to then work as expected.
> 
> ```
> import nest_asyncio
> nest_asyncio.apply()
> ```

I improved the situation here by refactoring away the need to call a coroutine from a synchronous context in the `BacktestEngine`.
https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/backtest/engine.pyx#L1000, so there should not an issue with nested event loops anymore.

cjdsellers:
@artnaz I think the actionables here to merge the diff are:

- Remove the new parquet data
- Remove the need for `nest_asyncio`

From that point it'll be great to incorporate your other tidy ups to the notebook, alternatively I can handle this on my end and close the PR.

Please let me know your preference when you get the change :pray:

artnaz:
Thanks @cjdsellers, I'll try to do this around this weekend.

cjdsellers:
Hey @artnaz 

Just a heads up I applied a couple of your docs additions here f0174b6d0c48a893918b1b41e027b3635a98a0d8.
Will close this one for now, thanks again!

PR #1177, opened 2023-07-17
Experiment ustr

### LINKED ISSUE #1166: String interning to reduce memory allocations/usage and simplify data structures
The engine uses string identifiers in many places to identify and check that two pieces of data are the same or to retrieve and print them for logging purposes. These strings are introduced by the loaded data (network, disk) or by user defined strategy.

The strings are stored in structures that cross ffi boundaries, are often cloned and can move between threads. They are wrapped in Box Arc to make the above safe and cheap.

```rust
#[repr(C)]
#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[pyclass]
pub struct Venue {
    pub value: Box<Arc<String>>,
}
```

However the strings are never really sliced, or deleted and are very similar across many objects. We can exploit this and significantly reduce memory allocations and make the structures lighter by using interned strings that are identified by `Copy`able keys. With such an implementation an identifier would look something like this.

```rust
#[repr(C)]
#[derive(Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[pyclass]
pub struct Venue {
    pub value: InternedStringKey,
}
```

This value can easily cross ffi boundaries without extra indirection and can be copied without needing a deep clone. More importantly only String value be allocated for all uses of it throughout the engine. All uses of it will refer to the lightweight InternedStringKey.

We can expect significant improvement in memory usage by comparing with baselines tests created in #1099 

---

There are many potential libraries for interning strings. These are the requirements -
* Thread safe keys - the engine uses multiple threads and async code extensively
* `Copy`able keys - So that they can be copied without a deep clone
* FFI support - It can be passed across ffi without extra indirection like `Box`
* Interned key - get back string using key, use key for equality comparison
* Intern store is append only - There is no mechanism to remove interned strings. It is expected that the intern store will be discarded after a full run.
* Intern store supports high throughput with concurrent execution - Ideally it will use a HashStore with key level read-write locking, since we'll only need to acquire the write lock once when writing the string. All other string access will be read access that don't need exclusive access.

Strings are accessed and created all across the codebase and it will be very cumbersome to pass the intern store everywhere. It is very likely that the best solution is to make intern store global and mutable using something like once_cell/lazy_static (which are now in std).

Which global static cell primitive to use is still unclear. The most general is OnceLock, however we might be able to do better if we can identify that mutable access is needed only in certain sections of the code (disk, network, etc).

twitu:
# Pull Request

Use ustr library to intern strings. Use interned strings for identifiers.

Closes #1166 

## Type of change

- [x] Huge Refactor

Experimental


twitu:
# Bug Report

Pickling and unpickling QuoteTick, TradeTick, SyntheticInstrument does not recreate the same InstrumentId, even though the textual representation is the same. This is after switching to Ustr style implementation in #1177 

### Expected Behavior

assert tick == unpickled should pass

### Actual Behavior

```
tests/unit_tests/model/test_tick.py 0x558e4a503a18, 0x558e4a4739b8
AUD/USD, SIM
0x558e4c80c3f8, 0x558e4ccec738
AUD/USD, SIM
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `u!("AUD/USD")`,
 right: `u!("AUD/USD")`', model/src/data/quote_api.rs:63:5
 ```

Equality fails because the InstrumentId's component Symbol and Venue store different pointers even though they point to the same textual string. This shouldn't happen because the strings are being interned.

1. Checkout commit - https://github.com/nautechsystems/nautilus_trader/pull/1177/commits/c1ed1bff6705d444bb400a20b3d2a1cf0f0e7c7c
2. Build and run `pytest tests -sk "test_pickling_round_trip_results_in_expected_tick"`

### Specifications

- OS platform: Linux
- Python version: 3.10.6
- `nautilus_trader` version: https://github.com/nautechsystems/nautilus_trader/pull/1177/commits/c1ed1bff6705d444bb400a20b3d2a1cf0f0e7c7c

twitu:
As expected this change shows significant improvement in bench-marking results.

Before ustr | Rust | Python
-- | -- | --
single stream (10M) | 2.4s | 3.2s
multi stream (72M, 61 files) | 23s | NA

After ustr | Rust | Python
-- | -- | --
single stream (10M) | 1s | 1.56
multi stream (72M, 61 files) | 18s | NA

A standard back test run timing went from ~2.35 to ~1.6s

PR #1191, opened 2023-08-05
Fix a connection issue with the IB client and add support for tags in bracket orders

dkharrat:
# Pull Request

1. Fixes an intermittent connection issue in the IB client when reading the socket. Sometimes an empty result is returned from the socket, which causes the `not (len(buf) > 0)` condition to be true and resetting the connection. By removing that condition, the read from the socket is retried again in the outer loop, which usually succeeds.
2. Adds support for providing `tags` for bracket orders.  The tags were hardcoded and couldn't be modified.

## Type of change

I tested it by running a live trading session locally.


cjdsellers:
Hi @dkharrat 

Thanks so much for the contribution! Great work on the connection fix. 

For the tags we may have some users relying on those, it's probably better to provide separate parameters for `entry_tags`, `sl_tags`, and `tp_tags` -- with defaults which match the current hard codings. 

Thoughts?

rsmb7z:
Hi @dkharrat 
You can easily create any order/combination directly within Strategy and have more granual control. You would essentially need
`self.order_factory.generate_client_order_id()` and `self.order_factory.generate_order_list_id()`.

@cjdsellers 
Maybe we can relax it `tag or 'ENTRY'` (existing static tag) as the tag argument is with `None`. I personally use the above approach though.

nautilus_trader/adapters/interactive_brokers/client/client.py
```diff
@@ -547,8 +547,9 @@ async def _socket_connect(self):
             # sometimes I get news before the server version, thus the loop
             while len(fields) != 2:
                 self._client.decoder.interpret(fields)
+                await asyncio.sleep(1)
                 buf = await self._loop.run_in_executor(None, self._client.conn.recvMsg)
-                if not self._client.conn.isConnected() or not (len(buf) > 0):
+                if not self._client.conn.isConnected():
```
rsmb7z, line 552: Removing `or not (len(buf) > 0)` would lead to errors.

dkharrat: what errors does it lead to exactly? From my testing, sometimes the socket returns an empty result on the initial connection, which means this condition results in forcibly disconnecting the client. Future retries behave the same way, eventually timing out, so the client is never able to connect successfully.

By removing the condition, it keeps reading from the socket until it receives something (or `self._client.conn.isConnected()` is false, so it should be safe. Let me know if I'm missing something and if you have any other suggestions.

dkharrat:
> Hi @dkharrat
> 
> Thanks so much for the contribution! Great work on the connection fix.
> 
> For the tags we may have some users relying on those, it's probably better to provide separate parameters for `entry_tags`, `sl_tags`, and `tp_tags` -- with defaults which match the current hard codings.
> 
> Thoughts?

Good idea! I updated the code to provide separate params for each sub-order. I kept the default as `None` for consistency with the other order methods, since the hardcoded values were not really used for any specific purpose within Nautilus. However, if you feel it's better to keep the hard-coded defaults, I'm happy to update it.

nautilus_trader/adapters/interactive_brokers/client/client.py
cjdsellers: I'm happy to merge as is, lets wait to hear back from @rsmb7z though.

rsmb7z: To add brief background on the issue, this section of code is from original IBAPI [here](https://github.com/nautechsystems/nautilus_ibapi/blob/main/ibapi/client.py#L168) and you will find that `or not (len(buf) > 0)` is not there as well same as you done.

- However if you review the code this section is executed only once on connection but the reason they have used loop is to handle the problem where TWS or Gateway may not send the initial expected packet. 
- As in their original comments in next line, if there is 0 buffer received the code would stuck in infinite loop and this is the issue was being experienced during writing of adapter [mainly tested and running on Windows and AWS Ubuntu both Worksation and Gateway]. So `or not (len(buf) > 0)` added to avoid the connection being stuck infinite. The reason is this reconnection would be called by watchdog during any sort of disconnection and we want it to try again instead of getting stuck in infinite loop.

So while it maybe working for you by removing this section, it may bring up the issue for rest including the automated reconnection and disconnections caused by IB's nightly resets etc. I have not experienced any sort of issue on Windows/Ubuntu. However if you can provide exact method to reproduce, that would really be helpful. Are you trying to connect the TWS/Gateway while it is starting?

dkharrat: I've been able to reproduce it pretty consistently. I'm running it on WSL (Ubuntu) and a local TWS/Gateway after it's been started for a while. It's never able to connect despite multiple attempts and it eventually times out after 90s.

I did notice that if I add a sleep before reading from the socket, it works. So, it might be because the client is reading too quickly before the socket is fully ready. I can change the solution to use a sleep instead, which also works. What do you think @rsmb7z ?

rsmb7z: There something is wrong during your initial connecting that's why it fails to connect to sub-sequent tries. This is problem on TWS side that if socket idoesn't receive data nicely you may need to restart the TWS. 
Mostly people use docker(https://github.com/UnusualAlpha/ib-gateway-docker) for Gateway in production so needs to be logged/authenticated once a week. So in this case gateway is inside docker and application is outside. Your case seems to be inverse. Can you try adding sleep before `TradingNode.run()` and see if it works.

dkharrat: I just tried it again to confirm, and noticed that the issue is only reproducible when using TWS. It is not reproducible using the IB Gateway docker. I'm using the default TWS configurations and haven't modified anything. Adding a sleep before `TradingNode.run()` did not help.

 I think it's a timing issue, since adding a `sleep` before attempting to read from the socket resolves it. I tried adding a `sleep` at various points in `client.py` and noticed that the point where it fixes it is right after [this line](https://github.com/nautechsystems/nautilus_trader/blob/32b0e70d0805326c2cdc317f62e62b69e64f4da1/nautilus_trader/adapters/interactive_brokers/client/client.py#L538). This could be because `sendMsg` is async the client attempting to read from the socket before the message is sent? I can update the change to add a sleep before reading from the socket and keep the `len(buf) > 0` check. If you have other suggestions on how to fix it, please let me know.

rsmb7z: Hi @dkharrat,
Are you using TWS `latest` or `stable` and which version?
Can you please try this solution, if this works we can proceed as this will not compromise existing functionality. In case doesn't work, you can try increasing the counter value.
```
            counter = 0
            while len(fields) != 2:
                if counter >= 10:
                    break
                counter += 1
                
                self._client.decoder.interpret(fields)
                buf = await self._loop.run_in_executor(None, self._client.conn.recvMsg)
                if not self._client.conn.isConnected():
```

dkharrat: I'm using the `stable` version of TWS. I've updated the code to limit the number of retries, similar to what you suggested.

rsmb7z: Awsome! @cjdsellers okay from my side to merge.

cjdsellers:
Great work on this everyone!

@dkharrat looks like you just need to run the pre-commit for some minor black formatting:
```
pre-commit install
pre-commit run --all-files
```

or alternatively
```
make pre-commit
```

dkharrat:
> Great work on this everyone!
> 
> @dkharrat looks like you just need to run the pre-commit for some minor black formatting:
> 
> ```
> pre-commit install
> pre-commit run --all-files
> ```
> 
> or alternatively
> 
> ```
> make pre-commit
> ```

Done!

PR #1195, opened 2023-08-06
Add doc

objectin:
# Pull Request

Adding doc using ```cargo doc --target-dir doc --no-deps```

## Type of change

- [ v ] New feature (non-breaking change which adds functionality)


cjdsellers:
Hi @objectin 

Thanks for your efforts here, and for your interest in contributing!

We do actually publish the [Rust docs](https://docs.nautilustrader.io/develop/core/) already, except they're currently fairly hidden and unlinked (mostly because they're very incomplete right now).
https://github.com/nautechsystems/nautilus_trader/blob/develop/.github/workflows/docs.yml#L46

Are there some particular areas you were interested in helping out with?

PR #1231, opened 2023-09-07
Improved limit order backtesting

objectin:
# Pull Request

Improved limit order backtesting, worked with @DSmyungjin 

## Type of change

Implemented Better Limit Order Backtesting (Now works with partial filled orders and dynamically changes order qty after a trade tick also for limit order backtesting)

- [ v ] New feature (non-breaking change which adds functionality)

## Details
For a more sophisticated l2 orderbook backtest, some modifications were made.

The modified backtest reads the qty information from the book and stores it inside the order object (pre_book_qty) when a limit order is accepted on the exchange.

When a tradetick occurs, the pre_book_qty is calculated first when performing the execution operation, creating a queue-like effect in the l2 orderbook.

We will be working on more sophisticated L2 backtests in the future.

order -> add pre_book_qty value to the object

book -> add get_quantity_for_price method - returns qty to the order book, based on price

BookOrder -> add pre_book_qty value to object

ladder -> simulate order fills pre_order_qty, if it exists, so that this value is subtracted first


cjdsellers (commented):
Hi @objectin 

Thanks so much for looking at this and for the contirbution. I think enhancing the realism for L2 order books is very valuable, and improves the realism of backtests.

Overall I think this solution would work, however its very backtest specific and changes two of the core structures `OrderBookDelta` and `BookOrder`. We can't change the schema for these as they're used in the data storage layer for writing arrow/parquet, and for live data feeds where `pre_book_qty` isn't a concept, or at least not something which would be reflected in "raw" order book data.

Do you think we could juggle a few things so we don't have to carry `pre_book_qty` on these existing structures, and instead move these to a map with some appropriate key, and params which we pass into the backtest specific methods?

This is just my first initial thought, and I'm happy to bounce around any ideas you may have with the above additional requirements.

Many thanks!

limx0:
@objectin - I've made a potentially simpler + more flexible change in https://github.com/nautechsystems/nautilus_trader/pull/1242.

Would you mind taking a look and seeing if it fits your needs?

cjdsellers:
Now that we have `SimulationModule.pre_process` available as a short term solution, we could probably prune this PR back to just exposing the `get_quantity_for_price` method?

cjdsellers:
Hi @objectin

Are you happy for me to apply the changes exposing `get_quantity_for_price` straight onto `develop` branch, and then close this PR? many thanks!

cjdsellers:
Closing this for now, also added the `get_quantity_for_price` method from 99600240d81a7f8f53796b677cb3f2fbfd76f443.

PR #1233, opened 2023-09-08
Add check for ibapi version

rsmb7z:
# Pull Request

Added verification check for the ibapi version.

## Type of change

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -20,6 +20,7 @@ from decimal import Decimal
 from inspect import iscoroutinefunction
 from typing import Callable, Optional, Union
 
+import ibapi
 import pandas as pd
 
 # fmt: off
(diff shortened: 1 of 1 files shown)


nautilus_trader/adapters/interactive_brokers/client/client.py
```diff
@@ -80,6 +81,11 @@
 # fmt: on
 
 
+ibapi_version = "10.19.1"
```
cjdsellers, line 84: Just curious about the context behind hard coding a specific version, and version desyncs?

rsmb7z: Users would try to pip install ibapi which results in installation of [older official](https://pypi.org/project/ibapi/) version from pypi. Thought adding a check may help ease the troubleshooting. If it can somehow read from pyproject.toml file or other approach, without hard coding that would actually be more better.

cjdsellers: Right, I see the use case now -- I agree this is helpful. Your thought about parsing the version from the `pyproject.toml` is good, then we single source the version which is easier to maintain.

cjdsellers: I went ahead and added this check https://github.com/nautechsystems/nautilus_trader/commit/b88dc4cb7503282734998c060e81bba7568776e3 :pray: .

Issue #1238, opened 2023-09-11
Binance min_notional, max_notional returning None or 0.0


AnthonyVince:
# Bug Report

### Expected Behaviour
The following code should return values in the strategy.
Binance pre-trade checks verify orders against the notional.
Nautilus Risk engine checs against max_notional (not checked yet)

 
`self.cache.instrument(self.instrument_id).min_notional`
`self.cache.instrument(self.instrument_id).max_notional`

### Actual Behaviour
Currently when connecting to Binance testnet for USDt margined perpetuals I get Zero or None.

<img width="882" alt="Screenshot 2023-09-11 at 16 10 00" src="https://github.com/nautechsystems/nautilus_trader/assets/137524768/51b601d4-62d7-4a5f-af79-b9ff1276ad5f">

### Steps to Reproduce the Problem
```

import os
from nautilus_trader.adapters.binance.common.enums import BinanceAccountType
from nautilus_trader.adapters.binance.config import BinanceDataClientConfig
from nautilus_trader.adapters.binance.config import BinanceExecClientConfig
from nautilus_trader.adapters.binance.factories import BinanceLiveDataClientFactory
from nautilus_trader.adapters.binance.factories import BinanceLiveExecClientFactory
from nautilus_trader.config import CacheDatabaseConfig
from nautilus_trader.config import InstrumentProviderConfig
from nautilus_trader.config import LiveExecEngineConfig
from nautilus_trader.config import LoggingConfig
from nautilus_trader.config import TradingNodeConfig
from nautilus_trader.core.rust.common import LogColor
from nautilus_trader.live.node import TradingNode
from nautilus_trader.model.data import OrderBookDeltas
from nautilus_trader.model.identifiers import InstrumentId
from nautilus_trader.model.orderbook import OrderBook
from nautilus_trader.trading.strategy import Strategy


class SubscribeStrategy(Strategy):
    def __init__(self) -> None:
        super().__init__()
        self.instrument_id = InstrumentId.from_str("ETHUSDT-PERP.BINANCE",)
        self.depth = 50
        self.interval_ms = 10

    def on_start(self) -> None:
        self.instrument = self.cache.instrument(self.instrument_id)
        if self.instrument is None:
            self.log.error(f"Could not find instrument for {self.instrument_id}")
            self.stop()
            return
        self.subscribe_order_book_deltas(self.instrument_id, depth=50)
        self.subscribe_order_book_snapshots(
            self.instrument_id,
            depth=20,
[...]

PR #1239, opened 2023-09-11
#1238 attach position risk for parsing binance instrument

filipmacek:
# Pull Request

## Type of change


cjdsellers:
It's strange the API isn't symmetric and the `maxNotional` doesn't appear next to the `minNotional` filter for exchange information. Looking at the docs it indeed looks like this information is on the position risk response, so lets try this!

PR #1240, opened 2023-09-12
fix Money.from_raw

ghill2:


cjdsellers:
Good catch, and many thanks!

PR #1242, opened 2023-09-17
Add pre_process hook to SimulationModule

limx0:
# Pull Request

Adds a `pre_process` hook to the SimulationModule and calls inside the BacktestEngine prior to passing through to the matching engines.

Should allow more complicated matching logic to be applied.

### Example: Passive fills

An example use case is custom fills for limit (passive) orders, which could look something like: 

```python
class PassiveFillModule(SimulationModule):
    def pre_process(self, data):
        # On trade tick
        if isinstance(data, TradeTick):
            # Check for any open passive orders
            open_orders = [o for o in self.cache.orders_open(instrument_id=data.instrument_id) if isinstance(o, LimitOrder)]

            for order in open_orders:
                # Check if trade in cross with our order
                should_fill = (
                    (order.side == OrderSide.BUY and order.price >= data.price) or
                    (order.side == OrderSide.BUY and order.price <= data.price)
                )
                if should_fill:
                    # Fill order
                    matching_engine = self.exchange.get_matching_engine(data.instrument_id)
                    book = matching_engine.get_book()
                    book.update_trade_tick(data)
                    fills = matching_engine.determine_limit_price_and_volume(order)
                    matching_engine.apply_fills(
                        order=order,
                        fills=fills,
                        liquidity_side=LiquiditySide.MAKER,
                    )
```

## Type of change

Added test.

### DIFF
--- a/nautilus_trader/backtest/engine.pyx
+++ b/nautilus_trader/backtest/engine.pyx
@@ -1037,10 +1037,7 @@ cdef class BacktestEngine:
                     raw_handlers_count = raw_handlers.len
 
                 # Run pre-process for any simulation_modules
-                if isinstance(data, (OrderBookDelta, OrderBookDeltas, QuoteTick, TradeTick, Bar, InstrumentStatusUpdate)):
-                    venue = self._venues[data.instrument_id.venue]
-                    for module in venue.modules:
-                        module.pre_data(data)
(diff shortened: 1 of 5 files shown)


limx0:
Just a follow up comment regarding usage here - make sure you update the cache after making any modifications to orders or you may run into issues like I did. 

I'll try and follow up with some working examples

PR #1260, opened 2023-09-28
Add python socket tests

limx0:
Add python socket tests.


limx0:
Adding python side socket tests, similar to http & websocket client. 

Currently segfaulting, @twitu any ideas here?

PR #1272, opened 2023-10-07
Add RelativeStrengthIndex(RSI) indicator for Rust

filipmacek:
# Pull Request

- trait `MovingAverage` was implemented that exposed `update_raw` method
- implemented `RelativeStrengthIndex` struct in rust with tests
- new indicator module `momentum` was created where RSI indicator is placed


cjdsellers (commented):
LGTM!

Issue #1236, opened 2023-09-10
WebSocketClient.connect(): Ability to pass connection headers


skc-j:
# Feature Request for WebSocketClient.connect() with headers.

For some websockets, expect webscoket client to pass connection headers (such as "Origin", "Authorization", etc).

In the current version, WebSocketClient.connect() does not seem to accept headers (though tokio-tungstenite seems to have this ability to take header).

For websocket connections expecting such connection headers, I getting "Exception: Unable to make websocket connection because of error: HTTP error: 403 Forbidden".

PR #1282, opened 2023-10-18
Passing Headers to WebSocketClient.connect()

ruthvik125:
# Pull Request
For issue #1236 
-> Passing a hyper::Request to which headers can be inserted ,instead of a url ,in connect_async()

Type Of Change

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
 
       # Install build dependencies
       - name: Install build dependencies
-        run: python -m pip install --upgrade pip setuptools wheel pre-commit poetry==1.3.2
+        run: python -m pip install --upgrade pip setuptools wheel pre-commit poetry==1.5.1
 
       # Setup cached pre-commit
       - name: Setup cached pre-commit
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [x64]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest]  # windows-latest
         python-version: ["3.9", "3.10", "3.11"]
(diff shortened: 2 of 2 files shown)


ruthvik125:
We can use a hyper::Request to which we can insert headers and pass it in connect_async() to connect from the client. Is adding an extra parameter for headers: for every function that is calling connect() or connect_with_server() better , or is making a separate function for connecting with headers better?

cjdsellers:
> We can use a hyper::Request to which we can insert headers and pass it in connect_async() to connect from the client. Is adding an extra parameter for headers: for every function that is calling connect() or connect_with_server() better , or is making a separate function for connecting with headers better?

Hi @ruthvik125 
Thanks for the contribution here!
I would say an additional parameter with a `None` like default would be the best way to go.

nautilus_core/network/src/websocket.rs
```diff
@@ -359,6 +364,7 @@ impl WebSocketClient {
     #[staticmethod]
     fn connect(
         url: String,
+        headers:Vec<(&str,&str)>,
```
twitu, line 367: You'll need to take a `Vec<(String, String)>` because having lifetimes will be problematic in the Python API. And you want to store these headers in the `WebsocketInnerClient` so that they can be re-used in case of a reconnect.

nautilus_core/network/src/websocket.rs
```diff
@@ -86,7 +86,11 @@ impl WebSocketClientInner {
     /// Connects with the server creating a tokio-tungstenite websocket stream.
     #[inline]
     pub async fn connect_with_server(url: &str) -> Result<(MessageWriter, MessageReader), Error> {
```
twitu, line 88: You want to pass a clone of the stored `Vec<(String, String)>` here. Then consume and add the key value pairs to the request.

twitu (changes requested):

cjdsellers:
We also need to run the pre-commit checks:
```
make pre-commit
```
or
```
pre-commit run --all-files
```

Many thanks!

PR #1314, opened 2023-10-29
Add OptionsContract in rust with pyo3

filipmacek:
# Pull Request

- existing Python pyo3 tests for instruments refactoring for function based tests
- struct `OptionsContract` with pymethods and pyo3 python tests


cjdsellers (approved):
LGTM!

PR #1317, opened 2023-10-31
Add FuturesContract in rust with pyo3

filipmacek:
# Pull Request

- struct `FuturesContract` with pymethods and pyo3 python tests


cjdsellers (commented):
LGTM!

Issue #1098, opened 2023-05-07
Porting network layer to Rust


twitu:
This issue tracks the effort and discussion for porting the network layer to Rust.

The network module is relevant to the live trading functionality. Exchange adapters use network module clients to make requests to exchanges and pass received data to the engine. It has clients for 3 protocols - HTTP, websockets and raw socket.

The end goal is to port all three client implementations to rust and expose them as pyo3 modules to be directly consumed by the adapters. Further along it is expected that the adapters will also be ported and the intermediate pyo3 layer will no longer be necessary. Since the network module is IO heavy it is currently implemented as async and will be async in the rust version as well.

twitu:
The first implementation aims to port the HTTP client. The client is backed by a hyper client that is configured for HTTPS. The custom response returned by the client has the body, status and headers that have keys in `header_keys` field of the client. Both client and response can directly be accessed as python objects through pyo3.

```rust
#[pyclass]
pub struct HttpClient {
    client: Client<HttpsConnector<hyper::client::HttpConnector>>,
    header_keys: Vec<String>,
}

#[pyclass]
pub struct HttpResponse {
    #[pyo3(get)]
    pub status: u16,
    #[pyo3(get)]
    headers: HashMap<String, String>,
    #[pyo3(get)]
    body: Vec<u8>,
}
```

The live trading setup has many async tasks running. Many of them will continue to be implemented in Python for the time being. So the rust and python async runtimes will have to co-exist. Here we use pyo3-asyncio library to help interface these two runtimes. At a high level it wraps a rust async task in python async task, and the result returned by the rust task is converted into a python object.

![image](https://user-images.githubusercontent.com/23196890/236677105-0b3e4e8d-e607-4229-9b16-e4860e619bb8.png)

Note: this means for each request two async tasks are created and executed. The underlying rust task and the wrapping python task.

Currently this test passes.

```
import asyncio
from nautilus_network import HttpClient

async def make_request(client: HttpClient):
    return await client.request("get", "https://github.com", {})

if __name__ == "__main__":
    client = HttpClient()
    res = asyncio.run(make_request(client))
    assert res.status == 200
    assert len(res.body) != 0
```

rsmb7z:
> Further along it is expected that the adapters will also be ported and the intermediate pyo3 layer will no longer be necessary. 

Hi @twitu @cjdsellers 
[...]

PR #1322, opened 2023-11-05
Add optional ping message

ayush-sb:
# Pull Request
- Add optional `message` field in WebSocketConfig
- Send either `Message::Ping` or `Message::Text` in `WebSocketClientInner::spawn_heartbeat_task` depending on presence of `message`

### DIFF
--- a/nautilus_core/network/src/websocket.rs
+++ b/nautilus_core/network/src/websocket.rs
@@ -715,4 +715,71 @@ counter = Counter()",
         sleep(Duration::from_secs(1)).await;
         assert!(client.is_disconnected());
     }
+
+    #[tokio::test]
+    #[traced_test]
+    async fn message_ping_test() {
+        prepare_freethreaded_python();
+
+        let header_key = "hello-custom-key".to_string();
+        let header_value = "hello-custom-value".to_string();
+
+        let (checker, handler) = Python::with_gil(|py| {
+            let pymod = PyModule::from_code(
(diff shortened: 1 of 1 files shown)


filipmacek:
Looks alright. Can you also pls add tests for this

ayush-sb:
Does adding tests here mean making one test with a `String` and one `None` for `message`?

filipmacek:
Yeah, we want to test if the mock `TestServer` from [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_core/network/src/websocket.rs#L523) is receiving correct ping messages. Maybe @twitu can help you with that

twitu:
The changes look good. We just need to add a couple of tests and get it passing.
* You can verify the CI checks are passing locally by running `make pre-commit`
* The design is a bit complex but here's an overview. All the logic in the websocket.rs is for a websocket client only including the `spawn_read_task` and `spawn_heart_beat_task`. All the various parts create a client that can send hearbeats, send messages and reconnect if the connection breaks. All of these are automatically run and managed when a new websocket client is created, so these are not something that you need to change or manage in the tests. More details are here https://github.com/nautechsystems/nautilus_trader/issues/1098#issuecomment-1598206852
* The main logic to look into for adding a small new test is `impl TestServer`. So you can create a new server that stores the expected heartbeat message and when in the impl logic you can check for those messages.

PR #1347, opened 2023-11-16
Implement HullMovingAverage in Rust

ayush-sb:
# Pull Request

- Implement HullMovingAverage in Rust, with tests
- Add pyo3 exports

### DIFF
--- a/nautilus_core/indicators/src/average/hma.rs
+++ b/nautilus_core/indicators/src/average/hma.rs
@@ -1,0 +1,246 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2023 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
+//  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+//  You may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------------------------------
+
+use std::fmt::{Display, Formatter};
+
+use anyhow::Result;
+use nautilus_model::{
+    data::{bar::Bar, quote::QuoteTick, trade::TradeTick},
+    enums::PriceType,
+};
+use pyo3::prelude::*;
+
+use crate::{
+    average::wma::WeightedMovingAverage,
+    indicator::{Indicator, MovingAverage},
+};
+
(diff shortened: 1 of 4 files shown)


cjdsellers:
Great work @ayush-sb 

This is currently still marked as a draft, are you happy if we merge?

ayush-sb:
Filip asked me to work on this. I'm just waiting on him to review it, so I've marked it as draft

cjdsellers:
Thanks @ayush-sb if its ready for review then best to mark it open.

ayush-sb:
Thanks a lot Chris, will keep in mind next time 😀

filipmacek:
Looks really good to me. Good job with weights arithmetics in rust, that's perfect. And per your question, the naming of fields is also alright. I would name them just like that

nautilus_core/indicators/src/average/hma.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn handle_trade_tick(&mut self, tick: &TradeTick) {
+        self.update_raw((&tick.price).into());
+    }
+
+    fn handle_bar(&mut self, bar: &Bar) {
+        self.update_raw((&bar.close).into());
+    }
+
+    fn reset(&mut self) {
+        self.value = 0.0;
```
filipmacek, line 80: You also need to reset the state, of three moving averages here

nautilus_core/indicators/src/average/hma.rs
```diff
@@ -0,0 +1,246 @@
...
+        assert_eq!(indicator_hma_10.value, 1522.0);
+        assert!(indicator_hma_10.has_inputs);
+        assert!(!indicator_hma_10.is_initialized);
+    }
+
+    #[rstest]
+    fn test_reset(mut indicator_hma_10: HullMovingAverage) {
+        indicator_hma_10.update_raw(1.0);
+        assert_eq!(indicator_hma_10.count, 1);
+        indicator_hma_10.reset();
```
filipmacek, line 240: You could better test this, running few `update_raw` then `reset` and few again `update_raw` would fail, because three ma's didn't change state in reset method, and this test would fail

filipmacek:
It would be also good to add python tests for these pyo3 objects. If added it here for example for `SimpleMovingAverage` so you can follow it for inspiration https://github.com/nautechsystems/nautilus_trader/pull/1349
Also pls add typings in for your moving average here in `nautilus_pyo3.pyi` file.
And pls check you have any problems rounding float values from rust, before putting `round(..)` try investigating if it will be round properly, we started experiencing that value `1987.5` from rust is actually returned like `1987.5000000000002` ,so check it and report here. If you need any help let me know

cjdsellers:
So regarding the above from @filipmacek 

We actually uncovered an issue which was addressed from fd4280f41dd7019ed970d649f347e9963b7f233c. So I think we just need to close the loop on the Python side tests, and adding the type annotations to the `nautilus_pyo3.pyi` and make that part of the workflow on indicators going forward. Then happy to merge!

cjdsellers:
Thanks for completing the extra pieces @ayush-sb :pray:

PR #1351, opened 2023-11-18
Nautilus SQL schema init folder structure, scripts and POC sql cache database

filipmacek:
# Pull Request

- added `*.env` files to gitignore, so they are not picked up by git if they are created locally in source code folders
- created schema folder structure for Nautilus SQL data model
- added two new actions to `Makefile` and that are `init-db` and `drop-db` that will initialize Nautilus SQL schema to connected database ( mostly Postgres here). We will need those actions to resync the latest schema with the database. Migrations of schema will not be currently supported
- added corresponding cargo target for this scripts
- added two utils functions `init_db_schema` and `setup_test_database` that will be used both by cargo targets and tests when initializing testing sqlite database
- initial design for `SqlCacheDatabase` with `general` table for storing custom keys and values

### DIFF
--- a/nautilus_core/persistence/src/bin/init_db.rs
+++ b/nautilus_core/persistence/src/bin/init_db.rs
@@ -1,0 +1,28 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2023 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
+//  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+//  You may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------------------------------
+
+use anyhow::Result;
+use dotenv::dotenv;
+use nautilus_persistence::db::database::{init_db_schema, Database};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // load envs if exists
+    dotenv().ok();
+    let db = Database::new(None, None).await;
+    let sql_schema_dir = "../schema/sql";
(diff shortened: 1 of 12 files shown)


nautilus_core/persistence/src/bin/init_db.rs
```diff
@@ -0,0 +1,28 @@
...
+//  limitations under the License.
+// -------------------------------------------------------------------------------------------------
+
+use anyhow::Result;
+use dotenv::dotenv;
+use nautilus_persistence::db::database::{init_db_schema, Database};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // load envs if exists
```
twitu, line 22: > added two new actions to Makefile and that are init-db and drop-db that will initialize Nautilus SQL schema to connected database ( mostly Postgres here). We will need those actions to resync the latest schema with the database. Migrations of schema will not be currently supported

You can consider using sqlx built in migrate functionality which applies migrations to a db on startup. https://docs.rs/sqlx/latest/sqlx/macro.migrate.html

nautilus_core/persistence/src/db/database.rs
```diff
@@ -106,30 +108,52 @@ impl Database {
...
+                    e
+                )
+            });
         }
-        Database::new(Some(DatabaseEngine::SQLITE), Some("sqlite:test_db.sqlite")).await
     }
+    Ok(())
+}
+
+pub async fn setup_test_database() -> Database {
```
twitu, line 134: Consider using sqlx test attribute to setup test db for tests. https://docs.rs/sqlx/latest/sqlx/attr.test.html

nautilus_core/persistence/src/bin/init_db.rs
filipmacek: Yeah, good point. I know about this, but migration functionality will come at the later stages when we finish building Nautilus SQL data model, and when schema becomes more stable and less changing. And when there will actual db state outside nautilus which needs to be preserved but migrated to the latest schema.

nautilus_core/persistence/src/db/database.rs
filipmacek: Yeah, but best to have these manual function where you have more control what are you doing. Also this function could will be exported with pyo3 so that python code can also use it; specially useful in pyo3 python tests

nautilus_core/persistence/src/bin/init_db.rs
cjdsellers: Thanks for the heads up @twitu, this will be very useful at some point.

cjdsellers (approved):

Issue #1353, opened 2023-11-18
Incorrect ts_event and ts_init in historic Interactive Brokers daily bars


benjaminsingleton:
# Bug Report

### Expected Behavior
I expect `ts_init` and `ts_event` to reflect accurate unix nanoseconds for daily bars.

### Actual Behavior
Daily bars retrieved using the `HistoricInteractiveBrokersClient` have timestamps that look like: `20231116000000000`.

Full example: `Bar(SPY.ARCA-1-DAY-LAST-EXTERNAL,449.22,450.56,448.12,450.23,48072316,20231116000000000)`

Per the [IB Docs](https://interactivebrokers.github.io/tws-api/historical_bars.html): "formatDate, The format in which the incoming bars' date should be presented. Note that for day bars, only yyyyMMdd format is available."

### Steps to Reproduce the Problem

```
import asyncio
import datetime

from nautilus_trader.adapters.interactive_brokers.historic import HistoricInteractiveBrokersClient

async def main():
    client = HistoricInteractiveBrokersClient(port=4002, client_id=5)
    await client._connect()
    await asyncio.sleep(2)

    bars = await client.request_bars(
        bar_specifications=["1-DAY-LAST"],
        end_date_time=datetime.datetime(2023, 11, 17),
        duration="1 D",
        tz_name="America/New_York",
        instrument_ids=["SPY.ARCA"],
    )
    print(bars)


if __name__ == "__main__":
    asyncio.run(main())
```

### Specifications

- OS platform: Ventura 13.0
- Python version: 3.10.13
- `nautilus_trader` version: 1.180

cjdsellers:
Now closed from https://github.com/nautechsystems/nautilus_trader/pull/1354.

Many thanks for the fix @benjaminsingleton :pray:

PR #1354, opened 2023-11-18
Fix timestamps in Interactive Brokers daily bars

benjaminsingleton:
# Pull Request

References #1353

## Type of change

Retrieved daily and non-daily (hourly, etc.) bars with success.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -1260,13 +1260,54 @@ class InteractiveBrokersClient(Component, EWrapper):
         )
         return bar
 
-    @staticmethod
-    def _ib_bar_to_ts_init(bar: BarData, bar_type: BarType) -> int:
-        ts_init = (
-            pd.Timestamp.fromtimestamp(int(bar.date), tz=pytz.utc).value
-            + pd.Timedelta(bar_type.spec.timedelta).value
-        )
-        return ts_init
+    def _convert_ib_bar_to_unix_nanos(self, bar: BarData, bar_type: BarType) -> int:
+        """
+        Convert the date from BarData to unix nanoseconds.
+
+        If the bar type's aggregation is 14 (daily), the bar date is always returned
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @benjaminsingleton 

Thanks for these changes, I think this looks great. My own preference is for time stamping bars on the close also - do you think we should make this configurable for the IB client though, and retain the default time stamping at the open by default?

We also have a config option `DataEngineConfig.time_bars_timestamp_on_close` which is related/in play here.

benjaminsingleton:
The existing code actually only adjusts `ts_init` to reflect the closing timestamp while `ts_event` reflects the opening timestamp. By default, it should it be set the opposite way? For bars, does `ts_init` generally reflect the start of the bar period?

benjaminsingleton:
The code here actually only adjusts `ts_init` to reflect the closing timestamp while `ts_event` reflects the opening timestamp. Am I correct that when running backtests / live data that the `DataEngineConfig.time_bars_timestamp_on_close` will take care of converting `ts_event` to the closing timestamp? If so, then I don't think we need to modify the code here.

cjdsellers:
> Am I correct that when running backtests / live data that the `DataEngineConfig.time_bars_timestamp_on_close=True` will take care of converting `ts_event` to the closing timestamp? If so, then I don't think we need to modify the code here.

Correct, `True` is actually the default behavior - when `False` it will use the opening timestamp for `ts_event` (and `ts_init` is always close / object initialization).

I see, now that I've expanded all the code I see this is indeed just fixing the parsing for daily bars with no change to the `ts_init` timestamp adjustment - so I think this is correct as is too.

rsmb7z (approved):

rsmb7z:
> We also have a config option `DataEngineConfig.time_bars_timestamp_on_close` which is related/in play here.

This flag is not in effect within the adapter, may be added. However current behavior is `ts_event` bar opening time and `ts_init` bar closing time (for historical) and actual time in case of live. New addition just supports the timestamp difference from IB following same.

cjdsellers:
Thanks for the review @rsmb7z !

PR #1355, opened 2023-11-19
Unified core time for live and static use cases

twitu:
# Pull Request

Creates a unified concept of time in the core crate. Time can be live or static based on its mode. It uses AtomicU64 to make all time operations immutable. It also makes time cloneable, so that all components in the engine can share the same time.

Currently it uses `Rc` cloneable time for single thread. But it can be switched to thread-safe cloneable time.

## Type of change

All tests pass.


filipmacek:
Awesome job here @twitu. I really like unification of clocks and the removal of mutable references, and good thing that you also refactored generators to use this new `AtomicTime` struct

cjdsellers:
Great solution here :crab:

Issue #1358, opened 2023-11-20
Implement "cash quantity" order attribute to enable crypto trading on Interactive Brokers


benjaminsingleton:
# Feature Request

To trade cryptocurrencies on Interactive Brokers, you must set a "cash quantity" when placing an order. Cash quantity is an attribute of the [order](https://github.com/nautechsystems/nautilus_ibapi/blob/main/ibapi/order.py#L190). If you attempt to execute a standard market buy order, IB returns an error message: "You must set Cash Quantity for this order'". This error is currently uncaught by NT (can only be found in DEBUG logs), so it's only apparent that something is wrong because there is a warning that the client order id is not found.

The NT IB adapter currently does [not support](https://github.com/nautechsystems/nautilus_trader/blob/master/nautilus_trader/adapters/interactive_brokers/parsing/execution.py#L70) this order field, which prevents crypto trading.

Per the [docs](https://interactivebrokers.github.io/tws-api/basic_orders.html#:%7E:text=Cash%20Quantity%20Orders-,Order%20order%20%3D%20new%20Order()%3B,order.cashQty(cashQty)%3B,-Limit%20if%20Touched): "Forex orders can be placed in denomination of second currency in pair using cashQty field." The docs say "can" but for crypto, you _must_.

Another reference to cashQty on SO: https://stackoverflow.com/a/69400469

Depending on the amount of work involved in implementing this feature, NT should probably raise a NotImplementedError for crypto trading. I'm happy to take a crack at it, but it's not immediately clear to me how to add this attribute to a Nautilus `Order`, so it can then be passed into `IBOrder` before being submitted to IB.

Below are logs the show the error: "You must set Cash Quantity for this order"

```
2023-11-19T17:14:25.003813000Z [INF] TESTER-001.MyStrategy-000: <--[EVT] OrderInitialized(instrument_id=BTC/USD.PAXOS, client_order_id=O-20231119-1714-001-000-1, side=BUY, type=MARKET, quantity=250.00000000, time_in_force=IOC, post_only=False, reduce_only=False, quote_quantity=True, options={}, emulation_trigger=NO_TRIGGER, trigger_instrument_id=None, contingency_type=NO_CONTINGENCY, order_list_id=None, linked_order_ids=None, parent_order_id=None, exec_algorithm_id=None, exec_algorithm_params=None, exec_spawn_id=None, tags=None).
[...]

PR #1362, opened 2023-11-23
Define cryptocurrencies on IB as inverse instruments

benjaminsingleton:
# Pull Request

Addresses #1358. Crypto instruments on Interactive Brokers are [inverse](https://docs.nautilustrader.io/api_reference/model/instruments.html#nautilus_trader.model.instruments.base.Instrument.is_inverse) instruments. A `cashQty` must be specified in the `IBOrder`.

## Type of change

Ran simple strategy to buy every few seconds. No errors encountered.


cjdsellers:
Thanks @benjaminsingleton

Looking good to me :pray: , did you intend on adding more to this PR?

benjaminsingleton:
Still a WIP. These changes resolve the `You must set Cash Quantity for this order` error, but now I'm encountering a new error: `API does not support trading in fractions for CashQty orders.`. I've tried decimal, double, int data types for cashQty, and I'm still getting the same error, so I need to continue investigating.

cjdsellers:
> Still a WIP. These changes resolve the `You must set Cash Quantity for this order` error, but now I'm encountering a new error: `API does not support trading in fractions for CashQty orders.`. I've tried decimal, double, int data types for cashQty, and I'm still getting the same error, so I need to continue investigating.

My first thought would be we're somehow sending a type which represents fractions, but then I see you tried `int`. It makes sense that if you expressed an inverse quantity as USD, then cents wouldn't be acceptable. 

Thanks for the update!

benjaminsingleton:
Update -- I'm able to buy / sell crypto without making any changes to code. Turns out there are settings that were blocking API calls and unhelpful error messages. I'm now doing testing because there is some unusual behavior (e.g., fills not being properly recognized), but I expect to have this PR ready soon.

benjaminsingleton:
This is ready for review.

nautilus_trader/adapters/interactive_brokers/execution.py
```diff
@@ -687,7 +691,7 @@ def _on_account_summary(self, tag: str, value: str, currency: str):
 
         self._account_summary_loaded.set()
 
-    def _handle_order_event(
+    def _handle_order_event(  # noqa: C901
```
cjdsellers, line 694: Thanks for including the code too :+1:

cjdsellers (approved):
Looks like a good solution, many thanks!

PR #1368, opened 2023-11-24
Reorganize Rust order events

filipmacek:
# Pull Request

- separated all order events into its own file so its cleaner and better oranized
- also its will be easier to port these events with pyo3


cjdsellers (approved):
Great stuff, this follows similar patterns we use for identifiers and the orders themselves.

PR #1372, opened 2023-11-25
Reorganize instrument stubs in Rust

filipmacek:
# Pull Request

- moved all instrument stubs to separate file with `stubs` feature so then can be reused in other modules


cjdsellers (approved):
Better organized, cool :rocket:

PR #1383, opened 2023-11-30
Implement event OrderSubmitted in Rust

filipmacek:
# Pull Request

- finished event `OrderSubmitted` in rust and exported in with pyo3 and tested on the python side

### DIFF
--- a/nautilus_core/model/src/events/order/stubs.rs
+++ b/nautilus_core/model/src/events/order/stubs.rs
@@ -23,7 +23,7 @@ use crate::{
     enums::{ContingencyType, LiquiditySide, OrderSide, OrderType, TimeInForce, TriggerType},
     events::order::{
         denied::OrderDenied, filled::OrderFilled, initialized::OrderInitialized,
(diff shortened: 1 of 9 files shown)


nautilus_core/model/src/events/order/stubs.rs
```diff
@@ -40,7 +40,7 @@ pub fn order_filled(
     instrument_id_btc_usdt: InstrumentId,
     client_order_id: ClientOrderId,
 ) -> OrderFilled {
-    let event_id = UUID4::new();
+    let event_id = UUID4::from("16578139-a945-4b65-b46c-bc131a15d8e7");
```
cjdsellers, line 43: We could probably make this stub UUID4 value a constant too?

cjdsellers (approved):

cjdsellers:
Looks like just a couple of conflicts to resolve after merging the prior diff?

filipmacek:
Good points. Rebased, added uuid4 stub

PR #1390, opened 2023-12-03
WIP fsspec bump

limx0:


cjdsellers:
Synced with @limx0 and will revisit this at another time when we actually need a newer `fsspec` version.

PR #1414, opened 2023-12-15
Implement TA-Lib Integration

rsmb7z:
# Pull Request

Enables straightforward integration of TA-Lib into strategies. Useful for anyone looking to utilize TA-Lib indicators.

## Type of change

### DIFF
--- a/examples/backtest/fx_talib_gbpusd_bars_internal.py
+++ b/examples/backtest/fx_talib_gbpusd_bars_internal.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+# -------------------------------------------------------------------------------------------------
+#  Copyright (C) 2015-2023 Nautech Systems Pty Ltd. All rights reserved.
+#  https://nautechsystems.io
+#
+#  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+#  You may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+# -------------------------------------------------------------------------------------------------
+
+import time
+
+from nautilus_trader.backtest.engine import BacktestEngine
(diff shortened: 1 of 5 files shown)


cjdsellers:
Thanks so much @rsmb7z

This has been a highly requested feature for years.

I'm thinking this will be a bit of a hit to test coverage for the files under the `indicators` subpackage though. I wonder if there's some basic smoke tests we can setup to exercise a reasonable amount of the code? We don't have to fully exhaust every edge case, even if that would be the ideal.

Thoughts?

rsmb7z:
Hi @cjdsellers 
Without a doubt, rigorous validation tests are crucial for ensuring quality and reliability. I'm currently developing tests for personal use, but I'd be more than happy to share them. To ensure they work seamlessly in a Continuous Integration (CI) environment, here are the necessary steps:

1. **Integration with Poetry**: TA-Lib can be added as a dependency in the Poetry configuration by inserting `ta-lib = "^0.4.28"` into the pyproject.toml file. It should be noted that the package name differs from the importing syntax, where `import talib` is used in the Python code.
2. **Installing the TA-Lib C Library**: The installation process on Windows presents certain challenges, and while automation is possible, a ready-made script for it is not currently available. For Linux systems, however, the following script has been effective. It facilitates the installation of the TA-Lib C Library, ensuring smooth operation of tests on Linux-based Continuous Integration (CI) systems:

    ```bash
    cd /home/ubuntu && wget http://prdownloads.sourceforge.net/ta-lib/ta-lib-0.4.0-src.tar.gz
    cd /home/ubuntu && tar -xzf ta-lib-0.4.0-src.tar.gz
    cd /home/ubuntu/ta-lib/ && ./configure --build=$(uname -i)-unknown-linux-gnu --prefix=/usr && make && sudo make install
    ```

    **Note**: This script is provided as is. Feel free to modify it or adjust the directory paths as needed for your CI environment.

By following these steps, we should be able to integrate TA-Lib into our CI pipeline effectively. If there are any specific requirements or constraints in our CI setup, please let me know, and I can adjust the approach accordingly.

cjdsellers:
Yes, I think we'll need to add `ta-lib` as an optional dependency, I could sort this out post merge and we could do a follow up diff to add tests?

rsmb7z:
> Yes, I think we'll need to add `ta-lib` as an optional dependency, I could sort this out post merge and we could do a follow up diff to add tests?

Sounds perfect.

cjdsellers (approved):

cjdsellers:
Given the dependency on the binary too, and the fact this is an optional install - I'm happy if the tests just run on Linux too.

PR #1423, opened 2023-12-22
risk engine: reset the state when starting a new backtest

davidsblom:
# Pull Request

Reset the state of the risk engine. Otherwise, all new orders are rejected when starting a new backtest.

## Type of change

Local  running of multiple backtests.


cjdsellers:
Great catch, thanks again @davidsblom :pray:

Issue #1421, opened 2023-12-21
Upgrade hyper dependency to 1.1.0


twitu:
# Feature Request

Hyper has finally stabilized with a major version release. Nautilus uses hyper in the `nautilus-network` module and this needs to be upgraded to the latest `1.1.0` version. However the hyper upgrade breaks some of its public apis and parts of nautilus logic that rely on it need to be updated.

https://hyper.rs/guides/1/upgrading/
https://github.com/hyperium/hyper/releases

cjdsellers:
Now completed from #1429.

Many thanks again @ayush-sb and @twitu :pray:

PR #1429, opened 2023-12-29
Rewrite client and server using reqwest and axum

ayush-sb:
# Pull Request

This PR is to upgrade the network module to hyper 1.1.0.

As mentioned in #1421, hyper 1.1.0 has had many breaking changes. The `Client` struct no longer exists, and requests are now made by a [`SendRequest`](https://docs.rs/hyper/latest/hyper/client/conn/http1/struct.SendRequest.html) which has a different implementation for HTTP1 and HTTP2. However, this is quite low-level for our needs. Initialization of a `SendRequest` object also requires a handshake which causes issues like initialization becoming async, resulting in having to make `fn default()` and other such methods async. Also, some exchanges only support HTTP1, but the HTTP1 SendRequest struct is not cloneable at the moment.

For now, [reqwest](https://github.com/seanmonstar/reqwest) provides a very similar Client object, with roughly the same functionality. It is also cloneable. Another option is moving to the [legacy Client](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Client.html) in the `hyper-util` crate. However, reqwest seems to do the job for now, and we can stay with it until there are some more changes to `hyper` that make it easier to implement for our needs.


ayush-sb:
The currently implemented reqwest client passes all existing tests. The next step is to rewrite the test server using hyper 1.1.0 and update all the tests.

ayush-sb:
The hyper upgrade has also caused breaking changes for servers, such as the removal of `Server` struct and moving towards a lower-level interface in general. The hyper-util crate also does not provide any legacy Server implementation.

[axum](https://github.com/tokio-rs/axum) provides a simple implementation for servers which is sufficient for our needs. While this is a comparatively heavier dependency, it's kept as a dev dependency as it's only used for tests. Moreover it simplifies the test code a lot. If there are any performance issues down the line, the server can be rewritten using a different crate such as [warp](https://github.com/seanmonstar/warp) or [tower-hyper](https://github.com/tower-rs/tower-hyper).

twitu (approved):
Reqwest appears to provide a much cleaner abstraction! Looks good to go 🚀 .

cjdsellers (approved):

PR #1447, opened 2024-01-12
Implement event AccountState in Rust pyo3

filipmacek:
# Pull Request

- continue rust `AccountState` event and export it with rust pyo3 and test it properly


nautilus_core/model/src/python/types/balance.rs
```diff
@@ -66,7 +66,7 @@ impl AccountBalance {
     }
 
     #[pyo3(name = "to_dict")]
-    fn py_to_dict(&self, py: Python<'_>) -> PyResult<PyObject> {
+    pub fn py_to_dict(&self, py: Python<'_>) -> PyResult<PyObject> {
```
cjdsellers, line 69: I think the `pub` access modifier will only affect Rust, anything in `pymethods` will be exposed to Python.

cjdsellers (approved):

Issue #1435, opened 2024-01-05
Logging V2 - optimizations and performance profiling


twitu:
# Feature Request

The first draft of logging implementation in #1431 implements the log framework for a custom nautilus logger. While functionally complete, it still needs a few changes and tests.

1. Performance benchmark. Since logging is used everywhere measure perf change between old and new logging will be critical
    * Logging specific benchmark - log million or so calls with some randomness in input and measure time taken. Log only to stderr to minimize effect of buffers and io on time
    * Full backtest - Measure time difference for any full backtest with both loggers
2. Refactor the logging.rs logic into separate writers for stdout, stderr and file. Which will make it easier to filter messages and extend with more writers if needed. `LogWriter`[^1] trait from `flexi_logger` is a good reference for the kind of features we want from a writer.

[^1]: https://docs.rs/flexi_logger/latest/flexi_logger/writers/trait.LogWriter.html

cjdsellers:
The improved writers have now landed, many thanks @twitu and @ayush-sb :pray: 

A related issue is https://github.com/nautechsystems/nautilus_trader/issues/1461, however closing this issue for now as completed.

PR #1452, opened 2024-01-16
Refactor writers for new logger

ayush-sb:
# Pull Request

WIP: refactor `logging.rs` logic to have separate writers for each stream #1435


cjdsellers:
Hey Ayush, looks like the pre-commit needs to be run for this.

ayush-sb:
Hey Chris, I'll fix it in the next commit today

nautilus_core/common/src/logging.rs
```diff
@@ -496,24 +600,10 @@ impl Logger {
...
-                        Self::write_stderr(&mut err_buf, &line);
-                        Self::flush_stderr(&mut err_buf);
-                    } else if line.level <= stdout_level {
-                        let line =
-                            Self::format_console_log(timestamp, &line, trader_id, is_colored);
-                        Self::write_stdout(&mut out_buf, &line);
-                        Self::flush_stdout(&mut out_buf);
-                    }
-
                     if fileout_level != LevelFilter::Off {
```
twitu, line 603: This can be stored in the FileWriter itself and be checked as part of enabled.

nautilus_core/common/src/logging.rs
```diff
@@ -496,24 +600,10 @@ impl Logger {
...
                         if Self::should_rotate_file(&file_path) {
```
twitu, line 604: This logic too can be pushed inside the FileWriter.

nautilus_core/common/src/lib.rs
```diff
@@ -35,3 +37,12 @@ pub mod python;
 
 #[cfg(feature = "redis")]
 pub mod redis;
+
+pub trait LogWriter {
+    /// Writes a log line.
```
twitu, line 42: Perhaps putting this in a separate module makes more sense. The implementation of the specific loggers stdout, stderr etc can also go in the same module.

This will also keep the logging module focused on the logging infrastructure rather than the IO part of it.

nautilus_core/common/src/logging.rs
```diff
@@ -481,9 +585,9 @@ impl Logger {
 
             match event {
                 LogEvent::Flush => {
-                    Self::flush_stderr(&mut err_buf);
-                    Self::flush_stdout(&mut out_buf);
-                    file_buf.as_mut().map(Self::flush_file);
+                    stdout_writer.flush();
```
twitu, line 588: This should not be needed anymore since the writers should flush when dropped. When the a Flush log even is received logic should simply break out of the loop.

PR #1454, opened 2024-01-16
Add log level to InteractiveBrokersHistoricClient

benjaminsingleton:
# Pull Request

The `InteractiveBrokersHistoricClient` does not support different log levels, which makes troubleshooting IB API issues difficult. This PR provides that functionality.

## Type of change

- [X] Tested locally with `INFO` and `DEBUG` levels set and observed different log messages.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/historic/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/historic/client.py
@@ -35,6 +35,7 @@ from nautilus_trader.common.clock import LiveClock
 from nautilus_trader.common.component import MessageBus
 from nautilus_trader.common.logging import Logger
 from nautilus_trader.common.logging import LoggerAdapter
+from nautilus_trader.common.logging import log_level_from_str
 from nautilus_trader.core.datetime import dt_to_unix_nanos
 from nautilus_trader.core.datetime import unix_nanos_to_dt
 from nautilus_trader.model.data import Bar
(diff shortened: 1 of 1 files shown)


nautilus_trader/adapters/interactive_brokers/historic/client.py
```diff
@@ -59,11 +60,12 @@ def __init__(
         port: int = 7497,
         client_id: int = 1,
         market_data_type: MarketDataTypeEnum = MarketDataTypeEnum.REALTIME,
+        log_level: str = "INFO",
```
cjdsellers, line 63: We could also use `LogLevel` here, which could then be passed straight through to the `Logger`?

benjaminsingleton: I initially had it that way but then thought it might be more user-friendly to make it a string -- similar to `log_level` in `LoggingConfig`.

cjdsellers: Yep, I had a similar thought too. The only reason its that way for `LoggingConfig` is so that it can be serialized, ideally we'd use a stronger type (and could add a msgspec encoding/decoding hook as such). I thought it might save you the trouble of importing the `log_level_from_str` function.

In any case, this works too - the above just another approach :ok_hand:

rsmb7z: > The only reason its that way for `LoggingConfig` is so that it can be serialized

@cjdsellers enums are supported now in `msgspec` by default so you may plan for strict typing in NautilusConfig.

PR #1468, opened 2024-01-24
IB instrument ID parsing changes

fhill2:
# Pull Request

@rsmb7z
@benjaminsingleton 

The current implementation of `ib_contract_to_instrument_id`raises ValueError for many FUT & FOP contracts available on IB.

Specifically, the regex matching does not support and parse localSymbols:
- with no spaces between symbol, letter month, year, `SXAH24`
- with strikes as an int (no remainder) `DVIP40000L`
- in exchange specific notation, mexder=`DVCE91MR24`, oms=`4TLSN4L`
- with multiple spaces (in rare cases), `ISP   24L20`

Here are some other issues with the parsing I've encountered:
- contracts that expire in year 203X, with localSymbol formats containing year as a single digit, get parsed as year 202X
- OSE.JPN as exchange, a period in the exchange should be replaced to properly form an InstrumentId object

I've added test cases to this PR to show the failing localSymbol formats, and hopefully to open a discussion on how to improve it.

## Proposed Changes

This PR includes my proposed changes. All FUT & FOP Instruments have their InstrumentId changed to the value of the contract's localSymbol. The localSymbol uniquely identifies the contract and therefore the instrument in the nautilus environment, and the regex parsing is removed.

Some changes to [interactive_brokers/historic/client.py](https://github.com/nautechsystems/nautilus_trader/blob/990a3f8c89b1f3f0f3572c0eb93e3813538485e6/nautilus_trader/adapters/interactive_brokers/historic/client.py#L453) `request_ticks()` `request_bars()` would have to be considered, so the user can load instruments on request using IBContract() without a localSymbol.

## Type of change

- [ x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -137,7 +137,7 @@ class InteractiveBrokersClient(
         # ConnectionMixin
         self._connection_attempt_counter: int = 0
         self._contract_for_probe: IBContract = instrument_id_to_ib_contract(
-            InstrumentId.from_str("EUR/CHF.IDEALPRO"),
+            InstrumentId.from_str("EUR.CHF=CASH.IDEALPRO"),
         )
 
         # MarketDataMixin
--- a/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
+++ b/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
@@ -81,18 +81,11 @@ RE_OPT = re.compile(
 )
 RE_IND = re.compile(r"^(?P<symbol>\w{1,3})$")
 RE_FUT = re.compile(r"^(?P<symbol>\w{1,3})(?P<month>[FGHJKMNQUVXZ])(?P<year>\d{2})$")
-RE_FUT_ORIGINAL = re.compile(r"^(?P<symbol>\w{1,3})(?P<month>[FGHJKMNQUVXZ])(?P<year>\d)$")
 RE_FUT2 = re.compile(
     r"^(?P<symbol>\w{1,4})(?P<month>(JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))(?P<year>\d{2})$",
 )
-RE_FUT2_ORIGINAL = re.compile(
-    r"^(?P<symbol>\w{1,4}) *(?P<month>(JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)) (?P<year>\d{2})$",
-)
 RE_FOP = re.compile(
     r"^(?P<symbol>\w{1,3})(?P<month>[FGHJKMNQUVXZ])(?P<year>\d{2})(?P<right>[CP])(?P<strike>.{4,5})$",
-)
-RE_FOP_ORIGINAL = re.compile(
-    r"^(?P<symbol>\w{1,3})(?P<month>[FGHJKMNQUVXZ])(?P<year>\d) (?P<right>[CP])(?P<strike>.{4,5})$",
 )
 RE_CRYPTO = re.compile(r"^(?P<symbol>[A-Z]*)\/(?P<currency>[A-Z]{3})$")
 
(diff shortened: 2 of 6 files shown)


cjdsellers (approved):
These changes look good to me!

cjdsellers:
Thanks @fhill2 

Looks like the pre-commit just needs to be run.

To install the git hook so that it runs automatically on commit:
```
pre-commit install
```

Then to run ad-hoc try either:
```
make pre-commit
```

Or alternatively:
```
pre-commit run --all-files
```

cjdsellers:
Pre-commit is fine now, thanks for that.

Looks like just two test fails in CI:
```
FAILED tests/integration_tests/adapters/interactive_brokers/test_providers.py::test_load_futures_contract_instrument - AttributeError: 'NoneType' object has no attribute 'id'
FAILED tests/integration_tests/adapters/interactive_brokers/test_providers.py::test_contract_id_to_instrument_id - AssertionError: assert {174230596: I...'CLZ3.NYMEX')} == {174230596: I...CLZ23.NYMEX')}
  Differing items:
  {174230596: InstrumentId('CLZ3.NYMEX')} != {174230596: InstrumentId('CLZ23.NYMEX')}
  Full diff:
  - {174230596: InstrumentId('CLZ23.NYMEX')}
  ?                              -
  + {174230596: InstrumentId('CLZ3.NYMEX')}
===== 2 failed, 4344 passed, 162 skipped, 59 warnings in 433.26s (0:07:13) =====
```

cjdsellers:
@benjaminsingleton @rsmb7z 

Thoughts on the new symbol format, and inclusion of the `=` separator?

benjaminsingleton:
Taking a look and will share some thoughts soon!

benjaminsingleton:
@fhill2 Thanks for all the thought and work that went into this PR.

My first reaction when looking at the code was wondering why we don't just make the IBKR `localSymbol` match the `InstrumentId.symbol` (except for replacing `.` with `-`). I'm not sure I understand why we are removing spaces from the IBKR `localSymbol`, which predates your PR, unless it's required (thoughts @rsmb7z?). Are spaces allowed in `InstrumentId.symbol` or does that cause problems elsewhere, @cjdsellers? It seems like it would be a lot easier to just keep both symbols the same, so we can avoid having to use regex to convert back and forth. 

I can see the benefits of including the `secType` in `InstrumentId.symbol`, but this change could impact a significant number of users and their data catalogs. If there's a way to avoid breaking changes for the majority of users, I think that would be preferred. What do you think about the following?

* Keep the existing `"symbol.venue"` pattern for `InstrumentId` but optionally allow users to specify a security type at the end, i.e. `"symbol.venue.sectype"`. To parse the `InstrumentId`, you can split on `.` and if it yields 3 parts, then we can use the last part to identify the security type otherwise we can use the existing parsing logic. If we went with this approach, I don't think this would introduce any breaking changes.

cjdsellers:
Thanks @benjaminsingleton

I had similar thoughts to yourself, I think your proposal is good - but I would alter this so that the venue is always at the end as the last component, which means we wouldn't break the existing parsing logic. This is also similar to the Databento convention for parent symbols i.e. `.FUT` for futures, and `.OPT` for options (although making IB and Databento match is out of scope for this right now).

Also to be considered later is using the four letter [ISO 10383 MIC codes](https://www.iso20022.org/market-identifier-codes) in place of IB's convention, e.g. `XNAS` for Nasdaq rather than `NASDAQ` (again, out of scope but just seeding a future discussion).

rsmb7z:
Hi @cjdsellers @benjaminsingleton 
The discussion regarding the symbology of InstrumentId, as previously noted in issue #861, highlights the reasons behind our current symbology. The aim was to maintain compatibility with existing systems. Here's a brief recap of subsequent developments:

1. Prior to my adaptation of the adapter, specifying an IBContract was mandatory for instrument loading. This method is still applicable and can be seen in our [Interactive Brokers example](https://github.com/nautechsystems/nautilus_trader/blob/master/examples/live/interactive_brokers/interactive_brokers_example.py#L90).
2. The use of a 2-digit year format was a deliberate choice to ensure compatibility over a span exceeding 10 years, for instance, differentiating between 2013 and 2023.
3. For Stocks, replacing spaces with '-' and eliminating spaces elsewhere were decisions made to avoid breaking existing functionality.
4. The InstrumentId is consistently derived from localSymbol as it guarantees uniqueness. For example, in the case of futures, the symbol might be `ES`, but the localSymbol could be `ESZ7`. The latter is essential for accurate queries since the former could yield multiple results.

Addressing the choice of symbology involves strategic decision-making and preference. To address the current issue, I propose the following:
- Implement a `strict_symbology` flag in the `InteractiveBrokersInstrumentProviderConfig`.
- When this flag is set to false, the system will continue with the current approach. If `true`, the system will use the format `localSymbol=secType.exchange`, where the localSymbol is utilized as is, without any parsing.
- In this scenario, the conversion to `IBContract` becomes straightforward, and the resulting NautilusSymbol will be formatted as 'localSymbol=secType'.

This approach offers a non-breaking change while providing flexibility to switch between a simplified or strict symbology framework.

rsmb7z:
I have submitted #1488 with proposed changes. If everybody agrees may be merged.
@fhill2 could you please test if setting `strict_symbology=True` works for you.

fhill2:
Thank you @rsmb7z I'll take a look and test this tonight.

cjdsellers:
Just checking what the plan is for this PR now? Was #1488 an alternative, or do we need to rebase and also land this one?

rsmb7z:
> Just checking what the plan is for this PR now? Was #1488 an alternative, or do we need to rebase and also land this one?

@cjdsellers that was replacement, so closing this. Thanks everyone.

Issue #1441, opened 2024-01-10
Issue serializing IB orders


pcgm-team:
I believe the bug can be recreated simply by sending any market orders on IB with streaming config defined.
```
2024-01-03T20:18:11.810040300Z [INF] TESTER-001.IntoViews-000: setting timer
2024-01-03T20:18:11.811402500Z [INF] TESTER-001.IntoViews-000: <--[EVT] OrderSubmitted(instrument_id=SCHW.NYSE, client_order_id=O-20240103-2018-001-000-2-E1, account_id=InteractiveBrokers-DU1435157, ts_event=1704313091810917901).
Could not convert '0.00' with type str: tried to convert to double
2024-01-03T20:18:11.874577200Z [WRN] TESTER-001.ExecEngine: Event message queue stopped with 1 message(s) on queue.
2024-01-03T20:18:16.911264400Z [WRN] TESTER-001.TradingNode: Timed out (5.0s) waiting for node to stop.
Status
------
DataEngine.check_disconnected() == False
ExecEngine.check_disconnected() == False
2024-01-03T20:18:16.911274800Z [INF] TESTER-001.TradingNode: DISPOSING...
2024-01-03T20:18:16.911440400Z [INF] TESTER-001.DataEngine: STOPPING...
2024-01-03T20:18:16.911449600Z [INF] TESTER-001.DataClient-InteractiveBrokers-001: STOPPING...
2024-01-03T20:18:16.912179100Z [INF] TESTER-001.DataClient-InteractiveBrokers-001: STOPPED.
2024-01-03T20:18:16.912589300Z [INF] TESTER-001.DataEngine: STOPPED.
2024-01-03T20:18:16.912596600Z [INF] TESTER-001.RiskEngine: STOPPING...
2024-01-03T20:18:16.912616000Z [INF] TESTER-001.RiskEngine: STOPPED.
2024-01-03T20:18:16.912619800Z [INF] TESTER-001.ExecEngine: STOPPING...
2024-01-03T20:18:16.912626300Z [INF] TESTER-001.ExecClient-InteractiveBrokers: STOPPING...
2024-01-03T20:18:16.912891500Z [INF] TESTER-001.ExecClient-InteractiveBrokers: STOPPED.
2024-01-03T20:18:16.913111600Z [INF] TESTER-001.ExecEngine: Canceling in-flight check task...
2024-01-03T20:18:16.913168600Z [INF] TESTER-001.ExecEngine: STOPPED.
[...]

PR #1469, opened 2024-01-25
Fix OrderUpdated arrow serialization

benjaminsingleton:
# Pull Request

References #1441 . `"trigger_price"` is incorrectly set to data type `pa.float64()` instead of `pa.string()`.

## Type of change

Tested locally. Awaiting validation from others.


cjdsellers:
Thanks Ben, good catch!

PR #1488, opened 2024-02-03
Support optional strict_symbology for IB

rsmb7z:
# Pull Request

The current symbology approach, while effective for a broad range of securities, encounters limitations with certain exchanges due to varied naming conventions that are not compatible with our existing parsing method. To address this, we are introducing a `strict_symbology` flag within the `InteractiveBrokersInstrumentProviderConfig`. When enabled (set to True), this flag mandates strict adherence to the `localSymbol=secType.exchange` format as stipulated by Interactive Brokers. This ensures comprehensive compatibility and eliminates parsing discrepancies.

For instance, with `strict_symbology` set to `True`, the symbology for instruments strictly follows the format provided by InteractiveBrokers, such as **EUR.USD=CASH.IDEALPRO**. Conversely, when set to `False`, the system applies a simplified symbology format, like **EUR/USD.IDEALPRO**. The default setting is False, which opts for the simplified symbology unless explicitly overridden.

## Type of change

Tests proposed by @fhill2 passing when `strict_symbology=True` is set in `InteractiveBrokersInstrumentProviderConfig`.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/config.py
+++ b/nautilus_trader/adapters/interactive_brokers/config.py
@@ -69,6 +69,12 @@ class InteractiveBrokersInstrumentProviderConfig(InstrumentProviderConfig, froze
 
     Parameters
     ----------
+    strict_symbology : bool, optional
+        Determines the symbology format used for identifying instruments. If set to True,
+        a strict symbology format is used, as provided by InteractiveBrokers where instrument symbols
+        are detailed in the format `localSymbol=secType.exchange` (e.g., `EUR.USD=CASH.IDEALPRO`).
+        If False, a simplified symbology format is applied, using a notation like `EUR/USD.IDEALPRO`.
+        The default value is False, favoring simplified symbology unless specified otherwise.
     build_options_chain: bool (default: None)
         Search for full option chain. Global setting for all applicable instruments.
     build_futures_chain: bool (default: None)
@@ -112,6 +118,7 @@ class InteractiveBrokersInstrumentProviderConfig(InstrumentProviderConfig, froze
             ),
         )
 
+    strict_symbology: bool = False
     load_contracts: frozenset[IBContract] | None = None
     build_options_chain: bool | None = None
     build_futures_chain: bool | None = None
--- a/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
+++ b/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
@@ -128,19 +128,23 @@ def contract_details_to_ib_contract_details(details: ContractDetails) -> IBContr
 
 def parse_instrument(
     contract_details: IBContractDetails,
+    strict_symbology: bool = False,
 ) -> Instrument:
     security_type = contract_details.contract.secType
+    instrument_id = ib_contract_to_instrument_id(
+        contract=contract_details.contract,
+        strict_symbology=strict_symbology,
+    )
     if security_type == "STK":
-        return parse_equity_contract(details=contract_details)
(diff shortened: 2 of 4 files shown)


rsmb7z:
**Python 3.12 (x64 ubuntu-latest)** failing because of runners disk space not available.

cjdsellers:
@rsmb7z I think this solution is looking good. Also, the disk space issue for Python 3.12 runners should be solved now if you want to rebase.

@benjaminsingleton and @fhill2 any further thoughts or comments on this one?

cjdsellers:
One thing I like about this solution is that the flag allows us to continue testing with both options.

I'm anticipating there may be further standardizations to come due Databento symbology, for instance OSS symbology for equity options https://docs.databento.com/examples/options/equity-options-introduction/overview

Which iirc IB also uses?

rsmb7z:
> One thing I like about this solution is that the flag allows us to continue testing with both options.
> 
> I'm anticipating there may be further standardizations to come due Databento symbology, for instance OSS symbology for equity options https://docs.databento.com/examples/options/equity-options-introduction/overview
> 
> Which iirc IB also uses?

InteractiveBrokers does not employ a proprietary symbology, which may give the impression of complexity compared to other available adapters. This is primarily due to the diverse range of securities it supports globally. The `IBContract.localSymbol` represents the unique symbol registered with the exchange, aligning with `Databento.raw_symbol`.
Regarding venue identification, InteractiveBrokers `IBContract.exchange` represents `ISO10383.ACRONYM`, while `Databento.exchange` utilizes `ISO10383.OPERATING_MIC`. This allows for the use of either MIC or ACRONYM as the venue in `Nautilus.InstrumentId`, though it requires maintaining parsing logic. To simplify, I propose offering a configurable option for this setting, with a minimum requirement that venues exclude spaces and dots.

However, symbology discrepancies may arise due to several factors:
- Exchange parsing necessitates access to the ISO10383 MIC list, available at [ISO20022](https://www.iso20022.org/sites/default/files/ISO10383_MIC/ISO10383_MIC.csv).
- Not all exchanges are covered by ISO10383, indicating that venue codes should not be limited to four characters (e.g., **IDEALPRO**).
- Symbology updates (e.g., old futures symbology **YM JUN 23** to new **YMH4**) may not be promptly reflected by brokers or data providers.

In the US, the broad array of securities and the adoption of updated naming conventions are effectively managed by a simplified symbology approach (currently implemented). However, for other regions undergoing transition to these conventions or some other reasons, implementing specific patterns may not be feasible. A scalable solution is to allow for the registration/loading of mappings for each Nautilus `InstrumentProvider`, where the parsing mechanism first checks the mapping before defaulting to the standard implementation.

Should there be interest, I am prepared to submit a demonstration PR in the coming weeks. This would aim to ensure seamless integration of Nautilus `InstrumentId` with both InteractiveBrokers and Databento, demonstrating the practical application of the proposed solutions.

benjaminsingleton:
All looks good to me!

And I'd definitely be interested to see this:

> Should there be interest, I am prepared to submit a demonstration PR in the coming weeks. This would aim to ensure seamless integration of Nautilus InstrumentId with both InteractiveBrokers and Databento

fhill2:
It's looking good to me.

The contract list I used to find the additional test cases, parse `ib_contract_to_instrument_id_strict_symbology` `instrument_id_to_ib_contract_strict_symbology` roundtrip successfully for all contracts. This was the same behaviour in my original PR. 

I'm interested to see the future developments, eg seamless integration across InteractiveBrokers / Databento and to see your solutions to it @rsmb7z

PR #1490, opened 2024-02-04
Implement Position component in Rust with all pnl/return calculations and tests

filipmacek:
# Pull Request
 
- full Rust port of `Position` component with pnl/returns calculation
- tests on Rust and Python side


cjdsellers (approved):
We'll have to reorganize the dependencies to get `Position` back into `model` I think, but this can be done post merge.

PR #1496, opened 2024-02-08
Optimize PyO3 `logger_log` call

twitu:
# Pull Request

Reduce allocations in the PyO3 `logger_log` call. The intention is to make it comparably performant to cython log calls and hence replace cython logger initialization with pyo3 logger initialization

## Type of change

- [x] Perf

CI tests. Needs more perf testing


twitu:
Based on results in https://github.com/nautechsystems/nautilus_trader/pull/1500, the changes don't make a significant enough performance improvement to par with Cython. As such the additional complexity of is not necessary.

PR #1499, opened 2024-02-14
Fix InteractiveBrokers historical bar data bug

benjaminsingleton:
# Pull Request

User "d2b" in Discord reported the following error:

```
Error on `_run_internal_msg_queue`: AttributeError("'coroutine' object has no attribute 'keywords'")
```

The issue originates from a bug in the `historicalDataUpdate` EWrapper override method of the `InteractiveBrokersClientMarketDataMixin` class. The root cause  is the incorrect usage of a `functools.partial` object, which encapsulates the `subscribe_historical_bars` function. The `functools.partial` was being called rather than simply accessing the `keywords` attribute.

## Type of change

Tested locally.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/common.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/common.py
@@ -14,6 +14,7 @@
 # -------------------------------------------------------------------------------------------------
 
 import asyncio
+import functools
 from abc import ABC
 from abc import abstractmethod
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


cjdsellers:
Looks like CI only failing due to a temporary GitHub runner glitch:
```
Error output:
No Python at '"C:\hostedtoolcache\windows\Python\3.11.7\x64\python.exe'
```

This normally clears after a few hours - I'll re-run the failing jobs later.

PR #1500, opened 2024-02-14
New logger perf test

ayush-sb:
# Pull Request

Temporary PR, only for testing new pyo3 logger. Will close after benchmarking.


twitu:
According to the performance tests Pyo3 logging is significantly worse off than Cython logging (between 20-60% slower). In such case, to emphasize performance logging must continue through Cython. Hopefully as more logic gets ported to Rust it becomes more realistic to fully move to Pyo3 based logging.

Still it will be useful to have a flag to easily switch between the two because rust side logging won't work when logging is initialized through python due to how static variables work during linking.

```
----------------------------------------------- benchmark: 1 tests -----------------------------------------------
Name (time in ms)          Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------
test_logging          111.2739  115.6815  113.4472  1.2138  113.3397  1.1149       2;1  8.8147      10           2
------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
================================================== 1 passed in 2.64s ===================================================

Pyo3 without bypass

---------------------------------------------- benchmark: 1 tests ---------------------------------------------
Name (time in ms)         Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------
test_logging          72.1569  73.7408  72.7025  0.4493  72.6305  0.5670       2;0  13.7547      10           2
---------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
================================================================================= 1 passed in 1.71s =================================================================================

Pyo3 with bypass

---------------------------------------------- benchmark: 1 tests ---------------------------------------------
Name (time in ms)         Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------
test_logging          60.5715  61.8973  60.9568  0.4172  60.8317  0.3002       2;1  16.4051      10           2
---------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
================================================================================= 1 passed in 1.42s =================================================================================

Cython with and without bypass
```

The results are a bit sus though because Cython shows no difference between bypassed and non-bypassed modes, while pyo3 shows significant difference.

PR #1502, opened 2024-02-16
Implement AverageTrueRange in Rust

rsmb7z:
# Pull Request

Implement AverageTrueRange in Rust with pyo3 tests.

## Type of change

Implemented pyo3 tests from existing implementation.


cjdsellers:
Thanks again @rsmb7z 

One small note is it'll cut down on the feedback loop of you run:
```
pre-commit install
```
Which will install a git hook to run automatically on every commit (and before pushing the PR).

PR #1505, opened 2024-02-19
Modfiy CSVBarDataLoader to allow passing any pandas arguments to the read_csv() method

rterbush:
read_csv() method.

# Pull Request

Change parameters for the load method in the CSVBarDataLoader to allow passing any pandas parameters to read_csv() and thereby make this much more flexible for different CSV file formats.

Not clear if this is a direction that the Nautilus team wants to go. If this is an acceptable change, it raises the question if we need a CSVTickDataLoader and CSVBarDataLoader. A CSVDataLoader with this flexibility could be used for any type/format of CSV data.

## Type of change

Limited testing on my end as I work to port some older data import scripts to latest API.

### DIFF
--- a/nautilus_trader/persistence/loaders.py
+++ b/nautilus_trader/persistence/loaders.py
@@ -14,7 +14,7 @@
 # -------------------------------------------------------------------------------------------------
 
 from os import PathLike
-
+from typing import Any
 import pandas as pd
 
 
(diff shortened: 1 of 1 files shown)


rterbush:
If wrangler.process() expects timestamp index then this will also need another slight change to make that index conversion.

cjdsellers:
Hey @rterbush

Thanks so much for this contribution!

Certainly making the CSV loaders more flexible is a good move, by adding optional parameters. We'll probably head the way of more specific loaders per integration (for example Databento, although they're DBN specific right now), but it will also be a good idea to keep some more generic and flexible loaders around also (such as your proposal here).

There's currently some test fails, are you able to see the CI output logs?

Cheers,

Chris

cjdsellers:
A `CSVDataLoader` may be appropriate, but then we'd want to make sure the schema of the pandas dataframe matches the intended Nautilus data type conversion. So something needs to be configurable here, either by using separate classes, or options through parameters.

Lets discuss further once you're back and have had a chance to look at CI and think on this further :pray:.

cjdsellers:
Added these changes in dbf776b61bff18586939cc055a45833b83e35c77 with credit to @rterbush, many thanks :pray:

Issue #1461, opened 2024-01-21
Flush logging when terminating process to avoid truncating logs


twitu:
# Feature Request

The current logging setup does not explicitly flush the logging setup when the process ends. This can cause the last few log lines to be truncated.

This can be solved using a guard pattern. A guard is created when the logging is initialized. This is typical done at the top level at the beginning of the application. The drop function for the guard flushes the logging setup and cleans up resources like threads by joining on them.

Since nautilus engine starts it's control flow in python. The guard must also derive `#[pyclass]`. And it's usage will look something like this.

```python
import nautilus.... logging

def main():
    guard = logger_init()
    # run engine
    return # guard is dropped here
```

cjdsellers:
Now completed https://github.com/nautechsystems/nautilus_trader/commit/4bdf54b1418aefc83845f0863def77d5d4c797f2.

PR #1519, opened 2024-02-27
Add Log Guard to avoid truncated logs

ayush-sb:
# Pull Request

WIP: Add a log guard to flush logs when dropped https://github.com/nautechsystems/nautilus_trader/issues/1461


twitu (commented):
LogGuard looks solid to me. You'll also want to export it through FFI as a Cython class as well, since currently logger is being initialized through Cython.

Also perhaps some of the tests are failing because the guard is not being assigned to a variable and hence dropped.

ayush-sb:
Closed because recent logger refactor caused too many merge conflicts.

PR #1531, opened 2024-03-10
Cash account pyo3 Rust Cython conversion

filipmacek:
# Pull Request
- some small fixes across the rust codebase
- refactored Rust pyo3 bindings function `from_dict` and `to_dict` for `CashAccount` what needed to be calculated manually
- main goal is to correctly convert object from/to Rust and Cython with dict approach and test 1:1 conversion


nautilus_core/accounting/src/account/cash.rs
```diff
@@ -158,7 +158,7 @@ impl DerefMut for CashAccount {
 
 impl PartialEq for CashAccount {
     fn eq(&self, other: &Self) -> bool {
-        self.id == other.id
+        self.id == other.id && self.events == other.events
```
cjdsellers, line 161: An account, like an order etc - should only be compared with its `id`.

nautilus_core/model/src/events/account/state.rs
```diff
@@ -93,7 +93,13 @@ impl PartialEq for AccountState {
     fn eq(&self, other: &Self) -> bool {
```
cjdsellers, line 93: As above, we only want to compare just the ID (like orders and other domain aggregates which contain state), as there are operations which just need to fetch the correct account and not further compare its state.

nautilus_core/model/src/python/types/balance.rs
```diff
@@ -61,17 +63,32 @@ impl AccountBalance {
...
     }
 
     #[pyo3(name = "to_dict")]
     pub fn py_to_dict(&self, py: Python<'_>) -> PyResult<PyObject> {
         let dict = PyDict::new(py);
         dict.set_item("type", stringify!(AccountBalance))?;
-        dict.set_item("total", self.total.to_string())?;
-        dict.set_item("locked", self.locked.to_string())?;
-        dict.set_item("free", self.free.to_string())?;
+        dict.set_item("total", self.total.as_f64().to_string())?;
```
cjdsellers, line 89: If we convert to floats, then we lose implicit precision with strings - which doesn't seem like a good idea?

tests/unit_tests/model/objects/test_balance_pyo3.py
```diff
@@ -44,9 +44,9 @@ def test_account_balance_to_from_dict():
     assert account_balance == AccountBalance.from_dict(result_dict)
     assert result_dict == {
         "type": "AccountBalance",
-        "free": "1500000.00 USD",
-        "locked": "25000.00 USD",
-        "total": "1525000.00 USD",
+        "free": "1500000",
```
cjdsellers, line 47: The currency code is probably redundant in the string, but we should retain the implicit precision. The `to_str` method provides this (then we would get e.g.  `"1500000.00"`). The other choice is to use floats, which I don't think would be a better option?

nautilus_core/accounting/src/account/cash.rs
filipmacek: Yes, I only added this becase tests we passing when i was comparing two different pyo3 CashAccount objects with different events, and `id` was only checked and it was equal, will revert this

cjdsellers (approved):

PR #1532, opened 2024-03-10
Improve IB client's connectivity resilience and component lifecycle

benjaminsingleton:
# Pull Request

Refer to #1533 and #1534 for details.

## Type of change

- Existing test suite passes and some new tests were added as well. Still, more live testing should be done by the community.


benjaminsingleton:
@rsmb7z In this PR, I removed the code associated with probing for connectivity because disconnect events now trigger a full reconnect. I wanted to verify with you before introducing this change. I'm not sure I've encountered a live example where probing for connectivity is required, but I'm sure you/others have. If you could share some thoughts, I'd find it helpful.

rsmb7z:
> @rsmb7z In this PR, I removed the code associated with probing for connectivity because disconnect events now trigger a full reconnect. I wanted to verify with you before introducing this change. I'm not sure I've encountered a live example where probing for connectivity is required, but I'm sure you/others have. If you could share some thoughts, I'd find it helpful.

Using the Client in historical mode negates the need for probing to verify connectivity. However, understanding that there’s no alternative method to check for connectivity issues between the Gateway/TWS and the IB servers is crucial. In such scenarios, even though the socket remains connected, detecting disconnections becomes challenging. Although there are signals like messages with reqId=-1, they are not entirely reliable and may not always indicate connectivity problems accurately. Given this context, it might be worthwhile to consider removing the probing mechanism and evaluating any improvements with the latest API version.

rsmb7z (approved):

tests/integration_tests/adapters/interactive_brokers/conftest.py
```diff
@@ -110,23 +134,27 @@ def data_client(mocker, data_client_config, venue, loop, msgbus, cache, clock):
...
         msgbus=msgbus,
         cache=cache,
         clock=clock,
     )
-    client._client.start()
+    client._client._is_ib_connected.set()
+    client._client._connect = AsyncMock()
+    client._client._eclient = MagicMock()
+    client._client._account_ids = {"DU123456,"}
+    # client._client.start()
```
cjdsellers, line 148: Did we mean to leave these in? i.e. is the test still testing what we want it to?

tests/integration_tests/adapters/interactive_brokers/conftest.py
```diff
@@ -136,16 +164,18 @@ def exec_client(mocker, exec_client_config, venue, loop, msgbus, cache, clock):
...
         clock=clock,
     )
-    client._client.start()
-    client._client.process_managed_accounts(accounts_list="DU123456,")
-    client._client.process_next_valid_id(order_id=1)
+    client._client._is_ib_connected.set()
+    client._client._connect = AsyncMock()
+    client._client._eclient = MagicMock()
+    client._client._account_ids = {"DU123456,"}
+    # client._client.start()
```
cjdsellers, line 178: As above

tests/integration_tests/adapters/interactive_brokers/client/test_client_error.py
```diff
@@ -16,10 +16,13 @@
...
 from unittest.mock import Mock
 
+import pytest
 
+
+@pytest.mark.asyncio
 def test_ib_is_ready_by_notification_1101(ib_client):
     # Arrange
-    ib_client._is_ib_ready.clear()
+    ib_client._is_ib_connected.clear()
```
cjdsellers, line 25: We could probably make a property or method to avoid "private" access, even in tests.

cjdsellers (approved):

AmosG:
any idea when this would be merged?
I think there is a bug with the current implmentation 
that this might solve

PR #1551, opened 2024-03-19
Fix Interactive Brokers client start

benjaminsingleton:
# Pull Request

Hotfix related to #1532. The `InteractiveBrokersClient` would automatically startup because there was a call to `_reset()` in the constructor, but that is generally considered an antipattern, so now it needs to be explicitly started.

## Type of change

- All tests pass


cjdsellers (approved):
Thanks for adding all of the type annotations too!

PR #1560, opened 2024-03-24
Conversion between instrument type and PyObject refactor

filipmacek:
# Pull Request

 - created new util function `convert_pyobject_to_instrument_type`  that gives enum `InstrumentType` for given `PyObject` based on the `instrument_type` property inside class
 - refactor various python function that uses this pattern (in accounting + position)
 - moved `convert_instrument_to_pyobject` from `adapters` crate to `model` crate


cjdsellers (approved):
:ok_hand:

PR #1568, opened 2024-03-29
Refine ts_init Query to Exclude Start Timestamp

rsmb7z:
# Pull Request

Updated the `ts_init` query condition to exclude the `start` timestamp (ts_init > {start_ts}), mitigating duplicate records in sequential queries. For flexibility, an optional flag for including/excluding the start timestamp maybe added in case keeping current use case is necessary.

### DIFF
--- a/nautilus_trader/persistence/catalog/parquet.py
+++ b/nautilus_trader/persistence/catalog/parquet.py
@@ -541,7 +541,7 @@ class ParquetDataCatalog(BaseDataCatalog):
 
         if start:
             start_ts = dt_to_unix_nanos(start)
-            conditions.append(f"ts_init >= {start_ts}")
+            conditions.append(f"ts_init > {start_ts}")
         if end:
             end_ts = dt_to_unix_nanos(end)
             conditions.append(f"ts_init <= {end_ts}")
--- a/tests/unit_tests/persistence/test_catalog.py
+++ b/tests/unit_tests/persistence/test_catalog.py
@@ -70,7 +70,7 @@ def test_catalog_query_filtered(
     assert len(trades) == 121
 
     trades = catalog_betfair.trade_ticks(start=1576875378384999936)
-    assert len(trades) == 121
+    assert len(trades) == 120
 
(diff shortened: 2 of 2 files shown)


nautilus_trader/persistence/catalog/parquet.py
```diff
@@ -541,7 +541,7 @@ def _build_query(
 
         if start:
             start_ts = dt_to_unix_nanos(start)
-            conditions.append(f"ts_init >= {start_ts}")
+            conditions.append(f"ts_init > {start_ts}")
```
cjdsellers, line 544: Surely we want to be inclusive of the start?

rsmb7z: Having end inclusive I don't think start should be included.

cjdsellers: Whats your reasoning here? I think this would be surprising behaviour to be start exclusive - normally the choice is `end` inclusive or exclusive (but `start` is always inclusive for most APIs).

cjdsellers: One reason we wouldn't want to make this change is Databento is `start` inclusive and `end` exclusive.

rsmb7z: I'm open to adopting the standard of making the end exclusive as well. My main concern arises when both start and end points are inclusive, as that's where potential issues of duplicate record.

cjdsellers: In that case, we just need one end exclusive - so I think better for that to be `end` for consistency.

It can still be surprising to some users though to have exclusive endpoints in ranges.

rsmb7z: Understood, it seems this issue wouldn't typically arise during normal use. Additionally, I've identified a similar situation elsewhere in the catalog. To sidestep the complexity of adding flags and the potential impact on other users, I'll close this matter for now.

cjdsellers: You make a good point about duplicate records _if_ the catalog is repeatedly queried, as start and end are both inclusive. I'll have a deeper think on this and look at the use cases.

PR #1570, opened 2024-03-30
bybit: add trade spot and trade option websocket msgspec schemas

davidsblom:
# Pull Request

Add msgspec schema's for trade messages for spot and option instruments, due to the `L` field being unique to futures and `id` being unique to options.

## Type of change

Locally listening to trade and quote ticks for the Bybit exchange.

### DIFF
--- a/nautilus_trader/adapters/bybit/data.py
+++ b/nautilus_trader/adapters/bybit/data.py
@@ -15,6 +15,7 @@
 
 import asyncio
 from collections import defaultdict
+from functools import partial
 
 import msgspec
 import pandas as pd
@@ -138,7 +139,7 @@ class BybitDataClient(LiveMarketDataClient):
         for instrument_type in instrument_types:
             self._ws_clients[instrument_type] = BybitWebsocketClient(
                 clock=clock,
-                handler=lambda x: self._handle_ws_message(instrument_type, x),
+                handler=partial(self._handle_ws_message, instrument_type),
                 base_url=ws_urls[instrument_type],
                 api_key=config.api_key or get_api_key(config.testnet),
                 api_secret=config.api_secret or get_api_secret(config.testnet),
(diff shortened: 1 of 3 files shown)


nautilus_trader/adapters/bybit/data.py
```diff
@@ -131,22 +133,24 @@ def __init__(
...
         # WebSocket API
         self._ws_clients: dict[BybitInstrumentType, BybitWebsocketClient] = {}
+        self._decoders: dict[str, dict[BybitInstrumentType, msgspec.json.Decoder]] = defaultdict(
+            dict,
+        )
         for instrument_type in instrument_types:
             self._ws_clients[instrument_type] = BybitWebsocketClient(
                 clock=clock,
-                handler=lambda x: self._handle_ws_message(instrument_type, x),
+                handler=partial(self._handle_ws_message, instrument_type),
```
davidsblom, line 142: strangely, this partial is needed instead of the lambda when subscribing to both spot and linear instruments. to me, not clear why. @cjdsellers  any ideas?

davidsblom: with the lambda, only 1 instrument type ticks are created.

cjdsellers:
Thanks for the help @davidsblom!

I think even more efficient if we just define some optional fields for the trade message, rather than the entirely separate messages per instrument. I'll merge as is though if this was working for you, and we can always refactor.

davidsblom:
Great! Optional fields are needed much cleaner.

The important fix is the use of the partial instead of a lambda. The lambda functions were being overwritten somehow when using multiple product types. Any idea how that is even possible?

davidsblom:
Found the answer here: https://stackoverflow.com/questions/50298582/why-does-python-asyncio-loop-call-soon-overwrite-data

When the lambda function is called, it will use the values bound by those names in the outer scope at that time of the call, not the values they had when the lambda was defined.

PR #1581, opened 2024-04-06
StopLimitOrder pyo3 Rust Cython conversion

filipmacek:
# Pull Request

- exported rust `StopLimitoOrder` with pyo3
- unified Rust and Cython Stop Limit order data type
- created `test_stop_limit_order_pyo3.py` test with added test `test_pyo3_cython_conversion` to check both dict matching and object equivalence


nautilus_core/model/src/orders/stop_limit.rs
```diff
@@ -93,7 +94,7 @@ impl StopLimitOrder {
                 instrument_id,
                 client_order_id,
                 order_side,
-                OrderType::LimitIfTouched,
+                OrderType::StopLimit,
```
cjdsellers, line 97: Good catch.

nautilus_core/model/src/orders/base.rs
```diff
@@ -968,8 +968,8 @@ impl OrderCore {
     }
 
     #[must_use]
-    pub fn init_event(&self) -> Option<&OrderEvent> {
-        self.events.first()
+    pub fn init_event(&self) -> Option<OrderEvent> {
+        self.events.first().cloned()
```
cjdsellers, line 972: Did this make things easier for the caller? (they can always clone too)

nautilus_trader/core/nautilus_pyo3.pyi
```diff
@@ -952,6 +952,7 @@ class LimitOrder:
     def is_primary(self) -> bool: ...
     @property
     def is_spawned(self) -> bool: ...
+    @classmethod
```
cjdsellers, line 955: Per discussion, we can't create an order from a dict snapshot.

nautilus_trader/model/orders/stop_limit.pyx
```diff
@@ -289,6 +292,42 @@ cdef class StopLimitOrder(Order):
             f"{emulation_str}"
         )
 
+    @staticmethod
+    cdef StopLimitOrder from_pyo3_c(pyo3_order):
```
cjdsellers, line 296: Per above.

cjdsellers (approved):

Issue #632, opened 2022-05-04
Improve commission handling 


limx0:
# Feature Request

This was raised on discord and warrants an issue. Currently we assume a flat maker/taker %  rate of commission, but there are quite a few different types of ways a broker or exchange may charge commission. 

A couple that come to mind: 
- flat fee (possibly tiered based on turnover or similar)
- % of notional  (possibly tiered based on turnover or similar)
- % of notional with a minimum per trade or period.

It looks like a `CommissionModel` used to exist [here](https://github.com/nautechsystems/nautilus_trader/blob/9c9d35badd73a2d746fc40c29c9b9810da9b97f6/nautilus_trader/common/market.pyx) which would be a good place to start. 

Ideally we would incorporate some sort of `registration` method where downstream users can easily register new models if desired.

rsmb7z:
I was looking for some solution on commission handling as currently only percentage based model is supported. I found below, it is nicely done and covers most of the possibilities. Can we implement this please, specially IB from here.

https://www.quantconnect.com/docs/v2/writing-algorithms/reality-modeling/transaction-fees/supported-models
https://github.com/QuantConnect/Lean/blob/5611b5508ada70dda96cc2b917e9475400f781e3/Common/Orders/Fees/InteractiveBrokersFeeModel.cs

limx0:
Perfect. @cjdsellers loves C#

limx0:
On a serious note - great find @rsmb7z , that will serve as a great reference and agreed, we should basically just implement the same thing.

limx0:
And just dumping some thoughts that I had the other day for this ticket. 

We should likely have some sort of backtest module that handles commission adjustments on a portfolio level. Many venues have discounts or rebates for volume, and we will want to model this in our backtesting. Individual trades should likely still "pay" the full commission, but we may want some model that adjusts or refunds commission at the portfolio level at the end of some period (say monthly or whatever is applicable) as a backtest is running.

cjdsellers:
Modeling fee tiers could also be part of the commission model, especially since its likely to be a plug in component to a `SimulatedExchange`. So likely better to handle there than the `Portfolio`, even though I agree its a portfolio level concern.

cjdsellers:
> Perfect. @cjdsellers loves C#

Ah, the memories...

dkharrat:
I just ran into this issue myself :)
Some brokers have a fixed commission, so supporting this model would be very helpful. Any plans to address this soon?

[...]

PR #1584, opened 2024-04-09
Implement CommissionModel

rsmb7z:
# Pull Request

This PR introduces an enhancement to the CommissionModel, as outlined in issue #632, maintaining the default behavior while integrating a fixed price model. It further establishes a foundation for developing adapter-specific commission models.

## Type of change

Existing tests passed and new tests added.

### DIFF
--- a/nautilus_trader/adapters/sandbox/execution.py
+++ b/nautilus_trader/adapters/sandbox/execution.py
@@ -22,6 +22,7 @@ import pandas as pd
 from nautilus_trader.backtest.exchange import SimulatedExchange
 from nautilus_trader.backtest.execution_client import BacktestExecClient
 from nautilus_trader.backtest.models import FillModel
+from nautilus_trader.backtest.models import InstrumentSpecificPercentCommissionModel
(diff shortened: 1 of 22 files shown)


cjdsellers:
Just following up on this.

I've completed a rename of `CommissionModel` to `FeeModel`, just because commissions are only one component of a fee model - I left the `get_commission` method name unchanged though.

I also think you named `InstrumentSpecificPercentCommissionModel` well, however the direction we're taking the platform will be to take the fees off the instrument definitions - as they are more applicable at the venue listing level.

So I renamed that to `MakerTakerFeeModel`.

We could expand the `FeeModel` to include funding rates/fees (or that could be a separate model), but you did a great job of wiring this through, making the next step of removing fees from instruments much easier :pray:.

cjdsellers:
I also took your suggestion of renaming `FixedCommissionModel` to `FixedFeeModel`, I think its cleaner.

Now on `develop` from c7c1a9efa6e9b4e5ffcaa6948b9599f0cb62731b.

PR #1586, opened 2024-04-10
Fix and Refine TradeTickDataWrangler.process_bar_data

rsmb7z:
# Pull Request

Followup on previous PR to fix call to `_create_side_if_not_exist`.

## Type of change

### DIFF
--- a/nautilus_trader/persistence/wranglers.pyx
+++ b/nautilus_trader/persistence/wranglers.pyx
@@ -572,32 +572,24 @@ cdef class TradeTickDataWrangler:
         # Ensure index is tz-aware UTC
         data = as_utc_index(data)
 
-        # Determine the Aggressor Side based on Close vs Open
-        if "side" not in data and "buyer_maker" not in data:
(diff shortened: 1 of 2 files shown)


cjdsellers:
Something we could consider is adding `default_volume: float = 1_000_000.0` similar to the quote tick wrangler. Often trades can be for a small size which doesn't represent the liquidity which may be available nearly immediately after the effect of the trade.

Alternatively, we could improve the fill dynamics in the order book to slip to the next price level if top-of-book size is exceeded?

rsmb7z:
> Alternatively, we could improve the fill dynamics in the order book to slip to the next price level if top-of-book size is exceeded?

I think something using config flag etc for the MatchingEngine would be more appropriate to ignore or relax the liquidity conditions for backtesting purpose so this way original volume is not lost and still accessible within Strategy.
Someone still wanting to use custom volume can modify in the input dataframe for `TradeTickDataWrangler`. The case for `QuoteTickDataWrangler` is bit different because there `volume` is not always available.

PR #1597, opened 2024-04-20
Fix method from_pyo3_c in Currency pair for lot_size conversion 

filipmacek:
# Pull Request

- in `CurrencyPair` class `lot_size` field should use `Quantity.from_raw_c` for the correct conversion


cjdsellers:
Good catch.

cjdsellers (commented):
LGTM.

PR #1599, opened 2024-04-20
Add from_str pyo3 method in identifier_for_python macro

filipmacek:
# Pull Request

- added `from_str` static pyo3 function in `identifier_for_python` macro and added typings  in `nautilus_pyo3.pyi` file


cjdsellers (approved):
I wonder if this might be redundant with the constructor though?

PR #1600, opened 2024-04-20
Finish Instrument trait and implement sqlx trait FromRow for instruments

filipmacek:
# Pull Request

- finalize `Instrument` Rust trait so that it contains its union from all instrument struct fields and can later represent SQL table `instrument`
- implement `FromRow` sqlx trait for Postgres that will help with conversion from Postgres to native rust type


cjdsellers (approved):
I'll merge this but then move the `FromRow` impls and database dependencies out of `nautilus_core` and into `infrastructure` for now.

PR #1608, opened 2024-04-24
Betfair bugfix: Reload cache as a dict of venue and client order ids

imemo88:
# Pull Request
Reload venue order and client order id cache correctly after a reconnection event. Please see #betfair (discord) where the issue highlighted a bit better

## Type of change
[x] Bug fix

In production


cjdsellers:
Hi @imemo88

Thanks so much for the fix!

Looks like we just have pre-commit failing so far. You can run pre-commit with:

```
make pre-commit
```
or
```
pre-commit run --all-files
```
You can also run the following to install the git hook to ensure it always runs on local commits too:

```
pre-commit install
```
It looks like some auto-formatting from `black` will occur, then you can push another commit.

limx0 (approved):

PR #1612, opened 2024-04-26
Add parsable raw string representation of Money object

filipmacek:
# Pull Request

Currently `to_str` function of `Money` object returns a string representation of a number with an underscore. For example, we return `10_000.00 USD` instead of more native `10000.00 USD`(and easier to parse by other systems).  
This looks right in logging where its easier to see bigger numbers with this separator but poses a problem for parsing with Rust where we cannot deserialize it back to the `Money` type. A lot of these `from_dict` functions rely on `serde_json` to convert it back to native Rust type, and they fail on `Money` type deserialization
So `to_raw_str` was created to represent this more parsing-friendly format and it was added across `to_dict` method of instruments


filipmacek:
Closing, as it was resolved by Rust parsing fix

PR #1616, opened 2024-04-28
Fix Portfolio registration for IB

rsmb7z:
# Pull Request

Fixes portfolio bug for InteractiveBrokers.

## Type of change


cjdsellers:
Thanks for the fix @rsmb7z :pray:.

I initially thought using `isinstance` and the `IB_VENUE` venue constant would be better - but then realized we'd be depending on the IB adapter (in the wrong direction).

So I think this is a good solution for now, and still a bit of a temporary hack before we make the portfolio more flexible in this area.

PR #1622, opened 2024-05-02
Fix IBKR reconnection after gateway/TWS disconnection

benjaminsingleton:
# Pull Request

This pull request addresses an issue with the Interactive Brokers (IBKR) reconnection mechanism. Previously, if the connection to the IB gateway or TWS was lost, either due to the nightly restart or a disrupted connection, the system was unable to successfully reestablish the connection. The changes introduced in this PR resolve this bug, ensuring that the connection can be properly restored after such interruptions.

## Type of change

Passes all integration tests. Also worked as anticipated while paper trading.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -135,7 +135,7 @@ class InteractiveBrokersClient(
 
         # ConnectionMixin
         self._connection_attempts: int = 0
-        self._max_connection_attempts: int = int(os.getenv("IB_MAX_connection_attempts", 0))
+        self._max_connection_attempts: int = int(os.getenv("IB_MAX_CONNECTION_ATTEMPTS", 0))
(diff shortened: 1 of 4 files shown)


cjdsellers:
I'm not as familiar with IB and the gateway, but would it be reasonable to bump up that `asyncio.wait_for` timeout to 30 seconds?

rsmb7z (approved):
LGTM.

rsmb7z:
> I'm not as familiar with IB and the gateway, but would it be reasonable to bump up that `asyncio.wait_for` timeout to 30 seconds?

A 15-second interval is acceptable as Docker container downtimes are typically infrequent, and the Gateway restart process itself only takes about 2-3 seconds. Additionally, the nightly reset recovery is generally quicker for live accounts than for paper accounts on the IB side. However, it might be beneficial to make this interval configurable. We could introduce an argument that allows users to adjust the timing based on the specifics of their environment.

PR #1623, opened 2024-05-04
Add from_raw_arrays_to_list method for Bar

rsmb7z:
# Pull Request

Add from_raw_arrays_to_list method for Bar.

## Type of change

### DIFF
--- a/nautilus_trader/model/data.pxd
+++ b/nautilus_trader/model/data.pxd
@@ -168,6 +168,20 @@ cdef class Bar(Data):
     )
 
     @staticmethod
+    cdef list[Bar] from_raw_arrays_to_list_c(
+        BarType bar_type,
+        uint8_t price_prec,
+        uint8_t size_prec,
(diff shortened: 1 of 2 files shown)


cjdsellers:
I suppose this offers some advantages when called from Python to avoid a for loop?

Otherwise I'd imagine performance would be fairly close just calling `from_raw_c` in a loop from Cython?

cjdsellers:
Might be good to add one test too?

rsmb7z:
> I suppose this offers some advantages when called from Python to avoid a for loop?
> 
> Otherwise I'd imagine performance would be fairly close just calling `from_raw_c` in a loop from Cython?

Yes correct, by having loop in Cython and passing the Numpy view is much faster straight from Python. I did some experiment with Wranglers and could be moved to Python offering same performance as currently in Cython. I think OrderBook is pending and then we could remove Cython for Wranglers.

PR #1626, opened 2024-05-04
Fix IBKR resubscribing after disconnection

benjaminsingleton:
# Pull Request

As reported in the Interactive Brokers Discord channel, resubscribing to subscriptions after a reconnect only worked for 1 of multiple subscriptions. This PR fixes that issue and also refactors the lifecycle methods, which are synchronous functions that need to call asynchronous functions. 

## Type of change

Tests pass and validated with live trading.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -365,7 +365,7 @@ class InteractiveBrokersClient(
         """
         if self.is_running:
             self._degrade()
-        if not self._is_ib_connected.is_set():
+        if self._is_ib_connected.is_set():
             self._log.debug("`_is_ib_connected` unset by `_handle_disconnection`.", LogColor.BLUE)
             self._is_ib_connected.clear()
         await asyncio.sleep(5)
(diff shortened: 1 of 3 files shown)


nautilus_trader/adapters/interactive_brokers/client/client.py
```diff
@@ -162,14 +162,14 @@ def _start(self) -> None:
...
         if not self._loop.is_running():
             self._log.warning("Started when loop is not running.")
-            self._loop.run_until_complete(self._startup())
+            self._loop.run_until_complete(self._async_start())
         else:
-            self._create_task(self._startup())
+            self._create_task(self._async_start())
 
-    async def _startup(self):
+    async def _async_start(self):
```
cjdsellers, line 171: Do we have synchronous versions of the start and stop methods? otherwise the `_async` prefix is a little redundant, and usually a postfix when there is also a synch version.

benjaminsingleton: `InteractiveBrokersClient` inherits from `Component`, which implements the `start()`, `stop()`, `reset()` and `resume()` synchronous methods and also defines `_start()`, `_stop()`,  `_reset()` and `_resume` as abstract synchronous methods to be overridden in subclasses. Because `InteractiveBrokersClient` needs to execute async methods, we have to make them work with the synchronous base class methods. Let me know if you were thinking about a different approach.

I can definitely change the method names to follow the postfix convention.

rsmb7z (approved):

benjaminsingleton:
I tested this on live markets and got the expected result, so I'm ready to merge.

PR #1628, opened 2024-05-05
Fix timer cancel when it is already expired

davidsblom:
# Pull Request

Fix timer canceling when it is already expired.

## Type of change

## How has this change been tested

Paper trading with a throttler ran into an issue as discussed on Discord. Error is now resolved.

### DIFF
--- a/nautilus_core/common/src/timer.rs
+++ b/nautilus_core/common/src/timer.rs
@@ -342,9 +342,11 @@ impl LiveTimer {
     /// Cancels the timer (the timer will not generate an event).
     pub fn cancel(&mut self) -> anyhow::Result<()> {
         debug!("Cancel timer '{}'", self.name);
-        if let Some(sender) = self.canceler.take() {
-            // Send cancellation signal
-            sender.send(()).map_err(|e| anyhow::anyhow!("{:?}", e))?;
+        if !self.is_expired.load(atomic::Ordering::SeqCst) {
+            if let Some(sender) = self.canceler.take() {
+                // Send cancellation signal
+                sender.send(()).map_err(|e| anyhow::anyhow!("{:?}", e))?;
+            }
         }
         Ok(())
     }


nautilus_core/common/src/timer.rs
```diff
@@ -342,9 +342,11 @@ impl LiveTimer {
     /// Cancels the timer (the timer will not generate an event).
     pub fn cancel(&mut self) -> anyhow::Result<()> {
         debug!("Cancel timer '{}'", self.name);
-        if let Some(sender) = self.canceler.take() {
-            // Send cancellation signal
-            sender.send(()).map_err(|e| anyhow::anyhow!("{:?}", e))?;
+        if !self.is_expired.load(atomic::Ordering::SeqCst) {
```
cjdsellers, line 345: Does this actually fix the errors? because `self.canceler.take()` will take the canceler out of the option leaving behind `None` - so when its called again `sender` won't be Some.

cjdsellers: I just realized that the select could exit with `is_expired` set `True` before the timer is canceled, so also checking this `is_expired` value before going for the canceler would cover this.

davidsblom: Hmm strange. For me the error doesn’t happen again, but have not seen the actual error message yet. I’ll dive deeper

davidsblom: Ah indeed!

davidsblom: I’ll make the change and test it. Somewhat later today.

cjdsellers: I think your impl is good as it is - especially if you don't see the errors anymore.

davidsblom: Okay sure. Feel free to merge then if this is fine.

PR #1629, opened 2024-05-05
Fix venue_order_id issue for Sandbox

rsmb7z:
# Pull Request

Fixes `venue_order_id` error on order event.

```
ValueError(The 'self.venue_order_id' <class 'nautilus_trader.model.identifiers.VenueOrderId'> of PAXOS-1-001 was not equal to the 'event.venue_order_id' <class 'nautilus_trader.model.identifiers.VenueOrderId'> of PAXOS-1-002)
  File "nautilus_trader\\execution\\engine.pyx", line 1007, in nautilus_trader.execution.engine.ExecutionEngine._apply_event_to_order
  File "nautilus_trader\\model\\orders\\base.pyx", line 946, in nautilus_trader.model.orders.base.Order.apply
  File "nautilus_trader\\core\\correctness.pyx", line 306, in nautilus_trader.core.correctness.Condition.equal
```

## Type of change

### DIFF
--- a/nautilus_trader/adapters/sandbox/execution.py
+++ b/nautilus_trader/adapters/sandbox/execution.py
@@ -109,7 +109,7 @@ class SandboxExecutionClient(LiveExecutionClient):
         )
         self.exchange = SimulatedExchange(
             venue=sandbox_venue,
-            oms_type=oms_type,
+            oms_type=OmsType.HEDGING if oms_type == OmsType.NETTING else OmsType.NETTING,
             account_type=self._account_type,
             base_currency=self._currency,
             starting_balances=[self.balance.free],


nautilus_trader/adapters/sandbox/execution.py
```diff
@@ -109,7 +109,7 @@ def __init__(
         )
         self.exchange = SimulatedExchange(
             venue=sandbox_venue,
-            oms_type=oms_type,
+            oms_type=OmsType.HEDGING if oms_type == OmsType.NETTING else OmsType.NETTING,
```
davidsblom, line 112: So the oms_type is changed to hedging if netting is used? Feels a bit off but maybe I understand this wrong.

davidsblom (commented):
Thanks for looking into this as well!

nautilus_trader/adapters/sandbox/execution.py
rsmb7z: I believe either one of `SimulatedExchange` or `ExecutionClient` should be generating the `VenueId`. When we use `NETTING` for both None of them is generating IDs and alternatively both generate the IDs which doesn't work.

cjdsellers:
I'm not sure this is the right approach to take to fix the root cause?

The venue order ID should be generated by the matching engine only from `_generate_venue_position_id`, potentially running in sandbox mode creates a race condition here because we have this real time element external to the matching engine or any backtest engine loop.

Needs a little more investigating.

rsmb7z:
@cjdsellers Yes you are right, even though it reduced the errors in my case with above but they are still there. Will update if I find any further clue.

davidsblom:
Gave it another try as well in this pull request #1635. @rsmb7z would be interesting to see if it fixes your issue as well.

rsmb7z:
Fixed [here](https://github.com/nautechsystems/nautilus_trader/commit/ad9a515874da677c5acd7815056dfce21b926578).

PR #1631, opened 2024-05-06
Add psql cache database orders

filipmacek:
# Pull Request

- removed `sqlx` from root cargo project dependency (now its only present in the `infrastructure` crate)
- created Rust test utility function `entirely_equal` that accepts two structs with `Serialize` serde trait implemented to check for full string equality ( we want to check that all fields are the some and not use native `PartialEq` trait which checks only ids)
- rename enum `OrderEvent` to `OrderEventAny` to follow current established pattern with enum container for aggregation of similar types ( like order, events, instruments)
- added new Postgres database queries: `add_order` and `load_order`
- added two new Postgres tables in the schema `order` and `order_event`
- implemented `OrderEvent` trait for all order events to represent Postgres table `order_event`
- implemented conversion functions `convert_pyobject_to_order_any` and `convert_order_any_to_pyobject` that will help with transformation between pyo3 `PyObject` and `OrderAny`
- updated python `CachePostgresAdapter` possibility of adding and querying order
- pyo3 and Rust e2e tests


nautilus_core/Cargo.toml
```diff
@@ -50,7 +50,6 @@ tracing = "0.1.40"
 tokio = { version = "1.37.0", features = ["full"] }
 ustr = { version = "1.0.0", features = ["serde"] }
 uuid = { version = "1.8.0", features = ["v4"] }
-sqlx = { version = "0.7.4", features = ["postgres", "runtime-tokio"] }
```
cjdsellers, line 53: 👌

nautilus_core/cli/Cargo.toml
```diff
@@ -21,5 +21,4 @@ log = { workspace = true }
 clap = { version = "4.5.4", features = ["derive", "env"] }
 clap_derive = { version = "4.5.4" }
 dotenvy = { version = "0.15.7" }
-sqlx = {workspace = true }
```
cjdsellers, line 24: 👌

nautilus_core/common/src/cache/core.rs
```diff
@@ -2554,7 +2554,7 @@ mod tests {
             UnixNanos::default(),
         )
         .unwrap(); // TODO: Should event generation be fallible?
-        order.apply(OrderEvent::Submitted(submitted)).unwrap();
+        order.apply(OrderEventAny::Submitted(submitted)).unwrap();
```
cjdsellers, line 2557: I agree with the rename since the enum follows the `*Any` pattern.

nautilus_core/core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ serde_json = { workspace = true }
 ustr = { workspace = true }
 uuid = { workspace = true }
 heck = "0.5.0"
+pretty_assertions = {version = "1.4.0"}
```
cjdsellers, line 23: We could even make this a workspace level dev dependency?

nautilus_core/infrastructure/src/sql/models/orders.rs
```diff
@@ -0,0 +1,356 @@
...
+pub struct OrderPendingUpdateModel(pub OrderPendingUpdate);
+pub struct OrderRejectedModel(pub OrderRejected);
+pub struct OrderReleasedModel(pub OrderReleased);
+pub struct OrderSubmittedModel(pub OrderSubmitted);
+pub struct OrderTriggeredModel(pub OrderTriggered);
+pub struct OrderUpdatedModel(pub OrderUpdated);
+
+impl<'r> FromRow<'r, PgRow> for OrderEventAnyModel {
+    fn from_row(row: &'r PgRow) -> Result<Self, sqlx::Error> {
+        let kind = row.get::<String, _>("kind");
```
cjdsellers, line 59: This is fine for the PR, but there might be a better way to know the `kind` and if we must use strings prefer the `stringify!` macro.

nautilus_core/model/src/orders/market.rs
```diff
@@ -359,6 +347,18 @@ impl Order for MarketOrder {
         self.quantity = event.quantity;
         self.leaves_qty = self.quantity - self.filled_qty;
     }
+
+    fn events(&self) -> Vec<&OrderEventAny> {
```
cjdsellers, line 351: 👌

nautilus_trader/model/events/order.pyx
```diff
@@ -855,7 +856,7 @@ cdef class OrderDenied(OrderEvent):
             "client_order_id": obj.client_order_id.value,
             "reason": obj.reason,
             "event_id": obj.id.value,
-            "ts_event": obj.ts_init,
+            "ts_event": obj.ts_event,
```
cjdsellers, line 859: This is probably the right thing to do, but it'll be the same as `ts_init`, as there's no separate event timestamp for this one.

nautilus_trader/core/nautilus_pyo3.pyi
```diff
@@ -1973,6 +1985,8 @@ class OrderInitialized:
     @classmethod
     def from_dict(cls, values: dict[str, str]) -> OrderInitialized: ...
     def to_dict(self) -> dict[str, str]: ...
+    @property
+    def order_type(self) -> str: ...
```
cjdsellers, line 1989: What was the motivation to add this? it's not always easily available at event generation.

cjdsellers (approved):

nautilus_core/core/Cargo.toml
filipmacek: Yes, we cool. It will be probably used in other crates also

nautilus_trader/core/nautilus_pyo3.pyi
filipmacek: To separate them at runtime, so that we know how to properly transform them

PR #1644, opened 2024-05-16
Betfair: improved session renewal

imemo88:
# Pull Request

Two fixes:
1- Refresh the stream connection once a session is invalid (self._client.disconnect() has been replaced with self._disconnect() which disconnect both the client and stream)
2- Avoid multiple reconnect attempts if there are multiple INVALID_SESSION_INFORMATION errors in the current event queue

## Type of change

In Prod


limx0 (approved):

cjdsellers:
Thanks for the contribution @imemo88 :pray:.

PR #1649, opened 2024-05-18
Fix order event object conversion

filipmacek:
# Pull Request

After the recent changes https://github.com/nautechsystems/nautilus_trader/commit/2526448a5d9b2783b78d6468de31a81b7af992ef in order event  pyo3 conversion `getter` attribute was missing as you cannot use `.getattr` on function `type_str`. Instead of that we can use `__name__` or `__class__` attribute, and remove all `type_str` function from pyo3 classes

```
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: PyErr
 { type: <class 'TypeError'>, 
value: TypeError("'builtin_function_or_method' object cannot be converted to 'PyString'"), 
traceback: None }

```

### DIFF
--- a/nautilus_core/model/src/python/events/order/mod.rs
+++ b/nautilus_core/model/src/python/events/order/mod.rs
@@ -65,7 +65,8 @@ pub fn order_event_to_pyobject(py: Python, order_event: OrderEventAny) -> PyResu
 }
 
 pub fn pyobject_to_order_event(py: Python, order_event: PyObject) -> PyResult<OrderEventAny> {
-    match order_event.getattr(py, "type_str")?.extract::<&str>(py)? {
+    let class = order_event.getattr(py, "__class__")?;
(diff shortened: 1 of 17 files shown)


nautilus_core/model/src/python/events/order/mod.rs
```diff
@@ -65,7 +65,8 @@ pub fn order_event_to_pyobject(py: Python, order_event: OrderEventAny) -> PyResu
 }
 
 pub fn pyobject_to_order_event(py: Python, order_event: PyObject) -> PyResult<OrderEventAny> {
-    match order_event.getattr(py, "type_str")?.extract::<&str>(py)? {
+    let class = order_event.getattr(py, "__class__")?;
```
cjdsellers, line 68: 👌

cjdsellers (approved):
I think this is much cleaner.

PR #1650, opened 2024-05-18
Add PSQL cache database update orders

filipmacek:
# Pull Request

- implemented `update_order`PSQL cache database function in Rust
- PSQL schema updates on table `order_event`
- both Rust and pyo3 tests


nautilus_trader/model/orders/stop_market.pyx
```diff
@@ -337,12 +337,12 @@ cdef class StopMarketOrder(Order):
             client_order_id=init.client_order_id,
             order_side=init.side,
             quantity=init.quantity,
-            trigger_price=Price.from_str_c(init.options["trigger_price"]),
+            trigger_price=Price.from_str_c(init.options['trigger_price']),
```
cjdsellers, line 340: Single quotes?

nautilus_trader/model/orders/stop_market.pyx
```diff
@@ -337,12 +337,12 @@ cdef class StopMarketOrder(Order):
...
             trigger_type=trigger_type_from_str(init.options["trigger_type"]),
-            time_in_force=init.time_in_force,
-            expire_time_ns=init.options["expire_time_ns"],
             init_id=init.id,
             ts_init=init.ts_init,
+            time_in_force=init.time_in_force,
+            expire_time_ns=init.options["expire_time_ns"] if init.options["expire_time_ns"] is not None else 0,
```
cjdsellers, line 345: - Double indexing
- I think `expire_time_ns` should be an integer?
- Do we really need to reorder the params at this stage?

nautilus_core/model/src/orders/base.rs
```diff
@@ -656,6 +656,85 @@ impl OrderCore {
     pub fn init_event(&self) -> Option<OrderEventAny> {
         self.events.first().cloned()
     }
+
+    pub fn generate_order_submitted_event(
```
cjdsellers, line 660: As discussed, I think there is a better location for these helpers.

Not where and how we'll generate these events in the production code, and currently these are just for a test.

tests/unit_tests/model/test_position_pyo3.py
```diff
@@ -91,6 +91,7 @@ def test_position_to_from_dict():
                 "commission": "2.00 USD",
                 "currency": "USD",
                 "event_id": "038990c6-19d2-b5c8-37a6-fe91f9b7b9ed",
+                "info": {},
```
cjdsellers, line 94: Are we adding an info field here for `Position`? I don't see this on the Cython side

nautilus_trader/model/orders/stop_market.pyx
filipmacek: Good catch, probably mistake. Fixed it here https://github.com/nautechsystems/nautilus_trader/pull/1650/commits/7c7102f8d8e562ebcf70474133688729a25ea8d3

nautilus_core/model/src/orders/base.rs
filipmacek: Refactored here https://github.com/nautechsystems/nautilus_trader/pull/1650/commits/234a7355a9b82d2e96f5e7f8254eb2a613d91720

tests/unit_tests/model/test_position_pyo3.py
filipmacek: Yes, we are. Check the `to_dict` method of `OrderFilled` in cython, https://github.com/nautechsystems/nautilus_trader/blob/2526448a5d9b2783b78d6468de31a81b7af992ef/nautilus_trader/model/events/order.pyx#L4800

filipmacek: Thats why its here to match Cython basically

nautilus_trader/model/orders/stop_market.pyx
filipmacek: As you can see there nothing much changed order that `None` check was added, because it could happen that `init.options["expire_time_ns"] ` is None

filipmacek: Basically this 
```
expire_time_ns=init.options["expire_time_ns"],
```

became this 
```
expire_time_ns=init.options["expire_time_ns"] if init.options["expire_time_ns"] is not None else 0,
```

filipmacek: Personally wouldnt worry much about this, as you once said this was bad design to include some of this params into `option` dictionary.  But we have some trouble unifying it with Rust as there every param in `options` is its own field

tests/unit_tests/model/test_position_pyo3.py
cjdsellers: Isn't that `info` for the individual order fill though, what will go into `info` for the entire `Position`?

nautilus_trader/model/orders/stop_market.pyx
cjdsellers: I don't see anywhere it can be `None` as its typed as `uint64_t expire_time_ns`. But doesn't matter so much I can check this once merged.

cjdsellers: yeah, I think the `options` dict itself wasn't a good design choice.

filipmacek: `None` can come from Rust I think, and i don't see much problem fixing it here

tests/unit_tests/model/test_position_pyo3.py
filipmacek: Yes, this in `info` on order event not on the `Position`

PR #1653, opened 2024-05-19
Fix IB tick level historical data downloading

DracheShiki:
# Pull Request

Temporal fix for the bug that leads to endless loop at the end of day when downloading tick-level historical data from Interactive Brokers.

## Type of change

        if min_timestamp.floor("S") == max_timestamp.floor("S"):
            max_timestamp = max_timestamp.floor("S") + pd.Timedelta(seconds=1)
the original related code here only works in the following situation:
nt bug 1
but doesn't work in the following situation:
nt bug 2

Temporal fix. There must be better way than mine...


DracheShiki:
And i got no idea why pre-commit doesn't work while merging.
on my local maschine:
image

rsmb7z:
LGTM for now. We can later improve further with further testing.

rsmb7z (approved):

cjdsellers:
Many thanks @DracheShiki 🙏.

PR #1675, opened 2024-05-28
Filter SANDBOX_INSTRUMENTS to match the execution venue

fredmonroe:
# Pull Request

The current implementation of the sandbox execution adapter depends on a global variable called SANDBOX_INSTRUMENTS
In the IBKR example it mentions that all the instruments must be from the same venue
Indeed this is true and the sandbox adapter will fail if you have adapters for more than one venue.
This is problematic for IBKR because the venues are often different e.g. NASDAQ, NYSE, CME etc.
I have modified the execution adapter when its loading to filter the global SANDBOX_INSTRUMENTS to just those instruments that match the venue the adapter is being configured for.

## Type of change

I have a local sandbox designed to work with IB that loads instruments on multiple venues (ARCA, NASDAQ) because it trades multiple stocks that have different primary exchanges via IBKR

following the example in the repo, i create multiple sandbox execution adapters - one for each venue that exists in my list of instruments

### DIFF
--- a/nautilus_trader/adapters/sandbox/execution.py
+++ b/nautilus_trader/adapters/sandbox/execution.py
@@ -111,7 +111,7 @@ class SandboxExecutionClient(LiveExecutionClient):
             base_currency=base_currency,
             default_leverage=config.default_leverage,
             leverages=config.leverages or {},
-            instruments=self.INSTRUMENTS,
+            instruments=[i for i in self.INSTRUMENTS if i.venue == sandbox_venue],
             modules=[],
             portfolio=portfolio,
             msgbus=self._msgbus,


cjdsellers:
I've noticed this `SANDBOX_INSTRUMENTS` class variable. I think this is a reasonable work around, but ultimately we'll want to refactor this class variable away soon.

Thanks for the contribution here :pray:.

fredmonroe:
> I've noticed this `SANDBOX_INSTRUMENTS` class variable. I think this is a reasonable work around, but ultimately we'll want to refactor this class variable away soon.
> 
> Thanks for the contribution here 🙏.

maybe the sandbox config could be passed the instruments to use instead of a global? i was worried though that configs maybe need to be persisted and that might be too heavy

cjdsellers:
Potentially, or the simulated exchange doesn't necessarily need all instruments up front - but can generate any matching engines after start and when data arrives if they don't exist.

In the mix is a validation that an instrument must exist for market data being added, as that's a common enough user mistake.

I'm looking at this at the moment.

cjdsellers:
@fredmonroe [this change](https://github.com/nautechsystems/nautilus_trader/commit/d60733de87c3e82b78cbe13c452fc53194e28900) gets rid of the class var and makes the sandbox execution client more flexible to when/how instruments are added. Filtering for the correct venue will naturally happen too.

PR #1687, opened 2024-05-31
Betfair: Amend order qty support + a few minor fixes

imemo88:
# Pull Request
Apologies for a somewhat large PR. I tried to keep it small.

Feature:
- The Nautilus modifyOrder API supports price and quantity amendments. Betfair cannot modify both in one atomic transaction. The current BF adapter only support price amendments but this change will allow the user to modify price OR quantity using the modifyOrder API.

Fixes:
- ModifyOrder (price) failures did not correctly parse failures (ExecutionReportStatus.FAILURE vs InstructionReportStatus.FALURE)
- Multiple execution.connect() calls will spawn multiple update_account_state() coros. Moved create_task(update_account_state) out of the connect method AND converted it to a do-while method as the sleep should be called first in a while True-loop in case of exceptions.
- The rust logger? does not support ", exc_info=True)" in log messages so opted to use traceback.format_exc instead.
- SUBSCRIPTION_LIMIT_EXCEEDED: Treat it similar to MAX_CONNECTION_LIMIT_EXCEEDED as SUBSCRIPTION_LIMIT_EXCEEDED does not close the connection. The connection will stay broken presently.

## Type of change

In prod, but I suspect I may need to submit a few modifyOrder fixes once I get more runtime


limx0 (approved):

limx0:
LGTM - Thanks @imemo88 

@cjdsellers this is good to go from my end

limx0:
This following is causing an infinite loop in `test_account_statement`

```
class BetfairExecutionClient(LiveExecutionClient):
    def __init__() -> None:
        .... # omitted 
        self._init_scheduled_tasks()    <---- causing infinite loop

    def _init_scheduled_tasks(self) -> None:
        self.create_task(self.account_state_updates())
```

imemo88:
Thanks @limx0, I'll have a look at the suggestions!

cjdsellers:
Closed from #1751 and will also credit @imemo88 in the release notes.

Issue #1689, opened 2024-06-01
Documentation - clang is a "From Source" installation dependency


Troubladore:
# Bug Report
Nautilus installation "From Source" requires clang, which is not specified in the Installation docs.

### Expected Behavior
Docs call out clang dependency.

### Actual Behavior
No dependency identified

### Steps to Reproduce the Problem

1. Browse to https://docs.nautilustrader.io/getting_started/installation.html
2. Read the "From Source" section

### Specifications

- OS platform: Any
- Python version: Any
- `nautilus_trader` version: 1.194.0

Troubladore:
This isn't a bug, but I wasn't sure what label to use instead.

Troubladore:
Docs updated, thanks Chris!

PR #1690, opened 2024-06-01
Document clang Installation dependency

Troubladore:
Issue #1689

# Pull Request

Update Install doc to include clang dependency. Added validation step for rustc and clang.

## Type of change

- [ x ] This change requires a documentation update

- N/A this is a documentation-only change


cjdsellers:
I think Clang is either pre-installed on macOS or comes with xcode, also probably `brew` would be used to install?

I'm happy to merge these changes as is once the CLA is signed, and I'll make a couple of tweaks.

Many thanks @Troubladore!

PR #1691, opened 2024-06-03
Update InteractiveBrokersGatewayConfig docs

Troubladore:
# Pull Request
I've updated class parameter documentation in an attempt to clarify meanings. My assumption is that IB Gateway or TWS can be dockerized, and that the port used on the surface of the container to provide access to API services may vary, but that common defaults should be called out (4002/4001; 7497/7496).

This request may be updated as further information/changes are received from pending changes to the adapter config classes.

## Type of change

No code changes were made, so no testing was performed. Documentation changes were partially confirmed through [Discord chat](https://discord.com/channels/924497682343550976/949809536108228609/1247185313760022549), but should be verified for accuracy in interpretation.


cjdsellers:
Many thanks @Troubladore :pray:.

PR #1697, opened 2024-06-05
Add PYO3_PATH configuration to reduce spurious compiles

twitu:
# Pull Request

cargo performs frequent recompiles because of a difference in Python path between CLI and rust analyzer environments. [^1]

## Type of change

Ran cli cargo build and made then cargo check using rust analyzer in vs code, only relevant changes were compiled.

[^1]: https://github.com/PyO3/pyo3/issues/1708


cjdsellers:
Great idea, thanks @twitu

PR #1703, opened 2024-06-07
Add index instrument support for Interactive Brokers

rsmb7z:
# Pull Request

Add support for Index instrument.

## Type of change

Included some tests.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/common.py
+++ b/nautilus_trader/adapters/interactive_brokers/common.py
@@ -111,6 +111,7 @@ class IBContract(NautilusConfig, frozen=True, repr_omit_defaults=True):
         "CRYPTO",
         "CFD",
         "CMDTY",
+        "IND",
         "",
     ] = ""
     conId: int = 0
--- a/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
+++ b/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
@@ -18,7 +18,6 @@ import re
 import time
 from decimal import Decimal
 
-import msgspec
 import pandas as pd
 from ibapi.contract import ContractDetails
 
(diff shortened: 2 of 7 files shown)


cjdsellers:
Hey @rsmb7z thanks for the PR!

Some thoughts here:
- Should this be a type of `Instrument` if the products aren't tradable?
- If not tradable, then should we treat it similar to `SyntheticInstrument` which is a type of `Data`?
- If not tradable, then do we need the size precision and increment?
- What does the data look like coming from IB?

rsmb7z:
> * Should this be a type of `Instrument` if the products aren't tradable?
> * If not tradable, then should we treat it similar to `SyntheticInstrument` which is a type of `Data`?

Although the InstrumentContract is not tradable, it does not fall under the category of SyntheticInstrument since it originates directly from the trading venue rather than being internally generated by Nautilus. It should indeed be treated as a type of Instrument because, similar to tradable instruments, it would be managed via data subscriptions in the same way. We have already categorized it as `asset_class=AssetClass.INDEX`, which provides a clear distinction. However, adding a new enumeration value to InstrumentClass could help further clarify its classification, ensuring it is properly recognized within Nautilus.

> * What does the data look like coming from IB?

In the ibapi, definition is structured in a `IBContract` object, similar to how stocks, futures, and other financial instruments are represented.

> * If not tradable, then do we need the size precision and increment?

Although there may not be an immediate need for these details, I have included them because they are present in the IBContract and are required attributes for the Instrument class. This ensures consistency and completeness in how we handle all types of instruments, tradable or not.

cjdsellers:
> Although the InstrumentContract is not tradable, it does not fall under the category of SyntheticInstrument since it originates directly from the trading venue rather than being internally generated by Nautilus. It should indeed be treated as a type of Instrument because, similar to tradable instruments, it would be managed via data subscriptions in the same way.

We also use `SyntheticInstrument(Data)` in the `DataEngine` to manage data feeds, and its not a tradable instrument.

> Although the instrument is not tradable, the size precision and increment parameters are indeed necessary for processing index data precision.

Interesting, so the reason I ask is because I would have just expected prices for an index, are there sizes too? I would have thought these would indicate trades?

I'm also happy to merge as you've added support into the Interactive Brokers adapter which definitely seems necessary. We can potentially tweak / refine the index instrument after merge based on your answers to the above.

PR #1722, opened 2024-06-19
Fix docs to use markdown link instead of html

Pushkarm029:
## Type of change


cjdsellers:
Many thanks! we'll improve this markdown gen soon, in the mean time I think these are much better than broken links :pray:

Issue #1741, opened 2024-06-26
Add Http request_order_book_snapshot method


graceyangfan:
# Feature Request
Due to the high cost of subscribing to the order book, especially for strategies that run multiple instruments and are not high-frequency, it is important to consider the utility of micro-level order book data for strategies. Therefore, adding an HTTP method to fetch order book snapshots should be considered. I request the addition of a request_order_book_snapshot method to obtain this data within the model.

Example usage:
```
def on_historical_data(self, data):
    if isinstance(data, OrderBookDeltas):
        self.deltas = data

def on_bar(self, bar: Bar):
    self.request_order_book_snapshot(
        instrument_id=self.instrument_id,
        limit=5
    )
    # Process deltas

cjdsellers:
Hi @graceyangfan

Thanks for the contribution here.

I think this makes sense, that a trader may require a snapshot at some arbitrary frequency and depth - but not necessarily be maintaining a book the entire time with deltas.

Thankfully this wasn't too much trouble to implement with the flexible messaging system.

The only thing left here from what I can see is to add some test coverage for the `DataEngine`. You could use the `TestDataProvider.order_book_snapshot(...)` method.

Do you have an adapter implementation for Binance?

I can handle Databento and docs.

cjdsellers:
Hey @graceyangfan 

Did you end up getting your use case working for Binance?

graceyangfan:
@cjdsellers  it is now working normally on Binance. The extension to other exchanges might have to rely on other developers.

cjdsellers:
Closing - PR #1745 merged and confirmed working on Binance by @graceyangfan. Other exchange implementations can be tracked as separate adapter-specific issues if needed.

PR #1745, opened 2024-06-27
Add request_order_book_snapshot method

graceyangfan:
# Pull Request

 #1741 Add request_order_book_snapshot method 

## Type of change

test with live trading example 


cjdsellers:
Thanks for adding the tests @graceyangfan, much appreciated :pray:.

graceyangfan:
@cjdsellers In fact, in backtest mode, the request is not doing anything. It only works for live trading at the moment. To make it work in backtest mode, some additional work is still needed.

cjdsellers:
Yes, for historical requests to work for backtesting the `DataEngine` needs access to the catalog. There's a section which handles this for ticks already.

PR #1751, opened 2024-06-29
Feat/amend order qty

limx0:
- Continuation of #1687 with fixes.


cjdsellers:
Thank you @imemo88 and @limx0 :rocket:

PR #1754, opened 2024-06-29
Fix OrderMatchingEngine processing by book type for quotes and deltas

davidsblom:
…der books

# Pull Request

When the strategy subscribes to both quote deltas and order book deltas, the matching engine applies both quotes as well as deltas to the order book which is incorrect.
In case of the Bybit venue, quote ticks are pushed at a higher frequency than deltas. Hence, some strategies would like to get the latest top of the book quotes, and also build indicators on top of the full order book.

The matching engine should use a L1_MBP order book when subscribing to both quotes and deltas, because the quotes are sent earlier introducing a lookahead bias.

## Type of change

Locally running backtests

### DIFF
--- a/nautilus_trader/backtest/matching_engine.pyx
+++ b/nautilus_trader/backtest/matching_engine.pyx
@@ -369,7 +369,8 @@ cdef class OrderMatchingEngine:
         if is_logging_initialized():
             self._log.debug(f"Processing {repr(delta)}")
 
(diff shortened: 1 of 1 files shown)


cjdsellers:
Agreed, this change is good when running a backtest with both quotes and deltas. We only want to apply the deltas conditionally for L2 and L3 books.

cjdsellers (approved):

PR #1755, opened 2024-06-29
Refactor accounting in Rust

filipmacek:
# Pull Request

- To create `AccountAny` whole `accounting` crate content needed to be moved to `model` crate
-  `Box<dyn Account>>` is replaced with `AccountAny` across the Rust codebase


cjdsellers (approved):
With this move, we're choosing performance over abstraction.

A potentially better location is the `common` crate, but this remains to be decided as we can still afford to shuffle things around.

Issue #1732, opened 2024-06-24
Network clients can use `AtomicBool` instead of `<Arc<Mutex<bool>>`


twitu:
# Enhancement Request

The nautilus websocket and socket clients uses an `Arc<Mutex<bool>>` to syncronize between sender and receiver ends of the stream, particularly for when to disconnect. While Arc Mutex works, `AtomicBool` is the perfect fit for the role. [^1]

Changing the type of the `disconnect_mode` field in `websocket.rs` and `socket.rs` should lead the type system to show all the places where changes are needed.

[^1]: https://users.rust-lang.org/t/atomicbool-or-mutex-bool/9958

PR #1759, opened 2024-06-30
Make `websocket` and `socket` disconnect sync function post atomic bool changes

twitu:
# Pull Request

Socket disconnection operation is now sync after moving to atomic bool in #1732. 

## Type of change

pre-commit and rust tests were run


twitu:
The existing implementation works. A later PR can improve the tests and unskip the skipped once.

PR #1793, opened 2024-07-19
Fix reconnect for Bybit websockets max args

davidsblom:
…tion request

# Pull Request

The documentation of the websocket of bybit states: `You can input up to 10 args for each subscription request sent to one connection`
So, now when resubscribing to all data streams, multiple requests are used. Otherwise, the server responds with errors such as:
```
WebSocket error: BybitWsMessageGeneral(op='subscribe', topic=None, success=False, ret_msg='args size >10')
```

## Type of change

Subscribing to 30 data streams during live execution, and forcing a reconnection by switching off the network, and switching it on again.


cjdsellers:
Thanks for the fix! @davidsblom

PR #1794, opened 2024-07-20
Fix Bybit parsing of trade and quote ticks for websocket messages

davidsblom:
# Pull Request

Explicitly set the quote price, quote size, trade price and trade size precisions for the Bybit venue.

## Type of change

Subscribing to live trade data for the bybit venue. Asserting in strategy that the correct prize and size precision are used for all ticks and adding a unit test.


cjdsellers:
Thanks @davidsblom and for adding the test also!

cjdsellers (approved):

PR #1799, opened 2024-07-23
Add `ignore_quote_tick_size_updates` option for Interactive Brokers Data Client

sunlei:
# Pull Request

## Type of change

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/config.py
+++ b/nautilus_trader/adapters/interactive_brokers/config.py
@@ -167,7 +167,7 @@ class InteractiveBrokersDataClientConfig(LiveDataClientConfig, frozen=True):
         If True, will request data for Regular Trading Hours only.
         Only applies to bar data - will have no effect on trade or tick data feeds.
         Usually used for 'STK' security type. Check with InteractiveBrokers for RTH Info.
-    market_data_type : bool, default REALTIME
+    market_data_type : IBMarketDataTypeEnum, default REALTIME
         Set which IBMarketDataTypeEnum to be used by InteractiveBrokersClient.
         Configure `IBMarketDataTypeEnum.DELAYED_FROZEN` to use with account without data subscription.
     ignore_quote_tick_size_updates : bool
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @sunlei

Thanks for the contribution.

I'm interested to know what your end goal is with this change?

A couple of minor points:
- Maybe `quote_ticks_update_size` is a clearer name, as it relates to quote tick updates? -- and not ignoring the size entirely
- The config option should retain current behavior, so in your case `False` or if using `quote_ticks_update_size` would be True

I'm interested in hearing any thoughts from @rsmb7z @benjaminsingleton and others?

nautilus_trader/adapters/interactive_brokers/config.py
```diff
@@ -170,6 +170,9 @@ class InteractiveBrokersDataClientConfig(LiveDataClientConfig, frozen=True):
     market_data_type : bool, default REALTIME
         Set which IBMarketDataTypeEnum to be used by InteractiveBrokersClient.
         Configure `IBMarketDataTypeEnum.DELAYED_FROZEN` to use with account without data subscription.
+    quote_ticks_ignore_size : bool
```
rsmb7z, line 173: A more suitable name for this flag could be `ignore_quote_tick_size_updates`, with the following description:

"If set to True, the QuoteTick subscription will exclude ticks where only the size has changed but not the price. This can help reduce the volume of tick data. When set to False (the default), QuoteTick updates will include all updates, including those where only the size has changed."

rsmb7z:
@cjdsellers @sunlei,
Yes, the current behavior should be retained with `False` as default. Apart from the config flag name and description, everything else looks good.

sunlei:
@cjdsellers I would like to provide liquidity for some stocks, so need more timely BBO and depth data.

rsmb7z (approved):

nautilus_trader/adapters/interactive_brokers/config.py
sunlei: I have made changes.
Thank you for your suggestions.

cjdsellers:
Thanks everyone :pray: , I'll merge this now.

PR #1801, opened 2024-07-23
Implement FillModel in Rust

filipmacek:
# Pull Request

- created new module `models` in `backtest` crate which will contain models implementations like `FeeModel`, `LatencyModel` , `FillModel` etc
- implemented `FillModel` with random number generation from seed or from entropy


cjdsellers (approved):
We probably don't need cryptographically secure random numbers with `rand_chacha`, but thats a small detail.

PR #1802, opened 2024-07-24
Implement FixedFeeModel in Rust

filipmacek:
# Pull Request

- create a new trait `FeeModel` for handling commissions 
- implemented a `FixedFeeModel` with tests


nautilus_core/backtest/src/models/fee.rs
```diff
@@ -0,0 +1,152 @@
...
+#[derive(Debug, Clone)]
+pub struct FixedFeeModel {
+    commission: Money,
+    zero_commission: Money,
+    change_commission_once: bool,
+}
+
+impl FixedFeeModel {
+    pub fn new(commission: Money, change_commission_once: Option<bool>) -> anyhow::Result<Self> {
+        if commission.as_f64() < 0.0 {
```
cjdsellers, line 41: Something to remember is that Rust is much more efficient with calculations on the domain value objects, so we don't necessary need to cast into ints and doubles as often as the legacy Cython code.

PR #1813, opened 2024-07-28
Implement bar schema with PSQL cache database in Rust

filipmacek:
# Pull Request

- added new function `load_bars` in `CacheDatabaseAdapter ` trait
- added new variant `AddBar` in `DatabaseQuery` enum
- implemented `add_bar` and `load_bars` for `PostgresCacheAdapter`
- new enums in PSQL schema related to bar aggregation and price type
- tested in `test_cache_database_postgres.rs`

### DIFF
--- a/nautilus_core/model/src/data/bar.rs
+++ b/nautilus_core/model/src/data/bar.rs
@@ -299,7 +299,6 @@ pub struct Bar {
 
 impl Bar {
     /// Creates a new [`Bar`] instance.
-    #[must_use]
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         bar_type: BarType,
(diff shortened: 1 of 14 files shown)


nautilus_core/model/src/data/bar.rs
```diff
@@ -310,8 +309,8 @@ impl Bar {
         volume: Quantity,
         ts_event: UnixNanos,
         ts_init: UnixNanos,
-    ) -> Self {
-        Self {
+    ) -> anyhow::Result<Self> {
+        Ok(Self {
```
cjdsellers, line 313: We probably don't gain anything by returning a result? because `Bar` creation is infallible thanks to the domain objects.

cjdsellers (approved):

PR #1814, opened 2024-07-28
Fix unsubscribing from order book deltas

davidsblom:
# Pull Request

The order book subscription is kept alive forever instead of closed on request. The actor is subscribing to the order book deltas, but also the data engine which is maintaining it's own order book. So there are 2 subscribers. The DataEngine._setup_order_book  sets up the subscription, but doesn't unsubscribe from the subscription later.

The message bus exposes the number of subscriptions for a certain topic, and during the unsubscription process, the data engine checks if the internal order book subscription is the last one.

## Type of change

During development of dYdX adapter. The unsubscribe_order_book_deltas() was not called by the DataEngine.

### DIFF
--- a/nautilus_trader/common/component.pxd
+++ b/nautilus_trader/common/component.pxd
@@ -284,6 +284,7 @@ cdef class MessageBus:
     cpdef bint is_subscribed(self, str topic, handler)
     cpdef bint is_pending_request(self, UUID4 request_id)
     cpdef bint is_streaming_type(self, type cls)
+    cpdef int num_subscribers(self, str pattern=*)
 
     cpdef void dispose(self)
     cpdef void register(self, str endpoint, handler)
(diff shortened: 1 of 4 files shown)


nautilus_trader/common/component.pxd
```diff
@@ -284,6 +284,7 @@ cdef class MessageBus:
     cpdef bint is_subscribed(self, str topic, handler)
     cpdef bint is_pending_request(self, UUID4 request_id)
     cpdef bint is_streaming_type(self, type cls)
+    cpdef int num_subscribers(self, str pattern=*)
```
cjdsellers, line 287: This probably doesn't justify a separate method, because I think its enough to just be able to call
```
len(self.subscriptions(pattern))
```

cjdsellers (approved):

PR #1820, opened 2024-07-31
Improve message handler invoked testing

filipmacek:
# Pull Request

The idea was to test if an appropriate message handler was called with generic `&dyn Any` message object, for example in 
 the messagebus `send` method.  To do that custom `CallCheckMessageHandler` was defined which sets the flag to true if the handler is invoked. The most problem I faced was into downcasting to `Any`

- created the `stub` file which will contain these testing message handlers which will be used by other components across codebase
-  modified `MessageHandler` so that it inherits from `Any` trait,  which allows for type casting using `Any`'s methods like `downcast_ref`
- with that change, we better support type casting through `Any` and provide an `as_any` method for easier type casting.

### DIFF
--- a/nautilus_core/common/src/msgbus/mod.rs
+++ b/nautilus_core/common/src/msgbus/mod.rs
@@ -16,6 +16,7 @@
 //! A common in-memory `MessageBus` for loosely coupled message passing patterns.
 
 pub mod database;
+pub mod stubs;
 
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
I think this is good progress on testing and nailing down how handlers work a little more.

PR #1828, opened 2024-08-04
Fix customdataclass and add tests

faysou:
# Pull Request

Fixed customdataclass and added some tests

## Type of change

- This change requires a documentation update

Using unit tests and running my own code that uses this feature.

### DIFF
--- a/nautilus_core/backtest/src/matching_engine.rs
+++ b/nautilus_core/backtest/src/matching_engine.rs
@@ -218,6 +218,7 @@ impl OrderMatchingEngine {
     }
 
     // -- TRADING COMMANDS ----------------------------------------------------
+    #[allow(clippy::needless_return)]
     pub fn process_order(&mut self, order: &OrderAny, account_id: AccountId) {
         if self.core.order_exists(order.client_order_id()) {
             self.generate_order_rejected(order, "Order already exists".into());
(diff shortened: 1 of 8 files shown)


cjdsellers:
Many thanks @faysou

faysou:
@cjdsellers I've improved the code a bit more so a customdataclass always is a dataclass while maintaining the possibility to keept dataclass arguments for the decorator. Also I've fixed a problem in the test (using a field with a default factory was creating a problem).

cjdsellers:
I hadn't actually approved the PR yet, I just triggered CI to run which has to be done on every new commit you push until your first PR is merged.

I actually think it would be more consistent with the other `Data` classes if we made `ts_event` and `ts_init` required parameters, and without the leading underscores which also turn up in the dictionary representation. This is because `Data` isn't valid without correct values for these, they must be provided. I understand what you mean about "hidden" fields, but I think in this case its better to be consistent with the intent of the `Data` contract and the other uses of it?

faysou:
Ok I've managed to modify the code so _ts_init and _ts_init annotations are created automatically so it's used by other functions and I've modified the construction as well to accept ts_init and ts_event. The code will be a bit slower I suppose with this version as more operations are done to address the translation between ts_init and and ts_event in from_dict as well as in the constructor. Also the constructor will be less elegant as it will contain **kwargs instead of a fully formed constructor generated by dataclass.

cjdsellers:
Cool, this looks good to me and many thanks.

One question, are we still able to pass `ts_event` and `ts_init` as kwargs?

faysou:
Hi @cjdsellers yes it's possible to pass them as kwargs, they just need a default value which I've included in the default constructor.

Issue #1804, opened 2024-07-25
StreamingFeatherWriter unable to write only one type when using include_types


faysou:
# Bug Report

### Expected Behavior
I would like to only serialise some custom data published on the message bus.

### Actual Behavior
If I don't include other types of data used in my backtest like QuoteTick and Bars then I get such error:

`Error sending log event: [ERROR] BacktestNode: Error running backtest: ('quote_tick', 'ESM4 P5230.GLBX')`

I need to enter all data types used in my backtest in order to avoid an error for example like this:

```
streaming = StreamingConfig(
    catalog_path=streaming_catalog_path,
    fs_protocol="file",
    include_types=[QuoteTick, GreeksData, Bar]
)
```

I think that in the StreamingFeatherWriter.write method there should be an instruction like the below comparing the current type to write against self.include_types

```
if cls not in self.include_types:
   return
```

I don't have a short self-contained example to reproduce the bug. You could use a backtest using two types of data like QuoteTick and Bars and only include one of them in include_types.

PR #1833, opened 2024-08-06
Fix handling include_types option in StreamingFeatherWriter

faysou:
# Pull Request

As described in this issue below errors could happen in a backtest with the high-level API when using StreamingFeatherWriter with only one custom data type. 
https://github.com/nautechsystems/nautilus_trader/issues/1804

This was because the writer was still trying to write every types at run time, because a check now added was missing.

## Type of change

Tested on my own code and added a new test.


cjdsellers:
Many thanks for adding the test as well @faysou

PR #1835, opened 2024-08-07
Add optional timeout for HTTP requests in Rust

davidsblom:
# Pull Request

Set an optional timeout for HTTP requests instead of using no timeout. 
When an HTTP server does not respond, a live execution cannot be stopped by pressing CTRL-C. With a timeout, the request times out, and the live execution exits as expected. This also gives the possibility to retry requests from the adapter side.

## Type of change

By calling HTTP endpoints of dYdX. The testnet is very slow and sometimes never responds.

### DIFF
--- a/nautilus_core/network/src/http.rs
+++ b/nautilus_core/network/src/http.rs
@@ -19,6 +19,7 @@ use std::{
     collections::{hash_map::DefaultHasher, HashMap},
     hash::{Hash, Hasher},
     sync::Arc,
+    time::Duration,
 };
 
 use bytes::Bytes;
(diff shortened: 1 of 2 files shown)


nautilus_core/network/src/http.rs
```diff
@@ -78,6 +79,7 @@ impl InnerHttpClient {
         url: String,
         headers: HashMap<String, String>,
         body: Option<Vec<u8>>,
+        timeout_sec: Option<u64>,
```
cjdsellers, line 82: nit: I'd probably go with `timeout_secs`, but don't worry about updating that and running CI again - I'll update on `develop`.

cjdsellers (approved):

nautilus_core/network/src/http.rs
davidsblom: Okay perfect

PR #1845, opened 2024-08-12
Attempt to port python to rust

andrewcz:
# Pull Request

## Type of change
1st attempt at moving some python code over to rust 
look in nautilus_trader/analysis/statistic.rs

this has not been tested port of straight code

not test first pull request


cjdsellers:
Hi @andrewcz

So all Rust should be under `nautilus_core`. I've added an `analysis` crate where statistics should go.

We definitely don't need regex for any names.

I'd recommend you might want to start with making sure Rust and Python indicator implementations are exactly 1-to-1 in behavior and double check the test implementations. When I last checked this there were differences. I think this would be a much better entry point for you than starting with portfolio statistics which need some more design to establish the patterns we'll want to use.

andrewcz:
Cheers Chris! I’ll start there.
Best,
Andrew

Issue #735, opened 2022-08-20
Integrate Binance Futures hedge mode


cjdsellers:
# Feature Request

Integrate Binance Futures `HEDGE` mode support, so that Nautilus `OMSType.HEDGING` can utilize it.

https://www.binance.com/en/support/faq/360041819691

cjdsellers:
Now implemented from #1846.

PR #1846, opened 2024-08-12
Add Binance Futures HEDGE mode

DevRoss:
# Pull Request

This PR attempts to support  Binance Futures `HEDGE` mode by adding position_side parameters. #735 
Note: Since I'm not good at rust, these commits only contain changes to python and cython.

## Type of change

Turn on hedge mode and run the strategy


cjdsellers:
Hey @DevRoss

Thanks so much for the contribution here.

What you've done through the Binance adapter looks reasonable. I don't think we can accommodate such a large change back into the core model though, including events and orders. This is the source of many test fails.

Is there a way we can get this feature in without having to specify the position side on an order, which is specific to the way Binance handle their "HEDGING mode"?

Also from prior experience, large changes in adapters like this can lead to many bugs. So I'm immediately thinking about what we could do to improve test coverage for Binance before this goes in, any thoughts?

cjdsellers:
I think we can actually use the `position_id` for this?

DevRoss:
Using `position_id` is indeed better.😄

cjdsellers:
Hey @DevRoss

To clarify, I think the majority of the changes were good.

The better place to indicate the hedging position side is through the `position_id` which can include the `LONG` or `SHORT` component at the end (currently uses `BOTH` for netted positions).

DevRoss:
Hey @cjdsellers 
I reworked the code and added some unittests. However, I found the method for generating orders in OrderFactory does not provide a `position_id` parameter, thus the `position_id` can only be passed to the Executor by `submit_order` or `submit_order_list` method now.

This will raise a issue: When submitting multiple orders with different `position_id`, the strategy needs to call the `self.submit_order_list` multiple times.

My suggestion is to add a position_id parameter to the order generation method in OrderFactory. This should solve the above issue and make it adaptor's `_submit_order` simpler.

nautilus_trader/adapters/binance/common/enums.py
```diff
@@ -35,9 +35,21 @@
 from nautilus_trader.model.enums import TimeInForce
 from nautilus_trader.model.enums import TriggerType
 from nautilus_trader.model.enums import bar_aggregation_to_str
+from nautilus_trader.model.identifiers import PositionId
 from nautilus_trader.model.orders import Order
 
 
+@unique
+class BinanceFuturesPositionSide(Enum):
```
cjdsellers, line 43: I'm curious to know why we moved this to common, I had a quick look at the API docs and only saw hedging mode associated with futures?

cjdsellers:
Hey @DevRoss

Many thanks again! and for adding the additional tests and accommodating the `position_id` change.
At a stretch, would it be possible to add a small blub to the Binance adapter docs about how to configure for hedge mode? :pray:
https://github.com/nautechsystems/nautilus_trader/blob/develop/docs/integrations/binance.md

DevRoss:
Sure!

nautilus_trader/adapters/binance/common/enums.py
DevRoss: Yes, hedging mode is only associated with futures. Currently all the code for parsing position_id is in common module, and I moved it to the current folder. Maybe it's better to keep it as it is? Feel free to modify the code 🙂

PR #1850, opened 2024-08-14
Improve customdataclass constructor to use positional arguments 

faysou:
# Pull Request

Improved customdataclass constructor to be able to use positional arguments after ts_event and ts_init

## Type of change

Modified one test to check it's working as expected. Checked on my own code as well.


faysou:
For the record, if needed to make the constructor more convenient for an IDE, a pyi file can be used with a proper signature, as it's generated at run time.

PR #1853, opened 2024-08-17
Implement subscribe SignalData for Actor

faysou:
# Pull Request

Added functionalities to be able to use Signals in actors. Subscribe signal allows to subcribe to all or one signal of a given name which will be received by the on_data callback in an actor. is_signal is a method of Data to check that a received data in on_data is a signal or a signal of a give name.

Also added more possible basic value types to SignalData.

## Type of change

- New feature (non-breaking change which adds functionality

Tested on my own code and added tests


cjdsellers:
Many thanks @faysou

PR #1905, opened 2024-09-02
Update remaining aggregators to work with composite bars

faysou:
# Pull Request

Update other aggregators than time aggregator to work with composite bars ("ESU4.GLBX-2-MINUTE-LAST-INTERNAL@1-MINUTE-EXTERNAL" for example) With previous refactorings only _apply_update_bar needed to be implemented for each aggregator.

For the value aggregator, the average price of open, high, low and close is used instead of price when using trades or quotes for aggregation.

## Type of change

Many tests have been added, similarly to existing ones

### DIFF
--- a/nautilus_trader/data/aggregation.pxd
+++ b/nautilus_trader/data/aggregation.pxd
@@ -53,7 +53,7 @@ cdef class BarBuilder:
 
     cpdef void set_partial(self, Bar partial_bar)
     cpdef void update(self, Price price, Quantity size, uint64_t ts_event)
-    cpdef void update_bar(self, Bar bar)
+    cpdef void update_bar(self, Bar bar, Quantity volume, uint64_t ts_init)
     cpdef void reset(self)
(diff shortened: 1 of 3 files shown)


nautilus_trader/data/aggregation.pxd
```diff
@@ -53,7 +53,7 @@ cdef class BarBuilder:
 
     cpdef void set_partial(self, Bar partial_bar)
     cpdef void update(self, Price price, Quantity size, uint64_t ts_event)
-    cpdef void update_bar(self, Bar bar)
+    cpdef void update_bar(self, Bar bar, Quantity size, uint64_t ts_init)
```
cjdsellers, line 56: Do we need to separate `size` from `bar` and then have to do `update_bar(bar, bar.volume)` ?

faysou: update from size to volume variable name done

PR #1911, opened 2024-09-04
Use an error code retry list for dYdX

davidsblom:
# Pull Request

Use a error retry list for dYdX. The list of error codes can be expanded, but the documentation does not list all error codes.

## Type of change

Live example.

### DIFF
--- a/nautilus_trader/adapters/dydx/http/errors.py
+++ b/nautilus_trader/adapters/dydx/http/errors.py
@@ -20,6 +20,7 @@ from typing import Any
 
 from grpc.aio._call import AioRpcError
 
+from nautilus_trader.adapters.dydx.common.constants import DYDX_RETRY_ERRORS_GRPC
 from nautilus_trader.adapters.dydx.grpc.errors import DYDXGRPCError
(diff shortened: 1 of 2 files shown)


nautilus_trader/adapters/dydx/http/errors.py
```diff
@@ -54,7 +55,7 @@ def should_retry(error: BaseException) -> bool:
 
     """
     if isinstance(error, DYDXGRPCError):
-        return True
+        return error.code in DYDX_RETRY_ERRORS_GRPC
```
cjdsellers, line 58: Are there rate limit or network related error codes which we'd want to retry on?
I'll have a look at the dYdX API myself when I can get to it.

davidsblom: Yes there are rate limits. This list can be expanded for sure. As of now, we make a distinction between RPC errors, HTTP errors and also catch exceptions thrown by the grpcio library (unavailable errors). 

HTTP requests not use the retry manager yet.

PR #1927, opened 2024-09-11
Update average price in ValueBarAggregator when using bars

faysou:
# Pull Request

Modified price value to average of high low close instead of open high low close in the ValueBarAggregator when using bars as source data.
This is to avoid having the close value being used a second time in the next open value, also similarly to the VWAP price.

## Type of change

Tests have been updated to reflect the new formula.

Such syntax needed to be used in order to compare decimal numbers because of dividing by 3.

```
expected = Decimal("40001.11")
assert (
   aggregator.get_cumulative_value().quantize(expected, rounding=ROUND_HALF_UP) == expected
)
```

### DIFF
--- a/tests/unit_tests/data/test_aggregation.py
+++ b/tests/unit_tests/data/test_aggregation.py
@@ -14,6 +14,7 @@
 # -------------------------------------------------------------------------------------------------
 
 from datetime import timedelta
+from decimal import ROUND_HALF_UP
 from decimal import Decimal
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


tests/unit_tests/data/test_aggregation.py
```diff
@@ -14,6 +14,7 @@
 # -------------------------------------------------------------------------------------------------
 
 from datetime import timedelta
+from decimal import ROUND_HALF_UP
```
cjdsellers, line 17: I'm interesting to know the thinking behind `ROUND_HALF_UP` vs the standard `ROUND_HALF_EVEN`?

faysou: I didn't even know ROUND_HALF_EVEN existed, I just asked an LLM for helping me solve the the rounding problem. I guess I used something that worked int his case and was rounding correctly.

faysou: Related https://gerrytan.wordpress.com/2014/02/25/whats-the-deal-with-half-up-and-half-even-rounding/ although it doesn't matter in this case as it's for a test, you can change to half even if you want in another commit so people can do by analogy in the future.

PR #1934, opened 2024-09-13
Fix no log in backtest run if log not initialized

faysou:
# Pull Request

Allow some log messages to stdout in an edge case where the kernel of a backtest fails before the logging is initialized

## Type of change

The issue can be reproduced for example by not passing a correct type to a strategy config in cython (the type mismatch creates an error and if the logger is not initialized, then the backtestnode's call to Logger does nothing.)

### DIFF
--- a/nautilus_trader/backtest/node.py
+++ b/nautilus_trader/backtest/node.py
@@ -25,6 +25,7 @@ from nautilus_trader.backtest.engine import BacktestEngineConfig
 from nautilus_trader.backtest.results import BacktestResult
 from nautilus_trader.common.component import Logger
 from nautilus_trader.common.component import LogGuard
+from nautilus_trader.common.component import init_logging
 from nautilus_trader.common.config import ActorFactory
 from nautilus_trader.common.config import InvalidConfiguration
 from nautilus_trader.core import nautilus_pyo3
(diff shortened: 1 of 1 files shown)


nautilus_trader/backtest/node.py
```diff
@@ -161,6 +162,11 @@ def run(self) -> list[BacktestResult]:
             except Exception as e:
                 # Broad catch all prevents a single backtest run from halting
                 # the execution of the other backtests (such as a zero balance exception).
+                try:
+                    init_logging()
```
cjdsellers, line 166: There's also a specific function to check this as well:
```
from nautilus_trader.common.component import is_logging_initialized
```

I'll update straight to `develop` with a similar pattern using that, and credit you.

faysou: ok, fyi, I've just made the change you mentioned

cjdsellers: I should have mentioned, also fixed on `develop` branch now https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/backtest/node.py#L163.

faysou: no problem, just close the pull request please, I don't think I can do it

PR #1935, opened 2024-09-13
Add option greeks

faysou:
# Pull Request

Add the possibility to: 

1) compute option greeks with an actor, by converting the mid of a quote to an implied volatility, then computing Black-Scholes greeks (delta, gamma, vega theta).
2) cache computed greeks of one option, publish greeks on the message bus as well, so another actor can analyze data.
3) compute greeks at portfolio level in a strategy
4) save and load greeks to and from a catalog
5) use a varying monthly short term interest rate (this can be generalised for more precision)

## Type of change

Example notebook in python already there, tests in python to come.

The core functions in rust already have tests.


cjdsellers:
Many thanks for this contribution @faysou 🙏

faysou:
Thank you, I think it's a big step forward, I will try to improve it and make it more and more general over time, for now it works with futures and bars for updating the spot and, quote ticks for updating option prices.

I could make it more general system for updating interest rates using messages and another actor taking care of interest rates.

We'll see, improvements should be easier than the first batch of code for this feature.

PR #1952, opened 2024-09-20
Implement Order test data builder in Rust

filipmacek:
# Pull Request

- Replaced custom factory functions in `TestOrderStubs` (e.g., `market_order`, `limit_order`, etc.) with a builder pattern. This allows setting only the necessary properties, avoiding the need to set `None` for unused fields. This approach is more concise, reduces boilerplate and cleaner API for custom order creation
- refactored all the Rust tests to use `OrderAnyTestBuilder`.


cjdsellers (approved):
Great cleanup here :chart_with_upwards_trend:

PR #1976, opened 2024-10-01
Add mode option to ParquetDataCatalog.write_data

faysou:
# Pull Request

Add mode option to ParquetDataCatalog.write_data. This allows for example to download new historical data and append or prepend it to an existing catalog. The existing file will be loaded in memory, then written again followed or preceded by the new data. The default write_data_mode is the same behaviour as before, called "overwrite".

## Type of change

- New feature (non-breaking change which adds functionality

on prototype, add write_data_mode="append" here https://github.com/nautechsystems/nautilus_trader/blob/develop/examples/backtest/databento_option_greeks.py#L60 
Added test

### DIFF
--- a/nautilus_trader/persistence/catalog/parquet.py
+++ b/nautilus_trader/persistence/catalog/parquet.py
@@ -236,6 +236,7 @@ class ParquetDataCatalog(BaseDataCatalog):
         data_cls: type[Data],
         instrument_id: str | None = None,
         basename_template: str = "part-{i}",
+        mode: str = "overwrite",
         **kwargs: Any,
     ) -> None:
         if isinstance(data[0], CustomData):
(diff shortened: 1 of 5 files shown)


nautilus_trader/persistence/catalog/parquet.py
```diff
@@ -236,6 +236,7 @@ def write_chunk(
         data_cls: type[Data],
         instrument_id: str | None = None,
         basename_template: str = "part-{i}",
+        write_data_mode: str = "overwrite",
```
cjdsellers, line 239: This naming is explicit, but since we know this is already writing, maybe `mode` would be cleaner?

faysou: ok I've changed accordingly

cjdsellers (approved):

PR #1986, opened 2024-10-06
Close existing tasks when reconnecting WebSocketClient

davidsblom:
# Pull Request

Close the read task and heart beat task when reconnecting the websocket. When the websocket connection reconnects, old heart beat tasks are not canceled. The heart beat task does not finish when a network error occurs and keeps running in the background, even when a new connection with a second heart beat task is created.

For safety, the existing read task is also closed.

## Type of change

Live example for dYdX. Manually disconnect the network and enabling it again to trigger a reconnect.


twitu (commented):
Similar logic exists in the `shutdown` function. Perhaps that can be re-used here to perform a full cleanup of resources before reconnecting.

davidsblom:
> Similar logic exists in the `shutdown` function. Perhaps that can be re-used here to perform a full cleanup of resources before reconnecting.

Thanks! Simplified the change, and tested it again.

twitu (approved):

cjdsellers (approved):

PR #1990, opened 2024-10-07
Add support for bbo quotes in Databento adapter

faysou:
# Pull Request

Add support for bbo-1s and bbo-1m sampled quotes in databento adapter

## Type of change

usage example on prototype (use bbo-1s or bbo-1m here as schema https://github.com/nautechsystems/nautilus_trader/blob/develop/examples/backtest/databento_option_greeks.py)  , added tests

Note sure if something needs to be added for live databento data for this new data type, or how I would subscribe to live databento data from a live node.


cjdsellers (approved):
Great work, thank you for adding the tests as well!

PR #1996, opened 2024-10-11
Use Rust rate limiter for dYdX websockets

davidsblom:
# Pull Request

Use rust ratelimiter for dYdX websockets

## Type of change

pre-commit / live example


cjdsellers (approved):
There's also the option of just using a default quota if there's only one kind of sending rate limit (would be less lines of code), but this also works :+1:.

PR #1999, opened 2024-10-12
Add docs and fix rate limiting for websockets of dYdX

davidsblom:
# Pull Request

Add docs and fix rate limiting for websockets of dYdX

## Type of change

Live example

### DIFF
--- a/nautilus_trader/adapters/dydx/websocket/client.py
+++ b/nautilus_trader/adapters/dydx/websocket/client.py
@@ -171,6 +171,11 @@ class DYDXWebsocketClient:
     def _handle_ping(self, raw: bytes) -> None:
         """
         Handle ping messages by returning a pong message.
+
+        Parameters
+        ----------
+        raw : bytes
+            The received ping in bytes.
         """
         self._loop.create_task(self.send_pong(raw))
 
--- a/nautilus_trader/adapters/dydx/websocket/client.py
+++ b/nautilus_trader/adapters/dydx/websocket/client.py
@@ -529,8 +529,6 @@ class DYDXWebsocketClient:
         ----------
         msg : dict[str, Any]
             Dictionary to serialize as JSON message and send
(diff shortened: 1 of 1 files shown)


nautilus_trader/adapters/dydx/websocket/client.py
```diff
@@ -105,13 +105,28 @@ def is_disconnected(self) -> bool:
...
+
+        Parameters
+        ----------
+        item : tuple[str, str]
+            Topic name.
+
+        Returns
+        -------
+        bool
+            Whether the client is already subscribed to this topic.
```
cjdsellers, line 129: I think it's best to add only additional info relative to the return type for the description, in this case it's been covered in the main description already.

nautilus_trader/adapters/dydx/websocket/client.py
```diff
@@ -147,10 +157,26 @@ async def connect(self) -> None:
...
 
     def _msg_handler(self, raw: bytes) -> None:
+        """
+        Handle pushed websocket messages.
+
+        Parameters
+        ----------
+        raw : bytes
+            The received message in bytes.
+        """
```
cjdsellers, line 167: Small point, but we tend to leave a single blank line at the end of the docstring. The `docformatter` lint may just handle this when its turned back on.

cjdsellers (approved):

cjdsellers:
Thanks @davidsblom :pray:

nautilus_trader/adapters/dydx/websocket/client.py
davidsblom: Thanks for formatting the code already.

nautilus_trader/adapters/dydx/websocket/client.py
davidsblom: Makes sense. Thanks!

PR #2002, opened 2024-10-13
Add aggregation of bars from historical data

faysou:
# Pull Request

Add aggregation of historical bars. This is useful for using a an external source of historical data and computing up to date indicators at the start of a backtest for aggregated bars.

This change touches many parts. Also:
+ Add actor.request_aggregated_bars, that takes a list of bar_types with a commone underlying data and creates a DataRequest that can request quotes, trades or bars depending on what is needed.
+ Add start and end field in BacktestRunConfig as it's supported by the engine when using a backtest node. Queried backtest data also takes into account this field using a minimum or maximum of dates between BacktestDataConfig and BacktestRunConfig
+ Add clear_log_file_name in LoggingConfig in order to reset a log file before a backtest. This is useful when testing.
+ Add max_date_str and min_date_str functions to datetime.pyx in order to easily compute the max and min of dates in various formats (str, int)
+ Modified BarAggregator and especially time bar aggregator to be able to aggregate historical bars, quote or trade ticks
+ Added _handle_aggregated_bars in the data engine
+ added an example notebook that can be used to test and demonstrate the various aspects involved (aggregation of historical bars, updates of indicators)

## Type of change

Extensive testing using using example notebook included
Added a few tests


nautilus_trader/adapters/databento/data_utils.py
```diff
@@ -221,20 +221,60 @@ def databento_data(
...
+        Base path for the catalog.
+    write_data_mode : str, optional
+        Mode for writing data to the catalog. Default is "overwrite".
+
+    Returns
+    -------
+    dict
+        A dictionary containing:
+        - 'catalog': The loaded catalog object.
+        - 'nautilus_definition': Processed Databento definition data.
```
cjdsellers, line 254: Do we need the `'nautilus_'` prefix for these?

nautilus_trader/data/aggregation.pxd
```diff
@@ -63,6 +63,7 @@ cdef class BarAggregator:
     cdef Logger _log
     cdef BarBuilder _builder
     cdef object _handler
+    cdef object _backup_handler
```
cjdsellers, line 66: Whats the purpose of the backup handler?

nautilus_trader/core/datetime.pyx
```diff
@@ -301,3 +301,29 @@ cpdef str format_iso8601(datetime dt):
 
     cdef tuple dt_partitioned = tz_stripped.rpartition('.')
     return f"{dt_partitioned[0]}.{dt_partitioned[2][:3]}Z"
+
+
+def max_date(date1: str | int | None = None, date2: str | int | None = None) -> str | None:
```
cjdsellers, line 306: Should probably either return a `date` object or append the name with `_str` or similar to make this clearer.
Also, this function is actually handling timestamps?

nautilus_trader/common/component.pxd
```diff
@@ -288,7 +288,7 @@ cdef class MessageBus:
     cpdef void add_streaming_type(self, type cls)
     cpdef void send(self, str endpoint, msg)
     cpdef void request(self, str endpoint, Request request)
-    cpdef void response(self, Response response)
+    cpdef void response(self, Response response, bint remove_callback=*)
```
cjdsellers, line 291: Do you think the bar aggregation machinery is encroaching too much into response handling and correlation IDs here? 

I think this is only necessary because of the recursion which is being used later on. (are we also able to avoid recursion?)

Is there another way this can be handled by adding another class member?

nautilus_trader/common/actor.pyx
```diff
@@ -1397,15 +1398,18 @@ cdef class Actor(Component):
         await_partial : bool, default False
             If the bar aggregator should await the arrival of a historical partial bar prior
             to actively aggregating new bars.
+        notify_actor : bool, default True
```
cjdsellers, line 1401: It seems asymmetric to add this just for bars? When would an actor call `subscribe_bars` but not want to actually receive them on the bus?

Could this be custom functionality we're asking regular actors to accommodate?

nautilus_trader/adapters/databento/data_utils.py
faysou: Yes it means nautilus objects as opposed to databento objects. It's to make things clear as both types of objects are returned by the databento_data function.

nautilus_trader/data/aggregation.pxd
faysou: It's to temporarily save the original handler while another one is used by the batch aggregation after a data response from request_bars. This avoid moving the handler to another domain like the data engine.

nautilus_trader/core/datetime.pyx
faysou: In the code base timestamp is what is used to handle strings and ints indifferently as dates at config level. I decided to return a string so it's easier to debug at config/node level as dates are still strings there that will be converted again to timestamps.

nautilus_trader/common/component.pxd
faysou: Resolved, concatenating different aggregated bar types to a single list response instead.

nautilus_trader/common/actor.pyx
faysou: Removed as discussed

PR #2012, opened 2024-10-17
Added timeout from env variable to IB client requests

marcodambros:
# Pull Request

This very small PR include two changes:
1. It fixes a bug in IB data client logger, which always printed None instead of the instrument name
2. For some requests to IB, instead of using a default timeout of 60, the timeout can be set with an environment variable, which if does not exist, will default to 60.

## Type of change

The code was tested locally by building nautilus and using it as a dependency in a project.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/market_data.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/market_data.py
@@ -14,6 +14,7 @@
 # -------------------------------------------------------------------------------------------------
 
 import functools
+import os
 from collections.abc import Callable
 from decimal import Decimal
 from inspect import iscoroutinefunction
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @marcodambros

Thanks for the contribution!

(1) Seems like a good change.
(2) We might want to make this more first-class and incorporate the timeout interval into the configuration, maybe @rsmb7z has some thoughts?

nautilus_trader/adapters/interactive_brokers/client/market_data.py
```diff
@@ -390,7 +391,8 @@ async def get_historical_bars(
                 return []
             self._log.debug(f"reqHistoricalData: {request.req_id=}, {contract=}")
             request.handle()
-            return await self._await_request(request, timeout, default_value=[])
+            ib_timeout = int(os.environ.get("IB_REQUEST_TIMEOUT", timeout))
```
cjdsellers, line 394: Hi @marcodambros

If you can remove these lines sourcing the timeout from an env var, we can then merge this and I'll add a config option for the timeout.

cjdsellers:
Hi @marcodambros

I've gone ahead and fixed the subscription logging per this PR (and credited you in the release notes).

For the `timeout`, that's actually exposed as a `timeout` parameter already on `HistoricInteractiveBrokersClient.request_bars`, so no need for additional config as far as I can see?

cjdsellers:
Fixed with commit 52c3ba918d93725c22820167a0b6c52d164dde0e.

PR #2015, opened 2024-10-18
Remove mutable reference in cache database trait in Rust

filipmacek:
# Pull Request

- in the trait `CacheDatabaseAdapter`, we have relaxed the signature of some functions to use immutable references. This change is possible because the `load_*` functions are read-only, and the `add_*` functions just are send data into the channel, so nothing is mutated from this function
- the exceptions are load_currencies and load_positions because the `con.keys` function from Redis requires a mutable reference.


cjdsellers (approved):
:ok_hand:

PR #2035, opened 2024-11-01
Port RiskEngine to Rust

Pushkarm029:
# Pull Request
- Ports RiskEngine to Rust.
- Ports `calculate_fixed_risk_position_size` to Rust.

## TODOs:
- [ ] Portfolio.rs [in another PR]
- [ ] Find a way to verify successful executions in tests.

## Type of change

- `make cargo-test`


nautilus_core/common/src/throttler/mod.rs
```diff
@@ -42,11 +43,11 @@ impl RateLimit {
...
 /// or processing messages.
 #[derive(Clone)]
-pub struct Throttler<T, F> {
-    inner: Rc<RefCell<InnerThrottler<T, F>>>,
+pub struct Throttler<T, S, D> {
+    inner: Rc<RefCell<InnerThrottler<T, S, D>>>,
 }
 
-impl<T, F> Debug for Throttler<T, F>
+impl<T, S, D> Debug for Throttler<T, S, D>
```
twitu, line 50: Why do we need different parameters for Send and Drop when they are the same traits `Fn(T)`? Have you considered removing it?

nautilus_core/execution/src/messages/mod.rs
```diff
@@ -33,6 +33,7 @@ pub use self::{
 };
 
 #[derive(Clone, Debug, Display)]
+#[allow(clippy::large_enum_variant)] // TODO: why?
```
twitu, line 36: Yes this is a problem that will have to be fixed throughout the system. We'll tackle this once significant control flow is in Rust.

nautilus_core/risk/src/engine/tests.rs
```diff
@@ -0,0 +1,23 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------------------------------
+
+// use rstest::rstest;
+
+// -- TESTS -----------------------------------------------------------------------------------
```
twitu, line 18: Consider keeping the tests in the same module engine/mod.rs module itself since these are unit tests. Also if the engine does not require multiple files consider keeping the logic in a single `engine.rs` file.

nautilus_core/risk/src/engine/mod.rs
```diff
@@ -22,140 +22,916 @@
...
-where
-    C: Clock,
-{
+impl RiskEngine {
+    pub fn new(
+        config: RiskEngineConfig,
+        // portfolio: PortfolioFacade TODO: fix after portfolio implementation
+        clock: Rc<RefCell<dyn Clock>>,
+        cache: Rc<RefCell<Cache>>,
+        msgbus: Rc<RefCell<MessageBus>>,
```
twitu, line 74: Rc RefCell is equivalent to a Python pointer - a shared reference that is garbage collected. Common state between components will be shared this way.

nautilus_core/risk/src/engine/mod.rs
```diff
@@ -22,140 +22,916 @@
...
+        let msgbus1 = msgbus.clone();
+        let throttled_submit_order = Throttler::new(
+            config.max_order_submit.clone(),
+            clock.clone(),
+            "ORDER_SUBMIT_THROTTLER".to_string(),
+            Box::new(move |order: SubmitOrder| {
+                msgbus1
+                    .borrow_mut()
+                    .send(&Ustr::from("ExecEngine.execute"), &order);
+            }) as Box<dyn Fn(SubmitOrder)>,
```
twitu, line 85: The trick here is that the closure does not need the whole RiskEngine i.e. `&self` state. It only needs a reference to the message bus since its own job is to send a message. By cloning and moving the message bus into the closure we can avoid the inner outer pattern.

nautilus_core/risk/src/engine/mod.rs
```diff
@@ -22,140 +22,916 @@
...
+            None,
+        );
+
+        let throttled_modify_order = Throttler::<ModifyOrder, _, _>::new(
+            config.max_order_modify.clone(),
+            clock.clone(),
+            "ORDER_MODIFY_THROTTLER".to_string(),
+            Box::new(|order: _| {}) as Box<dyn Fn(ModifyOrder)>,
```
twitu, line 93: TODO

nautilus_core/common/src/throttler/mod.rs
Pushkarm029: Previously, I was registering two different functions with the same parameters, which was causing an error. The compiler requires both functions to be identical, so I made this change. However, since we're now using closures as callbacks, this is no longer needed, so I have removed it.

cjdsellers (approved):
Great work here!

PR #2044, opened 2024-11-07
Implement websocket reconnection retries

davidsblom:
# Pull Request

Keep retrying to establish a websocket connection, even when the reconnection initially fails. The current behaviour is that the while loop exists, and the connection is broken completely.

Live example, turning on and off the network connecting.

### DIFF
--- a/nautilus_core/network/src/python/websocket.rs
+++ b/nautilus_core/network/src/python/websocket.rs
@@ -36,7 +36,7 @@ fn to_websocket_pyerr(e: tokio_tungstenite::tungstenite::Error) -> PyErr {
 #[pymethods]
 impl WebSocketConfig {
     #[new]
-    #[pyo3(signature = (url, handler, headers, heartbeat=None, heartbeat_msg=None, ping_handler=None))]
+    #[pyo3(signature = (url, handler, headers, heartbeat=None, heartbeat_msg=None, ping_handler=None, max_reconnection_tries=3))]
     fn py_new(
         url: String,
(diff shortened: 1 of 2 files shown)


twitu (commented):
Is it a good idea to keep retrying the connection without end?

Perhaps it can be made configurable. Just like the heartbeat connection retry can take an Option<u64> to decide whether and how many times to retry.

davidsblom:
> Is it a good idea to keep retrying the connection without end?
> 
> Perhaps it can be made configurable. Just like the heartbeat connection retry can take an Option to decide whether and how many times to retry.

Yeah, thought about that too. Is the heartbeat connection retried when failed? Couldn't find it. Do you have a link to the source as inspiration?

twitu:
https://github.com/nautechsystems/nautilus_trader/blob/a4c166041049ca894107fc545396e62212212564/nautilus_core/network/src/websocket.rs#L53

https://github.com/nautechsystems/nautilus_trader/blob/a4c166041049ca894107fc545396e62212212564/nautilus_core/network/src/websocket.rs#L132-L142

It's used to configure the frequency of hearbeats. But the same pattern can also be used to configure how many times to retry connection and not to retry connection for the None case.

davidsblom:
@twitu  @cjdsellers  I've added the retry logic. Could you give it another review pass?

cjdsellers:
Many thanks @davidsblom :pray:

cjdsellers (approved):

PR #2058, opened 2024-11-13
Port Portfolio to Rust

Pushkarm029:
# Pull Request

## Type of change

### DIFF
--- a/nautilus_core/portfolio/src/portfolio.rs
+++ b/nautilus_core/portfolio/src/portfolio.rs
@@ -15,7 +15,6 @@
 
 //! Provides a generic `Portfolio` for all environments.
 
-// Under development
 // improve error handling: TODO
 
 use std::{
@@ -145,10 +144,7 @@ impl MessageHandler for UpdateAccountHandler {
     }
 }
 
-pub struct Portfolio {
-    clock: Rc<RefCell<dyn Clock>>,
-    cache: Rc<RefCell<Cache>>,
-    msgbus: Rc<RefCell<MessageBus>>,
+pub struct PortfolioState {
     accounts: AccountsManager,
     analyzer: PortfolioAnalyzer,
     unrealized_pnls: HashMap<InstrumentId, Money>,
@@ -158,15 +154,10 @@ pub struct Portfolio {
     initialized: bool,
 }
 
-impl Portfolio {
-    pub fn new(
-        msgbus: Rc<RefCell<MessageBus>>,
-        cache: Rc<RefCell<Cache>>,
-        clock: Rc<RefCell<dyn Clock>>,
-    ) -> Self {
+impl PortfolioState {
+    fn new(clock: Rc<RefCell<dyn Clock>>, cache: Rc<RefCell<Cache>>) -> Self {
         let mut analyzer = PortfolioAnalyzer::new();
 
-        // Register default statistics
         analyzer.register_statistic(Arc::new(MaxWinner {}));
         analyzer.register_statistic(Arc::new(AvgWinner {}));
         analyzer.register_statistic(Arc::new(MinWinner {}));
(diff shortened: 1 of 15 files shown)


nautilus_core/portfolio/src/portfolio.rs
```diff
@@ -76,94 +155,616 @@ impl Portfolio {
...
+        }
+
+        let orders_open = borrowed_cache.orders_open(None, Some(&quote.instrument_id), None, None);
+
+        // todo
+        // Initialize initial (order) margin
+        // let result_init = self.accounts.
     }
 
     pub fn update_account(&mut self, event: &AccountState) {
```
twitu, line 675: `update_account` only needs a reference to the cache. It does not need access to the full state of the Portfolio. We can avoid the inner outer pattern by moving a clone of the cache the update account closure. Just like we did message bus in RiskEngine.

twitu (commented):
btw why do we need the pub type Statistic to be Arc<dyn ...> when it's supposed to be used single threaded?

faysou:
@Pushkarm029 you can rebase your dev branch and do a force push instead of merging with the dev branch which adds a lot of unnecessary commits.

I use this shell script which is quite practical, it has an optional argument to reset one or more commits as well:

```bash
custom_rebase() {
    git fetch origin develop:develop

    git stash
    git rebase develop
    git stash pop

    local commits=${1:-0}
    if [ "$commits" -gt 0 ]; then
        git reset --soft HEAD~"$commits"
    fi
}
```

nautilus_core/portfolio/src/portfolio.rs
```diff
@@ -54,9 +159,128 @@ pub struct Portfolio {
...
+        Rc::try_unwrap(portfolio_rc)
+            .map_err(|_| "Failed to unwrap Portfolio")
+            .unwrap()
+            .into_inner()
+    }
+
+    fn register_message_handlers(
+        msgbus: &Rc<RefCell<MessageBus>>,
+        cache: &Rc<RefCell<Cache>>,
+        portfolio: &Rc<RefCell<Self>>,
```
twitu, line 212: We cannot create an rc refcell of self from inside the new function. We will have to use an inner outer pattern here.

Also no need to pass rc refcell by & reference. Pass it by cloning it.

nautilus_core/portfolio/src/portfolio.rs
```diff
@@ -71,138 +295,965 @@ impl Portfolio {
...
+        log::info!(
+            "Initialized {} open position{}",
+            open_count,
+            if open_count == 1 { "" } else { "s" }
+        );
     }
 
     pub fn update_quote_tick(&mut self, quote: &QuoteTick) {
-        todo!()
+        self.unrealized_pnls.remove(&quote.instrument_id);
```
twitu, line 783: update quote tick uses self fields that are not behind and Rc Refcell which is why it won't work by just passing the cache or the message bus. The other update functions too use parts of self that are not cloneable references. So it seems like an inner outer style must be implemented.

nautilus_core/portfolio/src/portfolio.rs
```diff
@@ -16,36 +16,141 @@
...
     clock: Rc<RefCell<dyn Clock>>,
     cache: Rc<RefCell<Cache>>,
     msgbus: Rc<RefCell<MessageBus>>,
-    accounts: HashMap<Venue, AccountAny>,
+    accounts: AccountsManager,
     analyzer: PortfolioAnalyzer,
-    // venue: Option<Venue>, // Added for completeness but was meant to be a "temporary hack"
     unrealized_pnls: HashMap<InstrumentId, Money>,
     realized_pnls: HashMap<InstrumentId, Money>,
     net_positions: HashMap<InstrumentId, Decimal>,
```
twitu, line 156: Perhaps for portfolio we can take a slightly different approach to inner outer. These lines are the actual inner state of the portfolio.  Perhaps they can be wrapped into their own struct, and rc ref celled on this Portfolio struct (which will act as outer).

```rust
struct PorfolioState {
    accounts: AccountsManager,
    analyzer: PortfolioAnalyzer,
    unrealized_pnls: HashMap<InstrumentId, Money>,
    realized_pnls: HashMap<InstrumentId, Money>,
    net_positions: HashMap<InstrumentId, Decimal>,
}

struct Porfolio {
  ...
  inner. Rc<RefCell<PortfolioState>>
```

This way you can clone the portfolio state and pass it to the closure and relevant functions without needing to wrap the whole portfolio struct which might be cumbersome. However, it will mean that all the functions that the closures call will have to be changed slightly to receive the portfolio state, message bus, cache as separate parameter.

```
update_quote_tick(&self, ,.....)
update_qutote_tick(cache, messagebus, portfolio_state, ....)

---

To abstract away the borrow and borrow mut calls. You can also consider an additional thin helper layer.

```rust
struct PorfolioWrapper {
   inner: Rc<RefCell<PortfolioState>>,
}

struct Porfolio {
    inner: PortfolioWrapper
 }
```

For PortfolioWrapper you can consider implementing something like these helper methods to easily access the inner fields like

```rust
impl PortfolioWrapper {
    fn accounts(&self) -> impl Deref<Target = &AccountsManager> {
        self.inner.borrow().accounts
    }
}
```

twitu (commented):
Yup, we need the inner outer pattern for some of the methods.

PR #2060, opened 2024-11-17
Add Python type stubs

Emsu:
# Pull Request

Includes typing files I use to run databento strategies

## Type of change


cjdsellers:
Hi @Emsu 

Thanks for the contribution! this was definitely needed.

Looks like some `pre-commit` errors in CI.

You can run with:
```
pre-commit run --all-files
```
or using Make:
```
make pre-commit
```

You can also install a git hook which will run it automatically on commits with:
```
pre-commit install
```

cjdsellers:
Hey @Emsu

I pulled this locally and had a go at fixing.

It turns out it's not just issues with pre-commit formatting, but there are hundreds of mypy typing issues of different sorts which could potentially take hours to untangle. Also, these stubs are for the Cython modules which we're actively working to remove, so it would end up being a large amount of work for a temporary improvement.

Ideally we'd want auto-generated type stubs as part of the build (similar to how `cbindgen` is working right now for the FFI). This is something I'm going to start looking into for the dynamic libs compiling from Rust which have no FFI (such as adapters). The current approach of hand crafting the single `nautilus_pyo3.pyi.` is definitely unsustainable.

Soon I'll be adding more re-exports and module declarations so we can at least start nailing down the public API ahead of pyo3 becoming the standard for Python bindings.

Thank you again for your work here, but closing this for now anticipating you won't have the bandwidth to complete this.
Feel free to continue the discussion here if you like, and if there's something I didn't consider or you figure out solutions then we can always reopen, or alternatively creating another PR is also fine.

PR #2069, opened 2024-11-22
Add support of stop limit and stop market orders for dYdX

Saransh-Bhandari:
This PR introduces support for **Stop Limit** and **Stop Market** order types in the `nautilus_trader` dYdX adapter. Key updates include:
- Added the `is_conditional_order` flag to identify conditional orders.
- Mapped new `DYDXOrderStatus.UNTRIGGERED` to `OrderStatus.ACCEPTED`.
- Updated `_submit_order_single` to support `STOP_LIMIT` and `STOP_MARKET` order types, including handling `trigger_price`.
- Introduced a new method `calculate_conditional_order_trigger_subticks` to compute trigger prices in subticks for conditional orders.
- Extended the `create_order` method to include `conditional_order_trigger_subticks` and `trigger_price`.


nautilus_trader/adapters/dydx/execution.py
```diff
@@ -1027,7 +1027,7 @@ async def _submit_order_single(self, order) -> None:
             )
             return
 
-        if dydx_order_tags.is_short_term_order:
+        if dydx_order_tags.is_short_term_order and not dydx_order_tags.is_conditional_order:
```
davidsblom, line 1030: Wondering why an additional order tag is required. Can we also use the order type directly?

nautilus_trader/adapters/dydx/execution.py
```diff
@@ -1071,11 +1078,18 @@ async def _submit_order_single(self, order) -> None:
         }
 
         price = 0
+        trigger_price = 0
```
davidsblom, line 1081: Perhaps merge the latest develop branch into your branch. I've made similar changes too but haven't gotten to thoroughly testing it yet.

davidsblom (commented):
Thanks for the PR! Much appreciated. Did you manage to test this with the testnet or production network?

cjdsellers:
Hi @Saransh-Bhandari and @Akshay-deqode

Thanks for the contribution! we just need the CLA signing to be able to merge (see the CLAassistant comment above).

nautilus_trader/adapters/dydx/execution.py
Akshay-deqode: we can create a condition with order type. but as the structure followed in the code with the order tag we used the same if required we can change it .
please let us know what you suggest.

nautilus_trader/adapters/dydx/execution.py
Akshay-deqode: yes we  merged the develop branch in to ours.
we are currently testing it.

davidsblom: Great! Thanks.

nautilus_trader/adapters/dydx/execution.py
```diff
@@ -1090,7 +1098,10 @@ async def _submit_order_single(self, order: Order) -> None:
         if order.order_type == OrderType.LIMIT:
             price = order.price.as_double()
         elif order.order_type == OrderType.MARKET:
-            price = 0
+            if order.side == OrderSide.BUY:
```
davidsblom, line 1101: I think we need to make this more generic. This the explanation from support of dydx:

There are a few possible reasons why the SDK allows setting a price for a market order:

1. Price slippage protection: By setting a price for a market order, you can limit the potential price slippage. For example, if you set a price of $100 for a market buy order, the order will only be executed if the market price is at or below $100. If the market price is above $100, the order will not be executed.

Akshay-deqode: yes actually I just added this to test if it working or not.
for a fix we can have a oracle price with +- 0.1% (configurable) slippage.

nautilus_trader/adapters/dydx/grpc/order_builder.py
```diff
@@ -240,6 +240,7 @@ def create_order(
         good_til_block: int | None = None,
         good_til_block_time: int | None = None,
         execution: OrderExecution = OrderExecution.DEFAULT,
+        conditional_order_trigger_subticks: int = 0,
```
davidsblom, line 243: I believe this one can be removed. It is overwritten by trigger_price.

nautilus_trader/adapters/dydx/common/common.py
```diff
@@ -24,6 +24,7 @@ class DYDXOrderTags(NautilusConfig, frozen=True, repr_omit_defaults=True):
 
     is_short_term_order: bool = True
     num_blocks_open: int = 20
+    is_conditional_order: bool = False
```
davidsblom, line 27: This flag is probably not necessary. Only checking the order type is more robust if the end user forgets to set this property.

Akshay-deqode: Yes , I feel so ,this can be simplified using the order type only I am updating it accordingly

davidsblom (approved):

PR #2072, opened 2024-11-24
Implement high-precision 128-bit value types

twitu:
# Pull Request

- Use 128-bit integers to represent higher precisions for value types.
- Use `high-precision` feature flag to switch between lower and higher precision.

[See RFC](https://github.com/nautechsystems/nautilus_trader/issues/2084).


nautilus_trader/model/objects.pyx
```diff
@@ -67,8 +73,10 @@ PRICE_MIN = RUST_PRICE_MIN
 MONEY_MAX = RUST_MONEY_MAX
 MONEY_MIN = RUST_MONEY_MIN
 
-FIXED_PRECISION = RUST_FIXED_PRECISION
-FIXED_SCALAR = RUST_FIXED_SCALAR
+HIGH_PRECISION = PRECISION == 16
```
twitu, line 76: It'll be a safer option to load this from the cdef HIGH_PRECISION value defined in model.pxd. Just in case the digits of precision change for high precision it'll very difficult to track down why this is failing.

cjdsellers: Also had a similar thought, had to add another constant generated at compile time, as `HIGH_PRECISION` was already occupied with the DEF, but now achieving what we were after:

```
HIGH_PRECISION = RUST_HIGH_PRECISION_MODE
```

nautilus_core/model/src/data/trade.rs
```diff
@@ -191,8 +186,9 @@ mod tests {
         types::{Price, Quantity},
     };
 
+    #[cfg(feature = "high-precision")] // TODO: Add 64-bit precision version of test
```
cjdsellers, line 189: We'll do another sweep of the tests once merged.

nautilus_core/model/src/orderbook/book.rs
```diff
@@ -812,12 +812,13 @@ mod tests {
...
         assert_eq!(
             book.get_avg_px_qty_for_exposure(qty, OrderSide::Buy),
-            (100.0, 10_000_000.0, 100.0)
-        );
-        assert_eq!(
-            book.get_avg_px_qty_for_exposure(qty, OrderSide::Sell),
-            (99.0, 10_101_010.0, 99.0)
+            (100.0, 0.01, 100.0)
         );
+        // TODO: Revisit calculations
```
cjdsellers, line 817: This was outputting an unscaled raw quantity, we probably want the scaled quantity.

nautilus_core/model/src/types/currency.rs
```diff
@@ -15,7 +15,7 @@
 
 //! Represents a medium of exchange in a specified denomination with a fixed decimal precision.
 //!
-//! Handles up to 9 decimals of precision.
+//! Handles up to 16 decimals of precision.
```
cjdsellers, line 18: Will need to tidy up the docs some more depending on precision mode.

nautilus_trader/core/correctness.pyx
```diff
@@ -1201,13 +1200,13 @@ class PyCondition:
         Condition.positive(value, param, ex_type)
 
     @staticmethod
-    def positive_int(int64_t value, str param, ex_type = None):
+    def positive_int(value: int, str param, ex_type = None):
```
cjdsellers, line 1203: Relaxed all of these C types, as they're coming from Python in non-performance critical function calls.
Avoids having to name the specific integer types in C.

filipmacek (approved):

PR #2088, opened 2024-12-06
Fix specifying the price for market orders on dYdX

davidsblom:
# Pull Request

Fix specifying the price for market orders on dYdX

## Type of change

Live example and unit tests


davidsblom:
Yes good idea. Maybe have two options to let a strategy set the price, and fall back to using a slippage parameter?

cjdsellers (approved):
Great, and thank you for adding the tests!

PR #2092, opened 2024-12-08
Avoid C++ keyword conflict by renaming `delete` method

sunlei:
# Pull Request

This PR resolves a compilation issue that occurs when the project is built in a C++ environment.

Because I use some C++ features in pyx.
If a `.pyx` file uses the following directive:

```cython
# distutils: language = c++
from nautilus_trader.model.book cimport OrderBook
```

This directive configures the compiler to use C++ instead of C. When combined with `cimport` statements that include modules from nautilus_trader, the `delete` method conflicts with the reserved C++ keyword `delete`, resulting in a syntax error during compilation.

Specifically, the method name `delete` is interpreted as a reserved keyword in C++, leading to the following error:

```
error: expected member name or ';' after declaration specifiers
   void (*delete)(struct __pyx_obj_15nautilus_trader_5model_4book_OrderBook *, struct __pyx_obj_15nautilus_trader_5model_4data_BookOrder *, uint64_t, int __pyx_skip_dispatch, struct __pyx_opt_args_15nautilus_trader_5model_4book_9OrderBook_delete *__pyx_optional_args);
   ~~~~   ^
```

To address this issue, I have renamed the method `OrderBook.delete` to `OrderBook.remove`.
Additionally, for consistency, I standardized related naming conventions by making the following changes:

- `BookAction.Delete` -> `BookAction.Remove`
- `orderbook_delete` -> `orderbook_remove`

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


cjdsellers:
Hi @sunlei

Unfortunately, we can’t accommodate such a significant breaking change for an unsupported language (C++).

Would it be possible to use aliasing for this custom build instead? Feel free to discuss further in this PR.

sunlei:
Thank you for your response. Let me clarify the situation further to explain the context and the reasoning behind this change.

I am working on implementing a trading strategy using Cython. For this, I need to import `Strategy` in my `mm.pyx` file like this:

```cython
from nautilus_trader.trading.strategy cimport Strategy
from mm.utils cimport a_cpp_func
```

Additionally, my `mm.pyx` file imports a function implemented in another .pyx file (`mm/utils.pyx`) that uses C++ features. To compile this, I have to set the compilation directive for the project to use C++ in `mm/utils.pyx`.

However, when C++ is enabled, the delete method in `OrderBook` conflicts with the reserved C++ keyword `delete`.

This error makes it impossible to build the project when `mm.pyx` depends on both `Strategy` and C++ features. Renaming the `delete` method (and related naming conventions) to avoid this reserved keyword seemed like the best way to address this issue comprehensively.

If the method name remains `delete`, users implementing strategies with Cython would be forced to use only C.

PR #2093, opened 2024-12-08
Add sccache in CI

sunlei:
# Pull Request

Use `sccache` to speedup CI

- https://github.com/mozilla/sccache

Execution stats in task `Post Run sccache-cache`.

# But

When using GitHub Actions cache as the storage for sccache, sccache frequently hits the service's rate limit, resulting in the following error: `Failed to restore: Cache service responded with 429`。

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -17,6 +17,7 @@ jobs:
     env:
       # https://github.com/Mozilla-Actions/sccache-action
       SCCACHE_GHA_ENABLED: "true"
+      SCCACHE_CACHE_MULTIARCH: 1
       RUSTC_WRAPPER: "sccache"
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @sunlei

Thanks for the contribution - I think this is a great idea!

That said, we want to maintain CI reliability and avoid introducing any 429 flakiness. Do you have any thoughts on how we might address this? Perhaps S3 might work as an alternative storage option?

sunlei:
1. `sccache` reaches the rate limit of the service, the build will continue, but the storage might not be performed. So 429 will not affect the execution of CI.
2. S3 may be able to solve the problem, but the read and write charges of S3 may incur higher costs.

cjdsellers:
Lets give it a go, many thanks.

PR #2102, opened 2024-12-12
Add additional API functionality for Bybit

sunlei:
# Pull Request

Add more API for Bybit.

## Type of change


cjdsellers (approved):
Thanks for the improvements!

One thing I'll call out is that it's difficult for me to verify the changes are correct from visual inspection, and without many automated tests. I'm not suggesting we spend a lot of bandwidth adding those tests right now, but something to consider.

I know you're actively using this adapter, which is the ideal scenario to maintain and improve them.

PR #2108, opened 2024-12-14
Use exchange as venue for databento instrument ids instead of GLBX

faysou:
# Pull Request

Use exchange as venue for databento instrument ids instead of GLBX

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update


nautilus_trader/adapters/databento/data.py
```diff
@@ -392,7 +392,7 @@ async def _subscribe_imbalance(self, data_type: DataType) -> None:
             live_client = self._get_live_client(dataset)
             live_client.subscribe(
                 schema=DatabentoSchema.IMBALANCE.value,
-                symbols=[instrument_id.symbol.value],
+                instrument_ids=[instrument_id],
```
cjdsellers, line 395: We also need to convert these instrument IDs to the `nautilus_pyo3.InstrumentId` version.

PR #2129, opened 2024-12-18
Add WebSocket API trading support for Bybit

sunlei:
# Pull Request

This PR introduces the ability to trade on Bybit using the [WebSocket API](https://bybit-exchange.github.io/docs/v5/websocket/trade/guideline).
The implementation is controlled by the following parameters in `BybitExecClientConfig`:

- `use_ws_trade_api` : bool, default False
	- Enable or disable the use of the WebSocket API for trading. Use WS API if `True` else HTTP API.
- `use_http_batch_api` : bool, default False
	- Bybit's WebSocket trading API currently does not support batch operations, only single-order handling. This parameter provides a toggle to use the HTTP batch API for order batch operations instead of the WebSocket API.
- `ws_trade_timeout_secs` : float, default 5.0
	- Specifies the timeout in seconds to wait for `ACK` message from the WebSocket trading API.
- `base_url_ws_private` : str, optional
	- Customizable WebSocket URL for private channel.
- `base_url_ws_trade` : str, optional
	- Customizable WebSocket URL for trading channel.

## Type of change

I replaced `nautilus_trader.adapters.bybit` and all its submodules with the local `monkey_patch.adapters.bybit` using `monkey patching`.
This setup has been running in my production strategy environment for about a week without any issues observed so far.

### DIFF
--- a/examples/live/bybit/bybit_ema_cross.py
+++ b/examples/live/bybit/bybit_ema_cross.py
@@ -80,7 +80,7 @@ config_node = TradingNodeConfig(
             api_key=None,  # 'BYBIT_API_KEY' env var
             api_secret=None,  # 'BYBIT_API_SECRET' env var
             base_url_http=None,  # Override with custom endpoint
-            base_url_ws=None,  # Override with custom endpoint
+            base_url_ws_private=None,  # Override with custom endpoint
             instrument_provider=InstrumentProviderConfig(load_all=True),
             product_types=[product_type],
             testnet=False,  # If client uses the testnet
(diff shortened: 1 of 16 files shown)


cjdsellers:
Many thanks @sunlei these are great changes!

filipmacek:
@sunlei Awesome job. Do you have any metrics around the round-trip latency difference between HTTP API and WSS API when placing orders? It would be interesting to compare them. @cjdsellers and I were actually talking about this a few weeks ago.

sunlei:
@filipmacek 

I haven’t measured the round-trip latency yet. Currently, the metrics in my strategy focus primarily on internal computation time and overall operation duration.

These statistics are written to a file, and then I use Vector to convert the JSON-formatted log files into the Logfmt format, which is eventually sent to VictoriaMetrics.

My original plan was to calculate the round-trip latency after implementing WebSocket-based order placement by tracking events like on_order_accepted, on_order_canceled, and on_order_submitted. Once I gather these statistics, I can share the results publicly.

PR #2144, opened 2024-12-25
Allow bar aggregators to persist after request_aggregated_bars 

faysou:
# Pull Request

Allow bar aggregators to persist after request_aggregated_bars for future subscriptions

## Type of change

- Breaking change: change of argument name from update_existing_subscriptions to update_subscriptions

Existing tests

### DIFF
--- a/nautilus_trader/data/aggregation.pxd
+++ b/nautilus_trader/data/aggregation.pxd
@@ -66,6 +66,7 @@ cdef class BarAggregator:
     cdef object _handler_backup
     cdef bint _await_partial
     cdef bint _batch_mode
+    cdef public bint is_running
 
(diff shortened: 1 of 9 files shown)


nautilus_trader/data/aggregation.pxd
```diff
@@ -66,6 +66,7 @@ cdef class BarAggregator:
     cdef object _handler_backup
     cdef bint _await_partial
     cdef bint _batch_mode
+    cdef public bint is_live
```
cjdsellers, line 69: Could something like `is_running` more clearly distinguish from an environment context? (not backtest vs live etc)

faysou: ok, I've modified to is_running

PR #2148, opened 2024-12-29
Refine parsing candles for dYdX

davidsblom:
# Pull Request

Refine parsing candles for dYdX

## Type of change

Unit tests

### DIFF
--- a/nautilus_trader/adapters/dydx/schemas/ws.py
+++ b/nautilus_trader/adapters/dydx/schemas/ws.py
@@ -98,15 +98,14 @@ class DYDXCandle(msgspec.Struct, forbid_unknown_fields=True):
         high_price = Price(Decimal(self.high), price_precision)
         low_price = Price(Decimal(self.low), price_precision)
         close_price = Price(Decimal(self.close), price_precision)
-        avg_price = Decimal("0.25") * (open_price + high_price + low_price + close_price)
-        volume = Decimal(self.usdVolume) / avg_price
+        volume = Quantity(Decimal(self.baseTokenVolume), size_precision)
         return Bar(
             bar_type=bar_type,
             open=open_price,
             high=high_price,
             low=low_price,
             close=close_price,
-            volume=Quantity(volume, size_precision),
+            volume=volume,
             ts_event=dt_to_unix_nanos(self.startedAt),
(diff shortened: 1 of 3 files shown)


nautilus_trader/adapters/dydx/schemas/ws.py
```diff
@@ -98,15 +98,14 @@ def parse_to_bar(
         high_price = Price(Decimal(self.high), price_precision)
         low_price = Price(Decimal(self.low), price_precision)
         close_price = Price(Decimal(self.close), price_precision)
-        avg_price = Decimal("0.25") * (open_price + high_price + low_price + close_price)
-        volume = Decimal(self.usdVolume) / avg_price
+        volume = Quantity(Decimal(self.baseTokenVolume), size_precision)
```
cjdsellers, line 101: This is an interesting choice, because most crypto API's give you both base and quote volumes - but our bar type only has one volume.

After looking around a bit, for spot markets the base volume is common, equities use units of shares, futures is contracts, derivatives markets are often notional value (in quote currency).

I'm still happy to merge this in if you find the base volume more useful in practice on dYdX, or do we want to consider further? I'm thinking a configuration, or per asset class choice - and specify in docs.

davidsblom: Hi Chris, don't have a preference here. What do the other adapters use? 

In develop branch, it is also using base volume by converting from quote to base. So I thought to simplify that. If derivatives markets use notional value, we could do that too but it would be a breaking change.

davidsblom: Making it configurable is indeed preferred I guess. I'll add that then and keep the current default base volume?

cjdsellers: I see, looking closer there is that `/ avg_price`.

I think yet another config option is a nice to have - I'm happy to merge as is for now?

davidsblom: Sure, feel free to merge if you're happy with the change. 

One question that pooped up, is what does the backtesting engine expect in terms of volume. I think base volume but not completely sure.

PR #2149, opened 2024-12-29
Refine imports for type hints in Bybit

sunlei:
# Pull Request

Refine imports for type hints in Bybit.

## Type of change

### DIFF
--- a/nautilus_trader/adapters/bybit/common/enums.py
+++ b/nautilus_trader/adapters/bybit/common/enums.py
@@ -13,11 +13,13 @@
 #  limitations under the License.
 # -------------------------------------------------------------------------------------------------
 
+from __future__ import annotations
+
 from enum import Enum
 from enum import unique
+from typing import TYPE_CHECKING
 
 from nautilus_trader.core.nautilus_pyo3 import PositionSide
-from nautilus_trader.model.data import BarType
 from nautilus_trader.model.enums import BarAggregation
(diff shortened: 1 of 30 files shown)


cjdsellers (commented):
Are we using `TYPE_CHECKING` for performance reasons here?
I'd be interested to know the improvements if you have any before-after figures.

sunlei:
@cjdsellers 

I haven’t conducted performance tests for this specific scenario, but theoretically, it should be beneficial:

- Isolate type-checking code.
- Avoid unnecessary runtime imports.
- Reduce the likelihood of circular references.

cjdsellers:
Sure, happy to merge - but thoughts are:

- We haven't experienced any circular import issues for Bybit, that I know of.
- Departs from the standard import patterns in most other Python modules.
- Could be considered boilerplate unless a strong reason for it?

A reason in favor is the potential runtime performance improvement, would be interesting to know because it could inform heading this way for other modules too.

sunlei:
@cjdsellers 

Thank you!

- Bybit previously encountered a circular reference issue in one of its PRs, which I resolved using TYPE_CHECKING.
- I randomly checked two projects, PyTorch and Flask, and found that TYPE_CHECKING is used quite extensively.
- In my own projects, TYPE_CHECKING often serves as boilerplate code.

- For further reference:
  - https://mypy.readthedocs.io/en/stable/runtime_troubles.html

Issue #2150, opened 2024-12-29
Slow `nautilus_trader.common.component.is_matching`


ryantam626:
# Bug Report

### Expected Behavior
Expected `nautilus_trader.common.component.is_matching` to be faster.

### Actual Behavior
`nautilus_trader.common.component.is_matching` is prohibitively slow when dealing with instruments with long names (e.g. Polymarket - which results in topic like `data.book.deltas.POLYMARKET.0xf6ea0a571a43a122471e1c7604d6bd8c267082a39758bdd8162dbfae0cd8d137-40739562057794550541187194569338873089092359730180442205123214215534855241164`)

### Steps to Reproduce the Problem

```py
from nautilus_trader.common.component import is_matching_py

topic = 'data.instrument.POLYMARKET.0x4234768b3ad78225ec4cacf078a98044a880d1f0abf51997f22da9670b4a95e4-73174005964727253088253639382577689462665540891931226338645962800718025375512'
pattern = 'data.instrument.POLYMARKET.0x034e00691eb26f6023c35d196e4da700d440e39263957b54ee613b42bac0b9cc-18339512646452261434884166324269813378889037675218398618252803207699512685316'

# ipyton timeit magic
%timeit is_matching_py(topic, pattern)
```

gives
```text
82.7 μs ± 5.7 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

### Further explaination
`82.7 μs` doesn't sound too bad when considered in isolation (also there is further optimisation if we don't cross Cython boundary - let's ignore that for the purpose of this ticket).

But if we want to trade every single instrument in Polymarket, then we could have about 30000+ instruments (this figure was correct when I grabbed all instruments from Polymarket for testing a while back - could be wrong now).

When we want to subscribe to a new topic, we would attempt to match the topic to be subscribed against all patterns in the message bus, of which is filled with long patterns like the above, this would take about `30000 * 82μs` = `2.46s` which is too expensive. 

### Specifications

- OS platform: Windows WSL (`uname -r` output: `5.15.167.4-microsoft-standard-WSL2`)
- Python version: 3.11.10
- `nautilus_trader` version: 1.205.0 (from github and built locally, debug build)

ryantam626:
Proposed fix:

```patch
From 823616a7b1c12f0dc4c654b6809c68a755f7197e Mon Sep 17 00:00:00 2001
From: Ryan Tam <ryantam626@gmail.com>
[...]

PR #2151, opened 2024-12-29
Improve message bus topic matching performance

ryantam626:
# Pull Request

Improve `nautilus_trader.common.component.is_matching` exact match performance as suggested in #2150.

Also adds some more test cases.

## Type of change

Bug fix (non-breaking change which fixes an issue)

Existing tests are almost sufficient.

For completeness, I have also added some more tests for
1. `?` pattern which we do support
2. `[seq]` and `[!seq]` pattern which do NOT support

### DIFF
--- a/nautilus_trader/common/component.pyx
+++ b/nautilus_trader/common/component.pyx
@@ -2596,7 +2596,17 @@ cdef class MessageBus:
         return subs_array
 
 
+cdef inline bint contains_wildcard(str topic_or_pattern):
+    return '?' in topic_or_pattern or '*' in topic_or_pattern
+
+
 cdef inline bint is_matching(str topic, str pattern):
+    topic_contains_wildcard = contains_wildcard(topic)
+    pattern_contains_wildcard = contains_wildcard(pattern)
+
+    if not topic_contains_wildcard and not pattern_contains_wildcard:
+        return topic == pattern
+
     # Get length of string and wildcard pattern
     cdef int n = len(topic)
     cdef int m = len(pattern)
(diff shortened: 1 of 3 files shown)


ryantam626:
Actually I am not certain if `topic_contains_wildcard = contains_wildcard(topic)` and its check is required.

I was under the impression that it's required, because of this test https://github.com/nautechsystems/nautilus_trader/blob/c53f87ac40f0907b4313d826c5e0637a30665d01/tests/unit_tests/msgbus/test_bus.py#L463 in particular.

However upon trying to add more test which flips all the existing tests' `topic` and `pattern` around, they all failed, which led me to think that `topic_contains_wildcard = contains_wildcard(topic)` is actually not required - if that's the case then I think this can almost be reduced to a `fnmatch.fnmatch` (well depending on if we want to support other shell-style patterns?)

cjdsellers:
Hey @ryantam626 

There is a chance the implementation is not correct. We only have to preserve the current functionality, and this will be revisited and optimized in Rust later (current impl [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_core/common/src/msgbus/mod.rs)).

ryantam626:
Whoops, forgot to follow the contributor guide, will fix the CI failures.

ryantam626:
Should be fixed now, sorry for the repeated force push 😓 

FWIW, I think it probably is correct - I just got the wrong end of the stick about the purpose of the `["*", "*", True]` test!

I will give this patch a spin and follow up with another PR if I find a better solution in the future.

nautilus_trader/common/component.pyx
```diff
@@ -2596,7 +2596,17 @@ cdef class MessageBus:
         return subs_array
 
 
+cdef inline bint contains_wildcard(str topic_or_pattern):
+    return '?' in topic_or_pattern or '*' in topic_or_pattern
+
+
 cdef inline bint is_matching(str topic, str pattern):
+    topic_contains_wildcard = contains_wildcard(topic)
```
cjdsellers, line 2604: Small Cython thing: if a local variable isn't typed with a `cdef` then the Python interpreter is needed to figure out the type. Will merge like this though and optimize on `develop`.

PR #2173, opened 2025-01-04
Data migration script and guide

twitu:
# Pull Request

Add data migration script and guide

## Type of change

Tested a round trip


faysou:
"are to applications in the nautilus_persistence crate". I think it should be two.

PR #2174, opened 2025-01-05
Implement trailing stop logic in Rust

DeevsDeevs:
# Pull Request

Implement trailing stop logic in Rust

## Type of change

Added lots of tests

### DIFF
--- a/nautilus_core/execution/src/trailing.rs
+++ b/nautilus_core/execution/src/trailing.rs
@@ -1,0 +1,14 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2025 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
+//  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+//  You may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------------------------------
--- a/nautilus_core/execution/src/trailing.rs
+++ b/nautilus_core/execution/src/trailing.rs
@@ -60,15 +60,15 @@ pub fn trailing_stop_calculate(
                     .unwrap_or_default(),
                 last,
             )?;
-            println!("temp_trigger_price: {:?}", temp_trigger_price);
-
-            println!("order.order_side(): {:?}", order.order_side());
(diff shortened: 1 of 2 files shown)


nautilus_core/execution/src/trailing.rs
```diff
@@ -0,0 +1,497 @@
...
+// -------------------------------------------------------------------------------------------------
+
+use nautilus_model::{
+    enums::{OrderSide, OrderType, TrailingOffsetType, TriggerType},
+    orders::{base::OrderError, OrderAny},
+    types::Price,
+};
+
+/// Provides trailing stop calculation functionality
+pub struct TrailingStopCalculator;
```
cjdsellers, line 23: Because this struct has no state, we can probably just make these free functions?

nautilus_core/execution/src/trailing.rs
```diff
@@ -0,0 +1,497 @@
...
+                        .map(|p| p.as_f64())
+                        .unwrap_or_default(),
+                    bid,
+                    ask,
+                )?;
+
+                match order.order_side() {
+                    OrderSide::Buy => {
+                        if order.trigger_price().is_none()
+                            || order.trigger_price().unwrap() > temp_trigger_price
```
cjdsellers, line 136: For these, consider something like `if let Some(trigger_price) = order.trigger_price()` which both checks if the price is `Some` and also gets you the inner value.

cjdsellers (commented):
Great progress @DeevsDeevs, many thanks!

nautilus_core/execution/src/trailing.rs
DeevsDeevs: I’m still learning Rust through these pull requests, so your suggestions are definitely welcome.
I’ll make sure to switch to this approach everywhere, appreciate the tip!

nautilus_core/execution/src/trailing.rs
DeevsDeevs: I initially thought the struct approach would help preserve the same API as the Cython version, but you’re right, changing it to three free functions makes sense.
I will add a prefix and name them something like trailing_stop_calculate

nautilus_core/execution/src/trailing.rs
```diff
@@ -0,0 +1,757 @@
...
+    }
+
+    Ok((new_trigger_price, new_price))
+}
+
+fn trailing_stop_calculate_with_last(
+    price_increment: Price,
+    trailing_offset_type: TrailingOffsetType,
+    side: OrderSide,
+    offset: f64,
```
cjdsellers, line 385: This was a `double` in Cython for performance reasons, but Rust can deal with this very efficiently and desugar the domain objects during compilation. So we could also try `offset` as `Price` (since we also use `Price` for increments and offsets elsewhere now).

I'm also happy to refactor to the above on `develop` branch if you might not be able to get back to this?

DeevsDeevs: I’ll refactor it right now, no problem at all

cjdsellers (approved):
👌

PR #2175, opened 2025-01-05
Refine sccache for CI

sunlei:
# Pull Request

Refine sccache for CI.

## Type of change

## Detail

This update replaces the `GHA mode` in sccache with `local mode`, storing the entire cache generated by sccache in the GHA cache. This approach prevents frequent writes to the GHA cache, which could otherwise lead to 429 errors. However, it does come at the cost of using more GitHub Actions storage space.

If use `GHA mode` and the GHA cache doesn’t trigger 429 errors or if an S3-like service is used, the cache can be utilized to its full potential, resulting in a higher cache hit rate. However, the network I/O involved might increase the total execution time and incur higher monetary costs.

Given these trade-offs, storing the entire sccache cache in the GHA cache currently seems to be the most optimal solution.

## Total Run Time

`4h 46m 35s` -> `3h 40m 49s`

### Before:

image

[link](https://github.com/nautechsystems/nautilus_trader/actions/runs/12613670666/usage)

### After:

image

[link](https://github.com/nautechsystems/nautilus_trader/actions/runs/12619041642/usage)

## Known Issues

The current cache key is defined as:

`${{ matrix.os }}-sccache-build-${{ hashFiles('**/Cargo.lock', '**/poetry.lock') }}`.

In certain cases, this may lead to reduced cache hit rates. The ideal solution would be to include the Python version in the cache key, such as:

`${{ matrix.os }}-py${{ matrix.python-version }}-sccache-build-${{ hashFiles('**/Cargo.lock', '**/poetry.lock') }}`

However, GitHub Actions has a total cache space limit of 10 GB. Including the Python version in the key would significantly increase space usage, potentially squeezing out other types of caches. Therefore, a compromise was made to balance cache efficiency and storage constraints.

## Tips

Can view sccache stats on the summary page.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,20 @@ jobs:
   pre-commit:
     name: pre-commit
     runs-on: ubuntu-latest
+    env:
+      # > --------------------------------------------------
+      # > sccache
+      # https://github.com/Mozilla-Actions/sccache-action
+      SCCACHE_DIRECT: "true"
+      SCCACHE_CACHE_MULTIARCH: 1
+      SCCACHE_DIR: ~/.cache/sccache
+      RUSTC_WRAPPER: "sccache"
+      CC: "sccache clang"
+      CXX: "sccache clang++"
+      # Incrementally compiled crates cannot be cached by sccache
+      # https://github.com/mozilla/sccache#rust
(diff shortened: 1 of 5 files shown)


cjdsellers:
Great stuff @sunlei, this is a welcome improvement to the long running CI jobs! many thanks :pray:

PR #2179, opened 2025-01-06
Refine WebSocket client re-subscribe log in Bybit

sunlei:
# Pull Request

Refine WebSocket reconnect for Bybit.

## Type of change

### DIFF
--- a/nautilus_trader/adapters/bybit/websocket/client.py
+++ b/nautilus_trader/adapters/bybit/websocket/client.py
@@ -193,8 +193,9 @@ class BybitWebSocketClient:
             if self._auth_required:
                 await self._authenticate()
 
-            # Re-subscribe to all streams
-            await self._subscribe_all()
+            if not self._is_trade:
+                # Re-subscribe to all streams, trade client does not need to do this
+                await self._subscribe_all()
 
             if self._handler_reconnect:
                 await self._handler_reconnect()
--- a/nautilus_trader/adapters/bybit/websocket/client.py
+++ b/nautilus_trader/adapters/bybit/websocket/client.py
@@ -193,9 +193,8 @@ class BybitWebSocketClient:
             if self._auth_required:
                 await self._authenticate()
 
-            if not self._is_trade:
-                # Re-subscribe to all streams, trade client does not need to do this
-                await self._subscribe_all()
+            # Re-subscribe to all streams
+            await self._subscribe_all()
 
             if self._handler_reconnect:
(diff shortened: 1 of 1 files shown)


nautilus_trader/adapters/bybit/websocket/client.py
```diff
@@ -295,7 +295,9 @@ async def _subscribe_all(self) -> None:
             self._log.error("Cannot subscribe all: not connected")
             return
 
-        self._log.info("Resubscribe to all data streams")
+        self._log.warning(
+            f"Resubscribing {self.channel_type} channel to `{len(self._subscriptions)}` streams",
```
cjdsellers, line 299: Would this be confusing for the execution client where it would log?

```
Resubscribing Trade channel to `0` streams
```

Also, should it be a warning? I don't think `_subscribe_all` is only called on reconnects and I think we have a reconnect specific log which could be a warning?

sunlei: 1. Initially, I added a condition to `.subscribe_all()` in this commit: https://github.com/nautechsystems/nautilus_trader/pull/2179/commits/20f087130c4b1ad43161f0cff5fcab079f20b7a1. However, I felt it might increase the complexity of understanding the code, so I switched to enhancing the logging in this section for better differentiation. As a result, it’s intentional to have `0` appear in the trade channel.

2. In Bybit, `_subscribe_all` is only called during reconnections. This is because Bybit’s subscription process involves sending a subscription request for each topic individually, meaning `_subscribe_all` is not invoked during the initial subscription.

3. Therefore, I’ve adjusted the log level to `warning`, making it easier to monitor in log collection systems.

sunlei: Alternatively, should I move this log line to `_reconnect_wrapper`?

cjdsellers: All good, understood on the issues and I'll have a look once on `develop` branch :ok_hand:

cjdsellers: Now that I've had a closer look, I think moving that log to `_reconnect_wrapper` does makes sense, as its where the reconnect logic is directly called from.

Adjusted [here](https://github.com/nautechsystems/nautilus_trader/commit/311dbdfc5e3d2471a4558d1c4fa928c2ce7af34f), but feel free to PR again if I have the wrong idea.

PR #2183, opened 2025-01-06
Add `enable_event_logging` for `StrategyConfig`

sunlei:
# Pull Request

Add `enable_event_logging` for `StrategyConfig`.

In high-frequency trading strategies, `<--[EVT]` type events can be particularly noisy. To address this, a toggle switch has been added for better control.

## Type of change

### DIFF
--- a/nautilus_trader/trading/config.py
+++ b/nautilus_trader/trading/config.py
@@ -60,6 +60,7 @@ class StrategyConfig(NautilusConfig, kw_only=True, frozen=True):
     external_order_claims: list[InstrumentId] | None = None
     manage_contingent_orders: bool = False
     manage_gtd_expiry: bool = False
+    enable_event_logging: bool = True
 
 
 class ImportableStrategyConfig(NautilusConfig, frozen=True):
--- a/nautilus_trader/trading/strategy.pxd
+++ b/nautilus_trader/trading/strategy.pxd
@@ -81,6 +81,8 @@ cdef class Strategy(Actor):
     """If contingent orders should be managed automatically by the strategy.\n\n:returns: `bool`"""
     cdef readonly bint manage_gtd_expiry
     """If all order GTD time in force expirations should be managed automatically by the strategy.\n\n:returns: `bool`"""
+    cdef readonly bint enable_event_logging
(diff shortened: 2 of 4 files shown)


nautilus_trader/trading/config.py
```diff
@@ -60,6 +60,7 @@ class StrategyConfig(NautilusConfig, kw_only=True, frozen=True):
     external_order_claims: list[InstrumentId] | None = None
     manage_contingent_orders: bool = False
     manage_gtd_expiry: bool = False
+    enable_event_logging: bool = True
```
cjdsellers, line 63: @sunlei I renamed this to a more neutral `event_logging` to avoid a redundant affirmative.

We could also consider making it something like `event_log_level` as well, since this new behavior still logs warning and above level events?

sunlei: There are similar types of logs, such as `[CMD]-->`, but they involve parts of the code I’m not very familiar with. So, I didn’t make any changes there and only adjusted the `<--[EVT]` logs for now. The idea of an `event_log_level` is great!

cjdsellers: Understood, this is good insight from someone running HFT strats "in the wild".

I think it might still be valuable to see the commands, or would you ideally want a similar config option to filter commands from the strategy as well?

sunlei: @cjdsellers

Yes, having such an option would be ideal. In my high-frequency trading environment:

1. These logs don’t hold significant value because the sheer volume makes them difficult to review, and their format can be challenging to manage. My approach is to compute key metrics directly, output them in Logfmt format, and write them to VictoriaMetrics for analysis. For log-related data, I retain only the most critical logs and a few custom logs, which are then stored in VictoriaLogs.

2. Time is incredibly valuable. From my measurements, a single log entry can incur a few microseconds of overhead. To minimize this, I strive to reduce the number of logs and instead compute cumulative metrics that are periodically output using timers.

ps: This is solely my perspective based on my specific strategy.

cjdsellers: It's true that these log messages can be expensive. Thanks for the insights.

PR #2188, opened 2025-01-07
Add adaptive_bar_ordering to BacktestVenueConfig

faysou:
# Pull Request

Add adaptive_bar_ordering to BacktestVenueConfig

Executes Low before High if Low is closer to Open, and vice versa

## Type of change

Attempted to add a test in test_matching_engine.py, left it as comment for further work

### DIFF
--- a/examples/backtest/databento_test_request_bars.py
+++ b/examples/backtest/databento_test_request_bars.py
@@ -6,7 +6,7 @@
 #       extension: .py
 #       format_name: percent
 #       format_version: '1.3'
-#       jupytext_version: 1.16.4
+#       jupytext_version: 1.16.6
 #   kernelspec:
 #     display_name: Python 3 (ipykernel)
 #     language: python
@@ -225,6 +225,13 @@ class TestHistoricalAggStrategy(Strategy):
 
 # %%
 # BacktestEngineConfig
+tested_market_data = "bars"  # or "quotes" or "trades"
+
+historical_start_delay = 10 if tested_market_data == "bars" else 2
+historical_end_delay = 1 if tested_market_data == "bars" else 0
+
+backtest_start = "2024-07-01T23:55" if tested_market_data == "bars" else "2024-07-02T00:00"
+backtest_end = "2024-07-02T00:10" if tested_market_data == "bars" else "2024-07-02T00:02"
 
 strategies = [
     ImportableStrategyConfig(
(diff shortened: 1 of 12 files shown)


stefansimik:
Hi @faysou ,

First, thank you again for implementing this feature so quickly! While testing it, I had a thought about the configuration name `dynamic_bar_execution` that I wanted to run by you.

Would you consider renaming it to `adaptive_bar_ordering`? Here's my reasoning:

1. "Ordering" more precisely describes what we're modifying - it's specifically about the sequence of OHLC price points within a bar, rather than the broader concept of "execution"
2. "Adaptive" better communicates that the ordering is based on a specific condition (price proximity) rather than just being dynamic/random

The new name would make the feature's purpose more immediately clear to other users who might benefit from this improvement in backtesting accuracy.

What are your thoughts on this? :-)

faysou:
Yes no problem, I'll make the change later today

stefansimik:
I have tested this new behavior for basic scenario and  I can confirm it works great 👍 

CleanShot 2025-01-07 at 14 47 53@2x

faysou:
Yes, I've saved you some time, but you have good ideas as well, I like that. Most people never have any ideas 😂

cjdsellers (approved):
Noted that the test is WIP and being actively worked on :ok_hand:

stefansimik:
> Noted that the test is WIP and being actively worked on 👌

I was thinking about test - it probably belongs here:
`tests/unit_tests/backtest/test_matching_engine.py`,

but have no idea, how could I technically verify from Python, that Bar was splitted into 4 prices as this is done in Cython code. I can't simply find any link to internal data, that could be checked in the test to verify.

Maybe one option would be to verify it by the external effects -> for examle  when we fire 2 limit order (1st near top of the bar, 2nd near bottom of the bar) and we can check in which order they are filled. By this approach could be confusing and and too far from original simple point we need to test.

cjdsellers:
@stefansimik I also think thats a good location for the test, and @faysou is currently working on one.

I think testing the external effects of setting the config option `True` and then processing bars and checking order fills would be enough.

The `OrderMatchingEngine` is indirectly tested via the `SimulatedExchange` which contains it, in many of those `test_exchange*` modules. Thats why there aren't many tests in `test_matching_engine.py`. (There was a refactoring years ago which extracted the matching engine out of the exchange, but test modules weren't renamed.)

@filipmacek is currently working on the Rust port for `OrderMatchingEngine`, and we'll make sure this time around its more thoroughly and directly tested.

I also just pushed [this refactoring](https://github.com/nautechsystems/nautilus_trader/commit/936f22e70ce681168a65732356eabefb7ca42ca3).
- Renamed to `bar_adaptive_high_low_ordering` to more clearly express the behavior.
- Reordered the params to group the bar config options and adjust the hierarchy of settings.

stefansimik:
Like the new name even more = 100% clear and self-descriptive. 👌👌

PR #2202, opened 2025-01-10
Port market order filling for OrderMatchingEngine in Rust

filipmacek:
# Pull Request

- implements market order fill flow in `OrderMatchingEngine`


cjdsellers (approved):
:ok_hand:

PR #2212, opened 2025-01-14
Port limit order filling for OrderMatchingEngine in Rust

filipmacek:
# Pull Request

- porting limit order flow in `OrderMatchingEngine`

### DIFF
--- a/nautilus_core/backtest/src/matching_engine/mod.rs
+++ b/nautilus_core/backtest/src/matching_engine/mod.rs
@@ -1439,7 +1439,7 @@ impl OrderMatchingEngine {
         msgbus.send(&msgbus.switchboard.exec_engine_process, &event as &dyn Any);
     }
 
-    fn generate_order_accepted(&self, order: &OrderAny, venue_order_id: VenueOrderId) {
+    fn generate_order_accepted(&self, order: &mut OrderAny, venue_order_id: VenueOrderId) {
         let ts_now = self.clock.get_time_ns();
         let account_id = order
             .account_id()
@@ -1458,6 +1458,8 @@ impl OrderMatchingEngine {
         ));
         let msgbus = self.msgbus.as_ref().borrow();
         msgbus.send(&msgbus.switchboard.exec_engine_process, &event as &dyn Any);
+
+        order.apply(event).expect("Failed to apply order event");
     }
 
     #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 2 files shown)


Smida-1999 (approved):

nautilus_core/backtest/src/matching_engine/mod.rs
```diff
@@ -822,9 +868,18 @@ impl OrderMatchingEngine {
             }
 
             // Move market back to targets
-            self.core.bid = self.target_bid;
-            self.core.ask = self.target_ask;
-            self.core.last = self.target_last;
+            if let Some(target_bid) = self.target_bid {
```
filipmacek, line 871: Cython version has `self._has_targets` variable, and Rust doesn't. But best to use native Option and check if the value is defined

nautilus_core/backtest/src/matching_engine/mod.rs
```diff
@@ -737,8 +738,53 @@ impl OrderMatchingEngine {
...
+            if order.liquidity_side().is_some()
+                && order.liquidity_side().unwrap() == LiquiditySide::NoLiquiditySide
+            {
+                order.set_liquidity_side(LiquiditySide::Taker)
+            }
+            self.fill_limit_order(order);
+        } else if order.time_in_force() == TimeInForce::Fok
+            || order.time_in_force() == TimeInForce::Ioc
+        {
+            self.cancel_order(order, None);
```
filipmacek, line 786: The problem here is that after `self.accept_order`  is called,  the corresponding `AcceptOrder` event is not applied anywhere, and it should be in `generate_order_accepted` function

nautilus_core/backtest/src/matching_engine/mod.rs
```diff
@@ -1242,6 +1458,8 @@ impl OrderMatchingEngine {
         ));
         let msgbus = self.msgbus.as_ref().borrow();
         msgbus.send(&msgbus.switchboard.exec_engine_process, &event as &dyn Any);
+
+        order.apply(event).expect("Failed to apply order event");
```
filipmacek, line 1462: Should this be added there? The reasoning is here https://github.com/nautechsystems/nautilus_trader/pull/2212/files#r1916905546

nautilus_core/backtest/src/matching_engine/mod.rs
cjdsellers: `self.accept_order(order)` will call the `ExecutionEngine` immediately which will apply the event along with other logic.

nautilus_core/backtest/src/matching_engine/mod.rs
cjdsellers: The event will be applied to the order in the `ExecutionEngine` following the completion of `msgbus.send(&msgbus.switchboard.exec_engine_process, &event as &dyn Any)`.

cjdsellers (approved):

PR #2218, opened 2025-01-19
Port stop market order filling for OrderMatchingEngine in Rust

filipmacek:
# Pull Request

- implement function `process_stop_market_order` in `OrderMatchingEngine`
- added `orderbook_delta_sell` fixture value which is heavily used in order matching engine tests for simplicity and smaller tests
- refactor how `price` is calculated in `determine_market_price_and_volume`, immediately setting the price as aggressively as possible.

### DIFF
--- a/nautilus_core/backtest/src/matching_engine/tests.rs
+++ b/nautilus_core/backtest/src/matching_engine/tests.rs
@@ -121,7 +121,12 @@ pub fn orderbook_delta_sell(instrument_eth_usdt: InstrumentAny) -> OrderBookDelt
     OrderBookDelta::new(
         instrument_eth_usdt.id(),
         BookAction::Add,
-        BookOrder::new(OrderSide::Sell, Price::from("1500.00"), Quantity::from("1"), 0),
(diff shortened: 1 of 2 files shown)


nautilus_core/backtest/src/matching_engine/tests.rs
```diff
@@ -896,6 +901,7 @@ fn test_matching_engine_valid_market_buy(
 #[rstest]
 fn test_process_limit_post_only_order_that_would_be_a_taker(
     instrument_eth_usdt: InstrumentAny,
+    orderbook_delta_sell: OrderBookDelta,
```
cjdsellers, line 904: It might be better to have a helper (or `OrderBookDeltaBuilder`) for creating the delta which would make the price of 1500 clearer, which is important for the behavior of the test (and we'll want to be able to change price and size of deltas often in tests).

I think this fixture style could be reserved for fairly static things, without special values which are important to the test (such as `cache`, etc)?

cjdsellers (approved):
:ok_hand:

PR #2222, opened 2025-01-19
Add DataList to better handle several instruments

faysou:
# Pull Request

Add DataList to better handle several instruments.
Created DataList class that inherits from Data and modified data engine and clients accordingly.

## Type of change

Existing tests should catch any problem

### DIFF
--- a/nautilus_trader/adapters/betfair/data.py
+++ b/nautilus_trader/adapters/betfair/data.py
@@ -38,6 +38,7 @@ from nautilus_trader.common.component import MessageBus
 from nautilus_trader.common.enums import LogColor
 from nautilus_trader.core.correctness import PyCondition
 from nautilus_trader.core.data import Data
+from nautilus_trader.core.data import DataList
 from nautilus_trader.live.data_client import LiveMarketDataClient
 from nautilus_trader.model.enums import BookType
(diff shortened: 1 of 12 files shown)


cjdsellers:
Hi @faysou 

I think the core of the issue as [reported](https://discord.com/channels/924497682343550976/924504069433876550/1329748384692572160) by the user is occurring downstream of adapter -> `DataEngine`, and during `MessageBus` publishing: probably in [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/common/component.pyx#L2615) and somewhat related to [this issue](https://github.com/nautechsystems/nautilus_trader/pull/2151).

There is an existing issue remaining of slow wildcard pattern resolution for subscriptions which can result in this behavior (I'll raise a separate issue about it). It will be interesting to know how the user in the above report is subscribing for the instruments.

Also, we know that 600,000+ CME Globex instrument definitions can be decoded and sent through the message bus in a couple of seconds, so I don't think grouping them all into a new container type will fix the above issue.

The `DataList` in this PR _might_ be useful, as a more generic `OrderBookDeltas`, but I don't actually think we need it right now? it would be increasing the footprint of the codebase during the Rust port without a clear gain.

PR #2225, opened 2025-01-21
Port stop limit order processing for OrderMatchingEngine in Rust

filipmacek:
# Pull Request

- implement `process_stop_limit_order` function in `OrderMatchingEngine` in Rust with tests


cjdsellers (approved):
Thanks for adding the tests too :ok_hand:

Issue #2236, opened 2025-01-23
Portfolio support for Bar data


stefansimik:
## Current Behavior
The Portfolio component doesn't fully support bar-based backtesting, leading to incorrect PnL calculations when using OHLCV bar data. Specifically:

- `Portfolio.unrealized_pnl()` returns `None` while positions are open
- `Portfolio.realized_pnl()` returns incorrect values (zero profit) after positions are closed
- The Portfolio module doesn't import or handle Bar data natively

## Proposed Solution
Implement bar support based on **Bar-to-Tick Conversion**:

- Convert each OHLCV bar into synthetic trade ticks
- Split each bar into 4 proportional trade events based on OHLC prices
- Distribute volume and timestamps across the synthetic ticks
- Leverage existing tick-based Portfolio functionality

## Reproduction
A minimal example demonstrating the current limitation is attached.

[run_backtest.py.zip](https://github.com/user-attachments/files/18524121/run_backtest.py.zip)

It is fully reproducible example in one file (including artificial bars data).
Just unzip and run.

Strategy setup is extremely simple:
- Just 12 artificial bars (1-min) with increasing prices
- Just 2 orders: BUY MARKET at 1st bar ....... and SELL MARKET at 7th bar  
- Commission $2.50 per contract

I've added two debug points in the code to make it easy to see empty values being returned from `Portfolio` object:
1. **Line 93** (during open position): `unrealized_pnl` returns `None` instead of profit
2. **Line 101** (after position close): `realized_pnl` returns `0` when there should be profit

## Why this is important

This is my main rationale for why **bar data** should be treated as a first-class citizen in Nautilus in the future.

Three key factors support this perspective:

1. **First experience**: Bar data is typically the first type of data that beginners encounter, as it's widely available and accessible. When these commonly available data sources don't function properly, it creates immediate problems, raises repetitive questions, and causes disillusionment during users' first experiences with Nautilus Trader. It's therefore crucial that these common usage scenarios work seamlessly.
2. **Cost-effectiveness**: The price disparity between 1-minute and tick data is substantial. Looking at EUR/USD futures from DataBento over a 5-year period illustrates this:
   * 1-minute bars: `9.03 USD`
   * Tick data: `92.59 USD` (10x more expensive)
[...]

PR #2239, opened 2025-01-23
Add bar support to compute Portfolio pnls

faysou:
# Pull Request

Add support for bars in Portfolio.
Subscribing to External bars only which should be the most granular as less granular bars aggregated from bars or quote ticks would probably be internal.

Fixing: https://github.com/nautechsystems/nautilus_trader/issues/2236

## Type of change

Added one test for bars in Portfolio

### DIFF
--- a/examples/backtest/synthetic_data_pnl_test.py
+++ b/examples/backtest/synthetic_data_pnl_test.py
@@ -1,0 +1,220 @@
+from datetime import UTC
+from datetime import datetime
+from decimal import Decimal
+
+from nautilus_trader.backtest.engine import BacktestEngine
+from nautilus_trader.backtest.models import PerContractFeeModel
+from nautilus_trader.config import BacktestEngineConfig
+from nautilus_trader.config import LoggingConfig
+from nautilus_trader.config import StrategyConfig
+from nautilus_trader.core.datetime import dt_to_unix_nanos
+from nautilus_trader.core.datetime import unix_nanos_to_dt
+from nautilus_trader.model.currencies import USD
+from nautilus_trader.model.data import Bar
+from nautilus_trader.model.data import BarSpecification
+from nautilus_trader.model.data import BarType
+from nautilus_trader.model.enums import AccountType
+from nautilus_trader.model.enums import AssetClass
+from nautilus_trader.model.enums import BarAggregation
+from nautilus_trader.model.enums import OmsType
+from nautilus_trader.model.enums import OrderSide
+from nautilus_trader.model.enums import PriceType
+from nautilus_trader.model.enums import TimeInForce
+from nautilus_trader.model.identifiers import InstrumentId
+from nautilus_trader.model.identifiers import Symbol
+from nautilus_trader.model.identifiers import Venue
+from nautilus_trader.model.instruments import FuturesContract
+from nautilus_trader.model.objects import Money
+from nautilus_trader.model.objects import Price
+from nautilus_trader.model.objects import Quantity
+from nautilus_trader.trading.strategy import Strategy
+
+
+def create_6E_instrument(venue: Venue) -> FuturesContract:
+    symbol = Symbol("6E")
+    return FuturesContract(
+        # Core identification parameters for the Euro FX futures contract
+        instrument_id=InstrumentId(symbol, venue),  # 6E is CME's code for EUR/USD futures
+        raw_symbol=symbol,  # Symbol as used on the exchange
+        asset_class=AssetClass.FX,  # Indicates this is an FX futures contract
+        currency=USD,  # Contract is denominated in USD
+        # Price and size specifications from CME
+        price_precision=5,  # 5 decimal places for EUR/USD pricing
+        price_increment=Price(
+            Decimal("0.00005"),
+            precision=5,
+        ),  # Minimum tick = 0.00005 ($6.25 value)
+        multiplier=Quantity(Decimal("125000"), precision=0),  # Each contract = 125,000 EUR
+        lot_size=Quantity(Decimal("1"), precision=0),  # Minimum trading size is 1 contract
+        # Contract specifications and expiration details
+        underlying="EUR/USD",  # The underlying forex pair
+        activation_ns=0,  # Contract start time (0 = active now)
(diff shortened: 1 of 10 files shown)


nautilus_trader/portfolio/portfolio.pyx
```diff
@@ -378,6 +380,21 @@ cdef class Portfolio(PortfolioFacade):
...
+            bar.bar_type.instrument_id,
+            bar.close,
+            bar.close,
+            Quantity.from_str("0"),
+            Quantity.from_str("0"),
+            bar.ts_init,
+            bar.ts_init,
+        )
+
+        self._cache.add_quote_tick(tick)
```
davidsblom, line 395: Wondering whether adding the quote to the cache can have any side effects when strategies load quotes from the cache.

faysou: it could be the case if quotes and external bars are used at the same time, but portfolio currently uses the cache to read quote so there wasn't another easy way.

my assumption is that if somone has access to quote, he doesn't need external bars and can aggregate quotes instead. another possibility would be to add an option to consider bars or not in portfolio.

stefansimik: Please think this through carefully, guys 🙏 
I'd really like to help you think through the implications, but I don't have deep enough knowledge at this moment.

Generally speaking, it's very important that we maintain the generic functionality of the entire framework and don't introduce any assumptions that would be difficult to explain or understand later (for example, assuming that when someone uses certain data, they "shouldn't" need other data). People can configure all sorts of data and combinations, and generally, it should always work.

Please just emphasize making sure we don't create any limitations (where certain types of data conflict with each other in the Cache).

The goal is for Portfolio to be able to work 100% with bar-data while simultaneously maintaining all original flexibility without additional limitations or assumptions.

I completely understand this isn't a trivial matter, which is why I'm really relying on you more experienced folks.
I'll help and support you as much as I can with thorough testing ❤️ 👍

faysou: you're right, I'll do something about it, I wouldn't like to write something with a logical flaw

zhaoyiming808: > Please think this through carefully, guys 🙏 I'd really like to help you think through the implications, but I don't have deep enough knowledge at this moment.
> 
> Generally speaking, it's very important that we maintain the generic functionality of the entire framework and don't introduce any assumptions that would be difficult to explain or understand later (for example, assuming that when someone uses certain data, they "shouldn't" need other data). People can configure all sorts of data and combinations, and generally, it should always work.
> 
> Please just emphasize making sure we don't create any limitations (where certain types of data conflict with each other in the Cache).
> 
> The goal is for Portfolio to be able to work 100% with bar-data while simultaneously maintaining all original flexibility without additional limitations or assumptions.
> 
> I completely understand this isn't a trivial matter, which is why I'm really relying on you more experienced folks. I'll help and support you as much as I can with thorough testing and documenting this new functionality ❤️ 👍

It is great to keep all components 100% available

nautilus_trader/live/config.py
```diff
@@ -105,6 +105,8 @@ class LiveExecEngineConfig(ExecEngineConfig, frozen=True):
         weights imposed by the necessary order status requests.
     qsize : PositiveInt, default 100_000
         The queue size for the engines internal queue buffers.
+    portfolio_bar_updates : bool, default True
```
cjdsellers, line 108: This isn't a live-specific option, as we may want to configure this for backtests as well.
Since we don't have a dedicated portfolio config, consider moving this to `ExecEngineConfig` for now.

examples/backtest/synthetic_data_pnl_test.py
```diff
@@ -0,0 +1,232 @@
...
+                quantity=Quantity.from_str("1"),
+                time_in_force=TimeInForce.GTC,
+            )
+            self.submit_order(order)
+
+    def on_stop(self) -> None:
+        pass
+
+
+class PerContractFeeModel(FeeModel):
```
cjdsellers, line 150: This is a built-in type now, so could import it as:
```
from nautilus_trader.backtest.models import PerContractFeeModel
```

nautilus_trader/portfolio/portfolio.pxd
```diff
@@ -45,13 +46,16 @@ cdef class Portfolio(PortfolioFacade):
     cdef dict[InstrumentId, Money] _realized_pnls
     cdef dict[InstrumentId, Decimal] _net_positions
     cdef set[InstrumentId] _pending_calcs
+    cdef dict[InstrumentId, Price] _last_bid_price
```
cjdsellers, line 49: Why are we adding this? (should be available through the cache already)

nautilus_trader/portfolio/portfolio.pyx
```diff
@@ -110,6 +112,8 @@ cdef class Portfolio(PortfolioFacade):
         self._realized_pnls: dict[InstrumentId, Money] = {}
         self._net_positions: dict[InstrumentId, Decimal] = {}
         self._pending_calcs: set[InstrumentId] = set()
+        self._last_bid_price: dict[InstrumentId, Price] = {}
```
cjdsellers, line 115: We're duplicating logic and storage:
https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/cache/cache.pyx#L2052

nautilus_trader/portfolio/portfolio.pyx
```diff
@@ -1247,24 +1269,24 @@ cdef class Portfolio(PortfolioFacade):
         return Money(total_pnl, currency)
 
     cdef Price _get_last_price(self, Position position):
-        cdef PriceType price_type
+        price: Price | None = None
```
cjdsellers, line 1272: As above, lets just keep the changes for this PR to minimally what is needed for bars to be able to update.

nautilus_trader/portfolio/portfolio.pyx
faysou: To avoid creating a data in the cache that could interfere with real quotes, that's what the messages above were about.

faysou:
I've taken into account all comment above and updated the code just now

PR #2255, opened 2025-01-26
[WIP] Support for selecting memory allocator

sunlei:
# Pull Request

Support for selecting memory allocator.

- Added as features:
  - `allocator-jemalloc`
  - `allocator-mimalloc`
- In one of my Rust projects with heavy network and computational I/O workloads, both `mimalloc` and `jemalloc` outperformed the system allocator:
  - `mimalloc + glibc` performed better than `jemalloc + glibc`
    - Both outperformed configurations using `+ musl`
  - Therefore, `mimalloc + glibc` is recommended.

## Type of change

## Issue

Since the `cargo` checks in `pre-commit` are run with `--all-features`, the logic for choosing between allocators fails to pass the checks. As a temporary workaround, the code related to `jemalloc` has been commented out. So:

- Is there a better approach to support a mutually exclusive choice between the two?  
- Or should we simply remove `jemalloc` and keep only `mimalloc`?
- Use tools like https://github.com/taiki-e/cargo-hack ?

### DIFF
--- a/nautilus_core/Cargo.lock
+++ b/nautilus_core/Cargo.lock
@@ -3041,7 +3041,6 @@ dependencies = [
  "strum",
  "sysinfo",
  "tempfile",
- "tikv-jemallocator",
  "tokio",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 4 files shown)


cjdsellers:
As discussed, will close for now and revisit later when more of the port is done :pray:
Also, feel free to open another PR if you manage to solve the issue.

PR #2260, opened 2025-01-27
Refactor data request interfaces into messages

faysou:
# Pull Request

Refactor data interfaces into messages

## Type of change

Updated existing tests

### DIFF
--- a/nautilus_trader/common/actor.pxd
+++ b/nautilus_trader/common/actor.pxd
@@ -26,8 +26,8 @@ from nautilus_trader.core.message cimport Event
 from nautilus_trader.core.rust.model cimport BookType
 from nautilus_trader.core.uuid cimport UUID4
 from nautilus_trader.data.messages cimport DataCommand
-from nautilus_trader.data.messages cimport DataRequest
 from nautilus_trader.data.messages cimport DataResponse
+from nautilus_trader.data.messages cimport RequestData
 from nautilus_trader.indicators.base.indicator cimport Indicator
 from nautilus_trader.model.book cimport OrderBook
 from nautilus_trader.model.data cimport Bar
@@ -182,6 +182,9 @@ cdef class Actor(Component):
         self,
         DataType data_type,
         ClientId client_id,
+        datetime start=*,
+        datetime end=*,
+        int limit=*,
         callback=*,
         bint update_catalog=*,
         dict[str, object] params=*,
@@ -209,7 +212,7 @@ cdef class Actor(Component):
     cpdef UUID4 request_order_book_snapshot(
         self,
         InstrumentId instrument_id,
-        int limit,
+        int limit=*,
         ClientId client_id=*,
         callback=*,
         dict[str, object] params=*,
@@ -219,6 +222,7 @@ cdef class Actor(Component):
         InstrumentId instrument_id,
         datetime start=*,
         datetime end=*,
+        int limit=*,
         ClientId client_id=*,
         callback=*,
         bint update_catalog=*,
@@ -229,6 +233,7 @@ cdef class Actor(Component):
         InstrumentId instrument_id,
         datetime start=*,
         datetime end=*,
+        int limit=*,
         ClientId client_id=*,
         callback=*,
         bint update_catalog=*,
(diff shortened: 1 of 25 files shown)


nautilus_trader/common/actor.pxd
```diff
@@ -251,14 +257,15 @@ cdef class Actor(Component):
...
+        int limit=*,
         ClientId client_id=*,
         callback=*,
         bint include_external_data=*,
         bint update_subscriptions=*,
         bint update_catalog=*,
         dict[str, object] params=*,
     )
-    cpdef bint is_pending_request(self, UUID4 request_id)
+    cpdef bint is_pending_request(self, UUID4 id)
```
cjdsellers, line 268: I think we should leave the param name as `request_id`, as `id` shadows the built-in and isn't as explicit.

[Also applies to many changes below]

nautilus_trader/common/actor.pyx
```diff
@@ -2089,6 +2098,13 @@ cdef class Actor(Component):
         ----------
         data_type : DataType
             The data type for the request.
+        start : datetime, optional
+            The start datetime (UTC) of request time range (inclusive).
+        end : datetime, optional
+            The end datetime (UTC) of request time range.
+            The inclusiveness depends on individual data client implementation.
```
cjdsellers, line 2105: "The inclusiveness depends on the individual data client implementation."

nautilus_trader/data/messages.pyx
```diff
@@ -177,25 +185,32 @@ cdef class Unsubscribe(DataCommand):
...
-    data_type : type
-        The data type for the request.
     callback : Callable[[Any], None]
         The delegate to call with the data.
     request_id : UUID4
         The request ID.
     ts_init : uint64_t
         UNIX timestamp (nanoseconds) when the object was initialized.
-    params : dict[str, object], optional
+    params : dict[str, object]
```
cjdsellers, line 213: Should we leave `params` as optional? or did you find it useful having a downstream guarantee there would be a dict available to pass through?

nautilus_trader/live/data_client.py
```diff
@@ -710,152 +704,64 @@ def unsubscribe_instrument_close(
...
-        start: pd.Timestamp | None = None,
-        end: pd.Timestamp | None = None,
-        params: dict[str, Any] | None = None,
-    ) -> None:
-        time_range_str = format_utc_timerange(start, end)
-        limit_str = f" limit={limit}" if limit else ""
-        self._log.info(f"Request {instrument_id} trades{time_range_str}{limit_str}", LogColor.BLUE)
+    def request_trade_ticks(self, request: RequestTradeTicks) -> None:
+        time_range_str = format_utc_timerange(request.start, request.end)
+        self._log.info(f"Request {request.instrument_id} trades{time_range_str}", LogColor.BLUE)
```
cjdsellers, line 746: The limit part of the log message was removed?

cjdsellers (commented):
Thanks for your work here @faysou 

I think this is a great improvement:
- Better symmetry with command messages on the execution side
- Stronger contract, as request commands must be valid when initialized in `Actor`
- Reduces the surface area of the `DataClient` interface, making them easier to work with

It's also a good sign that even with all the additional message types, there is still more code being removed here than added.

I think there is some change risk for the adapter clients, as test coverage isn't as strong there as the core of the system. I've reviewed them, but it would be great if we could get additional reviews when possible :pray: :
- @sunlei for Bybit
- @davidsblom for dYdX

davidsblom:
LGTM (dYdX changes)

sunlei:
LGTM (Bybit)

cjdsellers:
Thank you @davidsblom and @sunlei

After we resolve some of the comments above, I'm planning to merge this after the next release (which should be this week).

nautilus_trader/data/messages.pyx
faysou: I've noticed it's easier to develop when all params are mandatory, and this would be the case likely in rust anyway also these message types are only touched internally, built by actor.pyx.

Params can be None as argument, and it's replaced by an empty dict if None, so then downstream clients don't need to check if the dict is None.

nautilus_trader/common/actor.pyx
faysou: added the inclusiveness depends ... to start arguments as well

faysou:
I'm done with the fixes, I'll start writing on the subscribe refactoring in another commit of the same PR (as there are too many changes in the current commit to start from the current develop branch)

PR #2261, opened 2025-01-28
Port process modify for OrderMatchingEngine in Rust

filipmacek:
# Pull Request

- Implemented the `process_modify` function in the `OrderMatchingEngine` and the `update_order` function.
- Verified that a "modify rejected" event is correctly generated when the order does not exist.
- Bootstrapped several `update_`  functions to support pattern matching in the main `update_order function`, laying the groundwork for their porting in the next PRs.


cjdsellers (approved):
:ok_hand:

PR #2266, opened 2025-01-29
Refactor option greeks feature

faysou:
# Pull Request

Refactor option greeks feature

- integrate option greeks calculation method in actor so calculations can be done synchronously in a strategy
- maintained ability to publish greeks, store them in a catalog and load them from a catalog
- added helper function for subscribing to greeks so a callback stores cached greeks loaded from a catalog
- improved example so it doesn't need a delay to wait for loaded greeks when subscribing to them from a catalog, the trick is to add the greeks backtestdataconfig before other types of market data so it reaches actors before.
- ability to compute shocked greeks and prices by shifting spot, vol, or time to expiry, this allows to have estimate of how portfolio greeks could evolve by varying the relevant market data variables. taking into account Futures correctly for this.
- the reason to have the greeks functionality in actor, is that any subclass of actor, for example a class responsible for monitoring greeks, can easily access the functionality
- added price field to PortfolioGreeks
- Removed InterateRateData, keeping only InterestRateCurveData
- InterestRateProvider caches InterestRateCurveData so it's then accessible the greeks computation function
- Fixed a bug in the cache when using price with bars
- Added GreeksCalculator class for greeks related function, accessible via actor.greeks.portfolio_greeks for example

## Type of changs

Tested using databento_option_greeks.py notebook, updated


nautilus_trader/common/actor.pyx
```diff
@@ -3453,8 +3573,53 @@ cdef class Actor(Component):
...
+        None
+
+        """
+        used_handler = handler or self.cache_greeks
+        self._msgbus.subscribe(
+            topic=f"data.GreeksData.instrument_id={underlying}*",
+            handler=used_handler
+        )
+
+    def cache_greeks(self, greeks_data: GreeksData):
```
cjdsellers, line 3610: I think its a bit asymmetric to expose a helper just for greeks, and not the other data types?
Better if we added `add_greeks` to the cache at this point?

cjdsellers (commented):
I think this is a good change moving `greeks.py` from risk to model.

I'm still unsure if actor is the right location for greeks calculations though, as this makes every actor a specialized greeks calculator too.

nautilus_trader/cache/cache.pyx
```diff
@@ -19,15 +19,13 @@ import time
...
-from nautilus_trader.cache.config import CacheConfig
-from nautilus_trader.core.rust.model import PriceType as PriceType_py
-
 from cpython.datetime cimport datetime
 from cpython.datetime cimport timedelta
 from libc.stdint cimport uint8_t
 from libc.stdint cimport uint64_t
 
+from nautilus_trader.cache.config import CacheConfig
+from nautilus_trader.core.rust.model import PriceType as PriceType_py
```
cjdsellers, line 28: It's a non-obvious convention, but we standardize pure Python imports above Cython imports for various reasons, so the previous location was intentional.

nautilus_trader/common/actor.pyx
```diff
@@ -26,19 +26,22 @@ attempts to operate without a managing `Trader` instance.
 
 import asyncio
 from concurrent.futures import Executor
+from typing import Callable
 
 import cython
 
+from cpython.datetime cimport datetime
+from libc.stdint cimport uint64_t
```
cjdsellers, line 34: As above, previous location was correct (Cython will go soon, so thankfully we don't need to remember this for much longer).

nautilus_trader/risk/greeks.pyx
```diff
@@ -0,0 +1,260 @@
...
+from nautilus_trader.model.data cimport DataType
+from nautilus_trader.model.identifiers cimport InstrumentId
+from nautilus_trader.model.identifiers cimport StrategyId
+from nautilus_trader.model.identifiers cimport Venue
+from nautilus_trader.model.objects cimport Price
+from nautilus_trader.model.position cimport Position
+
+
+cdef class GreeksCalculator:
+    def __init__(self, msgbus: MessageBus, cache: CacheFacade, clock: Clock, logger: Logger):
```
cjdsellers, line 42: Ideally we follow the formatting convention of other Cython classes (hang params).

Also, need to add docstrings.

cjdsellers (commented):
Thanks @faysou this is nicely done.

faysou:
I'm done with the changes mentioned above.

Issue #2238, opened 2025-01-23
RFC: Reuse `PolymarketWebSocketClient` during non-initial subscription


ryantam626:
# Background

I use nautlius to record polymarket data, I subscribe to new instruments' order book as I become aware of them - which could be during a reload of instruments (i.e. not during init).

If the new batch of instruments I want to subscribe to is massive, then the number of new `PolymarketWebSocketClient` creation becomes massive as well, and it's quite easy to hit a rate limit error.

# Relevant code 

https://github.com/nautechsystems/nautilus_trader/blob/fafcc3091e5025e21e1af375939f6ed48650b20f/nautilus_trader/adapters/polymarket/data.py#L204-L220

Notice how in the event where `self._ws_client` is already established and connected, we then opt for always creating a new client and subscribe over another WS.

# Proposal

Change the meaning of `ws_connection_delay_secs` - make client always wait for this many seconds prior to connecting to WS.

ryantam626:
Something along the lines of this.

```patch

From 125e4e8ab8cee63784ecf4ba989225f4e1f36c00 Mon Sep 17 00:00:00 2001
From: Ryan Tam <ryantam626@gmail.com>
Date: Thu, 23 Jan 2025 20:13:42 +0000
Subject: [PATCH] feat: Gate and group polymarket subscription by configurable
 value

This is to avoid overwhelming polymarket by establishing many WS at
once, since polymarket will throw 429 at us.
---
 nautilus_trader/adapters/polymarket/config.py |  6 +-
 nautilus_trader/adapters/polymarket/data.py   | 80 +++++++++----------
 2 files changed, 40 insertions(+), 46 deletions(-)

diff --git a/nautilus_trader/adapters/polymarket/config.py b/nautilus_trader/adapters/polymarket/config.py
index 249625b3a..adfd02453 100644
--- a/nautilus_trader/adapters/polymarket/config.py
+++ b/nautilus_trader/adapters/polymarket/config.py
@@ -50,8 +50,8 @@ class PolymarketDataClientConfig(LiveDataClientConfig, frozen=True):
         The HTTP client custom endpoint override.
     base_url_ws : str, optional
         The WebSocket client custom endpoint override.
-    ws_connection_delay_secs : PositiveInt, default 5
-        The delay (seconds) prior to main websocket connection to allow initial subscriptions to arrive.
+    ws_connection_delay_secs : PositiveFloat, default 0.1
[...]

PR #2271, opened 2025-01-30
Group polymarket subscriptions by configurable delay

ryantam626:
# Pull Request

Allow user to configure the delay of non-initial Polymarket WS connection per #2238

This is to avoid overwhelming polymarket by establishing many WS at once, since polymarket will throw 429 at us.

We separate out the initial delay and the non-initial delay just in case users wanting more fine-grained control about the delay.

N.B. `ws_connection_delay_secs` has a new meaning now - this is no longer the config for the initial delay but for non-intiial WS connection delay - this might need to be communicated to end users.

## Type of change

Built and ran locally against polymarket.


cjdsellers:
Hi @ryantam626 

Great work, and thanks for the contribution :pray: 

I'll make sure to communicate the config meaning change in the release notes.

Issue #2275, opened 2025-02-01
Processing TradeTick from bar fails in backtest if bar.volume / 4 and instrument.size_increment both evaluate to < 1


miller-moore:
# Bug Report

When the order matching engine processes a trade tick from a bar with a small volume and the instrument size increment is less than 1, the creation of TradeTick fails on Condition.positive_int error because TradeTick enforces `size` to be a positive int. 

### Expected Behavior
Take `max(bar.volume / 4, instrument.size_increment, 1)` instead of just `max(bar.volume / 4, instrument.size_increment)`: https://github.com/nautechsystems/nautilus_trader/blob/234b4915c8b3500b42182f1143951971f38efe48/nautilus_trader/backtest/matching_engine.pyx#L646

### Actual Behavior
Sometimes the bar.volume / 4 combined with instrument.size_increment < 1 results in a float trade tick size of < 1, leading to the Condition.positive_int error on `size`.

### Steps to Reproduce the Problem

1. Create a bar whose volume is less than 4 and whose instrument.size_increment < 1
2. Run a backtest with this bar
3. See the error thrown by Condition.positive_int

### Specifications

- OS platform: MacOS 15.1
- Python version: 3.12
- `nautilus_trader` version: 1.211.0

faysou:
Thanks for the report, I've implemented this code recently, I'll fix it today or tomorrow, I know what to do.

faysou:
I don't understand why cdef double size_value = max(bar.volume.as_double() / 4.0, self.instrument.size_increment.as_double())  is not enough.

Can you please send an example reproducing the issues ?

stefansimik:
As I understood it, the code is OK, but just **checked condition** is failing: `Condition.positive_int`.
If such a condition is there, then result of the code can be decimal number (smaller than integer 1), on which condition `Condition.positive_int` fails...

But let's wait for MRE if needed 👍

miller-moore:
I don't have an MRE ready but I will try to work one up if you still need one after the following attempted explanation. 

Basically: 
- `bar.volume.as_double() / 4.0` can evaluate to a float smaller than 1
- `self.instrument.size_increment.as_double()` can also evaluate to a float smaller than 1

[...]

PR #2277, opened 2025-02-02
Fix _process_trade_ticks_from_bar for low volume

faysou:
# Pull Request

Fix _process_trade_ticks_from_bar for low volume
Floored size of tick at 1 if bar volume is not 0

Solving issue here https://github.com/nautechsystems/nautilus_trader/issues/2275

## Type of change

### DIFF
--- a/nautilus_trader/backtest/matching_engine.pyx
+++ b/nautilus_trader/backtest/matching_engine.pyx
@@ -643,7 +643,12 @@ cdef class OrderMatchingEngine:
         #         )
 
     cdef void _process_trade_ticks_from_bar(self, Bar bar):
-        cdef double size_value = max(bar.volume.as_double() / 4.0, self.instrument.size_increment.as_double())
+        cdef double bar_volume = bar.volume.as_double()
+        if bar_volume == 0.0:
+            return
+
+        # Note: we make the approximation that volume is at least 4 if it's not 0
+        cdef double size_value = max(bar.volume.as_double() / 4.0, self.instrument.size_increment.as_double(), 1.0)
         cdef Quantity size = Quantity(size_value, bar._mem.volume.precision)
 
         # Create base tick template


stefansimik:
Hi @faysou 😊,
thank you very much for your work on this issue, I really appreciate it 💚 🙏 .

I'd like to discuss an additional scenario regarding the splitting of 1-minute bars into ticks based on their OHLC price patterns:

1. **Single-Price Bar** (Line-like appearance)
   - All OHLC prices are identical
   - Bar visually appears as a horizontal line
   - Minimum volume must be 1 contract (traded at this single price)
   - Splitting logic: Generate only ONE tick with this price
2. **Two-Price Bar** (Rectangle without wicks)
   - Open equals Low AND Close equals High (or vice versa)
   - Bar visually appears as a filled rectangle without wicks
   - Minimum volume must be 2 contracts (at least 1 contract per price level)
   - Splitting logic: Generate TWO ticks using only High and Low prices
3. **Three-Price Bar**
   - Contains exactly 3 unique prices among OHLC values
   - Example: Open=High with unique Low and Close
   - Minimum volume must be 3 contracts (at least 1 contract per price level)
   - Splitting logic: Generate THREE ticks using the unique prices (High, Low, and either Open or Close)

**The key principle is:**

Number of unique OHLC prices determines both:

1. The minimum possible volume (1 contract per unique price)
2. How many ticks should be generated when splitting the bar

**Summary:**

* Bar with 1 unique price has minimum volume of 1 and splits into 1 tick (not 4)  
* Bar with 2 unique prices has minimum volume of 2 and splits into 2 ticks (not 4)
* Bar with 3 unique prices has minimum volume of 3 and splits into 3 ticks (not 4)
* Bar with 4 unique prices has minimum volume of 4 and splits into 4 ticks

This realistic approach allow splitting bar into realistic count of trade ticks with correct minimum volume for each tick.

What do you think about this exact / realistic simulation? Does it make sense from your perspective?

faysou:
Hi @stefansimik I would rather close the issue as it's been worked on by @cjdsellers . Also volume are not just integers, so I'm not sure the logic is simple.

stefansimik:
Understood and clear 😊 👍 
This is definitely nothing critical, just a tuning things and nice-to-have.
We can think about it anytime later if needed 😊👍

PR #2286, opened 2025-02-05
Use make target distclean instead of clean to run git clean

demonkoryu:
# Pull Request

Use make target `distclean` instead of clean to run git clean. Make target `clean` now only removes intermediate build files.
The `clean` target previously deleted non-committed files, which might not be expected of this make target from the description on https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html

## Type of change

Tried it locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,18 @@ build-wheel-debug:
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @demonkoryu 

Great stuff, thanks for the contribution and for straightening this out :pray:.

PR #2291, opened 2025-02-07
Refactor execution message handling in Rust

filipmacek:
# Pull Request

- moved `models` and `matching_engine` from `backtest` to `execution` crate
- consolidated all order management and execution components into separate modules in `execution` crate
- created fully typed and statically defined all `OrderMatchingCore` handlers: `FillMarketOrderHandler`, `FillLimitOrderHandler` and `TriggerStopOrderHandler` with two variants of implementation for `OrderMatchingEngine` and `OrderEmulator`
- redesigned `OrderMatchingCore` from closures to these handlers with proper setter functions
- introduce handler management objects called adapters which take care of hooking up predefined functions in both msgbus and handler objects. They connect all order execution-related components and take care of the lifecycle of handlers functions
- `OrderMatchingEngine` now fills resting limit orders from `OrderMatchingCore` when iterated if the market bid/ask moved
- `OrderEmulator` completely refactored with new handler pattern, component cleaner and checked, and state moved from `OrderEmulatorState` to main struct which simplifies component state change and improves code quality and makes it easier to read


nautilus_core/execution/src/order_emulator/adapter.rs
```diff
@@ -0,0 +1,114 @@
...
+pub struct OrderEmulatorAdapter {
+    emulator: Rc<RefCell<OrderEmulator>>,
+}
+
+impl OrderEmulatorAdapter {
+    pub fn new(
+        clock: Rc<RefCell<dyn Clock>>,
+        cache: Rc<RefCell<Cache>>,
+        msgbus: Rc<RefCell<MessageBus>>,
+    ) -> Self {
```
twitu, line 48: A lot constructs are wrapped by Rc RefCell but it adds to the cognitive load. Perhaps all such constructs can defined a wrapper type alias to make it easier to read.

pub type SharedCache = Rc<RefCell<Cache>>;

nautilus_core/execution/src/order_emulator/adapter.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    fn initialize_cancel_order_handler(emulator: Rc<RefCell<OrderEmulator>>) {
+        let handler = CancelOrderHandlerAny::OrderEmulator(emulator.clone());
+        emulator.borrow_mut().set_cancel_order_handler(handler);
+    }
+
+    fn initialize_modify_order_handler(emulator: Rc<RefCell<OrderEmulator>>) {
+        let handler = ModifyOrderHandlerAny::OrderEmulator(emulator.clone());
+        emulator.borrow_mut().set_modify_order_handler(handler);
```
twitu, line 76: Again a common pattern emerging from the Rc RefCell shared nature of the constructs. For core constructs, perhaps it's better to define a new wrapper struct and expose AsRef and AsMutRef on them. So that the explicit borrows can be reduced.

```
struct SharedOrderEmulator(Rc<RefCell<OrderEmulator>>)

impl AsRef<OrderEmulator> for SharedOrderEmulator {
   pub fn ... {
       self.0.borrow()
   }
}

File: nautilus_core/execution/src/matching_core/handlers.rs [hunk 3]
```diff
@@ -0,0 +1,100 @@
...
+        match self {
+            FillMarketOrderHandlerAny::OrderMatchingEngine(engine) => {
+                engine.borrow_mut().fill_market_order(&mut order.clone());
+            }
+            FillMarketOrderHandlerAny::OrderEmulator(emulator) => {
+                emulator.borrow_mut().fill_market_order(&mut order.clone());
+            }
+        }
+    }
+}
```
twitu, line 45: Depending on this the FillMarketOrderHandler is consumed it might be better to implement the `FillMarketOrderHandler` trait for the OrderMatchingEngine and OrderEmulator individually and store it as as Box<dyn FillMarketOrderHandler> or Rc<dyn FillMarketOrderHandler> that will reduce the number of enums and wrapper types created for this functionality.

twitu (commented):
I'm noticing that using the shared constructs and message handler style traits is not as ergonomic as it could be. There's a lot of cognitive overhead to Rc, RefCell, explicit borrows and wrapper enums that make it hard to understand the core logic.

I believe this is mostly because of the SharedMessageHandler is defined. I'll look into how it can be made a little more flexible either a type level or using proc macros. What do you think?

PR #2311, opened 2025-02-11
Refine bar aggregators in Rust

faysou:
# Pull Request

Refine rust bar aggregators

I've also done some changes to TestClock and LiveClock in order to be able to update a time alert without having to pass again a callback, which would be tricky if an object needs to pass a callback to itself.

## Type of change

Existing tests pass

### DIFF
--- a/crates/common/src/clock.rs
+++ b/crates/common/src/clock.rs
@@ -298,19 +298,25 @@ impl Clock for TestClock {
         callback: Option<TimeEventCallback>,
     ) -> anyhow::Result<()> {
         check_valid_string(name, stringify!(name))?;
+        let name_ustr = Ustr::from(name);
+
         check_predicate_true(
-            callback.is_some() | self.default_callback.is_some(),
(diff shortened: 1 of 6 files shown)


crates/common/src/clock.rs
```diff
@@ -298,19 +298,25 @@ impl Clock for TestClock {
...
         )?;
 
-        let name_ustr = Ustr::from(name);
         match callback {
             Some(callback_py) => self.callbacks.insert(name_ustr, callback_py),
             None => None,
         };
 
-        let ts_now = self.time.get_time_ns();
+        // This allows to reuse a time alert without updating the callback, for example for non regular monthly alerts.
```
cjdsellers, line 315: We could also push this logic [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/common/component.pyx#L337) down to Rust, as I think it is achieving the same thing.
(can be in a follow-up PR.)

PR #2321, opened 2025-02-15
Add backtest timer example

stefansimik:
# Pull Request

This is a new example for beginners: **How to use a Timer in NautilusTrader strategies**.

## Type of change

- [x] New feature (non-breaking change that adds functionality)

✅ The strategy was run and verified to work as expected  
✅ Pre-commit checks passed  

## Important notes:

1. I added a very small (1-month) sample of clean OHLCV data (UTC time encoded). The reason is that these new examples should be very beginner-friendly = they shoud be using the most common / accessible type of market data i.e. **1-minute OHLCV bars in text format**(rather than binary formats like Databento `.dbn`, which are not directly readable by beginners or withou having additional knowledge).  I chose this period 'JAN 2024' intentionally, because there is no rollover (to next expiration contract) in that period.
2. These 1-minute OHLCV text-based data are planned to be used in all future beginner-oriented examples, as they introduce no barriers or unnecessary complexity (such as subscribing to low-level data and aggregating it ... or reading binary format like databento .... or using `ParquetDataCatalog` at the beginning, which completely obfuscates the flow for beginners.).  
3. Since this type of data was not available in our repo, I added it in a highly storage-efficient way—zipped using MAX compression, reducing the size by a factor of 10 (original 1.8MB, zipped 206 KB). 
4. What is important  - these zipped CSV files can be **directly** read by pandas, eliminating the need for manual extraction. If anyone wants to inspect the underlying CSV data, they can easily unzip the files of course. This approach ensures easy accessibility while significantly saving storage space.  
4. Given the substantial storage savings, I believe using the ZIP format in justified cases should be allowed in the project. Since `.zip` files were initially excluded in `.gitignore`, I temporarily modified it to allow them, committed and then reverted the change after committing the data. 

**One last  thing to notice**
* I named the file: `6EH4.GLBX_1min_bars_20240101_20240131.csv.zip`
* i.e. it contains also **range of the data in the name**. Technically this is duplicate information, as everyone can check this in the underlying CSV file, but I found it useful and transparent to extract this info into the filename itself, so that information is immediately accessible, especially because I have used ZIP format, so checking this would require uncompressing data first.


examples/backtest/use_clock_timer/intro.md
```diff
@@ -0,0 +1,11 @@
+This is a simple example of how to use NautilusTrader's **Timer** feature in a strategy.
```
cjdsellers, line 1: Since we have dedicated directories for these examples, we could consider naming the file `README.md`? then when browsing on GitHub it will automatically render the markdown.

examples/backtest/use_clock_timer/run_backtest.py
```diff
@@ -0,0 +1,119 @@
...
+    engine_config = BacktestEngineConfig(
+        trader_id=TraderId("BACKTEST_TRADER-001"),
+        logging=LoggingConfig(
+            log_level="DEBUG",  # set DEBUG log level for console to see loaded bars in logs
+        ),
+    )
+    engine = BacktestEngine(config=engine_config)
+
+    # Step 2: Define exchange and add it to the engine
+    GLOBEX = Venue("GLBX")
```
cjdsellers, line 49: We're moving towards MIC codes for these venues (`XCME`). This is working for Databento except for continuous contracts.

examples/backtest/use_clock_timer/strategy.py
```diff
@@ -0,0 +1,83 @@
...
+from nautilus_trader.common.events import TimeEvent
+from nautilus_trader.core.datetime import unix_nanos_to_dt
+from nautilus_trader.model.data import Bar
+from nautilus_trader.model.data import BarType
+from nautilus_trader.trading.strategy import Strategy
+
+
+class SimpleTimerStrategy(Strategy):
+    TIMER_NAME = "every_3_minutes"
+    TIMER_INTERVAL = pd.Timedelta(minutes=33)
```
cjdsellers, line 30: Typo, should be `minutes=3`?

cjdsellers (commented):
Thanks for this example @stefansimik!

A couple of thoughts.

I think the only reason we didn't compress CSV's in the past (years ago) is because GitHub didn't allow it. This has probably changed now as I see we have a `.zip` file under `test_data/bybit/`. So I think it's a good idea to compress, but could we standardize on gzip with `.csv.gz` extensions?

Also, there's a larger discussion needed at some point on test data organization. As I mentioned previously, I think it was a mistake adding test data to be tracked by the repo, but there's a few other considerations here such as convenience, and reproducibility.

So if we do add a small amount of test data as we're doing here, which is intended to be reused across many examples, then I think thats fine.

But could we locate the file in a less nested location? as I can't see us adding significant amounts of test data under this structure `{venue}/{instrument_class}/{asset_class}`, and if we did - we'd want to consider using an approach more like `test_data/large` (depending on how and where it could be obtained), to avoid tracking all of the data files.

Thoughts?

examples/backtest/use_clock_timer/intro.md
cjdsellers: Could it be worth considering prefixes like `001_use_clock_timer` to sort the examples in a more logical order than alpha?

stefansimik:
Thank you very much Chris for all ideas and improvements.
I see the point in all of them and going to implement right now 👍

stefansimik:
As a separate thing, I had to rename folders of individual examples:
* from: `001_xxxxx`
* to: `examples_001_xxxxx`

to make MYPY checker happy. Originally it complained, because all examples have filenames with the same name like this:

CleanShot 2025-02-16 at 09 11 11@2x



I don't like the solution (renaming with prefix `example_`) because these examples are already contained in folder named `/examples`, so it is duplication. I would suggest, let's rename the folders back to format `001_example_name`, after we find proper fix/setting to make MYPY happy.

To reproduce the issue:
1. rename the 2 example folders back (so the prefix `example_` is removed)
2. run `pre-commit run --all-files` to popup the errors again

examples/backtest/use_clock_timer/intro.md
stefansimik: Done, but MYPY complained, after adding number at the beginning of the folder, so I had to fix it (make MYPY happy) with additional `examples_` prefix, as explained in later comment in this thread

examples/backtest/use_clock_timer/run_backtest.py
stefansimik: Implemented.

examples/backtest/use_clock_timer/strategy.py
stefansimik: Implemented.

cjdsellers:
Hi @stefansimik 

Understood on the naming issue, we'll figure something out.

Thanks for making the amendments -- will merge now :+1:

cjdsellers:
> Anyway, I understand both angles of view (it is not black and white), so I leave final decision up to you Chris,
> to choose, what you prefer more (i.e. 20% storage saving in all new archives VS. consistency of using *.gz).
> I am completely OK with both variants, just wanted to point out important things to consider for you.

This won't make a significant difference though, because we don't intend on storing large amounts of test data going forward.

It also depends on what we're doing with the file, as we use gzip for on-the-fly stream decompression and parsing in the `TardisCSVDataLoader`. In this example, it's working fine with `.zip` and there wouldn't be a benefit with gzip.

- `.zip` is an archive format that can contain multiple files and metadata but is less friendly for pure streaming decompression.
- `.gz` is a single-stream compression format designed for easy, incremental, on-the-fly (de)compression.

stefansimik:
Thank you Chris, 
with this example we have ironed out a lot small issues, so next similar PR will be smoother I believe 😊 👍

PR #2325, opened 2025-02-16
Add documentation for debugging with Rust

faysou:
# Pull Request

Add documentation for debugging with rust

### DIFF
--- a/docs/developer_guide/testing.md
+++ b/docs/developer_guide/testing.md
@@ -47,3 +47,21 @@ These tests are expensive to maintain (as they must be kept in line with any ref
 offer little to no benefit in return. The intention is for all abstract method
 implementations to be fully covered by tests. Therefore `pragma: no cover` should be judiciously
 removed when no longer appropriate, and its use *restricted* to the above cases.
+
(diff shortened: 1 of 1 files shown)


cjdsellers:
Thanks @faysou 

This is an interesting conundrum.

The `test` profile currently optimizes for compile times, and assumes using logs / prints for debugging. Where as the suggested profile here favors using a debugger (maybe in an IDE). I'll test out compile time differences because I'd rather not have this overhead to using a debugger if we can avoid it.

faysou:
I thought the same thing. I also tried to add another test config but the debugger in intellijidea uses the test profile by default, it's possible to use another profile but it's not practical

cjdsellers:
I tested the profile and unfortunately compile times are ~50% longer on my machine. So too significant to make this the default `test` profile I think.

faysou:
Ok no problem. I'll see if there's a way to use another profile as default for debugging in intellijidea, I didn't manage to do it yesterday.

faysou:
I think I've found a satisfying solution. I'm making a PR in a few minutes

PR #2327, opened 2025-02-16
Add backtest clock and timers example

stefansimik:
# Pull Request

Added example how to use Clock + Timer in strategies.

## Type of change

✅  Strategy works locally
✅  pre-commit passed
✅  we checked together with Chris (this is 2nd round, with all fixes applied)


cjdsellers:
Hi @stefansimik

Thank you again for all the follow-ups from #2321 :pray:

PR #2328, opened 2025-02-17
Enable debugging in cargo test profile

faysou:
# Pull Request

- Modify default test profile to enable debug symbols
- Add nextest profile to be equal to the previous test profile, disabling debug symbols.
- Modify existing make cargo-test comands to use nextest profile
- Add new make cargo-test commands to use the default modified test profile
- Modify documentation accordingly

This way there's no change in behaviour in CI when doing make cargo-test. Also make cargo-test-debug allows to compile everything using debug symbols so when using the debugger in intellijidea for example, there's no need to recompile everything.

This should improve the dev experience for people not familiar with nautilus by making debugging easy compared to before.

## Type of change

Existing rust tests still passing

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -11,3 +11,7 @@ test-group = 'serial-tests'
 [[profile.default.overrides]]
 filter = 'test(test_order_book)'
 slow-timeout = { period = "300s" }
+
+[profile.nextest]
+filter = "package(*)"
+cargo-options = ["--profile=nextest"]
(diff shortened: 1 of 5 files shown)


faysou:
I don't know why this commit always fails

cjdsellers:
That's also what I thought of, so every crate is compiled twice with two profiles which is blowing out disk space.
Maybe for this one commit you could manually remove any cached `target` dir?

faysou:
How to do that ?

PR #2329, opened 2025-02-17
Port handling of market-if-touched order for matching engine in Rust

filipmacek:
# Pull Request

- added `is_touch_triggered` function in `OrderMatchingCore`
- added full implementation for processing and updating of the market if touched orders in `OrderMatchingEngine` with two test cases


cjdsellers (approved):
:ok_hand:

PR #2346, opened 2025-02-22
Added new MRE example with artificial generated bars

stefansimik:
# Pull Request

It demonstrates a simplistic trading strategy with artificially generated market data,
making it perfect for quick reproducing issues.

## Type of change

✅  pre-commit passed
✅  strategy manually checked: works correctly and produces expected outputs in terminal

### DIFF
--- a/examples/other/minimal_reproducible_example/README.md
+++ b/examples/other/minimal_reproducible_example/README.md
@@ -1,0 +1,6 @@
+# NautilusTrader Minimal Reproducible Example
+
+This repository contains a minimal reproducible example (MRE) for the NautilusTrader trading framework.
+
+It demonstrates a simplistic trading strategy with artificially generated market data,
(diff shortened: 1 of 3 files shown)


stefansimik:
I checked and here is the same problem as in previous PR
`Error: Unhandled error: HttpError: Resource not accessible by integration`

So if one will be fixed, then all PRs will be OK.

faysou:
Why does your example involve http ?

stefansimik:
> Why does your example involve http ? And where does it ?

Hi @faysou ,
I have really no idea. Chris checked it and fixed something. He wrote, me it is not my fault and the some resource was missing.

So I closed the old problematic PR's and sent a new one's today and it works (Chris fixed it probably).

faysou:
Ah ok. You can update the branch instead.

stefansimik:
Thank you @faysou  for excellent tips ❤️ 🚀

PR #2356, opened 2025-02-25
Refine Rust date functions

faysou:
# Pull Request

Refine rust date functions


crates/core/src/nanos.rs
```diff
@@ -206,6 +206,12 @@ impl Display for UnixNanos {
     }
 }
 
+impl From<UnixNanos> for DateTime<Utc> {
```
cjdsellers, line 209: :ok_hand:

PR #2358, opened 2025-02-25
Refine adapters concept documentation

faysou:
# Pull Request

Refine adapters documentations

### DIFF
--- a/docs/concepts/adapters.md
+++ b/docs/concepts/adapters.md
@@ -123,6 +123,11 @@ A simplified version of the request handler implemented in a `LiveMarketDataClie
 and send it back to actors/strategies is for example:
 
 ```python
+# nautilus_trader/live/data_client.py
+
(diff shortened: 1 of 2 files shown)


docs/concepts/adapters.md
```diff
@@ -134,3 +139,27 @@ async def _request_instrument(self, request: RequestInstrument) -> None:
...
     self._handle_instrument(instrument, request.id, request.params)
 ```
+
+The `DataEngine` which is an important component in Nautilus links a request with a `DataClient`.
+For example a simplified version of handling an instrument request is:
+
+```python
+# nautilus_trader/data/engine.pyx
+
+self._msgbus.register(endpoint="DataEngine.request", handler=self.request)
```
cjdsellers, line 149: We should probably point out somewhere that this comes for free with the base classes, so it looks less complex.

Issue #2359, opened 2025-02-25
Incorrect handling of IB execution.time timestring when tz is UTC


FGU1:
# Bug Report

### Expected Behavior
No error upon execution of an order in live mode, with IB TWS / GW set to UTC.. 
order status change and opening of position visible in NT.

### Actual Behavior
`Handler task processor stopped.`
`"Error on '_run_msg_handler_processor': ValueError('not enough values to unpack (expected 2, got 1)') "`
No position opened in NT, no order status change.

### Steps to Reproduce the Problem

1. Set the TWS / GW timezone to UTC on a live account
2. Execute an order in NT, with debug log enabled

### Cause of the issue

IB's time format usually follow this format : "20250225 15:15:00 Universal"
When the tz is Universal, and mode is live (not paper) and only in this case, IB can also send it with this format : "20250225-15:15:00"
This 2nd format provoke this issue upon execution of an order.

### Specifications

- OS platform: Linux
- Python version: 3.11
- `nautilus_trader` version: Any
-  IB live account : the paper mode send in "normal" time format, while the live send the problematic time format so live testing is needed
- IB API version : issue validated from version 10.19 to 10.34

faysou:
I've just done a PR to fix this issue. You will need to test it works once it gets merged.

faysou:
I'll close my PR then ...

FGU1:
Your are right, your version is better. 
I just tested it in both live & paper mode and it is ok. 
I see no reason not to merge it.

PR #2360, opened 2025-02-25
Fix time parsing for IB adapter

faysou:
# Pull Request

Fix time parsing for ib adapter

In response to issue https://github.com/nautechsystems/nautilus_trader/issues/2359

## Type of change

Not tested, user will need to test it works. Likely fixes issue based on the time format provider by user.


FGU1:
Tested in paper & live environnement, ok for me.

PR #2365, opened 2025-02-26
Add catalog write modes

faysou:
# Pull Request

Add "new_file" catalog write mode

- Allows to append data to a catalog by creating a new file when using write_data without the partitioning optional argument , which uses the _fast_write function
- Also modified the update_catalog argument of data requests in actors to update_catalog_mode so a catalog write mode can be passed as argument
- Can be useful if you have data of various quality types and don't want to take the risk to degrade the quality of some existing data by appending/concatenating new data

## Type of change

Added a test, also tested with own code

### DIFF
--- a/nautilus_trader/common/actor.pyx
+++ b/nautilus_trader/common/actor.pyx
@@ -36,6 +36,7 @@ from nautilus_trader.common.config import ImportableActorConfig
 from nautilus_trader.common.executor import ActorExecutor
 from nautilus_trader.common.executor import TaskId
 from nautilus_trader.common.signal import generate_signal_class
+from nautilus_trader.persistence.catalog.types import CatalogWriteMode
 
 from cpython.datetime cimport datetime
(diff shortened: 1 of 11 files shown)


nautilus_trader/common/actor.pyx
```diff
@@ -2079,7 +2079,7 @@ cdef class Actor(Component):
         datetime end = None,
         int limit = 0,
         callback: Callable[[UUID4], None] | None = None,
-        bint update_catalog = False,
+        str update_catalog_mode = None,
```
cjdsellers, line 2082: We could also create a `CatalogWriteMode` enum (or similar)? just to lock the typing and chance of errors down a little more. A regular Python enum would work (doesn't have to be one which starts in Rust and needs 1000 lines of code).

cjdsellers (approved):
:ok_hand:

Issue #2367, opened 2025-02-26
Improve Binance adapter to handle greater number of websocket data streams


himanshu2406:
# Bug Report

### Expected Behavior
According to [Binance Docs ](https://developers.binance.com/docs/derivatives/usds-margined-futures/websocket-market-streams)
**A single connection can listen to a maximum of 200 streams.**

### Actual Behavior

According to : nautilus_trader\adapters\binance\websocket\client.py
```python
    async def _subscribe(self, stream: str) -> None:
        if stream in self._streams:
            self._log.warning(f"Cannot subscribe to {stream}: already subscribed")
            return  # Already subscribed

        self._streams.append(stream)

        while self._is_connecting and not self._client:
            await asyncio.sleep(0.01)

        if self._client is None:
            # Make initial connection
            await self.connect()
            return

        msg = self._create_subscribe_msg(streams=[stream])
        await self._send(msg)
        self._log.debug(f"Subscribed to {stream}")
```

This seems to subscribe to **all streams using a single connection** (self.connect() is called only once during init).

### Steps to Reproduce the Problem

1. try subscribing to 300 instruments for binance future during on_start()
2. it'll start throwing an error / banning the IP due to connection behaviour not complying with binance restrictions.

### Specifications

NA : functional issue.

himanshu2406:
@cjdsellers if it's okay with you, can i pick this up to help contribute ?

cjdsellers:
Hi @himanshu2406 

Thanks for reaching out and for the offer of a PR.

What you're suggesting is reasonable and your help on improving this with a PR would be much appreciated 🙏.

PR #2369, opened 2025-02-27
Improve Binance websocket connections management

lidarbtc:
# Pull Request

As explained in issue #2367, the adapter creates a new connection for each subscription. I changed this so that the adapter will reuse the same connection until there are 200 connections. Once the limit is reached, a new connection will be made.

## Type of change

- [O] New feature (non-breaking change which adds functionality)
- [O] This change requires a documentation update

I have many pairs to trade in Binance, just like #2367. I test approximately 250 instruments. Before the change, it was heavily rate limited and had connection problems. After the change, I confirmed that it was not rate-limited and works with over 250 instruments.

### DIFF
--- a/nautilus_trader/adapters/binance/websocket/client.py
+++ b/nautilus_trader/adapters/binance/websocket/client.py
@@ -16,7 +16,7 @@
 import asyncio
 from collections.abc import Awaitable
 from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


cjdsellers:
Many thanks @lidarbtc

Nicely done :ok_hand:.

Issue #2372, opened 2025-02-27
Add Multi Processing Feature Flag in BacktestNode


himanshu2406:
# Feature Request

### Problem Statement

AFAIK, the trade engine processes data bar by bar to ensure accuracy and prevent look-ahead bias, simulating a realistic live-trading scenario even in backtesting—which is great.
However, this all happens in a single thread. This can be verified by running a multi-asset backtest and checking your CPU usage—you’ll notice only one core being utilized. 
This makes sense since node.run() executes in a single thread, and event-driven architectures generally don’t benefit from parallelization unless bars across assets are synchronized via IPC, which would be overkill for the trade engine.

However , If we want to fully utilize all cores and speed up processing across multiple instruments, multiprocessing seems like a solid approach. It preserves the event-driven structure while maximizing resource usage.

### Proposed Solution

- In case of multi-asset backtests , we can provide a feature flag for enabling multiprocessing by adding an extra optional input parameter for BacktestNode 
- This will then take in multiple configs (each initialized with their own subset of different instruments) and then run them in parallel using Python's Processing Pool. 
- The results for all individual node runs will be compiled together and returned to the caller once the parallel execution is complete. 

### Current Implementation for the same solution : 

```python
with multiprocessing.Pool(NUM_PROCESSES) as pool:
        results = pool.starmap(run_backtest, [(split, i) for i, split in enumerate(instrument_splits)])

where run_backtest : 
def run_backtest(instrument_subset, process_id):
 <code to initialize BacktestRunConfig>
 node = BacktestNode(configs=configs)
 results = node.run()
 return results

and instrument_splits : 
NUM_PROCESSES = min(8, num_instruments)
instrument_splits = [instruments[i::NUM_PROCESSES] for i in range(NUM_PROCESSES)]
```

### Additional Context

The current solution / workaround works as expected - however @cjdsellers , if you think this seems like a valid enhancement that other users running multi-asset backtests could benefit from I'd be happy to open a PR to add this. 
According to my experiments , using multi-processing to compute instruments in subsets distributed across parallel processes VS all instruments on single process -> the first approach gave me over a 8-10x performance boost when compared with the latter

himanshu2406:
as discussed , this functionality shall remain abstracted to let user decide how he wants to handle the parallelization perspective 
cc : @cjdsellers

PR #2374, opened 2025-02-27
Add Parallel Config Processing to BacktestNode

himanshu2406:
# Pull Request

As shown in #2372 , `BacktestNode` currently has no way to run backtests in parallel for multiple configs even though parallelization in this part of code wouldn't effect the event-driven architecture but will simply give a performance boost. 

I've hence added a `run_parallel()` function that can be used as a drop in replacement for run() when working with multiple configs / instruments.
Since this functionality is abstracted out in a separate function, current functionality of run won't be affected and shall work same as before. 

## Type of change

Earlier snippet (configs needed to be wrapped externally to run backtests in parallel) : 

```python
with multiprocessing.Pool(NUM_PROCESSES) as pool:
        results = pool.starmap(run_backtest, [(split, i) for i, split in enumerate(instrument_splits)])

where run_backtest : 
def run_backtest(instrument_subset, process_id):
 <code to initialize BacktestRunConfig>
 node = BacktestNode(configs=configs)
 results = node.run()
 return results

and instrument_splits : 
NUM_PROCESSES = min(8, num_instruments)
instrument_splits = [instruments[i::NUM_PROCESSES] for i in range(NUM_PROCESSES)]
```

New Snippet : 

```python
NUM_PROCESSES = min(8, num_instruments)  # Use up to 8 processes (adjust based on CPU cores)
instrument_splits = [instruments[i::NUM_PROCESSES] for i in range(NUM_PROCESSES)]
configs = []

for i, instrument_subset in enumerate(instrument_splits):
 <code to initialize BacktestRunConfig>
 configs.append(run_config)


node = BacktestNode(configs=configs)
all_results = node.run_parallel(max_workers=NUM_PROCESSES)
```

Both the results are verified to be the same.

### DIFF
--- a/nautilus_trader/backtest/node.py
+++ b/nautilus_trader/backtest/node.py
@@ -50,6 +50,7 @@ from nautilus_trader.model.objects import Currency
 from nautilus_trader.model.objects import Money
 from nautilus_trader.persistence.catalog.parquet import ParquetDataCatalog
 from nautilus_trader.persistence.catalog.types import CatalogDataResult
+import concurrent.futures
 
 
 class BacktestNode:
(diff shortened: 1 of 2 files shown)


himanshu2406:
as discussed , this functionality shall remain abstracted to let user decide how he wants to handle the parallelization perspective
cc : @cjdsellers

PR #2375, opened 2025-02-27
Refactor reconciliation interfaces into messages

faysou:
# Pull Request

Refactor order report interfaces into messages

Similar refactor as for data messages.
Using messages makes it clearer to understand where interfaces between component are located.

This also opens up the way to call execution client's order report functionalities from other components through messages if necessary in the future.

## Type of change

- Refactor

Adapted existing tests


cjdsellers:
Hi @faysou 

Thanks again for the contribution!

I think this is a good change because it makes the client interfaces even more standardized after you went to the trouble of refactoring the other data interfaces to messages to match the style of the execution side :+1:.

PR #2377, opened 2025-02-27
Fix catalog path error on windows

faysou:
# Pull Request

Fix catalog path error on windows

## Type of change


cjdsellers:
Many thanks for the fix @faysou

PR #2387, opened 2025-02-28
Refactor repetitive code in examples

stefansimik:
# Pull Request

Hi Chris,
this is a sample of how I've designed the refactoring of `/examples` to minimize repetitive code writing.

The key was to maintain:
* The simplicity of examples (easy to follow code)
* Accessibility of the code (no complex dependencies, easy to find)
* Self-contained examples (at least in the sense that all relevant code will be inside `/examples` folder)

If you can find a better location for the `demo_data_provider.py` utility, I'm happy to make adjustments. Where I've placed it was the best solution I could come up with for now.

I deliberately refactored just one example: `examples/backtest/example_02_use_clock_timer`, so you can evaluate the result. When you confirm it's okay, I'll consistently update all scenarios.

(Note: Very first example `example_01_load_bars_from_custom_csv` should stay as it is (long code), because the CSV loading itself is the main point of the whole example, i.e. extracting it would not make sense.
But all others from example 2 and on... will be refactored.

I extracted everything that was truly repetitive into `demo_data_provider.py`, while other elements that aren't included there I determined should remain inside all the `run_example.py` files.

I am open to any suggestions.

## Type of change

Refactoring of existint code to reduce writing duplicated code in examples.

* ✅  Refactored strategy still works by running simple: 
   * Run this from project NT folder:  `python examples/backtest/example_02_use_clock_timer/run_example.py`
* ✅  pre-commit passed


cjdsellers:
Hi @stefansimik

Great, this is moving in the direction I was thinking :+1: .

Just a small suggestion - we could shorten the module name for the data provider slightly (to `data_provider` ?), as it's already nested under `examples`.

stefansimik:
> Just a small suggestion - we could shorten the module name for the data provider slightly (to `data_provider` ?), as it's already nested under `examples`.

Yeah I like that too, it is better / simpler 👍
(and no need to narrow down the name, as there only single central generic data provider for now)

Update pushed ✅ 
and the refactorign of remaining examples will follow up soon.

PR #2391, opened 2025-03-01
Add test for time-bar aggregation

stefansimik:
# Pull Request

With @faysou , we made an excellent team today and divided the tasks between us.

* while @faysou was focusing on the fix [PR#2930](https://github.com/nautechsystems/nautilus_trader/pull/2390)
* I prepared the test in the meantime, that should ensure any similar problem will detected easily in the future

This test fails and indicates problem when composite bars aggregation stops working.

## Type of change

Just new test

* **before** @faysou's fix - the test "successfully" fails and indicates that there's any problem with bar aggregation that stopped working
* **after** @faysou's fix - the test runs smoothly


faysou:
Thanks for this, the test looks good. I was thinking it's hard to piece together all components together for testing this properly, but the way you does with the low level API and using the experience from your examples works well instead.

PR #2400, opened 2025-03-03
Add custom event with msgbus example

stefansimik:
# Pull Request

New tutorial style example for beginners -Hhow to use: Custom event with msgbus.

# Thing to note:

I have used standard `@dataclass` and it works without visible problems.
Is that OK, or is there any motivation and benefits using `@customdataclass` instead.

As it is now in the example, probably no data are serialized (they live just in memory),
so this implementation looks like a quick & easy way how to publish/subscribe data/events,
when there is no need for serialization... (or is it always required to handle serialization in all cases?)

I got some hints from @faysou , that `@customdataclass` should be used to allow serialization
like here: https://nautilustrader.io/docs/nightly/concepts/data/#option-greeks-example,
but is this increased complexity always required, or is it completely OK use this lightweight
version and having simple code, when we don't expect any serialization in our strategy?

In this example, I inherited from `nautilus_trader.core.message.Event`.
* Is that valid approach or is there some reason to inherit from `nautilus_trader.core.Data`?
* Or what if we don't inherit from anything at all - I tested it and the strategy still works!

Just asking, to understand if there are some strong requirements (or reasonable recommendations) how to interact with msgbus, or we can simply publish/subscribe to the msgbus what-ever object we want (no need for any subclass).
After testing it, it looks there is really no need to subclass anything and strategy still works great! 😊 

## Type of change

New example added.

✅  Strategy generates expected output in Terminal
✅  pre-commit passed

### DIFF
--- a/examples/backtest/example_09_custom_event_with_msgbus/README.md
+++ b/examples/backtest/example_09_custom_event_with_msgbus/README.md
@@ -1,0 +1,19 @@
+# Example: Custom events and Message Bus Publish/Subscribe Pattern
+
+This example demonstrates:
+* how to create and use custom events with the message bus in a **NautilusTrader** strategy.
+* how to create a custom event, publish it when specific conditions are met and subscribe to handle these events.
+
+This pattern is not only useful for proper event-driven communication between different parts of your trading system,
+but also for generating and receiving events within the same Actor/Strategy.
+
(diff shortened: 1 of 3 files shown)


faysou:
I suppose serialisation allows a message to be transported between nodes, but it's true that for simple cases and if you use the message bus directly there's no need for added complexity. Maybe @cjdsellers can comment on this.

PR #2410, opened 2025-03-05
Add docs for messaging styles

stefansimik:
# Pull Request

New documentation: Messaging styles

Created based on hints from @faysou , @cjdsellers , @demonkoryu  🙏 
Thank you all guys.

## Type of change

New documentation

New documentation added

✅  precommit passed

## TODO for Chris 
( @cjdsellers  )

1. New link should be placed **after the existing** "message_bus.md" file since it builds upon those concepts while introducing additional messaging patterns.
2. Double check please and make proper corrections, where needed 🙏 

The navigation structure would look like:

```
   - Architecture
   - Actors
   - Backtesting
   - Data
   - Message Bus
   - **Messaging Styles** (new content)
   - Strategies
   - ...
 ```
 
 **Why this placement?**

1. The content directly relates to both the message bus and actor concepts - serving as a bridge between these topics.
2. The existing "Message Bus" documentation focuses primarily on the technical implementation details of the message bus itself, while this new content focuses on practical usage patterns and comparisons between different messaging approaches.

### DIFF
--- a/docs/concepts/messaging_styles.md
+++ b/docs/concepts/messaging_styles.md
@@ -1,0 +1,201 @@
+# Messaging styles
+
+NautilusTrader is an **event-driven** framework where components communicate by sending and receiving messages.
+Understanding the different messaging styles is crucial for building effective trading systems.
+
+This guide explains the three primary messaging patterns available in NautilusTrader:
+
+| **Messaging Style** | **Purpose** | **Best For** |
+|:---|:---|:---|
+| **MessageBus - Publish/Subscribe to topics** | Low-level, direct access to the message bus | Custom events, system-level communication |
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @stefansimik

Thanks for the great doc! we're really getting there with adding implicit / tribal knowledge into the docs.

I'd suggest this should go either in the message bus concept guide, or the actor concept guide though? I don't think we need an entirely separate guide for this just yet? (I can move this after merge if you like? to avoid churning CI.)

On terminology, it's an interesting differentiation uncovered here: messaging _pattern_ (e.g., pub/sub, req/res, point-to-point), vs the messaging _style_ within a pattern.

stefansimik:
Thank you Chris,
feel free to relocate after merge please, if this is quick & easy for you.
Otherwise, I can do this of course 😊 👍

PR #2411, opened 2025-03-05
Refine update catalog docstrings

faysou:
# Pull Request

Refine update catalog docstrings

Remove end and update catalog must not be set together actor condition

Fixed default catalog in update catalog


nautilus_trader/common/actor.pyx
```diff
@@ -2113,6 +2113,8 @@ cdef class Actor(Component):
             Recommended catalog write modes are:
             - UpdateCatalogMode.MODIFY (appends or prepends new data to the catalog by concatenating it to the existing unique parquet file).
             - UpdateCatalogMode.NEWFILE (appends new data to the catalog by creating a new parquet file).
+            It is assumed that the data in a catalog for an instrument id is contiguous in time.
```
stefansimik, line 2116: typo: continuous

faysou: no it's not a typo

PR #2413, opened 2025-03-05
Refine Makefile to install uv venv

faysou:
# Pull Request

Refine Makefile to install uv venv

Add additional instructions to allow installation of a uv venv, the default behaviour, that is now hard to access when using --active

Having make install and make install-uv for example allows both possibilities, installing in a pyenv venv or installing a uv venv in the nautilus folder, that can be then activated with source .venv/bin/activate

## Type of change

- Improvement

Tested it works as expected

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,19 @@ install-debug:
 install-just-deps:
 	uv sync --active --all-groups --all-extras --no-install-package nautilus_trader
 
+# Install instructions that also set up a uv venv inside the nautilus directory
+.PHONY: install-uv
+install-uv:
+	BUILD_MODE=release uv sync --all-groups --all-extras
+
+.PHONY: install-debug-uv
+install-debug-uv:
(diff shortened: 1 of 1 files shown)


cjdsellers:
I'd have to go back and study the uv docs some more.

But I'm thinking we offer opinionated best practices for the Makefile, to keep it smaller and simpler. If someone wants to do something unusual or support a non standard use case, they can just run the command directly themselves?

Or do you find yourself needing both versions of the target yourself?

faysou:
Well it can become the standard to install a uv venv, the only reason I'm not using it at the moment is that it's not supported by intellijidea ultimate for tests. But as soon as it's supported I'll switch to it.

I didn't find how to install a uv venv with --active, so I thought it would be good to have both options.

Once a uv venv is activated then using --active for other commands works. It's only for installing that removing --active is useful.

PR #2428, opened 2025-03-09
Add logic to handle fractional volume for historical bar #2427

Endura2024:
# Pull Request

Stock reverse split caused some bar's volume < 1, which resulted ValueError('Invalid `value` for quantity: 0.24 was rounded to zero due to size increment 1 and size precision 0') was thrown. The fix is to round fractional volume before passing it to instrument.make_qty

## Type of change

Tested with script to call HistoricInteractiveBrokersClient.request_bars with specific date range, e.g. 2023-01-17 that as <1 volume on some bars to compare behavior before and after code change

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/market_data.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/market_data.py
@@ -650,6 +650,10 @@ class InteractiveBrokersClientMarketDataMixin(BaseMixin):
         ts_event = await self._ib_bar_to_ts_event(bar, bar_type)
         # used to be _convert_ib_bar_date_to_unix_nanos
 
+        # handle fractional quantity that likely caused by stock reverse split
+        if instrument.size_precision == 0 and instrument.size_increment >= 1 and bar.volume < 1:
+            bar.volume = round(bar.volume)
+
         return Bar(
             bar_type=bar_type,
             open=instrument.make_price(bar.open),


cjdsellers:
Hi @Endura2024

Thanks for the report.

Does a fractional volume suggest the `size_precision` needs to be greater than zero, to match the data?
The issue might actually be further upstream here with how the instrument is being defined.

Endura2024:
@cjdsellers  I assume change size_precision can fix the issue as well, but I don't know whether change size_precision  is a safe fix given most live data should have volume as integer. fractional volume is result of stock reverse split and only occurs for old historical data. Please advise what is the right solution to handle fractional volume in historical bar.

Thanks!

stefansimik:
After reading the thread and @cjdsellers' explanation, let me share my thoughts:

The situation with fractional volumes in historical data is **valid but rare**, occurring specifically with:
* Reverse stock splits only
* Historical data only (when volumes are back-adjusted)

Since this represents a real scenario that can't be ignored, I see three potential approaches:
1. Allow fractional volumes globally (which would reduce appropriate validation for 99% of other instruments)
2. Create a configurable setting to allow fractional volumes (default to `False`)
3. Continue disallowing fractional volumes (current behavior, not good)

Option 2 seems like the most logical choice - adding a parameter like `allow_fractional_volume` would maintain validation for most cases while handling these specific historical data scenarios.

I haven't examined exactly where/at what level this setting should be implemented (whether at a global configuration level or as an extension to the instrument definition) - if you have ideas, please share and we can easily choose the optimal approach if this solution makes sense to you.

faysou:
I don't think this would be doable as in various places of the system like bar aggregation and order execution, bar volume is checked against instrument properties. I think the best think that can be done is to round the volume to size increment property of an instrument.

stefansimik:
Mhm, it looks like we are facing a challenge with supporting two scenarios simultaneously:
to support 2 worlds at once:
* One before the stock reverse split
* A second after the split has occurred

It would be interesting to see, how other platforms deal with this problem 😊

cjdsellers:
After considering this some more, it might be best _not_ to support the handling of fractional volumes due to back adjustments from capital events?

I think the goal of backtesting with Nautilus is to create an event driven simulation which is as close to what would have happened live as possible. If your models need adjusted prices, then I think they should be fed separately and directly, rather than flowing through the entire system.

You also want to know your strategy/system handles capital events causing large price changes well. You can't know what the adjustment factor is until after the event occurs as well (so there is arguable some look ahead bias by using them to adjust).

This approach would keep things as simple as possible.

stefansimik:
Makes perfect sense too 👍 
Sometimes keeping things simple for 99.9% situations is better,
than being able to handle 100% with added complexity

Endura2024:
@cjdsellers 
I read the comments. Though overall reverse split is not frequent events, it happens not rarely to leveraged and inverse ETFs, small cap biotech stocks, as they usually have 10:1 reverse split and coupled with low trading volume.  The fix I did mostly impact historical bar code path. Please let me know whether the change acceptable or other solution is preferred?

Thanks!

cjdsellers:
@Endura2024

The fix is mostly a work around which introduces incorrect data by inflating the volume to the minimum size increment.
Per the above, I don't think its worth the added complexity to accommodate the use case?

I'd suggest not using adjusted data for backtesting with Nautilus -- only for model research, validation and direct inputs for evaluation.

PR #2433, opened 2025-03-11
Fix unbound variable for Bybit

sunlei:
# Pull Request

Fix unbound variable issue for Bybit and optimize with local variable for better performance.

## Type of change


sunlei:
Additionally, local variables can be used in other adapters to improve performance. I will address this in a separate PR for consistency.

PR #2434, opened 2025-03-11
Use local variables for better performance

sunlei:
# Pull Request

- Use local variables for better performance
- Fix several unbound variable issues

## Type of change

### DIFF
--- a/nautilus_trader/adapters/betfair/execution.py
+++ b/nautilus_trader/adapters/betfair/execution.py
@@ -86,10 +86,14 @@ from nautilus_trader.model.events.order import OrderFilled
 from nautilus_trader.model.identifiers import AccountId
 from nautilus_trader.model.identifiers import ClientId
 from nautilus_trader.model.identifiers import ClientOrderId
+from nautilus_trader.model.identifiers import InstrumentId
+from nautilus_trader.model.identifiers import StrategyId
 from nautilus_trader.model.identifiers import TradeId
+from nautilus_trader.model.identifiers import TraderId
 from nautilus_trader.model.identifiers import VenueOrderId
 from nautilus_trader.model.objects import Currency
 from nautilus_trader.model.objects import Money
+from nautilus_trader.model.objects import Price
 from nautilus_trader.model.objects import Quantity
 from nautilus_trader.model.orders import Order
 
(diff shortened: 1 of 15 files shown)


cjdsellers:
Hi @sunlei

Thanks for the PR and for separating this from the initial fix.

This is a very large change, which has a high risk of bugs due lower test coverage across the live adapters.
We might be looking at a very marginal benefit here for performance, and arguably readability?

Also, some of the handler functions can be very short where assigning local variables becomes a bit redundant and noisy.

I'll wait for some thoughts from others.

CC: @faysou

sunlei:
@cjdsellers 
This PR primarily reverts the changes introduced in #2375 related to how variables are accessed inside functions. Therefore, it can be diffed against b0036b3 (git diff b0036b3 aec8089 nautilus_trader/adapters).

Additionally, it includes fixes for several unbound variable issues.

cjdsellers:
Hi @sunlei 

Thanks for the additional context.

If you've gone to the trouble of diffing the changes then that helps to reduce the chance of bugs.

As you say, attribute access does have some cost in Python, even for Cython types (less than a microsecond but it can add up).

I find the code without the continual `command.` dereferencing _slightly_ more readable as well.

sunlei:
@cjdsellers 

Alternatively, I can split this PR into several smaller PRs or commits to reduce the complexity of the review.

faysou:
The goal of the refactoring of the code was to reduce the complexity of the code by keeping the variables defined in the messages. I don't think this PR helps or why to reverting to the situation of before is useful especially as commands are not often called. You don't subcribe or request data or reports often. So it just adds a lot of line codes for nothing fromy point of view, these codes are not part of a tight loop

sunlei:
@faysou

Requests and data subscriptions indeed aren't called frequently.
The main goal of this PR is to optimize high-frequency operations like order placement. Other modifications are made for consistency.

PS: Personally, I think that repeatedly extracting values from `command` increases cognitive load. For example, some parts currently have unbound variable bugs due to this approach.

cjdsellers:
After considering this some more, I think we could go with a compromise here where we extract the command attributes into local variables just for the Bybit methods which you deem are performance critical @sunlei?

Also if you could retain the unbound variable bug fixes that would be great.

I think a handful of attribute accesses on a Cython implemented type still add up to less than a microsecond, so if we're optimizing at the 10's of mics to millisecond level in Python then there are probably other more expensive things to be looking at - which also might not be worth it if usability is impacted (such as reducing logging).

One nice property of the command message which hasn't been mentioned in this issue yet is that the attributes are immutable thanks to Cython's `readonly` qualifier, and its easy to see which parameters are entering the method through the command.

sunlei:
I'm abandoning this PR.  

I can submit a separate PR with modifications specific to Bybit.  

However, I still stand by my opinion that when a value from `command.` needs to be accessed multiple times, using a local variable is preferable. It improves readability, reduces cognitive load, and lowers the likelihood of bugs.  

I'm not sure if PyCharm can reliably detect whether a specific attribute of `command.` exists, but using local variables allows IDEs and linting tools to better verify variable existence.  

Additionally, does adding 1-3 assignment statements at the beginning of a function really increase the likelihood of bugs?

PR #2435, opened 2025-03-11
Remove printing of toto

davidsblom:
# Pull Request

Remove printing of toto

Not tested


davidsblom:
@faysou  I've removed printing of `toto`. Is there a todo in that unsubscribe method still?

Issue #1063, opened 2023-03-24
Decide and enforce a data retention policy for the Cache


micapumono:
# Feature Request

To my understanding, all of the events that happen to the cache to build up state are saved. They are then replayed on startup in reconciliation to rebuild state. For continuous trading, it is important to at some point start forgetting trades, orders, flat positions. 

I would suggest something along the lines of:

- Forget flat positions - unless they hold some useful info
- Forget trades that led to a flat position 
- Forget orders that are now closed
- I'm not sure if commands are saved somewhere (nor why) but forget those after some time too

Things to consider:

Much of these items would still be useful for post trade analysis, the request is only that the live trading client cache doesn't remember all these things, they can still be stored in the database just not loaded up on startup. Perhaps they could be moved to a different key like prepending the key with HISTORICAL: vs LIVE:

For each of these things, it might still be worth having them around for some time, someone might have use cases for 
"what are my trades in the last N minutes" or 
"let's send a maker order to the order book that has historically had the most successful maker fills"

So instead of strict rules policy like the above, maybe every item should be stored for at least N minutes. This could be simpler as well.

cjdsellers:
I like this direction, and the functionality is definitely needed.

I believe the main use case here is for continuous trading, we want to avoid an unbounded increase in memory consumption as more and more objects / events are held in the cache.

I don't think we should immediately forget/drop out of memory those objects you've listed above though. The information is too useful to query for all kinds of operations, and the space required in memory is not _that_ high, and there would be some overhead bringing 'forgotten' objects back (even from Redis).

I think there should be a configurable max lookback though, and objects in the cache which fall outside that are then automatically 'trimmed' out of memory. Do we need configuration per object type though? that may not be wise as some objects rely on others being available for correct state.

micapumono:
So were you thinking a strict max look-back applied on the database side? How would this work for dependent objects such as the events that are part of a position or for positions that are not flat but very old.

Troubladore:
Might this follow a "garbage collection" framework?

Would it be possible to inspect indicators and actors for active references, to understand what the dependencies are?

[...]

PR #2440, opened 2025-03-12
Clear cache temporary workaround

sunlei:
# Pull Request

Add `drop_accounts_on_reset` & `drop_positions_on_reset` for `CacheConfig` (default `True` to retain current behavior).

In the case of a large number of orders, `.reset()` can be used to clear the cache to prevent memory usage from growing.

A toggle can be provided to control whether accounts and positions information is retained.

This is especially important for account information, as it is essential for the strategy's execution. Otherwise, it will trigger a `no account registered for` error.

## Type of change

### DIFF
--- a/crates/common/src/cache/config.rs
+++ b/crates/common/src/cache/config.rs
@@ -37,6 +37,10 @@ pub struct CacheConfig {
     pub flush_on_start: bool,
     /// If instrument data should be dropped from the cache's memory on reset.
     pub drop_instruments_on_reset: bool,
+    /// If account data should be dropped from the cache's memory on reset.
+    pub drop_accounts_on_reset: bool,
+    /// If position data should be dropped from the cache's memory on reset.
+    pub drop_positions_on_reset: bool,
     /// The maximum length for internal tick deques.
     pub tick_capacity: usize,
     /// The maximum length for internal bar deques.
@@ -57,6 +61,8 @@ impl Default for CacheConfig {
             use_instance_id: false,
             flush_on_start: false,
             drop_instruments_on_reset: true,
+            drop_accounts_on_reset: true,
+            drop_positions_on_reset: true,
             tick_capacity: 10_000,
             bar_capacity: 10_000,
             save_market_data: false,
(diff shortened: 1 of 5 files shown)


cjdsellers:
Hi @sunlei 

Thanks for the PR.

My initial impression is that this increases the chance of misconfiguration, say if you did enabled these options for repeated backtest runs - retaining instruments is not as critical, unless you had many versions which change over time.

Resetting the cache also clears the order state index's, so if the system is controlling active orders there will be a desync.
So we wouldn't want this blanket cache clearer with filters approach to become standard or recommended?

[This issue](https://github.com/nautechsystems/nautilus_trader/issues/1063) is related.

Also being discussed here, and in various other threads on discord:
https://discord.com/channels/924497682343550976/1343560318650945597

So I think this needs a more permanent solution, and I think it might be clearer to approach from the other direction and explicitly purge certain objects from the cache, rather than dropping all with a filter?

I'd be hesitant to merge this if we think this might not be the final solution for managing the cache object accumulation, because we'd want to remove these config options again after a better solution arrives - which would be a breaking change.

Thoughts?

sunlei:
@cjdsellers 

This is indeed a temporary workaround for the caching issue, and I didn't take backtesting into account since I rarely use it.

As we discussed on Discord, I have a solution that works for my own use case.

The reason I submitted this PR is that I recently saw someone else encountering the caching issue on Discord again, so I thought a small change could provide a temporary fix.

I also understand your perspective, so it's totally fine if this doesn't get merged. :)

cjdsellers:
Hi @sunlei 

Thanks, I think this is a decent workaround for object accumulation in the cache when trading live.

Lets leave this open for visibility for now, until we have a more permanent solution. I'll notify on discord that people can try this approach if they like.

cjdsellers:
Closing this as we now have the following cache purging methods available on `develop` branch:

- https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/cache/cache.pyx#L763
- https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/cache/cache.pyx#L790
- https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/cache/cache.pyx#L817
- https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/cache/cache.pyx#L862

Explanation in docs here:

https://nautilustrader.io/docs/nightly/concepts/live#memory-management

PR #2442, opened 2025-03-12
Update Rust parquet catalog with recent features

faysou:
# Pull Request

Update rust parquet catalog with recent features

Added different kinds of adding data to an existing parquet file as well as the ability to consolidate several parquet files into one to match what is already done in python, and timestamp_bound

## Type of change

Not tested yet

### DIFF
--- a/tests/test_data/large/checksums.json
+++ b/tests/test_data/large/checksums.json
@@ -1,5 +1,5 @@
 {
-  "databento_mbo_xnas_itch.csv": "sha256:5481466e7776659f7f093b1ca8b05c7e88b8d586036692089cca5decc7965099",
+  "databento_mbo_xnas_itch.csv": "sha256:ac66f2775ebbfd71b5ac1e62f47df5d151150b18323eaeb4bcc88b330ee7a456",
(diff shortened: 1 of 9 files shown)


tests/test_data/large/checksums.json
```diff
@@ -1,5 +1,5 @@
 {
-  "databento_mbo_xnas_itch.csv": "sha256:5481466e7776659f7f093b1ca8b05c7e88b8d586036692089cca5decc7965099",
+  "databento_mbo_xnas_itch.csv": "sha256:ac66f2775ebbfd71b5ac1e62f47df5d151150b18323eaeb4bcc88b330ee7a456",
```
cjdsellers, line 2: It's odd this checksum changed, probably a partial download unless Databento data has recently changed.
We're using it for an order book test which isn't fully completed, so not asserting anything.

I'll redownload myself after merge and see if the checksums have changed.

cjdsellers: I also get the new checksum after redownloading, so the data changed - which we expected might happen at some point.

Also highlights one of the benefits of these checksums.

cjdsellers: Looks like it was a partial or corrupted download, as I get the original checksum.

faysou:
Yes I must have commited by accident, something not practical is that make cargo-test changes this file, so often I need to make a rollback of it. Something to fix if possible

PR #2444, opened 2025-03-12
Improve InstrumentProvider error handling when loading

davidsblom:
# Pull Request

Catch Exceptions when loading the instruments.

This is a follow up from https://github.com/nautechsystems/nautilus_trader/pull/2437. The state variable `self._loading` needs to be reset to `False`. Hence, the previous fix still resulted in the fact that the instruments were not being reloaded when a Bybit error occured.

## Type of change

Live example

### DIFF
--- a/nautilus_trader/adapters/bybit/data.py
+++ b/nautilus_trader/adapters/bybit/data.py
@@ -33,7 +33,6 @@ from nautilus_trader.adapters.bybit.common.enums import BybitEnumParser
 from nautilus_trader.adapters.bybit.common.enums import BybitProductType
 from nautilus_trader.adapters.bybit.common.parsing import get_interval_from_bar_type
 from nautilus_trader.adapters.bybit.common.symbol import BybitSymbol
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @davidsblom 

Thanks for the PR!

In this case is the Bybit error recoverable on a retry, and not a bug in the adapter?

davidsblom:
Sure, that would be another way to fix it. Right now, the Bybit adapter doesn't use a retry manager. We could do that too if you prefer that.

cjdsellers:
I think this is a good change just because the provider clearly isn't loading anymore if it's handled an error and stopped processing the instruments.

I was thinking of a scenario where we're continually retrying on a bug, but I don't think that's the case here?

davidsblom:
The Bybit adapter should try every hour to fetch the instruments. As of now, there is no retry manager used in case of http errors. This fix just logs the error, and then one hour later the instruments are fetched again. 

Could follow up with a retry manager for the Bybit adapter.

PR #2448, opened 2025-03-13
Some minor performance optimizations

sunlei:
# Pull Request

Some minor performance optimizations.

## Type of change

### DIFF
--- a/nautilus_trader/adapters/bybit/execution.py
+++ b/nautilus_trader/adapters/bybit/execution.py
@@ -390,9 +390,7 @@ class BybitExecutionClient(LiveExecutionClient):
                 client_order_id = None
 
         self._log.info(
-            f"Generating OrderStatusReport for "
-            f"{repr(client_order_id) if client_order_id else ''} "
-            f"{repr(venue_order_id) if venue_order_id else ''}",
+            f"Generating OrderStatusReport for {repr(client_order_id) if client_order_id else ''} {repr(venue_order_id) if venue_order_id else ''}",
         )
         try:
             bybit_symbol = BybitSymbol(instrument_id.symbol.value)
@@ -665,8 +663,7 @@ class BybitExecutionClient(LiveExecutionClient):
 
         if order.is_closed:
             self._log.warning(
-                f"`CancelOrder` command for {command.client_order_id!r} when order already {order.status_string()} "
-                "(will not send to exchange)",
+                f"`CancelOrder` command for {command.client_order_id!r} when order already {order.status_string()} (will not send to exchange)",
             )
             return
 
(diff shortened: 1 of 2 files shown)


nautilus_trader/adapters/dydx/execution.py
```diff
@@ -739,30 +739,34 @@ async def generate_position_status_reports(
...
+            ws_message_channel = ws_message.channel
+            ws_message_type = ws_message.type
+
+            if ws_message_type == "channel_data":
+                if ws_message_channel == "v4_block_height":
+                    self._handle_block_height_channel_data(raw)
+                elif ws_message_channel == "v4_subaccounts":
+                    self._handle_subaccounts_channel_data(raw)
+                elif ws_message_channel == "v4_markets":
+                    self._handle_markets(raw)
```
davidsblom, line 754: Perhaps add an `else` block which logs an error when a message with an unknown channel is received.

nautilus_trader/adapters/dydx/execution.py
```diff
@@ -739,30 +739,34 @@ async def generate_position_status_reports(
...
+            elif ws_message_type == "subscribed":
+                if ws_message_channel == "v4_block_height":
+                    self._handle_block_height_subscribed(raw)
+                elif ws_message_channel == "v4_subaccounts":
+                    self._handle_subaccounts_subscribed(raw)
+                elif ws_message_channel == "v4_markets":
+                    self._handle_markets_subscribed(raw)
```
davidsblom, line 761: Perhaps add an `else` block here too which logs an error when a message with an unknown channel is received.

davidsblom (commented):
Thanks for your changes!

sunlei:
@davidsblom Added.

nautilus_trader/adapters/bybit/execution.py
```diff
@@ -884,13 +882,12 @@ async def _submit_order_list(self, command: SubmitOrderList) -> None:
...
                 submit_orders.append(batch_order)
 
-            now_ns = self._clock.timestamp_ns()
             for order in batch_submits:
                 self.generate_order_submitted(
                     strategy_id=order.strategy_id,
                     instrument_id=order.instrument_id,
                     client_order_id=order.client_order_id,
-                    ts_event=now_ns,
+                    ts_event=self._clock.timestamp_ns(),
```
cjdsellers, line 890: If we're optimizing for performance, then this creates a separate method call for every order.
I think it's also clearer from an event perspective that the orders in the batch were submitted together?

sunlei: My initial intention for modifying this was to improve the measurement of relevant data.

However, after reading your comment and giving it more thought, I agree that performance is more important.

So, I'll submit a PR later to revert to the original logic.

PS: Sorry for the back-and-forth changes in this PR.

PR #2454, opened 2025-03-16
Refine Rust catalog file operations

faysou:
# Pull Request

Refine rust catalog file operations
Made it work in python

## Type of change

Added python tests

### DIFF
--- a/crates/persistence/src/backend/catalog.rs
+++ b/crates/persistence/src/backend/catalog.rs
@@ -15,7 +15,7 @@
 
 use std::path::PathBuf;
 
-use anyhow::Result;
+use anyhow::{Result, anyhow};
 use datafusion::arrow::record_batch::RecordBatch;
 use heck::ToSnakeCase;
 use itertools::Itertools;
@@ -26,10 +26,8 @@ use nautilus_model::data::{
 };
 use nautilus_serialization::{
     arrow::{DecodeDataFromRecordBatch, EncodeToRecordBatch},
-    parquet::{
-        ParquetWriteMode, combine_data_files, min_max_from_parquet_metadata,
-        write_batches_to_parquet,
-    },
+    enums::ParquetWriteMode,
+    parquet::{combine_data_files, min_max_from_parquet_metadata, write_batches_to_parquet},
 };
 use serde::Serialize;
 
@@ -87,7 +85,6 @@ impl ParquetDataCatalog {
         let _ = self.write_to_parquet(bar, None, None, None, write_mode);
     }
 
-    #[must_use]
     pub fn write_to_parquet<T>(
         &self,
         data: Vec<T>,
@@ -95,19 +92,17 @@ impl ParquetDataCatalog {
         compression: Option<parquet::basic::Compression>,
         max_row_group_size: Option<usize>,
         write_mode: Option<ParquetWriteMode>,
-    ) -> PathBuf
+    ) -> Result<PathBuf>
     where
         T: GetTsInit + EncodeToRecordBatch + CatalogPathPrefix,
     {
         let type_name = std::any::type_name::<T>().to_snake_case();
         Self::check_ascending_timestamps(&data, &type_name);
-
-        let batches = self.data_to_record_batches(data);
-        let batch = batches.first().expect("Expected at least one batch");
-        let schema = batch.schema();
-        let instrument_id = schema.metadata.get("instrument_id");
-        let path =
-            path.unwrap_or_else(|| self.make_path(T::path_prefix(), instrument_id, write_mode));
+        let batches = self.data_to_record_batches(data)?;
+        let schema = batches.first().expect("Batches are empty.").schema();
+        let instrument_id = schema.metadata.get("instrument_id").cloned();
+        let new_path = self.make_path(T::path_prefix(), instrument_id, write_mode)?;
+        let path = path.unwrap_or(new_path);
 
         // Write all batches to parquet file
         info!(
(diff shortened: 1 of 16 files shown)


tests/unit_tests/persistence/test_catalog_pyo3.py
```diff
@@ -0,0 +1,200 @@
...
+from nautilus_trader.core.nautilus_pyo3 import Symbol
+from nautilus_trader.core.nautilus_pyo3 import Venue
+from nautilus_trader.persistence.catalog import ParquetDataCatalog
+
+
+AUDUSD_SIM = InstrumentId(Symbol("AUD/USD"), Venue("SIM"))
+ONE_MIN_BID = BarSpecification(1, BarAggregation.MINUTE, PriceType.BID)
+AUDUSD_1_MIN_BID = BarType(AUDUSD_SIM, ONE_MIN_BID)
+
+bar1 = Bar(
```
cjdsellers, line 22: The `TestDataStubs` could be useful here for stub bars: 

https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/test_kit/stubs/data.py#L197.

nautilus_trader/core/nautilus_pyo3.pyi
```diff
@@ -3480,6 +3486,48 @@ class PostgresCacheDatabase:
     def update_order(self, order: object) -> None: ...
     def update_account(self, account: Account) -> None: ...
 
+
+class PyParquetDataCatalogV2:
```
cjdsellers, line 3490: No need to prefix with `Py` here.

crates/serialization/src/parquet.rs
```diff
@@ -75,30 +68,17 @@ pub fn write_batches_to_parquet(
...
-            for batch in batches {
-                if batch.schema() != existing_schema {
-                    let cast_batch = batch.clone().with_schema(existing_schema.clone())?; // Attempt to cast.
-                    cast_batches.push(cast_batch);
-                } else {
-                    cast_batches.push(batch.clone()); // No cast needed, just clone.
-                }
-            }
+            let mut combined = Vec::with_capacity(existing_batches.len() + batches.len());
+            let batches: Vec<RecordBatch> = batches.to_vec();
```
cjdsellers, line 72: This is a nice simplification.

crates/persistence/src/python/catalog.rs
```diff
@@ -43,45 +46,105 @@ impl PyParquetDataCatalogV2 {
...
-            .write_to_parquet(data, None, None, None, write_mode);
+    pub fn write_bars(
+        &self,
+        data: Vec<Bar>,
+        write_mode: Option<ParquetWriteMode>,
+    ) -> PyResult<String> {
+        self.inner
+            .write_to_parquet(data, None, None, None, write_mode)
+            .map(|path| path.to_string_lossy().to_string())
+            .map_err(|e| PyIOError::new_err(format!("Failed to write bars: {}", e)))
```
cjdsellers, line 93: We can standardize on format strings here, rather than interpolation of `e` outside the string.

crates/persistence/src/backend/catalog.rs
```diff
@@ -294,42 +288,47 @@ impl ParquetDataCatalog {
         start: Option<UnixNanos>,
         end: Option<UnixNanos>,
         where_clause: Option<&str>,
-    ) -> datafusion::error::Result<QueryResult>
+    ) -> Result<QueryResult>
```
cjdsellers, line 291: We should leave all non built-in `Result` types qualified.

cjdsellers (approved):
Many thanks for all this work :ok_hand:

tests/unit_tests/persistence/test_catalog_pyo3.py
faysou: I need some py03 types instead of standard ones here, I could add a similar stubs file for pyo3 though.

crates/persistence/src/python/catalog.rs
faysou: Ok I'll do that

crates/persistence/src/backend/catalog.rs
faysou: It was to make it compatible with the calling function that uses anyhow

tests/unit_tests/persistence/test_catalog_pyo3.py
cjdsellers: Ah yes, in that case the way you've done it is the easiest path for now then.

crates/persistence/src/backend/catalog.rs
faysou: I'll try again to see if it works when keeping data fusion result in the signature

nautilus_trader/core/nautilus_pyo3.pyi
faysou: Ok I'll do that, I used an LLM for this and didn't notice

faysou:
I've updated things above. It should be ready after the CI finishes.

crates/persistence/src/backend/catalog.rs
faysou: I think it's better to keep the anyhow result as standard, the type of the error is not lost in the anyhow result, it's a box dyn errorr that can be converted back to whatever is needed.

cjdsellers: If the actual result type changed then thats fine, which is where using `anyhow::Result` is clearer.

PR #2457, opened 2025-03-18
Implement add venue for BacktestEngine

filipmacek:
# Pull Request

- removed the local msgbus dependency and used a global approach for order matching engine and simulated exchange, and fixed tests where each test needed to call `register_message_bus` function
- initialized cace, clock and exec engine in `NautilusKernel`
- changed clock type for order matching engine and fix tests where ES instrument is used (setting the test clock current time to be higher than ES activation ns)
- implemented `add_venue` for `BacktestEngine`
- added TODO comment in `iterate` function of `OrderMatchingEngine` were generic clock were introduced and we lost `set_time` function which needs to be added later

### DIFF
--- a/crates/backtest/src/engine.rs
+++ b/crates/backtest/src/engine.rs
@@ -19,23 +19,28 @@
 
 //! The core `BacktestEngine` for backtesting on historical data.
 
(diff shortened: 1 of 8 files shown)


cjdsellers (approved):
:ok_hand:

PR #2460, opened 2025-03-19
Finish global mesage bus design in Rust

filipmacek:
# Pull Request

- removed MessageBus dependency in existing Rust components, so now all Messagebus related call are from global functions
- refactored `get_message_bus` function to initialize the global message bus if it's not set in lock so that we don't have to initialize msgbus for each test manually
- refactored `MessagingSwitchboard` to also work with global message bus and added static non-parametrized methods like `exec_engine_process` , exec_engine_execute`, `data_engine_process` and `data_engine_process`


cjdsellers (approved):
Good progress :ok_hand:

PR #2499, opened 2025-04-07
Handle websocket errors for dYdX

davidsblom:
# Pull Request

Handle websocket errors for dYdX. When an internal error occurs, the websocket is not subscribed to the channel. Hence, the client tries to subscribe again after a delay.

## Type of change

Live example


cjdsellers:
Many thanks @davidsblom 

Let me know if you find copilot useful, or annoying 😄.

PR #2503, opened 2025-04-10
Improve Rust Port

Pushkarm029:
# Pull Request

## Type of change


cjdsellers:
This PR will be used to reference the main changes for when we attempt this again.

As discussed with Pushkar - there are breaking changes we don't want to introduce just yet and the database adapter interface also needs another round of redesign.

PR #2510, opened 2025-04-13
Remove catalog basename_template argument

faysou:
# Pull Request

Remove catalog basename_template argument

This argument is not used in general and can cause problems if modified wih update_catalog in data requests, so better to remove it.

## Type of change

Existing test passing


davidsblom:
@faysou thanks for this change. Was wondering why the basebame_template has been removed though. I've been using it to write to S3 buckets and it becomes very slow if all files are listed on every call.

faysou:
Hi @davidsblom it was creating too many edge cases. You can use NEWFILE instead with catalog.write. Is it what you mean by listing all files ?

davidsblom:
Yes indeed. NEWFILE is counting all the existing files which is extremely slow with s3 buckets containing mant small files, mainly due to storing a new file every hour.

faysou:
Ok. I'll do a commit to undo the change.

PR #2532, opened 2025-04-18
Revert removing basename_template

davidsblom:
# Pull Request

Revert removing basename_template and force a refresh of the fsspec cache when listing file

## Type of change

- [x] Revert a breaking change and bug fixes

Unit tests

### DIFF
--- a/docs/concepts/data.md
+++ b/docs/concepts/data.md
@@ -537,6 +537,28 @@ The following example shows the above list of Binance `OrderBookDelta` objects b
 catalog.write_data(deltas)
 ```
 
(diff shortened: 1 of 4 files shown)


faysou:
that's the problem I had when using basename template there can be a lot of edge cases, but I understand your point as well.

faysou:
Thanks for the rewrite. I've been working intensively on other features, and catalogs are not a priority to me right now, so it's good you can improve the code to match what you need.

davidsblom:
Thanks for the review. Appreciatie it.

PR #2535, opened 2025-04-19
Implement Block primitive for blockchain domain model

filipmacek:
# Pull Request

- added `defi` feature in Makefile cargo tests
- implemented `Block` struct with correct parsing of all the relevant fields (hex number, UnixNanos timestamps, etc) with helper `rpc` wrappers
- tested on data from live RPC node for blockchains Ethereum, Base, Polygon, Arbitrum

### DIFF
--- a/crates/model/src/defi/block.rs
+++ b/crates/model/src/defi/block.rs
@@ -1,0 +1,372 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2025 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
+//  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+//  You may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+//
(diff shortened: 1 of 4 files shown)


crates/model/src/defi/block.rs
```diff
@@ -0,0 +1,372 @@
...
+
+use nautilus_core::{UnixNanos, datetime::NANOSECONDS_IN_SECOND};
+use serde::{Deserialize, Deserializer};
+use ustr::Ustr;
+
+/// Represent Ethereum-compatible compatible blockchain block with essential metadata
+#[derive(Debug, Clone, Deserialize)]
+pub struct Block {
+    /// The unique identifier hash of the block
+    hash: Ustr,
```
twitu, line 26: Ustr is best used for values that have a limited number of unique values e.g. instrument id. BTC has about 1M blocks, others probably more which means 1M unique global allocations that will remain alive for the duration of the program. It means that even if block information is streamed and aggregated the blocks unique hash will remain in the global memory.

cjdsellers: Thanks @twitu I'll tweak this after merge.

cjdsellers (approved):

PR #2545, opened 2025-04-21
Add BacktestDataIterator to backtest engine

faysou:
# Pull Request

+ Add possibility to add on the fly data to the backtest engine by querying data catalogs defined with DataCatalogConfig when receiving a Subscribe command.
+ Two ways to iterate through data exist:
    + If only the usual backtest list exists, the iteration happens with the same logic as before
    + If more than one backtest list exists after adding an arbitrary number of data to the backtest data, a binary priority queue is used to know which data should be used next, prioritising by ts_init and queue/data priority (we want to pop all elements from the same queue/data list before moving to the next one)
+ There is also the possibility to append or prepend data subscriptions to the existing stream by attributing a positive or negative priority as next integer
+ There is also the possibility to stream data for a given duration before it gets renewed, so if a lot of data exists but won't be needed for a long time, there is no need to load a lot of data before it gets unsusbscribed.
+ Also added possibility to request backtest data during a backtest using a LiveMarketDataClient adapter, live subsriptions being disabled as it wouldn't make sense.

## Type of change

+ Existing tests pass
+ Added tests of iterator and also a test comparing messages with and without using backtest data
When not using backtest data, data is queried on the fly to DataCatalogs
+ Tested data requests from live market client using an included example that works with databento


Newportgogogadit:
> # Pull Request
> 
> + Add possibility to add on-the-fly data to the backtest engine by querying data catalogs defined with DataCatalogConfig when receiving a Subscribe command.
> + Two ways to iterate through data exist:
> + If only the usual backtest list exists, the iteration happens with the same logic as before
> + If more than one backtest list exists after adding an arbitrary number of data to the backtest data, a binary priority queue is used to know which data should be used next, prioritising by ts_init and queue/data priority (we want to pop all elements from the same queue/data list before moving to the next one)
> + There is also the possibility to append or prepend data subscriptions to the existing stream by attributing a positive or negative priority as the next integer
> + There is also the possibility to stream data for a given duration before it gets renewed, so if a lot of data exists but won't be needed for a long time, there is no need to load a lot of data before it gets unsusbscribed.
> + Also added possibility to request backtest data during a backtest using a LiveMarketDataClient adapter, live subsriptions being disabled as it wouldn't make sense.
> 
> ## Type of change
> 
> - New feature (non-breaking change which adds functionality)
> 
> ## How has this change been tested?
> 
> + Existing tests pass
> + Added tests of iterator and also a test comparing messages with and without using backtest data
> When not using backtest data, data is queried on the fly to DataCatalogs
> + Tested data requests from live market clients using an included example that works with databento

PR #2548, opened 2025-04-22
Add mark price subscription for Binance Futures

bartolootrit:
# Pull Request

Binance futures support of `Strategy::subscribe_mark_prices` and `Strategy::on_mark_price`.

## Type of change

It was tested manually on Binance

### DIFF
--- a/nautilus_trader/adapters/binance/futures/data.py
+++ b/nautilus_trader/adapters/binance/futures/data.py
@@ -33,6 +33,7 @@ from nautilus_trader.common.providers import InstrumentProvider
 from nautilus_trader.core.correctness import PyCondition
 from nautilus_trader.model.data import CustomData
 from nautilus_trader.model.data import DataType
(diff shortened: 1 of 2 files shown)


nautilus_trader/adapters/binance/futures/data.py
```diff
@@ -157,3 +158,11 @@ def _handle_mark_price(self, raw: bytes) -> None:
         )
         generic = CustomData(data_type=data_type, data=data)
         self._handle_data(generic)
```
cjdsellers, line 160: Good call leaving this in, so we don't immediately break by changing to the first class `MarkPriceUpdate`.
I'll make sure to deprecate this one in the release notes for a cycle.

cjdsellers (approved):
Hi @bartolootrit 

I was hoping someone would pick this up, many thanks for the contribution!

PR #2593, opened 2025-05-03
Improve validations for StopLimitOrder

### LINKED ISSUE #2529: Standardize validations for orders
# Order validations

There are many order types in Nautilus that need standardized validations.

Currently only a couple of orders have standardized validations:
- [x] [MarketOrder](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/market.rs#L131)
- [x] [LimitOrder](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/limit.rs#L92)

We need to add `new_checked` methods which return a `Result` for the remaining order types:
- [x] [LimitIfTouched](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/limit_if_touched.rs)
- [x] [MarketIfTouched](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/market_if_touched.rs)
- [x] [MarketToLimit](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/market_to_limit.rs)
- [x] [StopLimit](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/stop_limit.rs)
- [x] [StopMarket](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/stop_market.rs)
- [x] [TrailingStopMarket](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/trailing_stop_limit.rs)
- [x] [TrailingStopLimit](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/trailing_stop_market.rs)

These should be accompanied by tests in the same module as the order to verify that the validations exist and are correct.

nicolad:
* **Implemented `new_checked`** for `StopLimitOrder`, bringing it in line with `MarketOrder`, `LimitOrder`, `LimitIfTouchedOrder`, and `MarketToLimitOrder`.
* **Dropped the side-specific trigger/limit inequality**
  *BUY* orders may now use either `trigger ≥ limit` **or** `trigger ≤ limit`; the mirror applies to *SELL*.
  This unblocks “limit-if-touched” and similar strategies that were previously rejected.
* Added complete **domain-level validations**:

  * quantity > 0
  * price > 0
  * trigger\_price > 0
  * optional `display_qty ≤ quantity`
  * `TimeInForce::Gtd` ⇒ non-zero `expire_time` required
* **Legacy `new` now wraps `new_checked`**, preserving the historical *panic-on-error* behaviour for external callers.
* Extended rust-doc explaining the two legal stop-limit styles (break-out vs. mean-reversion).
* Added a concise `Display` implementation for easier debugging/logging.
* **Unit-tests (4)** live in the same module and cover the happy path plus each validation branch.

### Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update

Fixes #2529

***

Note: this PR takes into consideration changes introduced in this commit https://github.com/nautechsystems/nautilus_trader/commit/c44604c060becc664e57fc4b85079447593ddfbe

### DIFF
--- a/crates/model/src/orders/stop_limit.rs
+++ b/crates/model/src/orders/stop_limit.rs
@@ -64,6 +64,10 @@ pub struct StopLimitOrder {
 
 impl StopLimitOrder {
     /// Creates a new [`StopLimitOrder`] instance.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the order is invalid.
     #[allow(clippy::too_many_arguments)]
     pub fn new_checked(
         trader_id: TraderId,
--- a/crates/model/src/orders/stop_limit.rs
+++ b/crates/model/src/orders/stop_limit.rs
@@ -19,7 +19,10 @@ use std::{
 };
 
 use indexmap::IndexMap;
-use nautilus_core::{UUID4, UnixNanos};
+use nautilus_core::{
+    UUID4, UnixNanos,
+    correctness::{FAILED, check_predicate_false},
+};
 use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use ustr::Ustr;
@@ -35,7 +38,10 @@ use crate::{
         AccountId, ClientOrderId, ExecAlgorithmId, InstrumentId, OrderListId, PositionId,
         StrategyId, Symbol, TradeId, TraderId, Venue, VenueOrderId,
     },
-    types::{Currency, Money, Price, Quantity},
+    types::{
+        Currency, Money, Price, Quantity, price::check_positive_price,
+        quantity::check_positive_quantity,
+    },
 };
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -59,7 +65,7 @@ pub struct StopLimitOrder {
 impl StopLimitOrder {
     /// Creates a new [`StopLimitOrder`] instance.
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub fn new_checked(
         trader_id: TraderId,
         strategy_id: StrategyId,
         instrument_id: InstrumentId,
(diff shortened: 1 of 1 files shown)


crates/model/src/orders/stop_limit.rs
```diff
@@ -509,3 +589,126 @@ impl Display for StopLimitOrder {
...
+
+    use crate::{
+        enums::{OrderSide, OrderType, TimeInForce, TriggerType},
+        instruments::{CurrencyPair, stubs::*},
+        orders::{Order, builder::OrderTestBuilder},
+        types::{Price, Quantity},
+    };
+
+    #[rstest]
+    fn buy_breakout_ok(_audusd_sim: CurrencyPair) {
```
nicolad, line 608: @cjdsellers let me know what do you think about this type of test, should I leave, remove or add more like this?

nicolad: the reason I am asking is because I haven't seen smth like this here

https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/orders/market.rs#L539

crates/model/src/orders/stop_limit.rs
```diff
@@ -58,8 +64,12 @@ pub struct StopLimitOrder {
 
 impl StopLimitOrder {
     /// Creates a new [`StopLimitOrder`] instance.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the order is invalid.
```
nicolad, line 70: added this because got this error [here](https://github.com/nautechsystems/nautilus_trader/actions/runs/14813508656/job/41591161735)
image

cjdsellers (approved):

PR #2602, opened 2025-05-05
Improve validations for TrailingStopLimitOrder

nicolad:
Strengthen *server-side* validation for `TrailingStopLimitOrder`.

* Introduced `new_checked`, a fallible constructor that enforces:
  * **Quantity must be positive** (`check_positive_quantity`).
  * **display_qty ≤ quantity**.
  * **GTD orders require a non-zero `expire_time`**.
* Existing `new` now delegates to `new_checked` and panics on failure, preserving the public API while centralising validation.
* Added 4 focused unit tests in `trailing_stop_limit.rs` to cover the happy path and each new invariant.
* Updated docs/comments to reflect the new behaviour.

These checks eliminate an entire class of runtime errors where an invalid order could slip through initialisation and fail deep in the matching or execution layers.
***

Related: https://github.com/nautechsystems/nautilus_trader/issues/2529

### DIFF
--- a/crates/model/src/orders/trailing_stop_limit.rs
+++ b/crates/model/src/orders/trailing_stop_limit.rs
@@ -21,6 +21,8 @@ use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use ustr::Ustr;
 
+use nautilus_core::correctness::{FAILED, check_predicate_true};
+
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
:ok_hand:

Issue #2618, opened 2025-05-09
Bug in handling modify_order() for a trailing stop order


hope2see:
# Bug Report

## Expected Behavior

A request of a not-triggered trailing stop order's quantity should be accepted, and processed. 

### Actual Behavior

The request is rejected with this event : 

> OrderModifyRejected(instrument_id=USD/JPY.SIM, client_order_id=O-19700101-000000-000-None-1, venue_order_id=SIM-1-001, account_id=SIM-001, reason='TRAILING_STOP_LIMIT SELL order new trigger stop px of 12.000 was in the market: bid=13.000, ask=14.000', ts_event=0)

### Steps to Reproduce the Problem

The problem can be tested with the code below. 
The event can be caught when handler is added. 

```

   def test_trailing_stop_limit_order_modify(
        self,
    ) -> None:
        # Arrange: Prepare market
        tick = TestDataStubs.quote_tick(
            instrument=USDJPY_SIM,
            bid_price=13.0,
            ask_price=14.0,
        )
        self.exchange.process_quote_tick(tick)
        self.data_engine.process(tick)
        self.portfolio.update_quote_tick(tick)

        trailing_stop = self.strategy.order_factory.trailing_stop_limit(
            instrument_id=USDJPY_SIM.id,
            order_side=OrderSide.SELL,
            quantity=Quantity.from_int(200_000),
            activation_price=None,
            trailing_offset_type=TrailingOffsetType.PRICE,
            trailing_offset=Decimal("1.0"),
            limit_offset=Decimal("1.0"),
            trigger_type=TriggerType.BID_ASK,
        )
        self.strategy.submit_order(trailing_stop)
        self.exchange.process(0)

        # When activation_price is set to None, it defaults to the bid price, 
        # and trigger_price is set by appling trailing_offset. 
        assert trailing_stop.is_activated
        assert trailing_stop.activation_price == Price.from_str("13.000")
        assert trailing_stop.trigger_price == Price.from_str("12.000")
        assert not trailing_stop.is_triggered
        assert trailing_stop.status == OrderStatus.ACCEPTED

        # Request modify_order
        new_quantity=Quantity.from_int(100_000)
        self.strategy.modify_order(trailing_stop, new_quantity) 
        self.exchange.process(0) 

        # Assert
        assert self.strategy.cache.orders()[0] == trailing_stop
        assert self.strategy.cache.orders()[0].quantity == new_quantity # --> Fails! 

```

### Cause 

[...]

PR #2619, opened 2025-05-09
Fix modifying and updating trailing stop orders

hope2see:
# Pull Request

This PR addresses [Issue #2618 ](https://github.com/nautechsystems/nautilus_trader/issues/2618),
and, also another issue w.r.t. modifying unactivated trailing stop orders.

**NOTE** 
Based on my (imperfect) analysis, there are roughly three types of updates (modifications) for trailing stop orders. (Please correct me if anything is missing or inaccurate.)
1. Trailing in response to market changes:  
Updates to the `trigger_price` and/or `price`.
2. Explicit modification requests from `Strategy.modify_order()`:  
Currently, only `quantity` is allowed to be modified.
3. Contingent order updates: 
Only the `quantity` needs to be updated.

The existing _update_trailing_stop_order() method previously handled only case (1). It has now been renamed to _trail_stop_order().

To support both case (2) and case (3), new methods _update_trailing_stop_market_order() and _update_trailing_stop_limit_order() have been added.

## Type of change

- [v] Bug fix (non-breaking change which fixes an issue)

- passed all the existing pytest 
- A new test 'test_trailing_stop_limit_order_modify' is added in test_exchange_trailing_stops.py


cjdsellers:
Hi @hope2see

Thanks for the PR to address this!

The test makes it easy to pin-point the bug, I was also able to reproduce locally.

Your analysis is correct on the three different ways a trailing stop order might be updated. I think your changes more clearly separate the logic to accommodate all of these scenarios. We could probably tighten up the typing on the Rust side once we sweep back through there when execution becomes a focus (data is the focus right now).

Once merged I'll add another test to confirm the buy-side logic.

PR #2524, opened 2025-04-17
Improve consistency of environment variables for build

twitu:
# Pull Request

Inconsistent environment variables cause frequent re-builds between
* make build-debug
* rust analyzer cargo check
* rust analyzer cargo test
* pre-commit
* cargo in cli

```
   0.359245532s  INFO prepare_target{force=false package_id=pyo3-build-config v0.24.1 target="build-script-build"}: cargo::core::compiler::fingerprint:     dirty: EnvVarChanged { name: "VIRTUAL_ENV", old_value: None, new_value: Some("/home/twitu/Code/nautilus_trader/.venv") }

       Dirty pyo3-build-config v0.24.1: the env variable VIRTUAL_ENV changed
[pyo3-build-config 0.24.1] cargo:rerun-if-env-changed=VIRTUAL_ENV
```

This PR adds a dry run feature that sets the appropriate environment variables in the terminal using `make build-dry-run`. It also prints the appropriate settings for vscode rust analyzer that can be set manually in `.vscode/settings.json`. These are the main settings.

```
{
  "rust-analyzer.check.extraEnv": {
    "CC": "clang",
    "CXX": "clang++",
    "VIRTUAL_ENV": "/home/twitu/Code/nautilus_trader/.venv"
  },
  "rust-analyzer.runnables.extraEnv": {
    "CC": "clang",
    "CXX": "clang++",
    "VIRTUAL_ENV": "/home/twitu/Code/nautilus_trader/.venv"
  },
  "rust-analyzer.cargo.features": "all",
  "rust-analyzer.check.features": "all"
}
```

This may not be consistent across OS and architecture. You can debug your build by running

```
export CARGO_LOG="cargo::core::compiler::fingerprint=info"
export RUST_LOG=trace
cargo build -vv <flags>
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How has this change been tested?

* make build-debug
* dummy change in common crate - triggers cargo check
* make build-debug (200 secs)
* make consistent build environment
* dummy change in common crate - triggers cargo check
* make build-debug (48 secs)

PR #2640, opened 2025-05-16
Improve development debug builds with cranelift backend

twitu:
# Pull Request

## Summary

Use cranelift backend and nightly toolchain codegen_backend feature for much faster debug builds
* fresh make build debugs: 3m 45s -> 2m 42s
* incremental cargo check dummy change in common crate -> 50 s -> 12 s

Downside is that introduces nightly toolchain into build process. There does not appear to be a clean way of separating nightly toolchain by profile. For e.g. nightly for dev and stable for release.

Rust analyzer needs to be updated to use this backend and it's settings need to use nighly toolchain.

```
rustup override set nightly
rustup component add rust-analyzer
rustup override set stable
```

```
    "rust-analyzer.check.extraEnv": {
        "RUSTUP_TOOLCHAIN": "nightly",
        "VIRTUAL_ENV": "/home/twitu/Code/nautilus_trader/.venv",
        "CC": "clang",
        "CXX": "clang++"
    },
    "rust-analyzer.cargo.extraEnv": {
        "RUSTUP_TOOLCHAIN": "nightly",
        "VIRTUAL_ENV": "/home/twitu/Code/nautilus_trader/.venv",
        "CC": "clang",
        "CXX": "clang++"
    },
    "rust-analyzer.runnables.extraEnv": {
        "RUSTUP_TOOLCHAIN": "nightly",
        "VIRTUAL_ENV": "/home/twitu/Code/nautilus_trader/.venv",
        "CC": "clang",
        "CXX": "clang++"
    },
    "rust-analyzer.runnables.extraArgs": [
        "--features",
        "default,python"
    ],
    "rust-analyzer.server.extraEnv": {
        "RUSTUP_TOOLCHAIN": "nightly"
    },
 ```

## Related Issues/PRs

#2524 

## Type of change

- [x] Documentation update

## Testing

Tested out various builds with feature flags.


faysou:
It doesn't seem it's possible to parse the codegen feature with fmt and clippy unless switching the whole toolchain to nightly. 
This could be possible locally and the modified file could be named cargo.toml.dev for example so locally we can switch the cargo.toml file and use the nightly rustup for faster compilations.

cjdsellers:
> It doesn't seem it's possible to parse the codegen feature with fmt and clippy unless switching the whole toolchain to nightly.

Correct, nightly is required - which we don't want to switch to for the specified toolchain in `rust-toolchain.toml`.

That's also the conclusion we came to - that this could provide some wins for local development only using nightly.
@twitu might have some ideas of how to tweak this so we can merge and easily switch to the cranelift backend just for local dev?

twitu:
Turns out having multiple toolchains at a project level config is not possible. There are parts of a solution hidden away in flags and environment variables and implicity arguments but none of them connect into a cohesive solution. Some options

* Define separate `Cargo.toml.dev` and use `--manifest-file` with each cargo command
* Pass cargo features as command line using `-Zcodegen-backend` but profiles cannot be configured to conditionally change backend for dev and test profiles
* No clean way to configure pre-commit hooks to use cargo nightly and extra args

I don't see any integrated solution with current tooling. Instead we can fallback to a low tech solution where the dev can apply a patch to their Cargo.toml, rust-toolchain.toml, build and dev and remove the changes before pushing.

twitu:
Added a pre-commit hook to check for nightly features and updated environment setup docs with instructions. Give it a try!

twitu:
The aws_lc library does not work with cranelift because it exports symbols in a way that only LLVM supports. This means that cranelift will not work when running the full system. However, it is still useful for faster local dev loops on crates and tests that are not related to the aws_lc which comes from rustls.

https://github.com/rust-lang/rustc_codegen_cranelift/issues/1520

PR #2643, opened 2025-05-16
Use faster message bus pattern matching logic from Rust

twitu:
# Pull Request

## Summary

Use fast matching logic in Cython through ffi

## Related Issues/PRs

#2634 

## Type of change

- [x] Perf improvement

### DIFF
--- a/crates/core/src/ffi/string.rs
+++ b/crates/core/src/ffi/string.rs
@@ -64,10 +64,10 @@ pub unsafe fn cstr_to_ustr(ptr: *const c_char) -> Ustr {
 ///
 /// Panics if `ptr` is null.
 #[must_use]
-pub unsafe fn cstr_to_bytes(ptr: *const c_char) -> Vec<u8> {
+pub unsafe fn cstr_to_bytes(ptr: *const c_char) -> &'static [u8] {
     assert!(!ptr.is_null(), "`ptr` was NULL");
     let cstr = unsafe { CStr::from_ptr(ptr) };
-    cstr.to_bytes().to_vec()
+    cstr.to_bytes()
 }
 
 /// Convert a C string pointer into an owned `Option<Ustr>`.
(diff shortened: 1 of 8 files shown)


crates/core/src/ffi/string.rs
```diff
@@ -64,10 +64,10 @@ pub unsafe fn cstr_to_ustr(ptr: *const c_char) -> Ustr {
 ///
 /// Panics if `ptr` is null.
 #[must_use]
-pub unsafe fn cstr_to_bytes(ptr: *const c_char) -> Vec<u8> {
+pub unsafe fn cstr_to_bytes(ptr: *const c_char) -> &'static [u8] {
```
cjdsellers, line 67: :ok_hand:

nautilus_trader/common/component.pyx
```diff
@@ -2694,36 +2695,7 @@ cdef inline bint contains_wildcard(str topic_or_pattern):
...
-    cdef int i
-    for i in range(1, n + 1):
-        for j in range(1, m + 1):
-            if pattern[j - 1] == '*':
-                t[i, j] = t[i - 1, j] or t[i, j - 1]
-            elif pattern[j - 1] == '?' or topic[i - 1] == pattern[j - 1]:
-                t[i, j] = t[i - 1, j - 1]
-
-    return t[n, m]
+    return is_matching_ffi(pystr_to_cstr(topic), pystr_to_cstr(pattern))
```
cjdsellers, line 2698: :broom: :chart_with_upwards_trend:

twitu:
30x faster cython matching by ffi to new rust logic :rocket:

stastnypremysl:
Hi, @twitu.

I discourage to use non-deterministically random chooses in the test. They might result in a unreproducible fails. Better more runs with deterministic random chooses, then fewer with indeterministic in my opinion.

I would recommend to set a random seed there.

PR #2658, opened 2025-05-21
Improve ergonomics of messaging topics, patterns, and endpoints

twitu:
# Pull Request

Use phantom types for Topic pattern for a more ergonomic and concise api. It also draws from design principle that both types have the same physical representation but have different validation properties. Also merged endpoints and topics since they are physically and functionally the same.

## Related Issues/PRs

#2646 

## Type of change

- [x] Refactor


cjdsellers:
Great stuff, thanks @twitu

PR #2662, opened 2025-05-22
Improve Weighted Moving Average (WMA) indicator parity with Cython 

nicolad:
Restores **feature-parity and behavioural equivalence** between the Rust `WeightedMovingAverage` implementation and the canonical Python/Cython reference.

---

## 1 · Why this matters — Context & Motivation

| Theme | Why it matters |
| ----- | -------------- |
| **Correctness ⇄ Consistency** | Mixed Python (quant research) + Rust/WASM (execution) stacks must emit identical WMA values; silent divergence leaks *P ∿ L*. |
| **Robust edge-case handling** | Zero-weight, sub-ε weight-sum or NaN inputs previously produced UB or silent drift. |
| **Simpler mental model** | Removes the redundant `has_inputs` flag and double-initialisation logic. |
| **Test-driven culture** | Full-parity *rstest* suite protects against future regressions and enables confident refactors. |

---

## 2 · What changed

| Concern | Rust *before* | Rust *after* (this PR) |
| ------- | ------------- | ---------------------- |
| Validation rules | Sparse, panicked on some edge cases only | ✨ Full validation in `new_checked`: `period > 0`, `weights.len() == period`, `∑wᵢ > ε` |
| Redundant state | `has_inputs` boolean tracked manually | **Removed** — `has_inputs()` now derives from `!inputs.is_empty()` |
| Weighted sum loop | Manual index + `reverse_weights` vec | Direct `zip(self.inputs.iter().rev(), self.weights.iter().rev())` (zero alloc) |
| Initialisation flag | Two competing assignments | Single canonical rule: `initialized = count() ≥ period` |
| Error messages | Generic | Descriptive, actionable (`"period must equal weights.len()"`) |
| Unit tests | 8 cases | 30 cases / 100 % line & branch coverage |
| Docs | Minimal | Elaborate `///` comments on invariants & panics |
| Performance note | – | No observable regression (same O(1) update); future SIMD possible ✈️ |

---

## 3 · Implementation notes

* **Validation constant** `EPS = f64::EPSILON` ensures weight-sum check does not break on FP rounding.  
* `weighted_average()` is *safe* because `self.weights.len() == self.period` is proven by construction.  
* Negative weights are tolerated *iff* the **total** is positive (common in certain signal schemes); covered by tests.  
* NaN inputs **poison** the output until `reset()` — matches Python behaviour.  
* `WeightedMovingAverage` is still `!Send + !Sync`; wrap in `Arc<Mutex<…>>` when sharing between threads.  

---

## 4 · Tests added / updated

| Test name | Purpose |
| --------- | ------- |
| `new_panics_on_zero_period` / `new_checked_err_on_zero_period` | Guards `period > 0` |
| `new_panics_on_zero_weight_sum` / `weight_sum_below_epsilon` variants | Enforces `∑wᵢ > ε` |
| `initialized_flag_transitions` | Canonical initialisation after `period` samples |
| `count_matches_inputs_and_has_inputs` | Verifies redundant flag removal |
| `weighted_average_with_non_uniform_weights` | Numerical accuracy with 1:2:3 weights |
| `negative_weights_positive_sum` | Mixed-sign weights sanity |
| `nan_input_propagates` | NaN poisoning rule |
| `single_period_returns_latest_input` | `period == 1` degenerates to *last price* |
| `value_with_sparse_weights` | Sparse weight-vector edge case |
| *+ 20 more robustness cases* → *30 total* / *100 % coverage*. |

### DIFF
--- a/crates/indicators/src/average/wma.rs
+++ b/crates/indicators/src/average/wma.rs
@@ -454,4 +454,23 @@ mod tests {
             assert!(res.is_err());
         }
     }
+
+    #[rstest]
+    fn single_period_returns_latest_input() {
+        let mut wma = WeightedMovingAverage::new(1, vec![1.0], None);
+        for i in 0..5 {
+            let v = i as f64;
+            wma.update_raw(v);
+            assert_eq!(wma.value(), v);
+        }
+    }
+
+    #[rstest]
+    fn value_with_sparse_weights() {
+        let mut wma = WeightedMovingAverage::new(3, vec![0.0, 1.0, 0.0], None);
+        wma.update_raw(10.0);
+        wma.update_raw(20.0);
+        wma.update_raw(30.0);
+        assert_eq!(wma.value(), 20.0);
+    }
 }
--- a/crates/indicators/src/average/wma.rs
+++ b/crates/indicators/src/average/wma.rs
@@ -43,7 +43,6 @@ pub struct WeightedMovingAverage {
     pub initialized: bool,
     /// Inputs
     pub inputs: Vec<f64>,
-    has_inputs: bool,
 }
 
 impl Display for WeightedMovingAverage {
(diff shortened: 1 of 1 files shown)


nicolad:
needs some attention from you @cjdsellers 👇 


| # | Weight rule / behaviour                                           | Rust WMA (`new_checked`)                | Python WMA                       | Parity |
|---|-------------------------------------------------------------------|-----------------------------------------|----------------------------------|--------|
| 1 | `weights.len() == period`                                         | ✔ Enforced (`check_predicate_true`)     | ✔ Enforced                      | ✅     |
| 2 | Σ weights > ε (`f64::EPSILON` / `np.finfo(float).eps`)            | ✔ Enforced                              | ✔ Enforced                      | ✅     |
| 3 | Negative weights allowed if Σ > ε                                 | ✔ Allowed                               | ✔ Allowed                       | ✅     |
| 4 | Weights optional (`None`) → SMA fallback                          | ✖ Not supported (must pass `Vec<f64>`)  | ✔ Supported (`weights=None`)    | ⚠️     |
| 5 | Warm-up uses **trailing** subset of weights (`weights[-n:]`)      | ✔ Yes (`weights[self.period-n..]`)      | ✔ Yes                           | ✅     |
| 6 | Guard that warm-up Σ weights > ε                                  | ✖ Open                                  | ✖ Open                          | ⚠️     |
| 7 | Invalid weights ⇒ runtime error                                   | `panic!` (in `new`), `Err` (checked)    | `ValueError`                    | ✅     |
| 8 | Degenerate case `period=1`, `weights=[1.0]` → passthrough         | ✔ Works                                 | ✔ Works                         | ✅     |
| 9 | NaN inside weights detected / rejected                            | ✖ No explicit guard (Σ check catches)   | ✖ Same                          | ✅     |

crates/indicators/src/average/wma.rs
```diff
@@ -142,21 +158,16 @@ impl MovingAverage for WeightedMovingAverage {
...
     fn update_raw(&mut self, value: f64) {
-        if !self.has_inputs {
-            self.has_inputs = true;
-            self.inputs.push(value);
-            self.value = value;
-            return;
-        }
         if self.inputs.len() == self.period {
-            self.inputs.remove(0);
+            self.inputs.pop_front();
```
cjdsellers, line 164: If we're using a `VecDeque` with a fixed capacity then this might not be necessary?

cjdsellers (commented):
Great stuff, and thanks for all the tests!

Removing `has_inputs` is a nice refinement.

On the suggestions, I don't think we need to support optional weights at this stage?

crates/indicators/src/average/wma.rs
nicolad: `VecDeque::with_capacity(n)` only **pre-allocates** space; it doesn’t enforce a hard cap.
If you keep pushing after `len == n`, the deque re-allocates and grows, so you must still `pop_front()` (or otherwise drop the oldest element) to keep the window at exactly `period` items.

nicolad:
audio explanation


https://github.com/user-attachments/assets/7ab95de4-1da4-4825-8cc3-f26c4c4d1a59

cjdsellers (commented):
thanks @nicolad :ok_hand:

PR #2667, opened 2025-05-25
Change VecDeque for fixed-capacity ArrayDeque in LinearRegression

nicolad:
### Why touch the buffer in `LinearRegression` at all?  
`LinearRegression` (LR) keeps a sliding window whose length (`period`) is known a-priori.  
With the current **`std::collections::VecDeque`** the queue *still* grows past that length: every `push_back` that exceeds capacity triggers a heap re-allocation and mem-copy. Those reallocations are

* wasted work — we immediately `pop_front` afterwards, and  
* a latency spike right on the trading hot-path.

**`arraydeque::ArrayDeque`**, in contrast, is “a circular buffer with fixed capacity … [that] can be stored directly on the stack”¹.  
Its capacity is a *const generic* baked into the type, so pushes never allocate and the window length is never exceeded.

| Aspect | `VecDeque` | `ArrayDeque<N>` |
|--------|-----------|-----------------|
| Capacity | Dynamic; grows geometrically | Compile-time fixed (`const N`) |
| Allocation | Heap (may reallocate) | **None** (stack/static) |
| Push/Pop cost | Amortised *O(1)* but occasional mem-copy | Strict *O(1)* |
| Enforced upper bound | Caller-dependent, not guaranteed | Compile-time assert; panic if `period > N` |
| Max window length | `usize::MAX` (practically unbounded) | Chosen at build time (`MAX_PERIOD = 16 384`) |
| Failure mode when full | Reallocate + copy | Programmer picks **Wrapping** (overwrite) or **Saturating** (reject) |


cjdsellers (approved):
:ok_hand: thanks @nicolad this is a good optimization.

nicolad:
image

PR #2693, opened 2025-06-06
Implement From<Pool> → CurrencyPair & InstrumentAny

nicolad:
Add two `impl From<Pool>` blocks:

* `Pool → CurrencyPair`
* `Pool → InstrumentAny`

### Why

* Downstream services (price-feed, risk, matching) already speak `CurrencyPair`/`InstrumentAny`.
* Reduces boilerplate in adapters: `CurrencyPair::from(pool)` vs manually building IDs, symbols, increments.

### DIFF
--- a/crates/model/src/defi/dex.rs
+++ b/crates/model/src/defi/dex.rs
@@ -18,7 +18,7 @@ use std::{borrow::Cow, fmt::Display, sync::Arc};
 use crate::defi::{amm::Pool, chain::Chain};
 use crate::identifiers::{InstrumentId, Symbol, Venue};
 use crate::instruments::{Instrument, any::InstrumentAny, currency_pair::CurrencyPair};
-use crate::types::{currency::Currency, price::Price, quantity::Quantity};
+use crate::types::{currency::Currency, fixed::FIXED_PRECISION, price::Price, quantity::Quantity};
 
 /// Represents different types of Automated Market Makers (AMMs) in DeFi protocols.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


crates/model/src/defi/dex.rs
```diff
@@ -96,3 +99,46 @@ impl Display for Dex {
         write!(f, "Dex(chain={}, name={})", self.chain, self.name)
     }
 }
+
+impl From<Pool> for CurrencyPair {
+    fn from(p: Pool) -> Self {
+        let symbol = Symbol::from(format!("{}/{}", p.token0.symbol, p.token1.symbol));
+        let id = InstrumentId::new(symbol, Venue::from("DEFI"));
```
cjdsellers: I think the venue should probably be the [identifier for the DEX](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/defi/dex.rs#L94).

crates/model/src/defi/dex.rs
```diff
@@ -96,3 +99,46 @@ impl Display for Dex {
...
+
+        let size_precision = p.token0.decimals;
+        let price_precision = p.token0.decimals.saturating_add(p.token1.decimals);
```
cjdsellers: This should probably be the `min(token1.decimals, FIXED_PRECISION)` (because we're currently limited to 16 rather than the full 18 which much of DeFi will need).

cjdsellers (commented):
Thanks for this useful impl @nicolad !

crates/model/src/defi/dex.rs
v9ai: done, thx

crates/model/src/defi/dex.rs
v9ai: done, thx

PR #2700, opened 2025-06-09
 Confirm parity between Python and Rust indicators

nicolad:
| №  | Topic                      | Reality in **Rust** (`AroonOscillator`)                          | Reality in **Py/Cython** (`aroon.pyx`)               | Status | Parity Note                                                                                                              |
|----|----------------------------|------------------------------------------------------------------|------------------------------------------------------|--------|---------------------------------------------------------------------------------------------------------------------------|
| 1  | **Constructor API**        | `try_new(period) -> Result<Self, String>` + `#[new]` wrapper     | `__init__(period: int)`                              | ✅     | Identical public signature: single positional `period` argument.                                                          |
| 2  | **Parameter validation**   | Returns `Err(PyValueError)` on 0 / > MAX_PERIOD (no panic)       | Raises `ValueError`                                  | ✅     | FFI boundary is safe on both sides.                                                                                       |
| 3  | **Window length**          | Maintains **`period + 1`** items                                 | Uses `deque(maxlen = period + 1)`                    | ✅     | Canonical rule fixed and documented.                                                                                      |
| 4  | **Warm-up / readiness**    | `initialized == total_count ≥ period + 1`                        | `ready == len(deque) ≥ period + 1`                   | ✅     | Identical logic.                                                                                                          |
| 5  | **Formula core**           | `100 × (period – idx)/period`                                    | Same maths (scales after division)                   | ✅     | Bit-wise identical after rounding.                                                                                        |
| 6  | **What a bar feeds**       | `high` & `low`                                                   | `high` & `low`                                       | ✅     | No divergence.                                                                                                            |
| 7  | **Counter exposure**       | Private field + `#[getter] fn count(&self) -> usize`             | `@property count` (read-only)                        | ✅     | Single, documented API on both sides.                                                                                     |
| 8  | **`has_inputs` flag**      | Manual bool; user should rely on `initialized`                   | Managed by base class; documented to prefer `initialized` | 🟡   | Behaviour is aligned; implementation detail still differs but is documented—no functional risk.                           |
| 9  | **Reset**                  | Clears deques, zeros fields                                      | Same                                                 | ✅     | Full parity.                                                                                                              |
| 10 | **Unit tests**             | rstest vectors incl. numeric parity                              | **Mirrored pytest suite** using same test vectors    | ✅     | CI will fail if either side drifts.                                                                                       |
| 11 | **Error handling**         | All public APIs return `Result`; only `debug_assert` inside      | Raises Python exceptions                             | ✅     | Zero panics, uniform error semantic.                                                                                      |
| 12 | **`Display` / `__repr__`** | `"AroonOscillator(period)"`                                      | `"AroonOscillator(period)"`                          | ✅     | Identical stringification.                                                                                                |
***


cjdsellers (approved):
:ok_hand:

PR #2707, opened 2025-06-11
Minor CI/CD and build system improvements

stastnypremysl:
# Pull Request
### CI/CD Workflow Enhancements:
* Added a new GitHub Actions workflow (`.github/workflows/copilot-setup-steps.yml`) to set up a build environment with Redis and PostgreSQL services, along with specific environment variables for debugging and firewall configurations, so Github Copilot can be better run over the repository.
* Updated branch triggers in `.github/workflows/build.yml` to include all branches for pull requests instead of just the `develop` branch. (This behavior is advantageous for longer developed branchess)

### Build System Improvements:
* Enhanced the `Makefile` clean targets by splitting `clean` into `clean-build-artifacts`, `clean-caches`, and `clean-builds`. This provides more granularity when removing build and cache files. Additionally, added commands to clean Rust and Python artifacts separately. --- Much more space is released, when `make clean` is called.

### Environment Configuration:
* Updated `.github/actions/common-setup/action.yml` to set `SCCACHE_CACHE_SIZE=4G` for better cache management based on package size and runner constraints.
* Improved handling of `CFLAGS` and `LDFLAGS` in `build.py` to append new flags instead of overwriting existing ones, ensuring compatibility with other configurations.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [x] Maintenance / chore


.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [master, nightly, develop]
   pull_request:
-    branches: [develop]
+    branches: ['*']
```
cjdsellers, line 7: I'd prefer to keep the build workflow scoped to just the `develop` branch, rather than every branch collaborators might push to.

This is to reduce CI churn: as we should only be committing against PR's opened on the `develop` branch after code has been tested locally and in preparation for merge (rather than incremental commits during development).

cjdsellers (commented):
Hi @stastnypremysl 

Thanks, I think these are good refinements overall

.github/workflows/build.yml
stastnypremysl: This is only for opened pull requests. There is no risk, that it will run on any collaborator branch.

This helps me to setup pipelines in a fork during the development without any actual harm to the upstream repo. (If no cross branch merge request happens with the exception of develop). 

If it was possible, I would be glad. But let me know, if not, so the branch may get merged sooner then later (I would like to avoid merge conflicts).

stastnypremysl:
@cjdsellers, the AI has problems with connection to Postgres. If you had any idea, where might be a problem, please tell me. I would like to have all test suites passing there, so AI may use tests as a validation of a code.

.github/workflows/build.yml
cjdsellers: Understood, so that addresses my concern about CI churn then.

Could you describe more about the purpose of the pipelines you're setting up in your fork?

cjdsellers:
> @cjdsellers, the AI has problems with connection to Postgres. If you had any idea, where might be a problem, please tell me. I would like to have all test suites passing there, so AI may use tests as a validation of a code.

I'd have to know what the actual error output is to be able to help. But as a starting point you could try `make start-services` locally, which will trigger docker compose to bring up the necessary infra.

.github/workflows/build.yml
stastnypremysl: I want to try use GitHub Copilot for development and I need the pipelines for the verification, that everything is ok, before putting it to merge request, where will it spam.

I am also used to using free computation power from GitHub for opensource testing, so I usually prefer commiting a lot and  testing on remote, when I am finding unknown test to fail, rather then doing it locally.

stastnypremysl:
> > @cjdsellers, the AI has problems with connection to Postgres. If you had any idea, where might be a problem, please tell me. I would like to have all test suites passing there, so AI may use tests as a validation of a code.
> 
> I'd have to know what the actual error output is to be able to help. But as a starting point you could try `make start-services` locally, which will trigger docker compose to bring up the necessary infra.

Thanks. I didn't know about `make start-services`. It might help me to eliminate the source of the timeout errors.

When I get to the concrete error again, I will write it to you.

.github/workflows/build.yml
cjdsellers: > I need the pipelines for the verification, that everything is ok

I think this is a good use of a pipeline like this, as you'd have high assurance of CI passing before opening a PR.

> I am also used to using free computation power from GitHub for opensource testing, so I usually prefer committing a lot and testing on remote, when I am finding unknown test to fail, rather then doing it locally

I'd recommend you run tests locally before committing to GitHub, as the "free" runners are low spec and take ~10x longer to compile and run the entire test suite depending on your own machine specs. So this would be a much slower feedback loop for local development.

PR #2714, opened 2025-06-16
Add submit order list support for sandbox

petioptrv:
# Pull Request

## Summary

This PR adds support for submitting order lists when using the sandbox execution client.

## Related Issues/PRs

N/A

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Release notes

## Testing

I attempted to add a unit test in the `tests/integration_tests/adapters/sandbox/test_execution.py` file, however, the current unit tests around sandbox execution are WIP, so I am not sure if my test case makes sense.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,7 @@ Released on TBD (UTC).
 - Added support for Pool liquidity updates for blockchain adapter (#2692), thanks @filipmacek
 - Added fill report reconciliation warning when discrepancy with existing fill (#2706), thanks @faysou
 - Added `bid_levels` and `ask_levels` for `OrderBook.pprint`
+- Added support for order-list submission in the sandbox execution client.
 
 ### Breaking Changes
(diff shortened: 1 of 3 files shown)


petioptrv:
@cjdsellers, please take a look at my test case. I know that the first throwaway event is the initialization of the order, but what about the subsequent ones? Also, would they be group by event (as I have them now) or by order?

cjdsellers:
Hi @petioptrv

Thanks for the contribution and good catch!

The anatomy of the test looks fine to me. If the test actually passes then its fine to unskip, otherwise if there is some obscure error you can't figure out then we can just merge and I'll have a look locally.

petioptrv:
@cjdsellers, I thought the issue I was seeing earlier was obscure, but on second glance, it now works, and I unskipped the previous test as well, since it passes too.

Issue #2704, opened 2025-06-10
IBRK: Partially filled bracket order and SL triggered


bartlaw:
# Bug Report

I have the following setup: bracket limit order with SL and TK market orders. When stop loss is triggered when the limit order is only partially filled, I see the following error:

```
[ERROR] TESTER-001.InteractiveBrokersClient-003: Error on '_run_msg_handler_processor': ValueError("'quantity' not a positive real, was 0.0")
```

After that position on NT side is corrupted.

I see the same behavior on the 1.218 version and 1.219 today's nightly build.

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates.
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

### Steps to Reproduce the Problem

1. Fire bracket order from NT with qty = 100
2. Modify/move the limit order in TWS until it is partially filled.
3. Move limit order away from current price.
4. Move the stop order until it is filled.

### Specifications

- OS platform: Docker
- Python version: 3.13.4
- `nautilus_trader` version: 1.219.0 docker nightly build

faysou:
Hello, can you please provide some MRE so I can have a look ? It's relatively easy for me to debug things, but I don't want to spend time reproducing the bug. Thanks

bartlaw:
Hi @faysou 
Attached code to reproduce. 90% of the time it will produce the bug. The trick is that the entry order must get partially filled (it is random). So if it does not work on the first run, just stop and try again.

[mre.zip](https://github.com/user-attachments/files/20693723/mre.zip)

faysou:
I've just made a PR to fix this, please test if it solves the issue.

bartlaw:
Do you cancel bracket orders with qty=0? I believe they are valid orders needed in case the partially filled limit entry order is filled.

faysou:
to be honest I've used augment code for this solution. if you need more, you can have a look at it and use the same tool.

bartlaw:
If it cancels bracket orders and leaves limit order unprotected without stop loss, it is not good. It is better if it just crashes.

faysou:
[...]

PR #2717, opened 2025-06-16
fixed IBKR partially filled bracket order and SL triggered bug

bartlaw:
# Pull Request

## Summary

Fixes problem from bug #2704

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/2704

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Release notes

## Testing

Manual testing with code from issue.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -67,6 +67,7 @@ Released on TBD (UTC).
 - Fixed last value updating for RSI indicator (#2703), thanks @bartlaw
 - Fixed gateway/TWS reconnect process for IBKR (#2710), thanks @bartlaw
 - Fixed Interactive Brokers options chain issue (#2711), thanks @FGU1
+- Fixed Partially filled bracket order and SL triggered for IBKR (#2704)
 
 ### Documentation Updates
 None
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @bartlaw 

Thanks for the fix!

Since we don't have much test coverage here, it would be ideal to either:
- Add a test
- Describe more clearly what the bug was and how the changes fix it

:pray:

bartlaw:
Hi @cjdsellers 

Thanks for the merge.
In real life, this bug will trigger when a bracket order is sent to IB and price bounces between the entry order and take profit order. 
First, it will fill the entry limit order partially and then the take-profit limit order. In this case, IB modifies total quantities of orders to prevent overfill. It seems like it is sending some strange reports to NT in this case.
I've tested it, and it seems to work with this fix, although I don't really understand why they send 0 total qty. Maybe as a way of saying that order is inactive.
Without this fix, positions are totally wrong on the NT side in this scenario.

In a related issue, I've attached code to reproduce it. It uses SL instead of TP, but the logic is the same.

Will take a look at test cases.

PR #2723, opened 2025-06-19
Improve implementation, validations and testing for Rust instruments

nicolad:
| **Cython / Python Method**              | **Rust Method in `Instrument`** | **Notes on Parity**                              |
|----------------------------------------|---------------------------------|--------------------------------------------------|
| `symbol` (property)                    | `symbol(&self) -> &str`         | Same return semantics (borrowed `&str`).         |
| `venue` (property)                     | `venue(&self) -> &str`          | Same return semantics.                           |
| `get_cost_currency()`                  | `cost_currency(&self) -> &Currency` | Renamed for Rust style; logic identical.    |
| `make_price(value)`                    | `make_price(&self, value: f64)` | Rounds using instrument price-precision.         |
| `make_qty(value, round_down=False)`    | `make_qty(&self, value: f64, round_down: bool)` | Round-down flag mirrors Python. |
| `notional_value(qty, price, use_quote)`| `notional_value(&self, qty, price, use_quote_for_inverse: bool)` | Handles linear & inverse paths. |
| `calculate_base_quantity(qty, last_px)`| `calculate_base_quantity(&self, qty, last_px)` | One-to-one logic.                                |

### DIFF
--- a/crates/model/src/instruments/mod.rs
+++ b/crates/model/src/instruments/mod.rs
@@ -35,9 +35,8 @@ pub mod stubs;
 use anyhow::ensure;
 use enum_dispatch::enum_dispatch;
 use nautilus_core::UnixNanos;
-use rust_decimal::Decimal;
+use rust_decimal::{Decimal, RoundingStrategy, prelude::*};
 use rust_decimal_macros::dec;
-use serde_json::{Value, json};
 use ustr::Ustr;
 
 // Re-exports
@@ -60,8 +59,7 @@ macro_rules! check_positive {
     };
 }
 
-#[allow(clippy::missing_errors_doc)]
-#[allow(clippy::too_many_arguments)]
+#[allow(clippy::missing_errors_doc, clippy::too_many_arguments)]
 pub fn validate_instrument_common(
     price_precision: i32,
     size_precision: i32,
@@ -89,33 +87,42 @@ pub fn validate_instrument_common(
     ensure!(margin_init >= dec!(0), "margin_init negative");
     ensure!(margin_maint >= dec!(0), "margin_maint negative");
 
-    if let Some(pi) = price_increment {
-        check_positive!(pi.as_f64(), "price_increment not positive");
+    if let Some(increment) = price_increment {
+        check_positive!(increment.as_f64(), "price_increment not positive");
         ensure!(
-            pi.precision as i32 == price_precision,
+            increment.precision as i32 == price_precision,
             "price_precision != price_increment.precision"
         );
     }
-    if let Some(l) = lot_size {
-        check_positive!(l.as_f64(), "lot_size not positive");
+    if let Some(lot) = lot_size {
+        check_positive!(lot.as_f64(), "lot_size not positive");
     }
-    if let Some(q) = max_quantity {
-        check_positive!(q.as_f64(), "max_quantity not positive");
+    if let Some(quantity) = max_quantity {
+        check_positive!(quantity.as_f64(), "max_quantity not positive");
     }
-    if let Some(q) = min_quantity {
-        ensure!(q.as_f64() >= 0.0, "min_quantity negative");
+    if let Some(quantity) = min_quantity {
+        ensure!(quantity.as_f64() >= 0.0, "min_quantity negative");
     }
-    if let Some(n) = max_notional {
-        check_positive!(n.as_f64(), "max_notional not positive");
+    if let Some(notional) = max_notional {
+        check_positive!(notional.as_f64(), "max_notional not positive");
     }
-    if let Some(n) = min_notional {
-        ensure!(n.as_f64() >= 0.0, "min_notional negative");
+    if let Some(notional) = min_notional {
+        ensure!(notional.as_f64() >= 0.0, "min_notional negative");
     }
-    if let Some(p) = max_price {
-        check_positive!(p.as_f64(), "max_price not positive");
+    if let Some(price) = max_price {
+        ensure!(
+            price.precision as i32 == price_precision,
+            "price_precision != max_price.precision"
+        );
     }
-    if let Some(p) = min_price {
-        ensure!(p.as_f64() >= 0.0, "min_price negative");
+    if let Some(price) = min_price {
+        ensure!(
+            price.precision as i32 == price_precision,
+            "price_precision != min_price.precision"
+        );
+    }
+    if let (Some(min), Some(max)) = (min_price, max_price) {
+        ensure!(min.as_f64() <= max.as_f64(), "min_price exceeds max_price");
     }
     Ok(())
 }
@@ -200,7 +207,7 @@ pub trait Instrument: 'static + Send {
     fn is_inverse(&self) -> bool;
     fn is_quanto(&self) -> bool {
         self.base_currency()
-            .map(|c| c != self.settlement_currency())
+            .map(|currency| currency != self.settlement_currency())
             .unwrap_or(false)
     }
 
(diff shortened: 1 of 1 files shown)


v9ai:
| Area                                | Status | What’s still needed (➜ “left”)                                                                                                   |
|-------------------------------------|:------:|-------------------------------------------------------------------------------------------------------------------------------|
| **Trait vs struct architecture**    | ✅     | Nothing – existing `Instrument` trait + concrete structs is sufficient.                                                       |
| **Core helpers** (`make_price`, `make_qty`, tick navigation) | ✅ | – |
| Auto-derive **`price_increment`** from `price_precision` when caller passes `None` | ❌ | Add default in each constructor (`10^-price_precision`).                                                                      |
| `_min_price_increment_precision` logic in `make_price`      | ⚠️ | Re-implement rounding to use *min*(price precision, tick-increment precision).                                                |
| **Common validations** parity with Cython (`validate_instrument_common`) | ⚠️ | Add guards for `max_price`, `max_notional` > 0 and cross-field “both-or-neither” conditions.                                   |
| Negative-path **unit tests** for *every* validation guard   | ❌ | `rstest` each failure case; assert panic messages.                                                                            |
| **`from_dict`** (serde) + round-trip `to_dict`/`from_dict` tests | ❌ | Implement via `serde_json::from_value`; test JSON⇄struct equality.                                                             |
| **PyO3 conversions** (struct ⇄ Python object)               | ❌ | Implement `#[pymethods]` `__getstate__/__setstate__`; write tests for Python↔Rust round-trip and dict conversions.            |
| Functional tests for **inverse & quanto** notional / cost-ccy | ❌ | Validate `calculate_notional_value` and `cost_currency` across instrument types.                                              |
| **Tick-scheme range tests** (min/max price limits)          | ❌ | Verify `next_bid/ask_prices` stop at bounds and respect custom schemes.                                                       |
| Decision on separate `Instrument` struct                    | ✅ | No dedicated struct needed – trait approach confirmed.                                                                        |

crates/model/src/instruments/mod.rs
```diff
@@ -151,31 +278,88 @@ pub trait Instrument: 'static + Send {
...
+            .collect()
+    }
+
+    fn next_ask_prices(&self, value: f64, n: usize) -> Vec<Price> {
+        (0..n)
+            .filter_map(|i| self.next_ask_price(value, i as i32))
+            .collect()
+    }
+
+    fn to_dict(&self) -> Value {
```
cjdsellers: We don't need a `to_dict()` in the Rust layer (we can implement these in the `python` module, many are done iirc)

crates/model/src/instruments/mod.rs
```diff
@@ -103,40 +217,53 @@ pub trait Instrument: 'static + Send {
...
-            let increment = 10f64.powi(-i32::from(self.size_precision()));
-            let rounded_value = (value / increment).floor() * increment;
-            Quantity::new(rounded_value, self.size_precision())
+        let inc = 10f64.powi(-(self.size_precision() as i32));
+        let rounded = if round_down.unwrap_or(false) {
+            (value / inc).floor() * inc
         } else {
-            // Use standard rounding behavior (banker's rounding)
-            Quantity::new(value, self.size_precision())
+            let factor = 10f64.powi(self.size_precision() as i32);
```
cjdsellers: Is all this additional code to ensure half-even rounding?

crates/model/src/instruments/mod.rs
```diff
@@ -151,31 +278,88 @@ pub trait Instrument: 'static + Send {
...
-        Quantity::new(value, self.size_precision())
+        Quantity::new(quantity.as_f64() / last_px.as_f64(), self.size_precision())
+    }
+
+    fn next_bid_price(&self, value: f64, n: i32) -> Option<Price> {
+        let px = if let Some(scheme) = self.tick_scheme() {
+            scheme.next_bid_price(value, n, self.price_precision())?
+        } else {
+            let inc = self.price_increment().as_f64();
+            if inc <= 0.0 {
```
cjdsellers: Prices can potentially be negative (options, spreads, commodities etc).

crates/model/src/instruments/mod.rs
```diff
@@ -190,109 +374,108 @@ pub const EXPIRING_INSTRUMENT_TYPES: [InstrumentClass; 4] = [
 mod tests {
     use rstest::rstest;
 
-    use crate::instruments::{CurrencyPair, Instrument, stubs::*};
+    use super::*;
+    use crate::{instruments::stubs::*, types::Money};
 
     #[rstest]
-    fn test_make_qty_standard_rounding(currency_pair_btcusdt: CurrencyPair) {
+    fn make_qty_rounding(currency_pair_btcusdt: CurrencyPair) {
```
cjdsellers: We'll want to use `rstest` cases here for the parametrization. We could have substantially more cases as well.

crates/model/src/instruments/mod.rs
```diff
@@ -151,31 +278,88 @@ pub trait Instrument: 'static + Send {
...
-                (quantity.as_f64(), self.quote_currency())
+        let use_quote = use_quote_for_inverse.unwrap_or(false);
+        if self.is_inverse() {
+            if use_quote {
+                Money::new(quantity.as_f64(), self.quote_currency())
             } else {
-                let amount =
-                    quantity.as_f64() * self.multiplier().as_f64() * (1.0 / price.as_f64());
-                let currency = self
+                let amt = quantity.as_f64() * self.multiplier().as_f64() * (1.0 / price.as_f64());
```
cjdsellers: I think `amt` and `ccy` as local variable names are a little terse in this context and a bit different to other parts of the codebase. Lets stick with `amount` and `currency` for now.

crates/model/src/instruments/mod.rs
v9ai: removed, thx

crates/model/src/instruments/mod.rs
nicolad: added tests, thx

v9ai: | Concern / Feature                               | **“Before” – simple `f64` version**                                                     | **“After” – `rust_decimal` + guards**                                                                                              | Why the extra code? |
|-------------------------------------------------|-----------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|---------------------|
| **Numeric type used**                           | Pure `f64` maths.                                                                       | `Decimal` (base-10 exact) → then back to `f64`.                                                                                    | Avoids binary-decimal drift. |
| **Rounding when `round_down = Some(true)`**     | `floor()` toward −∞.                                                                    | `RoundingStrategy::ToZero` (truncate toward 0).                                                                                    | Symmetric for neg & pos sizes; aligns with most exchange “round-down” rules. |
| **Default rounding (`None` / `false`)**         | *No rounding at all* (just passes `value`).                                             | `MidpointNearestEven` (half-even, “banker’s”).                                                                                     | **Yes, half-even rounding is now explicit** instead of implicit/absent. |
| **Half-even support on midpoint ties**          | ❌ (relies on `Quantity::new` or IEEE default).                                          | ✅ Guaranteed via `MidpointNearestEven`.                                                                                            | Adds determinism and auditability. |
| **Handling NaN / ∞**                            | Whatever `Quantity::new` does (implicit).                                               | `from_f64_retain` + `expect("non-finite")` – explicit panic.                                                                       | Fail fast & loud. |
| **Guard against sub-tick quantities**           | None.                                                                                   | Panics if positive input would round to < 10 % of one tick.                                                                        | Prevents zero-size orders. |
| **Cross-platform determinism**                  | Dependent on IEEE-754 quirks / FP env.                                                  | Pure integer math inside `Decimal` → deterministic.                                                                                | Reproducible replays. |
| **Performance**                                 | Minimal ops (cheap).                                                                    | Extra `Decimal` ops; still cheap but slower in tight loops.                                                                        | Usually negligible outside ultra-low-latency paths. |
| **Lines of code / readability**                 | Short, easy to scan.                                                                    | Longer, self-documenting; comments explain invariants.                                                                             | Clarity & safety trade-off. |
| **Primary reason for the added code?**          | —                                                                                       | **Not just half-even rounding**; also precision-safe conversion, non-finite guard, sub-tick guard, deterministic behaviour. | Half-even is one piece of the puzzle. |


### Is all this additional code only to ensure half-even rounding?

No. Ensuring midpoint-nearest-even rounding *is* an important part, but the extra lines also:

1. **Eliminate binary floating-point drift** by moving to exact base-10 arithmetic during rounding.
2. **Fail fast on NaN/∞** instead of relying on downstream checks.
3. **Guarantee deterministic behaviour** across platforms and compiler flags.
4. **Prevent “rounded-to-zero” sizes** that would be rejected by exchanges.

So half-even rounding is *necessary but not sufficient*—the additional safeguards and determinism are equally valuable in a trading context.

v9ai: let me know if there are any actions I need to take here ☝️

crates/model/src/instruments/mod.rs
v9ai: adjusted, thx

crates/model/src/instruments/mod.rs
v9ai: actioned, thx

PR #2733, opened 2025-06-22
Improve implementation, validations and testing for Rust instruments

nicolad:
| Area | Item / Sub-task | Status | Key Facts & Follow-ups |
|------|-----------------|--------|------------------------|
| **Implementation** | Port all helper methods from Cython `Instrument` → Rust (`make_price`, `make_qty`, tick helpers, notional math, base-qty math) | ✅ | Trait-default impls complete. New ergonomic wrappers **`try_make_price` / `try_make_qty` / `try_calculate_base_quantity`** return `anyhow::Result`, while legacy panicking versions call their `try_` cousins. |
| | **Expose `info` metadata in PyO3** | ⛔️ | Still missing – needs `HashMap<String, PyObject>` plus getter/setter. |
| | **Remove `to_dict`/`from_dict` from Rust** | ✅ | All serialization now lives exclusively in the Python shim. |
| | Tick-scheme abstraction (`Fixed`, `Crypto0_01`, …) | ✅ | Trait-based, `enum_dispatch`, case-insensitive `FromStr`. |
| | **Negative-price support** | ✅ | Validation no longer forbids `price < 0`; tests added. |
| | Half-even rounding retained, but internal maths replaced by `rust_decimal::round_dp_with_strategy` for clarity | ✅ | Extensive doc-comments added to explain why half-even is required in trading systems. |
| | Variable naming cleanup (`amount`, `currency`, …) | ✅ | Short forms purged. |
| | Centralised validation (`validate_instrument_common`) | ✅ | Uses `ensure!` + helper macro `check_positive!`; covers paired-field, precision and bound checks. |
| | Option/Future-specific domain checks (underlying / strike / expiry order) | ⛔️ | Deferred to follow-up. |
| **Testing** | Parameterised examples converted to `rstest` cases | ✅ | All scalar grids migrated; many new edge cases; names use full words. |
| | Property-based fuzzing (`proptest`) for price/qty factories | ✅ | 10 000 cases, 0.0001 → 1e8 range. |
| | Validation *negative-path* coverage | ✅ | 30 + panic / error-return assertions. |
| | PyO3 ⇄ Python round-trip tests | ⛔️ | Blocked on `info` exposure. |
| **Misc / Tooling** | Lint & CI (`cargo clippy -D warnings`, multi-OS) | ✅ | Clean as of latest push. |
| | Docs (mdBook chapter on TickScheme + inverse/quanto maths) | ⛔️ | Draft underway; will land after API stabilises. |
| | Helper fns now have `try_*` fallible variants | ✅ | Downstream crates can opt-in to error handling. |

### DIFF
--- a/crates/model/src/instruments/mod.rs
+++ b/crates/model/src/instruments/mod.rs
@@ -34,9 +34,11 @@ pub mod stubs;
 
 use std::{fmt, str::FromStr};
 
-use anyhow::{anyhow, bail, ensure};
+use anyhow::{anyhow, bail};
 use enum_dispatch::enum_dispatch;
 use nautilus_core::UnixNanos;
+#[allow(unused_imports)]
+use nautilus_core::correctness::{check_positive_u64, check_predicate_true, check_valid_string};
 use rust_decimal::{Decimal, RoundingStrategy, prelude::*};
 use rust_decimal_macros::dec;
 use ustr::Ustr;
@@ -61,7 +63,7 @@ pub fn default_price_increment(price_precision: u8) -> Price {
 
 macro_rules! check_positive {
     ($val:expr, $msg:expr) => {
-        ensure!($val > 0.0, $msg);
+        check_predicate_true($val > 0.0, $msg)?;
     };
 }
 
@@ -82,23 +84,23 @@ pub fn validate_instrument_common(
     max_price: Option<&Price>,
     min_price: Option<&Price>,
 ) -> anyhow::Result<()> {
-    ensure!(price_precision >= 0, "price_precision negative");
-    ensure!(size_precision >= 0, "size_precision negative");
+    check_predicate_true(price_precision >= 0, "price_precision negative")?;
+    check_predicate_true(size_precision >= 0, "size_precision negative")?;
     check_positive!(size_increment.as_f64(), "size_increment not positive");
-    ensure!(
+    check_predicate_true(
         size_increment.precision as i32 == size_precision,
-        "size_precision != size_increment.precision"
-    );
+        "size_precision != size_increment.precision",
+    )?;
     check_positive!(multiplier.as_f64(), "multiplier not positive");
-    ensure!(margin_init >= dec!(0), "margin_init negative");
-    ensure!(margin_maint >= dec!(0), "margin_maint negative");
+    check_predicate_true(margin_init >= dec!(0), "margin_init negative")?;
+    check_predicate_true(margin_maint >= dec!(0), "margin_maint negative")?;
 
     if let Some(increment) = price_increment {
         check_positive!(increment.as_f64(), "price_increment not positive");
-        ensure!(
+        check_predicate_true(
             increment.precision as i32 == price_precision,
-            "price_precision != price_increment.precision"
-        );
+            "price_precision != price_increment.precision",
+        )?;
     }
     if let Some(lot) = lot_size {
         check_positive!(lot.as_f64(), "lot_size not positive");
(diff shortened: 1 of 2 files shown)


crates/model/src/instruments/mod.rs
```diff
@@ -241,66 +352,128 @@ pub trait Instrument: 'static + Send {
...
     fn ts_init(&self) -> UnixNanos;
 
-    /// Creates a new [`Price`] from the given `value` with the correct price precision for the instrument.
+    fn _min_price_increment_precision(&self) -> u8 {
+        self.price_increment().precision
+    }
+
+    /// # Errors
+    ///
+    /// Returns `anyhow::Error` if the value is not finite or cannot be converted to a `Price`.
```
cjdsellers, line 361: We don't need to qualify the error type in the description here. _"Returns an error if ..."_ is enough.

crates/model/src/instruments/stubs.rs
```diff
@@ -556,7 +636,7 @@ pub fn betting() -> BettingInstrument {
     );
     let selection_id = 50214;
     let selection_name = Ustr::from("Kansas City Chiefs");
-    let selection_handicap = 0.0; // As per betting convention, no handicap
+    let selection_handicap = 0.0;
```
cjdsellers, line 639: This comment is still reasonably informative.

crates/model/src/instruments/stubs.rs
```diff
@@ -501,7 +581,7 @@ pub fn option_spread() -> OptionSpread {
         Symbol::from("UD:U$: GN 2534559"),
         AssetClass::FX,
         Some(Ustr::from("XCME")),
-        Ustr::from("SR3"), // British Pound futures (option on futures)
+        Ustr::from("SR3"),
```
cjdsellers, line 584: This comment is still reasonably informative.

crates/model/src/instruments/stubs.rs
```diff
@@ -464,7 +544,7 @@ pub fn option_contract_appl() -> OptionContract {
         InstrumentId::from("AAPL211217C00150000.OPRA"),
         Symbol::from("AAPL211217C00150000"),
         AssetClass::Equity,
-        Some(Ustr::from("GMNI")), // Nasdaq GEMX
+        Some(Ustr::from("GMNI")),
```
cjdsellers, line 547: This comment is still reasonably informative.

crates/model/src/instruments/mod.rs
```diff
@@ -584,4 +803,443 @@ mod tests {
...
+    #[case(0.999_999, true)]
+    #[case(1.000_0001, false)]
+    #[case(1.000_0001, true)]
+    #[case(1.234_5, false)]
+    #[case(1.234_5, true)]
+    #[case(2.345_5, false)]
+    #[case(2.345_5, true)]
+    #[case(0.000_999_999, false)]
+    #[case(0.000_999_999, true)]
+    fn quantity_rounding_grid(
```
cjdsellers, line 1172: Good amount of cases here 👌

crates/model/src/instruments/mod.rs
```diff
@@ -53,9 +56,14 @@ use crate::{
     types::{Currency, Money, Price, Quantity},
 };
 
+pub fn default_price_increment(price_precision: u8) -> Price {
+    let step = 10f64.powi(-(price_precision as i32));
+    Price::new(step, price_precision)
+}
+
 macro_rules! check_positive {
```
cjdsellers, line 64: This macro isn't necessary.

crates/model/src/instruments/mod.rs
```diff
@@ -76,23 +84,23 @@ pub fn validate_instrument_common(
     max_price: Option<&Price>,
     min_price: Option<&Price>,
 ) -> anyhow::Result<()> {
-    ensure!(price_precision >= 0, "price_precision negative");
-    ensure!(size_precision >= 0, "size_precision negative");
+    check_predicate_true(price_precision >= 0, "price_precision negative")?;
+    check_predicate_true(size_precision >= 0, "size_precision negative")?;
```
cjdsellers, line 88: We have more specialized correctness functions which could be used for many of these, e.g.
https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/data/trade.rs#L76
```
check_positive_quantity(size, stringify!(size))?;
```

PR #2736, opened 2025-06-25
Rust instruments: implemented and used check_positive_decimal

nicolad:
Introduce `check_positive_decimal` in **crates/core/src/correctness.rs** to enforce that `rust_decimal::Decimal` parameters are strictly positive. The helper replaces hand-rolled checks inside `validate_instrument_common`, tightening validation for fields such as `margin_init`. Comprehensive rstest cases cover positive, zero, and negative edge conditions.

### Impact

Non-breaking: only invalid (≤ 0) decimal inputs now fail early with a clear error message, improving safety while leaving existing valid workflows unchanged.

### Notes

* Added function‐level docs and `#[inline(always)]` hint.
* Extended unit-test suite in both `correctness` and `model::instruments` modules.

### DIFF
--- a/crates/core/src/correctness.rs
+++ b/crates/core/src/correctness.rs
@@ -23,6 +23,7 @@
 //! An [`anyhow::Result`] is returned with a descriptive message when the
 //! condition check fails.
 
+use rust_decimal::Decimal;
 use std::fmt::{Debug, Display};
 
 use crate::collections::{MapLike, SetLike};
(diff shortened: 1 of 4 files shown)


crates/core/src/correctness.rs
```diff
@@ -507,6 +508,19 @@ where
...
 }
 
+/// Checks the `Decimal` value is positive (> 0).
+///
+/// # Errors
+///
+/// Returns an error if the validation check fails.
+#[inline(always)]
+pub fn check_positive_decimal(value: Decimal, param: &str) -> anyhow::Result<()> {
+    if value <= Decimal::ZERO {
```
cjdsellers, line 518: Nicely implemented 👌

crates/model/src/instruments/mod.rs
```diff
@@ -85,7 +85,7 @@ pub fn validate_instrument_common(
         "size_precision",
     )?;
     check_positive_quantity(multiplier, "multiplier")?;
-    // TODO: check_positive_decimal
+    check_positive_decimal(margin_init, "margin_init")?;
     check_predicate_true(margin_init >= dec!(0), "margin_init negative")?;
```
cjdsellers: We could use `check_positive_decimal` for both of these now rather than the general `check_predicate_true`.

crates/core/src/correctness.rs
v9ai: thx

crates/model/src/instruments/mod.rs
v9ai: done, thx

cjdsellers (approved):
👌

PR #2748, opened 2025-07-01
Propagated start and stop in response data, start and stop are newly required parameters in request/response data and set new standard for start and stop boundaries handling

stastnypremysl:
Added also new logic to adapters (those, which expected None, are now filtering incomming responses from venues) and actors (to handle None requests - either by an error or by default value).

# Pull Request

## Summary

This pull request introduces a new `start` and `end` timestamp feature to various response types and their related methods, ensuring that data responses include a time range for the data they represent. This change impacts the core data structures, test cases, and client implementations.

### Core Data Structure Updates:
* Added `start` and `end` fields of type `UnixNanos` to response structs like `CustomDataResponse`, `InstrumentResponse`, `InstrumentsResponse`, `BookResponse`, `QuotesResponse`, `TradesResponse`, and `BarsResponse`. Updated their constructors to include these fields.

### Test Case Adjustments:
* Modified test cases (`test_request_instrument`, `test_request_instruments`, `test_request_quotes`, `test_request_trades`, `test_request_bars`, `test_request_book_snapshot`, `test_request_data`) to include `start` and `end` timestamps when creating response objects.

### Client Implementation Updates:
* Updated methods in the `LiveDataClient` trait to include `start` and `end` timestamp parameters when sending responses (e.g., `send_instrument_response`, `send_book_response`, `send_instruments_response`, etc.).

### Mock Data Adjustments:
* Updated `MockDataClient` in `data_client.rs` to provide `start` and `end` timestamps when creating mock data responses. 

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Newly, any `request_data` except for `request_instrument`/`request_instruments` will require a start date, This should be later changed by adding sensible calculated default values.

## Release notes

## Testing


nautilus_trader/backtest/node.py
```diff
@@ -401,7 +401,16 @@ def _create_engine(self, run_config_id: str) -> BacktestEngine:
                 trade_execution=venue_config.trade_execution,
             )
 
-        # Add instruments
+        # Load instruments from configured catalogs
+        for catalog_config in engine_config.catalogs:
```
stastnypremysl: @cjdsellers @faysou Does this change makes sense please? I added it because of a failing test test_backtest_same_with_and_without_data_configs, but I am not sure, whether it is meaningful.

faysou: No, I don't think we should load all instruments of all catalogs by default into the cache, the existing code is already loading instruments corresponding to BacktestDataConfigs.

faysou:
https://github.com/nautechsystems/nautilus_trader/blob/e8299748092e061960dfd81982dd51ffafb2ae0e/nautilus_trader/common/actor.pyx#L2794

The signatures of functions requiring start should be changed so they don't default to None.

You will be able to modify these kind of instructions in engine.pyx, as the start and end are now contained in response

```python
if update_catalog:
    start = response.params.get("request_ts_start")
    end = response.params.get("request_ts_end")
```

will become

```python
if update_catalog:
    start = response.start
    end = response.end
```

I can modify these request_ts_start things after the merge.

Apart from these small comments, looks ok to me

stastnypremysl:
> https://github.com/nautechsystems/nautilus_trader/blob/e8299748092e061960dfd81982dd51ffafb2ae0e/nautilus_trader/common/actor.pyx#L2794
> 
> The signatures of functions requiring start should be changed so they don't default to None.

@faysou, I would  like to keep them defaulting to the None, because in the future, there might be some logic in `actor.pyx`, how default `start` is given. (e.g. 100 bars for time based aggregation, if None is given) --- The main improvement here is in the consistency across all venues.

cjdsellers:
I'm happy to leave the defaults unchanged for now and we can revisit this later.

nautilus_trader/backtest/node.py
cjdsellers: I think we should stick to existing behavior for now, also because the catalog should be supporting multiple versions of each instrument eventually.

nautilus_trader/common/actor.pyx
```diff
@@ -2405,6 +2405,50 @@ cdef class Actor(Component):
             handler=self.handle_signal,
         )
 
+# -- VALIDATIONS -------------------------------------------------------------------------
+
+    cdef tuple _validate_datetime_range(
```
cjdsellers, line 2410: On the next pass we might want to extract the concept of a time range into a `TimeRange` type, since `start` and `end` travel through the system so often.

nautilus_trader/common/actor.pyx
```diff
@@ -2560,16 +2609,11 @@ cdef class Actor(Component):
...
-
-        if end is not None:
-            Condition.is_true(end <= now, "end was > now")
+        # Handle None values for start by replacing with current UTC time
+        if start is None:
+            start = self.clock.utc_now()
 
-        if start is not None and end is not None:
-            Condition.is_true(start <= end, "start was > end")
+        start, end = self._validate_datetime_range(start, end)
```
cjdsellers, line 2626: If both `start` and `end` are `None`, then shouldn't we use the same `now` timestamp for both - which infers a point request / latest snapshot (rather than a tiny difference between the two which is latency between `clock.utc_now()` calls?)

nautilus_trader/data/client.pxd
```diff
@@ -167,8 +169,8 @@ cdef class MarketDataClient(DataClient):
 
 # -- DATA HANDLERS --------------------------------------------------------------------------------
 
-    cpdef void _handle_instrument(self, Instrument instrument, UUID4 correlation_id, dict[str, object] params)
-    cpdef void _handle_instruments(self, Venue venue, list instruments, UUID4 correlation_id, dict[str, object] params)
-    cpdef void _handle_quote_ticks(self, InstrumentId instrument_id, list ticks, UUID4 correlation_id, dict[str, object] params)
-    cpdef void _handle_trade_ticks(self, InstrumentId instrument_id, list ticks, UUID4 correlation_id, dict[str, object] params)
-    cpdef void _handle_bars(self, BarType bar_type, list bars, Bar partial, UUID4 correlation_id, dict[str, object] params)
+    cdef void _handle_instrument(self, Instrument instrument, UUID4 correlation_id, datetime start, datetime end, dict[str, object] params)
```
cjdsellers, line 172: These started as `cdef` as no need to expose a Python wrapper, then at some stage changed to `cpdef` just for ease of development and very minimal performance overhead.

I imagine you're thinking of changing back to `cdef` due access mod and performance (I don't think it's necessary at this stage in cythons service life?)

nautilus_trader/data/messages.pyx
```diff
@@ -2388,6 +2388,10 @@ cdef class DataResponse(Response):
         The response ID.
     ts_init : uint64_t
         UNIX timestamp (nanoseconds) when the object was initialized.
+    start : datetime
+        The start datetime (UTC) of response time range (inclusive).
+    end : datetime
+        The end datetime (UTC) of response time range.
```
cjdsellers: Should we mention inclusiveness for `end` as well?

nautilus_trader/common/actor.pyx
faysou: Could be a good idea

nautilus_trader/data/messages.pyx
faysou: In the data engine when handling responses I assume that we query [start, end] intervals, ie. both inclusive.

cjdsellers: Yes, I think we're standardizing on inclusive end, so should we explicitly document that in the `end` description - for symmetry?

```
The end datetime (UTC) of response time range (inclusive).
```

nautilus_trader/data/client.pxd
stastnypremysl: I have done it due to the reasons you mentioned. It was just an easy change as I was passing by.

Should I revert it?

nautilus_trader/data/messages.pyx
stastnypremysl: Fixed.

nautilus_trader/common/actor.pyx
stastnypremysl: I agree, that sounds like a good idea.

nautilus_trader/common/actor.pyx
stastnypremysl: Fixed.

Btw. I have changed the adapters, so they expect the difference, so this should work even without the change. But it is cleaner this way.

stastnypremysl:
@cjdsellers, please swap `*` for `-` during squashing the commits. I am not sure, I am able to rebase it now, after the merge from the develop was done.

I will also add a comment to the description.

nautilus_trader/backtest/node.py
stastnypremysl: Okey. I will fix the test other way.

stastnypremysl: I've implemented a breaking fix to how instruments are loaded. More in description of the ticket.

faysou:
You  can soft reset until HEAD~n and then do a single commit. Rebasing is what to do with PRs not merging.

nautilus_trader/data/client.pxd
cjdsellers: No need to revert, but thought I should provide context on why they were changed to `cpdef`.

nautilus_trader/adapters/binance/data.py
```diff
@@ -542,12 +543,18 @@ async def _unsubscribe_bars(self, command: UnsubscribeBars) -> None:
...
     async def _request_instrument(self, request: RequestInstrument) -> None:
-        if request.start is not None:
+        # Check if start/end times are too far from current time
+        now = self._clock.utc_now()
+        now_ns = dt_to_unix_nanos(now)
+        start_ns = dt_to_unix_nanos(request.start)
+        end_ns = dt_to_unix_nanos(request.end)
+
+        if abs(start_ns - now_ns) > 10_000_000:  # More than 10ms difference
             self._log.warning(
```
cjdsellers, line 553: Databento use the `start` and `end` for instrument definition requests:

https://databento.com/docs/api-reference-historical/timeseries/timeseries-get-range?historical=python&live=python&reference=python

nautilus_trader/adapters/binance/data.py
```diff
@@ -542,12 +543,18 @@ async def _unsubscribe_bars(self, command: UnsubscribeBars) -> None:
...
 
     async def _request_instrument(self, request: RequestInstrument) -> None:
-        if request.start is not None:
+        # Check if start/end times are too far from current time
+        now = self._clock.utc_now()
+        now_ns = dt_to_unix_nanos(now)
+        start_ns = dt_to_unix_nanos(request.start)
+        end_ns = dt_to_unix_nanos(request.end)
+
+        if abs(start_ns - now_ns) > 10_000_000:  # More than 10ms difference
```
cjdsellers, line 552: What is the purpose of this 10ms check?

nautilus_trader/adapters/binance/data.py
```diff
@@ -570,25 +577,20 @@ async def _request_trade_ticks(self, request: RequestTradeTicks) -> None:
...
             )
         else:
             # Convert from timestamps to milliseconds
-            start_time_ms = None
-            end_time_ms = None
-            if request.start:
-                start_time_ms = int(request.start.timestamp() * 1000)
-            if request.end:
-                end_time_ms = int(request.end.timestamp() * 1000)
+            start_time_ms = int(request.start.timestamp() * 1000)
```
cjdsellers, line 592: We could use the `millis_to_nanos` function here, but can be done in a follow-up.

cjdsellers: We could use the `secs_to_millis` function here, but can be done in a follow-up.

[edit] I just noticed you did this below :+1:

stastnypremysl:
> You can soft reset until HEAD~n and then do a single commit. Rebasing is what to do with PRs not merging.

@faysou, after your comment I dived in into research of git and found out `git rebase -i --rebase-merges HEAD~N` option. The additional steps aren't required now.

nautilus_trader/adapters/binance/data.py
stastnypremysl: `start_ns` and `end_ns` are being set in live trading with a small time hole between `now_ns` in the current context and `start_ns` and `end_ns`.

## Example - negative check in live trading:
* `start_ns=10`
* `end_ns=10`
* `now_ns=1_056`

The user probably called `request_instrument` with `None` for both `start` and `end`. `now` in `request_instrument` is `10` and until the code reaches the second `clock.now_ns()`, the time is already `1_056`. Probably no reason to show the warning.

## Example - positive check in live trading:
* `start_ns=100_000_000`
* `end_ns=200_000_000`
* `now_ns=330_086_916`

User is evidently requesting for historical instrument data. The warning about meaninglessness of this action is raised now.

nautilus_trader/adapters/binance/data.py
stastnypremysl: Databento don't have this check integrated.

nautilus_trader/adapters/binance/data.py
stastnypremysl: Done.

PR #2757, opened 2025-07-07
Add PyO3 config script to prevent unnecessary rebuilds

stastnypremysl:
Further changes:
- Built wheels are newly in dist folders separated by category (build, debug)
- Minor refinements of the Makefile

## Purpose
Currently, when the `make build` is run twice under the same repository, it triggers a rebuild even without any change, because of PyO3 detects the environment change. This PR solves it.

Also, the dist folders have been separated to ease the recognition of which wheel file was build with which profile.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [x] Maintenance / chore

## Release notes


stastnypremysl:
As described in [environment_setup.md](https://github.com/nautechsystems/nautilus_trader/blob/4863fc5f0c4a97fb39e10f8c4160e22513084b03/docs/developer_guide/environment_setup.md#L4), this problem is solvable by setting PYO3_PYTHON environment variable. I would suggest to set it up directly in .venv/bin/activate. 

This presents much better solution, then the MR here, which makes codebase more complex. Therefore closing it.

PR #2766, opened 2025-07-10
Allow getting any number of historical bars from the dydx adapter

DeirhX:
## Summary

Removes the hardcoded limit of 100 historical bars (dydx API can do 1000) and automatically splits larger historical bar requests into multiple API calls so you can get the entire range.

Note: there might be other limits, such as rate limits, that could prevent getting a larger chunk of history. I tested only on reasonable ones so far.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

I didn't modify the tests; let me know if I should add a case specific to a larger timespan that would trigger this new feature

### DIFF
--- a/nautilus_trader/adapters/dydx/data.py
+++ b/nautilus_trader/adapters/dydx/data.py
@@ -933,10 +933,8 @@ class DYDXDataClient(LiveMarketDataClient):
         return nautilus_instrument_id
 
     async def _request_bars(self, request: RequestBars) -> None:
-        max_bars = 100
+        max_bars = 1000
         limit = request.limit
-        if limit == 0 or limit > max_bars:
-            limit = max_bars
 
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @DeirhX 

Thanks so much for the contribution! :+1: 

It looks like just the pre-commit is failing, you can fix this by running the following and diffing any changes:
```
make pre-commit
```
or (required for Windows)
```
pre-commit run --all-files
```

Also consider running the following to install the git hook which will run the pre-commit automatically on every commit:
```
pre-commit install
```

DeirhX:
Thanks, I thought my pre-commit applies to your repo as well, but I was wrong. Will explore the Nautilus coding style and adapt accordingly.

PR #2773, opened 2025-07-11
Add integration tests for dYdX bar-partitioning and large-history handling

nicolad:
This PR introduces a full test-suite that validates the new bar-partitioning logic used by `DYDXDataClient` to fetch large spans of historical data safely and efficiently.

**Key points**

* **Partitioning boundaries** – verifies when requests should/shouldn’t be split across multiple HTTP calls.
* **Time-frame coverage** – checks minute, 5-minute, 15-minute, hourly, 4-hour and daily bar calculations.
* **Chunk sizing** – asserts correct chunk counts for 1 000–3 000 bar ranges.
* **Partial-bar handling** – guarantees that incomplete candles are excluded yet still returned separately.
* **Error resilience** – simulates API failures and rate-limiting delays to ensure graceful degradation.
* **Limit enforcement** – confirms global `limit` is honored across all partitions.


cjdsellers:
Great work @nicolad

PR #2777, opened 2025-07-12
Fixed limit handling in dydx request_bars

DeirhX:
Fix to https://github.com/nautechsystems/nautilus_trader/pull/2766 so that limits specifying the maximum number of bars work correctly in both cases.


DeirhX:
Found out that my last commit broke the case of having to partition the request. Sorry for that.

cjdsellers:
@DeirhX thanks for the fix :pray:

PR #2790, opened 2025-07-15
Fix typo in logging in dydx adapter

DeirhX:
A very simple fix of a typo.


cjdsellers:
Thanks for the fix @DeirhX, good catch

DeirhX:
No problem. The order book doesn't work for me, so I'll look into that next 😼

PR #2808, opened 2025-07-23
Fix parallel pytest tests and enable them

stastnypremysl:
- Add it to the Makefile and CI/CD pipeline
- Add tmp_path isolation for catalog including tests to prevent shared directory race conditions
- Add xdist_group markers to database integration tests to prevent race conditions
- Make dydx execution test deterministic by using fixed UUID instead of random (precondition for parallel pytest-xdist to work)
## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore


faysou:
It runs much faster now, thank you

PR #2811, opened 2025-07-24
Add Python memory leak detection for tests

stastnypremysl:
- Add memory tracking fixture to pytest that monitors RSS memory usage and tracemalloc
- Configure via environment variables:
  - MEMORY_TRACKING_ENABLED_PY (default: false)
  - MEMORY_LEAK_THRESHOLD_BYTES_PY (default: 1024 * 1024 * 10 bytes)
- Add parallel `make pytest-memory-tracking` target to run tests with tracking enabled
- Report memory leaks with initial, final, increase in MB, tracemalloc stats and snapshot with a test failure

## Type of change

- New feature for tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -372,6 +372,11 @@ else
 	uv run --active --no-sync pytest --new-first --failed-first --tb=short -n logical --dist=loadgroup
 endif
 
+.PHONY: pytest-memory-tracking
+pytest-memory-tracking:  #-- Run Python tests with memory tracking enabled
+	$(info $(M) Running Python tests with memory tracking enabled...)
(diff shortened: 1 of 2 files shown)


stastnypremysl:
There are 31 probable memory leak test failures with >10 MiB heap size diff.

image

cjdsellers (commented):
Thanks for this @stastnypremysl, it'll help identify mem leaks.

Lets keep it separate from the CI builds for now, and we can consider incorporating it into the `nightly` run later.

stastnypremysl:
Hi, @cjdsellers. The CI/CD part was removed. When CI/CD passes, it should be ready to be merged.

PR #2813, opened 2025-07-26
Dydx-normalize-instruments

DeirhX:
# Pull Request

## Summary

I believe we should always normalize the instrument id, even when populating the order book. Without this change, I was getting different IDs inserted in the book compared to other uses of this instrument.

## Type of change

- [X] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

Unfortunately, there is no test for this adapter.


davidsblom:
Thanks for opening the pull request. Do you know what the underlying issue is? Looking at the diff, I don't see a reason why it didn't work. Thanks!

DeirhX:
Hey David, I'll try to supply a more detailed description once I get back to working on dydx integration. 
I remember that while my instrument names were "DOGE-USDC", internally, they were represented as "DOGE-USDC-PERP". I had no problem getting live and historic bars, but subscribing orderbook deltas and receiving updated ended with an exception as it was reaching for a different symbol name.

DeirhX:
This was where I was getting the exception without this modification: https://github.com/nautechsystems/nautilus_trader/blob/c8a8100238c1f839a59f42c5df56bc810f0fd8dd/nautilus_trader/adapters/dydx/data.py#L625

But I understand it can have potentially other, undesired effects, so we can leave it as draft for the time being.

davidsblom:
Thanks! All instruments should be named COIN-USD-PERP in case spot trading is introduced later by dYdX.

cjdsellers:
Seems like there might be an underlying issue with dYdX symbol naming conventions / normalization?

cjdsellers:
@DeirhX and @davidsblom 

What are your thoughts on this one, should we close for now?

davidsblom:
Sure, probably better to solve the underlying issue.

Issue #2767, opened 2025-07-10
Adding .pyi stub files for .pyx Cython modules


woung717:
## Context
G'day, I managed to generate .pyi stub files from .pyx files using the [cythonbuilder](https://github.com/mike-huls/cythonbuilder) project, after making few edits to the code. There are more files that need to be checked, but these seem like low-hanging fruit for now.

Thanks!

### Considerations

I know that porting to Rust is an ongoing process, but it would be useful to have an interim solution. Could I submit a pull request with the added .pyi files? Would the .pyx (interfaces) files be changed further?

cjdsellers:
Hi @woung717

Thanks for reaching out.

If you run `make pre-commit` is mypy OK with your changes? this was the blocker last time this was attempted (resulted in ~500 errors which was untenable to fix).

woung717:
@cjdsellers Thanks for letting me know! I’ll run the command and check for any issues.

faysou:
You could use an AI agent to fix mypy issues as well.

woung717:
I've tested several solutions:

* [cythonbuilder](https://github.com/mike-huls/cythonbuilder): Couldn't generate complete stubs for complex `.pyx` files. Output is broken in places. Fails `pre-commit mypy` checks.
* PyCharm cached stubs: All stub information is well extracted, but still fails `pre-commit mypy` with many errors.
* `mypy stubgen`: Doesn't seem to support `.pyx` files.

The most common issues are with converting Cython-style type hints and correctly identifying things like `@staticmethod`s.

Still, even beyond that, there are other specific issues that prevent `mypy` from passing. One major problem is Cython's implicit type conversion—for example, when a function expects an `InstrumentId` but receives a `str`. Unless we modify the calling `.py` files to avoid these implicit conversions, adding `.pyi` files will always lead to `mypy` failures.

It seems like there is no viable static analysis-based stub generator for cython projects. Personally, I prefer to rely on static analysis for code generation to maintain consistency, but given the limitations, using an LLM as @faysou suggested might not be a bad idea. I tested several complex cases (previously failed) and the results were good. That said, any generated stubs would definitely need testing.

In my opinion, managing the stubs in a separate project—like how `pandas-stubs` does it—might be a reasonable way to sidestep `mypy`. I think we need to hear @cjdsellers 's opinion on this.

faysou:
I think that given the current goal is to remove cython there's no point in thinking too much about it now. Better focus on the rust migration.

[...]

PR #2819, opened 2025-07-29
[Draft] Adding .pyi stub files for .pyx Cython modules 

woung717:
# Pull Request

## Summary
@cjdsellers 
This is a draft PR for feedback before moving forward. 
I added a stub file for the Account class. If you have any suggestions for improvements or ways to align it with the current code base, please let me know.

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/2767

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [x] Maintenance / chore

## Documentation

## Release notes

## Testing


faysou:
This looks interesting. You would need to convert the comments to English though.

cjdsellers:
Hi @woung717

The number of pyi files is surprising, but I suppose this matches the modules one-to-one.

You'll need to change the imports to use the main `nautilus_trader` package rather than the pyo3 bindings, as they're different types.

- Are the docstrings necessary given we're mostly going for import resolution here?
- What is your actual workflow for producing these? I'm concerned about maintainability
- Does pre-commit still pass when these are present, or they have to be removed every time before running it?

woung717:
@cjdsellers Yeah, quite a lot.

> You'll need to change the imports to use the main nautilus_trader package rather than the pyo3 bindings, as they're different types.

Got it. However, I think it should import .pyi stub symbols, since the type checkers (Pyright and Pyrefly) used in IDEs other than PyCharm can't resolve Cython (.pyx) symbols in .pyi files. It's only for import resolution, not implementation, so I think it will work.

> Are the docstrings necessary given we're mostly going for import resolution here?

Docstrings in .pyi files provide useful documentation for VSCode and other IDE users. I don't see any reason to avoid them. (Except stub files can be bloated.) Below is an example with .pyi stubs (with docstring) from VSCode.
image

> What is your actual workflow for producing these? I'm concerned about maintainability

 I generated stubs using LLM with detailed instructions, then corrected minor errors in the generated stubs. I wrote a .pyx ↔ .pyi match validation script for this. I think it can be used in the pre-commit pipeline.

> Does pre-commit still pass when these are present, or they have to be removed every time before running it?

Unfortunately, it still caused errors, so it had to be removed before running mypy. 
(Note: Mypy errors occur because it doesn’t fully understand Cython’s syntax differences from Python, such as implicit type casting. Given the size of this codebase, making stubs compatible with both Cython and Python is nearly impossible.)

cjdsellers:
> Docstrings in .pyi files provide useful documentation for VSCode and other IDE users. I don't see any reason to avoid them. (Except stub files can be bloated.)

- Yes, bloated stub files
- Increased overhead keeping docs aligned in multiple places
- Increased size of diffs on docs changes (applies to code changes also)
- Small point, but worth mentioning: [PEP 561](https://peps.python.org/pep-0561/) _“Stub files should not contain docstrings or any runtime logic, as they are purely for type checking.”_

I agree that an advantage is intellisense in IDE's etc.

> Got it. However, I think it should import .pyi stub symbols, since the type checkers (Pyright and Pyrefly) used in IDEs other than PyCharm can't resolve Cython (.pyx) symbols in .pyi files. It's only for import resolution, not implementation, so I think it will work.

I don't think we should refer to `pyo3` paths for legacy Cython imports in pyi's meant for the Cython package? I think `.pyi` stubs can work for Cython modules, as we do have some, e.g. https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/core/uuid.pyi

> Mypy errors occur because it doesn’t fully understand Cython’s syntax differences from Python, such as implicit type casting

Cython supports implicit coercion of Python objects (such as `int`, `float`, or `str`) to compatible C primitive types, but only in C-specific contexts, like when assigning to a cdef-declared variable or passing to a C function. In contrast, when dealing with Python contexts (such as untyped variables or those declared as cdef object), Cython does not introduce any extra type coercions; it simply relies on Python's standard runtime behavior for dynamic typing.

i.e. it won't implicitly convert `InstrumentId` to a string because `InstrumentId` implements `__str__`, it's strongly typed and so raises a `TypeError` in this situation (regardless, you've mentioned this a couple of times but I don't think this is an issue related to type stubs, unless you can point to a concrete example - as this only makes Cython _more_ flexible, as it can coerce additional Python types into a C primitive in some cases).

> making stubs compatible with both Cython and Python is nearly impossible.

I don't fully agree with this, stubs can be compatible with both, the typing just has to be what Cython can actually accept. The overhead of writing the `.pyi` files was a reason why they weren't written and maintained from the outset.

Out of interest, with the new stubs present - what is the behavior when auto importing e.g. `TradeTick` where we have pyo3 stubs [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/core/nautilus_pyo3.pyi) and potentially a `data.pyi` next to `data.pyx`? is this the cause of some of the mypy errors? (I can't recall exactly what the errors were now.)

Thinking about this further, we're facing a large increase in complexity with potentially three different versions of stubs:
- PyO3 types from https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/core/nautilus_pyo3.pyi
- Generated stubs for v2 under `/python/nautilus_trader/`
- New stubs being discussed here under a top level `stubs/` (which are moved into the package at build time only?)

woung717:
@cjdsellers 
> I agree that an advantage is intellisense in IDE's etc.

I think intellisense provide from stub is another important feature that improves the dev experience with between resolution. However, I also agree that maintaining redundant docstrings may not be negligible. In that case, we could exclude the docstrings from the stubs this time and improve upon them in the next step. E.g. injecting cython-side docstrings into stub files at build time.

> > Got it. However, I think it should import .pyi stub symbols, since the type checkers (Pyright and Pyrefly) used in IDEs other than PyCharm can't resolve Cython (.pyx) symbols in .pyi files. It's only for import resolution, not implementation, so I think it will work.
> 
> I don't think we should refer to `pyo3` paths for legacy Cython imports in pyi's meant for the Cython package? I think `.pyi` stubs can work for Cython modules, as we do have some, e.g. [`develop`/nautilus_trader/core/uuid.pyi](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/core/uuid.pyi)

I got that .pyi stubs should import paths other than pyo3. What I mean is that symbols from Cython code (.pyx files) cannot be recognized in pure Python (since cimport is not possible in .pyi files). Therefore, .pyi stubs (which use Python syntax) need to import each other rather than importing .pyx or .pyo3 symbols. Am I explaining clearly?

> I don't fully agree with this, stubs can be compatible with both, the typing just has to be what Cython can actually accept. The overhead of writing the `.pyi` files was a reason why they weren't written and maintained from the outset.

Sure, it's totally possible to write stubs that are compatible with both, but my point is that, within the scope of this PR, it takes too much time to add them manually by investigating the codebase and applying the rules that Cython can accept — especially given the current size of the codebase and transitional phase of the project. The current stubs were generated solely from the corresponding .pyx files (symbols).

> Out of interest, with the new stubs present - what is the behavior when auto importing e.g. `TradeTick` where we have pyo3 stubs [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/core/nautilus_pyo3.pyi) and potentially a `data.pyi` next to `data.pyx`? is this the cause of some of the mypy errors? (I can't recall exactly what the errors were now.)
>
> Thinking about this further, we're facing a large increase in complexity with potentially three different versions of stubs:

When I tested by placing a stub (`atr.pyi` in this case) that have redundant symbol in pyo3 stubs, it didn't introduce any mypy errors. About the auto importing (if it means in IDE) it shows a list to choose which symbol should be imported since this two symbols have different import path. IDE and mypy will refer the symbols from actually imported stub that stated in the code.

> * Generated stubs for v2 under `/python/nautilus_trader/`

Does it mean existing pyi stubs such as `nautilus_trader/core/uuid.pyi`? Do you have any case that causes to concern? 
And I think we might need to prevent importing from the `stubs/` path when developing the nautilus_trader project itself.

By the way, what is the purpose of the pyo3 stub? As far as I know, PyO3 is used for binding Rust implementations. However, I found that some Cython code imports those symbols from the stub.

woung717:
@cjdsellers Stub writing is almost done. (Docstrings removed, imports changes to non-pyo3)
I need to add steps for copying stub files on build and stub file validation on pre-commit. Can you give me some guide? (What to see, where to start, etc..)

cjdsellers:
Hi @woung717 

> I got that .pyi stubs should import paths other than pyo3. What I mean is that symbols from Cython code (.pyx files) cannot be recognized in pure Python (since cimport is not possible in .pyi files). Therefore, .pyi stubs (which use Python syntax) need to import each other rather than importing .pyx or .pyo3 symbols. Am I explaining clearly?

Yes, I'm following now - this is why we need the type stubs because while Cython produces the extension modules which have a pure Python API, Python still needs the additional type information.

> Sure, it's totally possible to write stubs that are compatible with both, but my point is that, within the scope of this PR, it takes too much time to add them manually by investigating the codebase and applying the rules that Cython can accept — especially given the current size of the codebase and transitional phase of the project. The current stubs were generated solely from the corresponding .pyx files (symbols).

Understand your point. Or you could conditionally add the type information and build it up incrementally? There's nothing special about Cython, as it will just produce extension modules you then write the type stubs for as if it was Python (the same as how we've done it for the PyO3 stubs I'll explain in a moment - i.e., you just use the type(s) Cython can actually accept - or we expect the user to pass, if it can coerce additional types then that's just a bonus).

> About the auto importing (if it means in IDE) it shows a list to choose which symbol should be imported since this two symbols have different import path.

Makes sense, thanks.

> Does it mean existing pyi stubs such as nautilus_trader/core/uuid.pyi?

No, any stub files not under the `python/` directory are for the current "legacy" v1 system and the top-level `nautilus_trader/` package, per the location.

> By the way, what is the purpose of the pyo3 stub? As far as I know, PyO3 is used for binding Rust implementations. However, I found that some Cython code imports those symbols from the stub.

It's because PyO3 will produce a dynamic library which needs the type stubs in exactly the same way as Cython. It just means we can then have some static analysis. There is a lot of code via the `pyo3` crate (and `nautilus_pyo3` lib it produces) which is being actively used in the current "legacy" v1 system as well.

I had a look at the PR, do you refer to the `stubs.` subpackage to avoid moving the stub files into the main package? If they _were_ moved into the main package then you could just e.g. `import nautilus_trader.core.uuid` directly, per previous example. [edit: just read your last message, I think it's better the type stubs are moved into the main package as you suggest. Though I'm not sure of the best way to achieve this and keep the pre-commit passing].

Do imports then resolve to these stub files, or to the module itself? e.g., as pictured below; `UUID4` resolves to the actual module location because the `uuid.pyi` stub is placed beside the `uuid.pyx` module.

image
image

cjdsellers:
I still have some concerns over maintainability here.

You mentioned stub validation on pre-commit, so this will be an automated way of _checking_ that the stub file remains in line with the implementation? Recalling some of the refactorings we've done in the past and even recently, there would be an enormous increase in maintenance overhead here to manually keep these stubs consistent (@faysou could probably attest to this too). Thankfully cbindgen has been auto-generating the C headers needed for the Rust FFI this whole time, otherwise the burden of having to manually maintain that would have been untenable and probably removed.

Acknowledging that ideally stubs _should_ have been added and maintained from the outset - I'm also questioning the advantage of adding these so late in the service life of Cython (no concrete timeline here but in "single digit months" at this point). There could be some emergent complexity on top of the known complexity of these additional stubs, right at the projects "complexity peak" with the v2 build/package now running alongside in the same branch (decisions there a related but separate topic/issue).

Adding these type stubs, if they require manual updating (even with AI assistance), shifts some burden and pain from users (who might then be able to resolve imports with more ease) and onto maintainers / developers of the platform.

Also acknowledging and appreciating the amount of work you have put into this, and that I could/should have formed this opinion sooner, I'm currently leaning against merging this unless stub generation is completely automated (deterministically - and not via LLM's), as it will be with `pyo3-stub-gen` for the v2 package. I also wouldn't want you to investing further effort into attempting to get automatic stubs working either.

We can still discuss the above though.

woung717:
@cjdsellers I understand your situation. It would be nice to know how long the current Cython base will last. I thought the Rust migration was going on a separate branch, and that no more massive changes would happen on the Cython codebase.

Once I've also tried generating stubs using the Cython parser, which was successful for the most part, but implementing the type inference for type hints and imports would take too much time and effort for the limited timespan of the codebase (as you mentioned).

In any case, it's good to hear that the complete Rust migration will be finished soon. Kudos!

woung717:
And, it is okay for me as I can use for myself, anyway. (And that was my initial motivation.) Then, is it okay to maintain the stubs only in my own repository, so that people who want to download and configure it manually?

sunlei:
@woung717

Maybe you could build your own package, something like:
- https://github.com/typeddjango/django-stubs
- https://github.com/python/typeshed

cjdsellers:
@woung717 

Thanks for your understanding.

Yes, at least you can use your work locally. All I would request is, lets not publish anything official with the `nautilus` name to PyPI, but keeping it based from a repo you are working on would be great :+1:

PR #2822, opened 2025-08-01
Refine request of instruments

faysou:
# Pull Request

## Summary

- Add "keep_last_instrument" as params for RequestInstruments
- Allow to query Instruments with a ts_init bigger than now if end is not explicitly specified
- Improve make pytest to display failed tests when cancelling early
- Reduced the time of the longest running tests for faster dev iterations

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [x] Maintenance / chore

## Documentation

## Release notes

## Testing


nautilus_trader/adapters/betfair/parsing/core.py
```diff
@@ -133,13 +133,13 @@ def betting_instruments_from_file(
                 if mc.market_definition:
                     market_def = msgspec.structs.replace(mc.market_definition, market_id=mc.id)
                     mc = msgspec.structs.replace(mc, market_definition=market_def)
-                    instruments = make_instruments(
+                    new_instruments = make_instruments(
```
cjdsellers, line 136: Good catch

nautilus_trader/common/actor.pyx
```diff
@@ -2570,13 +2572,10 @@ cdef class Actor(Component):
             The instrument ID for the request.
         start : datetime, optional
             The start datetime (UTC) of request time range.
-            If ``None`` then will be replaced with the current UTC time.
-            Should be left-inclusive (start <= value), but inclusiveness is not currently guaranteed.
-            The `start == end` case is allowed to request only the initial state.
+            The inclusiveness depends on individual data client implementation.
```
cjdsellers, line 2575: I think these are more accurate param descriptions right now. We have not fully committed to consistent inclusiveness across all integrations, and are also not achieving it just yet even if that might be a future goal.

nautilus_trader/common/actor.pyx
```diff
@@ -2596,34 +2595,29 @@ cdef class Actor(Component):
...
+            If `end` is not `None` and > current timestamp (now).
         ValueError
-            If `start` is > `end`.
+            If `start` and `end` are not `None` and `start` is >= `end`.
         TypeError
             If `callback` is not `None` and not of type `Callable`.
 
-        Notes
-        -----
-        When querying from a catalog, the engine will return the most recent instrument
```
cjdsellers, line 2609: I agree that it's more intuitive that we don't apply this post processing by default. We should keep the default request as a basic `start` to `end` time range query, as a user may just want any _changes_ to instrument definitions and not necessarily always include the latest instrument prior to `start`.

Having said that, it _could_ be useful to be able to include these "prior to the query range active instruments", but I think we should add this functionality when necessary, and it should be controlled by an additional param and not be the default.

nautilus_trader/data/engine.pyx
```diff
@@ -1547,19 +1543,6 @@ cdef class DataEngine(Component):
         self._log.warning(f"Cannot handle request: no client registered for '{request.client_id}', {request}")
 
     cpdef void _query_catalog(self, RequestData request):
-        """
```
cjdsellers, line 1550: Small point on code style - we typically don't use docstrings for documenting private methods, as they are meant for public facing docs. Any docs for these should be regular comments.

nautilus_trader/data/engine.pyx
```diff
@@ -1583,24 +1566,17 @@ cdef class DataEngine(Component):
         # We assume each symbol is only in one catalog
         for catalog in self._catalogs.values():
             if isinstance(request, RequestInstruments):
-                # For instruments requests, we omit the start parameter to ensure we find
```
cjdsellers, line 1586: As above, there are certain cases where handling a range query in this way is logically correct. However, this shouldn't be the default (it's correct to remove this comment to align with standard/previous behavior).

nautilus_trader/data/engine.pyx
```diff
@@ -1646,34 +1622,26 @@ cdef class DataEngine(Component):
                 f"data[-1].ts_init={data[-1].ts_init}, {ts_now=}",
             )
 
-        # Special handling for instruments requests to ensure proper filtering
-        if isinstance(request, RequestInstruments) and request.start is not None:
```
cjdsellers, line 1650: Correct to remove this post processing for now. We can bring it back later based on a parameter the user can control.

nautilus_trader/data/engine.pyx
```diff
@@ -1646,34 +1622,26 @@ cdef class DataEngine(Component):
...
+        if isinstance(request, RequestInstrument):
             if len(data) == 0:
                 self._log.error(f"Cannot find instrument for {request.instrument_id}")
                 return
 
-            # Input instruments are sorted by ts_init
-            data = data[len(data)-1]
+            data = data[-1]
+        elif isinstance(request, RequestInstruments):
+            keep_last_instrument = request.params.get("keep_last_instrument", True)
```
cjdsellers, line 1632: `only_latest` or `only_last` might be a more intuitive param name.

nautilus_trader/persistence/catalog/parquet.py
```diff
@@ -325,14 +325,6 @@ def _write_chunk(
         directory = self._make_path(data_cls=data_cls, identifier=identifier)
         self.fs.mkdirs(directory, exist_ok=True)
 
-        if isinstance(data[0], Instrument):
```
cjdsellers, line 328: Removing this is correct, as we _do_ want to be able to write multiple versions of an instrument - as they can change over time.

nautilus_trader/test_kit/stubs/persistence.py
```diff
@@ -77,6 +77,9 @@ def schema():
     @staticmethod
     def news_events() -> list[NewsEventData]:
         df = pd.read_csv(TEST_DATA_DIR / "news_events.csv")
+        # Use only first 5000 rows for faster testing (vs original 86,985 rows)
```
cjdsellers, line 80: We can probably move these explanatory comments into the PR description (for the next one).

cjdsellers (approved):
Thanks @faysou 👌

PR #2848, opened 2025-08-13
Fix venue_order_id handling for Polymarket order status request

DeirhX:
# Pull Request

## Summary

There are cases when reconciling polymarket orders doesn't fill in the command.venue_order_id, but it is obtained from the cache (see line 485). However, this case appears not to have been thought of on line 504, which uses the immutable command.venue_order_id (despite the fact that a new venue_order_id variable was created on line 485), leading to an exception even if we got the venue from the cache.

This PR allows the query to go through in case we had to get the venue_order_id from the cache.

## Type of change

- [X] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

Tested on my strategy that was previously randomly failing due to exceptions here (usually happening when there were some slow orders in-flight)

### DIFF
--- a/nautilus_trader/adapters/polymarket/execution.py
+++ b/nautilus_trader/adapters/polymarket/execution.py
@@ -480,7 +480,8 @@ class PolymarketExecutionClient(LiveExecutionClient):
     ) -> OrderStatusReport | None:
         await self._maintain_active_market(command.instrument_id)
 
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @DeirhX 

Good catch and thanks for the fix!

PR #2867, opened 2025-08-18
Relax conditions on start and end of instrument requests in adapters

faysou:
# Pull Request

## Summary

- Removed conditions on request instrument in adapters
- Put start as required parameter for time series requests
- Put start and end as Option in rust response messages to facilitate internal use to transport data in adapters

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing


stastnypremysl:
Hi, @faysou.

This change present in my opinion regression to an old behavior, where `start` and `end` could be None inside the internals. 

This is optional and up to @cjdsellers, but None for `start` and `end` are newly reserved for unified defaults across all adapters. `request_*` will provide automatically sensible defaults, which will be same across backtest and live trading, and across all venues. I can't also remember, whether there wasn't some bug, I was preparing fix for by this - you might fill me if you remember, since you were at the discussions.

I recommend not to merge this, as this also present the exact opposite of we agreed upon and presents discrepancy between backtest and live trading.

cjdsellers:
Hi @faysou

There is a difference here between `start` and `end` requirements for the actor request methods, and for the request/response messages.

I see we provide a `None` default for many `start` parameters in `Actor` (e.g. [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/common/actor.pyx#L2899)) when we later validate it as `not_none` [here](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/common/actor.pyx#L2545).

I recommend we keep `start` as required for both actor request methods _and_ the request/response messages, _except_ for the instrument requests - where that becomes clunky UX, because an instrument can be active prior to start, and so a user would probably expect it to be in the range - and how to determine `start` for "latest"? However, we only have one `DataResponse` type in Cython, and so can't have `start` as required in this case (so needs a little more discussion?).

@stastnypremysl The previous philosophy and implementation of this allowed for optional `start` and `end` which would then use the defaults of integration API's - this was much easier to reason about by reading the integration API docs (which is recommended), rather than NT then layering on some attempt at aligning every integration.

I understand that you favor predictability between backtest and live, but backtesting is against a simulated venue with some config options, and we have many live adapters with different API behaviors. Also, I thought you had a reason for requiring `start` and `end` on the messages for a future PR, but we still don't know what the reason for this was.

So I think we should reconsider the previous change to requiring `end` for the request message itself, where we set `end` to the current time now if it's `None`. I think `end` as `None` just means until the end of available data?

faysou:
I've just made start required in the actor methods for market data requests (ie. not for instruments). Regarding making start and end required in responses, especially rust ones, I think start and end as optional makes more sense as sometimes responses don't stem from requests when used internally in adapaters from what I've seen.

faysou:
I've had a number of issues because of assuming that start and end were required for request_instrument(s). I had only agreed to restrictions on start and end for non instrument requests when you did your PR @stastnypremysl. I've reverted the restrictions on request instrument in this PR and another PR, but the restrictions on start and end for the request of market data are unchanged.

stastnypremysl:
Hi, @faysou. Could you please specify the issue?

faysou:
Most brokers don't require start and end for requesting instruments. Also the way you modified the query of instruments with the catalog before lead to issues before that I had to fix. For example you inverted end with start. You probably had a logic in mind but it was counterintuitive compared to how requests work for other data types.

stastnypremysl:
@faysou @cjdsellers I am little sorry, I have headache again. I will understand, if you merge it sooner as now I am unusable. We might discuss a revert later, when I am well (if my Discord ban for reporting unspecified OWASP will be somehow resolved - I couldn't give a proof as it might have lead to immediate security breach of GitHub, if it is not false possitive).

faysou:
The changes in this PR are quite basic and my goal is for the library to be as good as possible.

PR #2879, opened 2025-08-22
Fix Bybit positions pagination to handle more than 20 positions

scoriiu:
- Add pagination support with 200 item limit per request (Bybit's maximum)
- Update BybitPositionResponseStruct to use BybitListResultWithCursor for pagination support
- Implement cursor-based pagination loop in query_position_info method
- Ensures all positions are fetched when user has more than 20 positions

Previously, the API would only return the default 20 positions. This fix ensures all positions are retrieved by paginating through the results.


cjdsellers:
Many thanks @scoriiu

PR #2880, opened 2025-08-22
Standardize zero PnL as Money instead of None when exchange rate missing

v9ai:
# Fix: Return zero PnL instead of None when exchange rate missing

## Summary
Fixes null pointer dereference in portfolio PnL calculation when exchange rate data is unavailable.

## Changes
- **`crates/portfolio/src/portfolio.rs`**: Return `Some(Money::new(0.0, currency))` instead of `None` in `calculate_realized_pnl`
- **`crates/portfolio/src/tests.rs`**: Added test for missing exchange rate scenario

## Problem/Solution
Previously returned `None` when exchange rate conversion failed → potential null pointer dereferences.
Now returns zero PnL as safe fallback while maintaining retry logic for pending calculations.

**Risk**: Low - defensive improvement, only affects edge case, maintains backward compatibility.

### DIFF
--- a/crates/portfolio/src/portfolio.rs
+++ b/crates/portfolio/src/portfolio.rs
@@ -1007,13 +1007,14 @@ impl Portfolio {
                     xrate
                 } else {
                     log::error!(
-                        // TODO: Improve logging
-                        "Cannot calculate realized PnL: insufficient data for {}/{}",
+                        "Cannot calculate realized PnL: insufficient exchange rate data for {}/{}, marking as pending calculation",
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hi @nicolad 

Thanks for highlighting this. There are still some design decisions to be made on the Rust side, considering this further since it's been highlighted here - it's probably nicer to be passing around a `Money` domain "object" to represent zero PnL, rather than `None`. So I think this is a good change.

PR #2882, opened 2025-08-23
Improve security for Credential struct

nicolad:
Harden BitMEX credential handling and improve ergonomics.

* Store `api_secret` as `Box<[u8]>` and derive `ZeroizeOnDrop` (skipping `api_key`) so secrets are wiped from memory on drop.
* Implement a custom `Debug` that redacts the secret and add `Clone` for `Credential`.
* `sign()` behavior and public constructor remain unchanged; no breaking API changes.
* Adds `zeroize` (with `alloc`/`zeroize_derive`) and a unit test asserting debug redaction.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,15 +1,14 @@
+# Global flags that work across all platforms
 [target.'cfg(all())']
 rustflags = [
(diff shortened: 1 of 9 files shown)


cjdsellers (approved):
👌 thanks @nicolad

PR #2895, opened 2025-08-26
Use precise integer arithmetic in decode_multiplier to prevent precision loss

nicolad:
The decode_multiplier function was using floating-point arithmetic (value as f64 / 1e9)
followed by string formatting, which introduced precision loss for financial calculations.
This could lead to incorrect multipliers affecting P&L, position sizing, and risk calculations.

Changes:
- Replace floating-point division with exact integer arithmetic
- Handle fixed-point scaling (1e-9) using integer operations only
- Preserve exact decimal representation without rounding errors
- Add comprehensive test coverage for edge cases
- Remove unused DATABENTO_FIXED_SCALAR constant

The fix ensures multipliers are decoded with perfect precision, maintaining consistency
with other decode functions that use exact integer arithmetic for financial values.

### DIFF
--- a/crates/adapters/databento/src/decode.rs
+++ b/crates/adapters/databento/src/decode.rs
@@ -39,8 +39,6 @@ use super::{
     types::{DatabentoImbalance, DatabentoStatistics},
 };
 
-const DATABENTO_FIXED_SCALAR: f64 = 1_000_000_000.0;
-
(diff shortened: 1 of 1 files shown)


cjdsellers:
Thanks for the improvement @nicolad 👌

PR #2898, opened 2025-08-26
Fix Bybit balance precision errors for high-value tokens

scoriiu:
When parsing balances from Bybit, tokens with very high values (like MOG with trillions in circulation) can have floating point precision issues where the API returns values that don't exactly add up (total - locked != free).
This fix uses raw integer arithmetic to calculate the free balance, ensuring the AccountBalance invariant is always satisfied.


cjdsellers:
Thanks for the fix @scoriiu :pray:

PR #2903, opened 2025-08-27
Improve Bybit balance precision by avoiding float conversion

scoriiu:
Parse balance values directly from strings to raw integers without intermediate float conversion. This preserves full decimal precision for high-value tokens like MOG.


cjdsellers:
Thanks @scoriiu, good changes

PR #2904, opened 2025-08-27
Fix Bybit WebSocket balance precision errors

scoriiu:
Extract string-to-raw conversion to centralized parsing utility. Fixes MOG and other high-value token balance errors from WebSocket updates by avoiding float conversion entirely.


cjdsellers:
Thanks @scoriiu :ok_hand:

PR #2905, opened 2025-08-27
Refactor SpreadQuoteAggregator

faysou:
Moved management of spread quote aggregators from data engine to BacktestMarketDataClient.


cjdsellers:
Thanks @faysou, this is a good change because it's backtest specific.

PR #2916, opened 2025-08-30
Implement Hyperliquid EIP-712 signing and nonce management

nicolad:
- Add signing module with EIP-712 signer for L1 and user actions
- Implement thread-safe nonce management with replay protection
- Add action canonicalization for Hyperliquid protocol

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,6 +115,7 @@ futures = { version = "0.3.31", default-features = false, features = ["std", "as
 futures-util = "0.3.31"
 heck = "0.5.0"
 hex = "0.4.3"
+hmac = "0.12.1"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -115,6 +115,7 @@ futures = { version = "0.3.31", default-features = false, features = ["std", "as
 futures-util = "0.3.31"
 heck = "0.5.0"
 hex = "0.4.3"
+hmac = "0.12.1"
```
cjdsellers: Lets continue to use `aws-lc-rs` for cryptography as its actively maintained and undergoing FIPS 140-3 validation.

v9ai: Done, thank you.

PR #2943, opened 2025-09-07
Fix metadata retention on deduplication

ms32035:
# Pull Request

## Summary

Unfortunately looks like pyarrow group by is losing the metadata attached to the table, so this patch fixes the issue

## Type of change

- [X] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

- [X] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)

## Release notes

## Testing

### DIFF
--- a/nautilus_trader/persistence/catalog/parquet.py
+++ b/nautilus_trader/persistence/catalog/parquet.py
@@ -687,9 +687,11 @@ class ParquetDataCatalog(BaseDataCatalog):
 
     @staticmethod
     def _deduplicate_table(table: pa.Table) -> pa.Table:
-        res = table.group_by(table.column_names).aggregate([])
-        res = res.replace_schema_metadata(table.schema.metadata)
-        return res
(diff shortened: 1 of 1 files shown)


faysou:
```python
def _deduplicate_table(table: pa.Table) -> pa.Table:
    deduped_data_table = table.group_by(table.column_names).aggregate([])

    result_table = pa.Table.from_arrays(
        deduped_data_table.columns, 
        schema=table.schema
    )

    return result_table
```

This function should produce correct column metadata, I think it would be better.

ms32035:
@faysou I wouldn't say so, because what you're proposing recreates the pyarrow table, whereas the current solution only creates a shallow copy so no memory rewrite for data occurs

faysou:
But the column metadata is wrong with your solution. Apparently from_arrays doesn't create a new copy of the data, just a copy of pointers, so it should be fine.

PR #2946, opened 2025-09-08
Fix historical data request race condition in DataEngine

lisiyuan656:
# Pull Request

## Summary

  Here's a breakdown of the issue:

   1. When you request historical bars that span both the catalog and the exchange, the DataEngine creates a "query group" to manage the multiple data responses (one from the catalog, and one or more from the exchange
      for the missing data).

   2. The DataEngine's _handle_query_group_aux method in nautilus_trader/data/engine.pyx is responsible for merging these responses. The problem is that the final merged DataResponse object incorrectly uses the start and
      end timestamps from the last partial response it receives.

   3. If the response from the exchange (for the more recent, missing data) arrives after the response from your local catalog, the final DataResponse will have the start and end times of only the exchange data.

   4. This incorrect time range is then used by the _handle_aggregated_bars_aux method, which causes it to only process the bars within that smaller, later time window, effectively discarding the data that was correctly
      read from your local catalog.

  Here are the changes I'll make to fix this:

   1. In nautilus_trader/data/engine.pyx, add a new dictionary _query_group_requests to the DataEngine's __init__ method to store the original requests.
   2. Modify _new_query_group to store the original RequestData object.
   3. Update _handle_query_group_aux to use the start and end from the stored original request in the final merged response.
   4. Update the _reset method to clear the new dictionary.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/nautilus_trader/data/engine.pxd
+++ b/nautilus_trader/data/engine.pxd
@@ -100,6 +100,7 @@ cdef class DataEngine(Component):
     cdef readonly dict[str, SnapshotInfo] _snapshot_info
     cdef readonly dict[UUID4, int] _query_group_n_responses
     cdef readonly dict[UUID4, list] _query_group_responses
+    cdef readonly dict[UUID4, RequestData] _query_group_requests
 
     cdef readonly dict[InstrumentId, str] _topic_cache_deltas
     cdef readonly dict[InstrumentId, str] _topic_cache_quotes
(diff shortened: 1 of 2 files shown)


faysou:
I agree, there's a bug. The fix looks fine to me.

cjdsellers:
Thanks for the fix @lisiyuan656

PR #2948, opened 2025-09-09
Update IB adapter to fetch orders from all clients using reqAllOpenOrders

dinana:
## Summary
This PR updates the Interactive Brokers adapter to align with other broker implementations by fetching ALL open orders from the broker, not just orders placed through the current API client session.

## Changes
- Changed `get_open_orders()` method to use `reqAllOpenOrders` instead of `reqOpenOrders`
- Updated method documentation to clarify the new behavior
- Updated corresponding test to verify the correct API method is called

## Context
Currently, the IB adapter only fetches orders placed through the current API client (client ID 0). This is inconsistent with how other broker adapters work:
- **Binance**: Uses `query_all_orders()` to fetch all orders
- **Bybit**: Uses `query_order_history()` to fetch all orders  
- **OKX**: Fetches all orders with configurable filters
- **BitMEX**: Fetches all orders using REST API

This inconsistency prevents proper order synchronization when using multiple trading interfaces with Interactive Brokers.

## Technical Details
The change from `reqOpenOrders` to `reqAllOpenOrders` means the adapter now fetches orders from:
- All API clients (regardless of client ID)
- TWS/IB Gateway GUI
- All other trading interfaces connected to the account

**Important limitation**: The IB API's client ID architecture has inherent limitations. When using client ID 0 (as NautilusTrader does), the adapter can see:
- Orders placed by client ID 0 itself
- Orders placed through TWS GUI (client ID -1)
- Orders from other API clients **only** when using `reqAllOpenOrders`

Without this change, orders from other API clients (client IDs 1, 2, etc.) would not be visible to NautilusTrader.

## Testing
- All existing IB adapter tests pass
- Updated the `test_get_open_orders` test to verify `reqAllOpenOrders` is called
- Manually tested with mock orders from different client IDs to verify filtering works correctly

## Breaking Changes
None. This change maintains backward compatibility while extending functionality.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -83,6 +83,7 @@ class InteractiveBrokersClient(
         host: str = "127.0.0.1",
         port: int = 7497,
         client_id: int = 1,
+        fetch_all_open_orders: bool = False,
     ) -> None:
         super().__init__(
             clock=clock,
@@ -97,6 +98,7 @@ class InteractiveBrokersClient(
         self._host = host
         self._port = port
         self._client_id = client_id
+        self._fetch_all_open_orders = fetch_all_open_orders
 
         # TWS API
         self._eclient: EClient = EClient(
--- a/nautilus_trader/adapters/interactive_brokers/client/order.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/order.py
@@ -39,6 +39,8 @@ class InteractiveBrokersClientOrderMixin(BaseMixin):
 
     """
 
+    _fetch_all_open_orders: bool
+
     def place_order(self, order: IBOrder) -> None:
         """
         Place an order through the EClient.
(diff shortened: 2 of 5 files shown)


faysou:
I think that this behaviour should be optional, there could be cases where several instance of nautilus for different underlyings are used and don't care about the orders of other client ids.

dinana:
The IBKR account (and hence the username and password) are the same for all client IDs as they're input in the ibgateway which can only handle a singular account, the experience is disfunctional as requesting positions or portfolio would get a whole account view, but if you would request orders for said positions it wouldn't be retrieved if you had entered them from another client id. 
Also if you want to employ different client IDs for different strategies to avoid duplication of order IDs for instance or for any other reason you'd have similar discrepencies.
This is also consistent with other broker adapters for other brokers/providers.

faysou:
But still from what you say, several client ids can coexist and manage their own positions which can be useful ?

dinana:
No, they will get positions for all client IDs, the reqPositions function doesn't have the same limitation as requesting orders. Hence the descripency.

faysou:
I think you can make the new behaviour optional, it could still be beneficial to have several client ids receiving their own order updates, even if they receive all positions.

cjdsellers:
Thanks for the contribution @dinana

mytdm88:
Lib\site-packages\nautilus_trader\adapters\interactive_brokers\client\client.py", line 26, in <module>
    from ibapi.commission_report import CommissionReport
ModuleNotFoundError: No module named 'ibapi.commission_report'  after upgraded the ibapi to the latest stable version 10.37.2 today.   So may I know which ibapi version should be used? Thanks.

faysou:
@cjdsellers the ibapi should be downgraded to what it was before, we don't need to ugrade yet as discussed previously.

mytdm88:
Hi All, I followed this https://pypi.org/project/nautilus_ibapi/  and installed the 10.37.2.  

The script can start but now I am getting the "2025-09-15T11:33:29.139413000Z [ERROR] DISCONNECT-DEMO.InteractiveBrokersClient-001: Connection failed
TypeError(make_msg() missing 2 required positional arguments: 'useRawIntMsgId' and 'text')
  File "..\Lib\site-packages\nautilus_trader\adapters\interactive_brokers\client\connection.py", line 57, in _connect
    await self._send_version_info()
  File "..\Lib\site-packages\nautilus_trader\adapters\interactive_brokers\client\connection.py", line 166, in _send_version_info
    msg = comm.make_msg(v100version)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^

I may downgrade to ibapi 10.30.1 if I need to.

PR #2954, opened 2025-09-11
Use filesystem in catalog for isfile and isdir

limx0:
## Summary

Use catalog filesystem instead of `Path`, which only works when `self.fs = file`

## Type of change
- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing


cjdsellers:
Thanks for the fix @limx0!

PR #2962, opened 2025-09-16
Reset BarAggregator state to avoid live data conflict when requesting historical data

lisiyuan656:
# Pull Request

## Summary

This bug happens when subscribe_bars() and request_aggregated_bars() are called at the same time. Since the DataEngine utilize the same aggregator for live and historical data, if the request_aggregated_bars() take time longer than a new live bar arrives, the aggregator would store ts_last/_last_close from the live bar, which prevents the historical data to be correctly aggregated.
This fix resets the aggregator's state at the beginning of a historical batch (start_batch_update) erases the "poisoned" state from the live feed, ensuring the historical data can be processed correctly.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/nautilus_trader/data/aggregation.pyx
+++ b/nautilus_trader/data/aggregation.pyx
@@ -312,6 +312,9 @@ cdef class BarAggregator:
         self.is_running = False # is_running means that an aggregator receives data from the message bus
 
     def start_batch_update(self, handler: Callable[[Bar], None], uint64_t time_ns) -> None:
+        self._builder.reset()
+        self._builder.ts_last = 0
+        self._builder._last_close = None
         self._batch_mode = True
         self._handler_backup = self._handler
         self._handler = handler


faysou:
You should call subscribe bars as a callback called after request aggregated bars, it doesn't make sense to have request aggregated bars and subscribe bars called at the same time in a live setting. In a backtest setting calls are sequential so it's ok.

lisiyuan656:
> You should call subscribe bars as a callback called after request aggregated bars, it doesn't make sense to have request aggregated bars and subscribe bars called at the same time in a live setting. In a backtest setting calls are sequential so it's ok.

How should i do this in a live setting? The callback signature seems to be Callable[[UUID4], None], which is not the same as subscribe_bars().
screenshot_1758029910
Also if i request several different bar_types just like the [example](https://github.com/nautechsystems/nautilus_trader/blob/8eea03c8847ec6d1fbf7e7bd1d8e316d890f9056/examples/backtest/notebooks/databento_test_request_bars.py#L182) here, how can i subscribe the different bars in a single callback?

faysou:
the callback can be any function that takes a uuid as argument, you can do anything with it.

lisiyuan656:
Ok, i got it. Thank you for the help! I'll close the pr for now.

Issue #2981, opened 2025-09-22
[Hyperliquid] Inital integration and Rust crate


cjdsellers:
- [x] REST API `HttpClient` credentials, signing
- [x] REST API models for data
- [x] REST API models for execution
- [x] WebSocket API credentials, signing
- [x] InstrumentProvider / definitions parsing
- [x] `DataClient` in Python
- [x] `DataClient` in Rust
- [x] `ExecutionClient` in Python
- [x] `ExecutionClient` in Rust
- [x] Basic market and limit orders
- [x] Conditional / advanced orders
- [x] Execution reconciliation

PR #2992, opened 2025-09-25
Implement Hyperliquid InstrumentProvider / definitions parsing

nicolad:

### DIFF
--- a/crates/adapters/hyperliquid/src/data/mod.rs
+++ b/crates/adapters/hyperliquid/src/data/mod.rs
@@ -13,8 +13,6 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
 
-//! Live market data client implementation for the Hyperliquid adapter.
-
 use std::{
     future::Future,
     sync::{
@@ -46,12 +44,9 @@ use nautilus_core::{
 };
 use nautilus_data::client::DataClient;
 use nautilus_model::{
-    currencies::CURRENCY_MAP,
     data::Data,
-    enums::CurrencyType,
-    identifiers::{ClientId, InstrumentId, Symbol, Venue},
-    instruments::{CryptoPerpetual, CurrencyPair, Instrument, InstrumentAny},
-    types::{Currency, Price, Quantity},
+    identifiers::{ClientId, InstrumentId, Venue},
+    instruments::{Instrument, InstrumentAny},
 };
 use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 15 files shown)


crates/adapters/hyperliquid/src/data/mod.rs
```diff
@@ -132,22 +230,51 @@ impl HyperliquidDataClient {
...
-        // For now, return empty list as placeholder
-        let meta = self
-            .http_client
-            .info_meta()
-            .await
-            .context("failed to load meta information")?;
+        let mut instruments = Vec::new();
+
+        // Load perpetual instruments
+        match self.http_client.get_perp_meta().await {
```
cjdsellers: We can keep this client method higher level (use `request_instruments`), rather than going for a more direct/inner method, i.e. the calls to the parsing function can go down inside the REST client.

nautilus_trader/adapters/hyperliquid/providers.py
```diff
@@ -15,84 +15,268 @@
...
+    Load spot and perpetual instruments from Hyperliquid's REST ``/info`` API.
     """
 
     def __init__(
         self,
-        client: Any | None = None,  # TODO: Replace with actual HyperliquidHttpClient when available
+        client: HyperliquidHttpClient,
         config: InstrumentProviderConfig | None = None,
+        *,
+        include_spot: bool = True,
```
cjdsellers: We typically have some kind of `HyperliquidProductType` or instrument type enum list for this rather than separate bools.

cjdsellers (commented):
Thanks @nicolad, I can see it coming together nicely

crates/adapters/hyperliquid/src/data/mod.rs
v9ai: Thanks, appreciate.

nautilus_trader/adapters/hyperliquid/providers.py
v9ai: Done, thanks.

crates/adapters/hyperliquid/src/python/mod.rs
```diff
@@ -24,6 +24,8 @@ use pyo3::prelude::*;
 
 /// Loaded as `nautilus_pyo3.hyperliquid`.
 #[pymodule]
-pub const fn hyperliquid(_: Python<'_>, _m: &Bound<'_, PyModule>) -> PyResult<()> {
+pub fn hyperliquid(m: &Bound<'_, PyModule>) -> PyResult<()> {
+    m.add_class::<http::PyHyperliquidHttpClient>()?;
```
cjdsellers: We don't need the Py prefix for the names we export to Python.

crates/adapters/hyperliquid/src/python/http.rs
```diff
@@ -12,3 +12,184 @@
...
+
+/// Python binding for HyperliquidInstrumentDef.
+#[pyclass(name = "HyperliquidInstrumentDef")]
+#[derive(Clone, Debug)]
+pub struct PyHyperliquidInstrumentDef {
+    pub(crate) inner: HyperliquidInstrumentDef,
+}
+
+#[pymethods]
+impl PyHyperliquidInstrumentDef {
```
cjdsellers: We don't need to export the raw types out to Python, only use in Rust for converting into Nautilus types.

crates/adapters/hyperliquid/src/python/http.rs
```diff
@@ -12,3 +12,184 @@
...
+
+use crate::http::{
+    client::HyperliquidHttpClient,
+    parse::{HyperliquidInstrumentDef, parse_perp_instruments, parse_spot_instruments},
+};
+
+/// Python binding for HyperliquidHttpClient.
+#[pyclass(name = "HyperliquidHttpClient")]
+#[derive(Debug)]
+pub struct PyHyperliquidHttpClient {
```
cjdsellers: It's gone for an inner-outer pattern here which isn't necessary - see other adapters (`okx`, `bitmex`, etc).

crates/adapters/hyperliquid/src/http/client.rs
```diff
@@ -187,6 +188,77 @@ impl HyperliquidHttpClient {
         serde_json::from_value(response).map_err(Error::Serde)
     }
 
+    /// Get complete perpetuals metadata.
+    pub async fn get_perp_meta(&self) -> Result<crate::http::models::PerpMeta> {
```
cjdsellers: The fully qualified `PerpMeta` isn't typical of what we're doing elsewhere.

crates/adapters/hyperliquid/src/data/mod.rs
```diff
@@ -132,22 +227,28 @@ impl HyperliquidDataClient {
...
-            .await
-            .context("failed to load meta information")?;
-
-        tracing::debug!(
-            "loaded {count} assets from Hyperliquid meta",
-            count = meta.universe.len()
-        );
+        let mut instruments = Vec::new();
+
+        match self.http_client.request_instruments().await {
```
cjdsellers: This should return `Vec<InstrumentAny>` so we can still push even more down into the HTTP client.

crates/adapters/hyperliquid/src/python/mod.rs
v9ai: Done, thanks.

crates/adapters/hyperliquid/src/python/http.rs
v9ai: Done, thank you.

crates/adapters/hyperliquid/src/python/http.rs
v9ai: Done, thx.

crates/adapters/hyperliquid/src/http/client.rs
v9ai: Appreciate, done.

crates/adapters/hyperliquid/src/data/mod.rs
v9ai: Change applied, thank you,

crates/adapters/hyperliquid/src/data/mod.rs
```diff
@@ -132,22 +130,24 @@ impl HyperliquidDataClient {
...
-            .http_client
-            .info_meta()
-            .await
-            .context("failed to load meta information")?;
-
-        tracing::debug!(
-            "loaded {count} assets from Hyperliquid meta",
-            count = meta.universe.len()
-        );
+        let instruments = match self.http_client.request_instruments().await {
```
cjdsellers, line 133: If you make it `anyhow::Result<Vec<InstrumentAny>>` then you could change to `self.http_client.request_instruments().await?` and the caller can handle the error how they like.

v9ai: Thanks for expanation

PR #2999, opened 2025-09-29
Implement Hyperliquid DataClient in Rust

nicolad:

### DIFF
--- a/crates/adapters/hyperliquid/src/data/mod.rs
+++ b/crates/adapters/hyperliquid/src/data/mod.rs
@@ -13,26 +13,22 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
 
-use std::{
-    future::Future,
-    sync::{
-        Arc, RwLock,
-        atomic::{AtomicBool, Ordering},
-    },
+use std::sync::{
+    Arc, RwLock,
+    atomic::{AtomicBool, Ordering},
 };
 
 use ahash::AHashMap;
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, anyhow};
+use chrono::{DateTime, Utc};
 use nautilus_common::{
     messages::{
         DataEvent,
         data::{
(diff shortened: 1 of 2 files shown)


crates/adapters/hyperliquid/src/data/mod.rs
```diff
@@ -103,63 +113,53 @@ impl HyperliquidDataClient {
...
+            instrument_refresh_active: false,
         })
     }
 
     fn venue(&self) -> Venue {
         *HYPERLIQUID_VENUE
     }
 
-    #[allow(dead_code)]
-    fn send_data(sender: &tokio::sync::mpsc::UnboundedSender<DataEvent>, data: Data) {
```
cjdsellers, line 114: We do still need something like this to get data back to the async runner.

crates/adapters/hyperliquid/src/data/mod.rs
```diff
@@ -44,26 +40,32 @@ use nautilus_core::{
...
-    },
-    config::HyperliquidDataClientConfig,
+    common::consts::HYPERLIQUID_VENUE, config::HyperliquidDataClientConfig,
     http::client::HyperliquidHttpClient,
-    websocket::client::HyperliquidWebSocketClient,
 };
 
+// Placeholder for WebSocket client until it's implemented
+#[derive(Debug)]
+pub struct HyperliquidWebSocketClient;
```
cjdsellers: Could import the real thing now and stub out any required methods if they're missing - interface looks correct below.

crates/adapters/hyperliquid/src/data/mod.rs
v9ai: Added back, thanks.

crates/adapters/hyperliquid/src/data/mod.rs
v9ai: Changed, thanks.

PR #3003, opened 2025-09-30
Implement Hyperliquid ExecutionClient in Python

nicolad:


nautilus_trader/adapters/hyperliquid/execution.py
```diff
@@ -45,19 +45,19 @@
 
 
 if TYPE_CHECKING:
```
cjdsellers: We can remove the `TYPE_CHECKING` as this Python layer is rather temporary and doesn't need it as either a typing workaround or perf optimization.

v9ai: Done, thank you.

PR #3013, opened 2025-10-02
Implement Hyperliquid ExecutionClient in Rust

nicolad:
Related: https://github.com/nautechsystems/nautilus_trader/issues/2981

### DIFF
--- a/crates/adapters/hyperliquid/src/execution/mod.rs
+++ b/crates/adapters/hyperliquid/src/execution/mod.rs
@@ -13,36 +13,336 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
 
-//! Hyperliquid execution adapter for Rust Trading Toolkit.
-//!
-//! This module contains the implementation of an execution adapter that provides
-//! connectivity to the Hyperliquid exchange for order management and trade execution.
-//!
-//! ## Overview
-//!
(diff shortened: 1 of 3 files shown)


crates/adapters/hyperliquid/src/execution/mod.rs
```diff
@@ -13,36 +13,336 @@
...
+    http::client::HyperliquidHttpClient,
+    websocket::client::HyperliquidWebSocketClient,
+};
+
+#[derive(Debug)]
+pub struct HyperliquidExecutionClient {
+    core: ExecutionClientCore,
+    config: HyperliquidExecClientConfig,
+    http_client: HyperliquidHttpClient,
+    ws_client: Option<HyperliquidWebSocketClient>,
```
cjdsellers, line 51: I don't think the `ws_client` needs to be optional, we can init at the same time as the HTTP client.

cjdsellers (approved):
Good progress @nicolad, thank you.

crates/adapters/hyperliquid/src/execution/mod.rs
v9ai: Will be refactored in a follow-up PR

cjdsellers: Thanks, I'll also standardize this across the adapters in a follow-up.

Issue #3031, opened 2025-10-04
FillModel derived extensions do not "walk the order book" when using bars data


idobz:
# Bug Report

## Confirmation

**Before opening a bug report, please confirm:**

- [v] I’ve re-read the relevant sections of the documentation.
- [v] I’ve searched existing issues and discussions to avoid duplicates.
- [v] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [v] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

### Expected Behavior

The expected behavior when using, for example, the OneTickSlippage FillModel, is for market orders to be fulfilled (putting aside the question of price).

### Actual Behavior

The actual behavior is that market orders are not fulfilled when the order book - as defined by the fill model - does not contain 'best_bid`/'best_sell' price BookOrder has 0 quantity

### Steps to Reproduce the Problem

Untar attached tar file. It contains a python file (with strategy and  main), and a small catalog.
Run the python code using an environment that has NT installed (note the version used)

When run with `FILL_MODEL = "Standard"` (the default setting in row 26) the orders get filled
When run with `FILL_MODEL = "OneTickSlippage"` (row 27) the orders are not filled.

### Specifications

- OS platform: Ubuntu
- Python version: 3.12.7
- `nautilus_trader` version: 1.221.0a20250921

[fillmodel.tar.gz](https://github.com/user-attachments/files/22698479/fillmodel.tar.gz)

cjdsellers:
Thanks for the thoughts and write up @idobz  

Now fixed from #3039.

idobz:
Hi @cjdsellers,

Will keep this channel open for a little longer, as I try to push through with the IB/Nautilus paper trading scenario. Hope that's ok

Some comments:
1. I manually place venue=None to circumvent the issues with this. Not sure what is lost by this, as I still manage to buy and sell 

2. I often get these errors:
2025-10-07T16:05:00.043941909Z [WARN] IB_LIVE-001.CX_spread: Order submitted: STOP_MARKET BUY CX  2025-10-07 16:05:00.009618458 current close: 9.065
2025-10-07T16:05:00.044863154Z [WARN] IB_LIVE-001.InteractiveBrokersClient-062: The price does not conform to the minimum price variation for this contract. (code: 110, req_id=142)
[...]

Issue #2194, opened 2025-01-08
Enhanced order-fill simulation in backtesting


stefansimik:
## Summary

Main goal of this proposal is to enhance the current `FillModel` implementation and allow complete flexibility in simulating order fills during backtesting. The proposal is based on this design proposal https://gist.github.com/stefansimik/fe9577649d8eb60dcc163830b003212d , that demonstrates the implementation approach through practical examples.

The core idea is to extend the existing L2 OrderBook-based processing with two key simulation points:
1. Market order fills - simulating how orders consume liquidity from the orderbook
2. Limit order fills - simulating partial fills when market price touches the limit price

## Background & Motivation
The current `FillModel` handles only very basic order-fill scenarios and has significant limitations in its ability to simulate more complex scenarios such as:
- Competition for liquidity
- Varying liquidity conditions across trading sessions
- Price impact for large orders
- Partial fills
- Order queue position simulation
- Complex L1 data simulation

## Proposed Solution
The proposal suggests extending the `FillModel` with two new methods that were analyzed and suggested by @cjdsellers as the best hooks for implementation:
- `.process_market_fills` - integrated at: https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/backtest/matching_engine.pyx#L1577
- `.process_limit_fills` - integrated at: https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/backtest/matching_engine.pyx#L1483

Key implementation details:
- Keep the current `FillModel` responsibility of determining if/how much of an order is filled
- Default to no-op implementations for backward compatibility

Key benefits:
1. Enables simulation of any imaginable order-fill scenario
2. Centralizes all order-fill simulation logic
3. Leverages existing L2 OrderBook processing
4. Maintains clear responsibilities of `FillModel`

## Historical Context
Currently, `FillModel` has only two primary settings (as documented in backtesting.md):

1. `prob_fill_on_limit` - Controls limit order simulation when price touches the order price
2. `prob_slippage` - Controls market order simulation and potential slippage

Importantly, all existing functionality can be easily replicated using the newly suggested model. The design proposal (see above link) includes specific examples demonstrating how to implement current behaviors, including probabilistic fills and slippage, while also enabling much more sophisticated simulation scenarios.

stefansimik:
@faysou found useful article:
https://quant.stackexchange.com/questions/38781/backtesting-market-making-strategy-or-microstructure-strategy

[...]

PR #3039, opened 2025-10-06
Add fallback to default order book for custom fill model

Hamish-Leahy:
fix(backtest): fallback to standard market fills when simulated book yields no fills

When using custom FillModel implementations like OneTickSlippageFillModel that
create simulated OrderBooks without liquidity at best bid/ask prices, market
orders would fail to fill entirely. This change adds a fallback mechanism that
reverts to the standard market fill logic when the simulated book produces no
fills, ensuring market orders always execute while still allowing custom models
to apply their intended slippage behavior when liquidity is available.

Fixes issue where OneTickSlippageFillModel prevented market order fills when
the simulated book contained no orders at best bid/ask prices (zero quantity
BookOrders), causing strategies to fail unexpectedly in backtesting.

- Modified determine_market_fills_with_simulation() in backtest engine
- Added fallback to determine_market_price_and_volume() when simulated fills empty
- Preserves existing behavior for models that provide liquidity
- No breaking changes to public API

### DIFF
--- a/nautilus_trader/backtest/engine.pyx
+++ b/nautilus_trader/backtest/engine.pyx
@@ -5022,12 +5022,17 @@ cdef class OrderMatchingEngine:
 
         if simulated_book is not None:
             # Use simulated OrderBook for fill determination
-            return simulated_book.simulate_fills(
+            fills = simulated_book.simulate_fills(
                 order,
                 price_prec=self.instrument.price_precision,
                 size_prec=self.instrument.size_precision,
                 is_aggressive=True,
             )
+            # If simulation produced no fills (e.g., custom model removed best levels),
+            # fall back to standard market logic to preserve expected behavior.
+            if not fills:
+                return self.determine_market_price_and_volume(order)
+            return fills
         else:
             # Fall back to standard logic
             return self.determine_market_price_and_volume(order)


faysou:
Thank you for this. Just for my understanding, why do fill models sometimes not produce some order book ?

Hamish-Leahy:
Thanks! In short, fill models can legitimately choose not to produce a simulated order book (returning None) to defer to the engine’s default logic for backward compatibility or when inputs are incomplete (e.g., no best bid/ask yet, halts/auctions). They can also intentionally synthesize a book without top-of-book liquidity to model specific scenarios—like the OneTickSlippageFillModel, which removes best levels to guarantee one-tick slippage. In the original issue, that design led to market orders not filling when the simulated book had no best levels; this PR adds a safe fallback to the standard market fill path when a model-produced book yields no fills, so market orders still execute while preserving the model’s intended slippage when liquidity exists.

Original Issue was Issue #3031

faysou:
ok, thank you for the explanation. but if no order book is produced by a fill model, wouldn't it be better to accept that a market order can't be filled immediately until liquidity returns ? I have observed before a live situation where a market order couldn't be filled despite a market being open, because there was no liquidity.

Or maybe that's an edge case situation and we could add a config option if needed to not default to the default behaviour if the order book is empty.

So your fix is likely good already.

Hamish-Leahy:
You are absolutely right. If a fill model produces no liquidity, deferring the market order until liquidity returns is a valid and realistic behavior and matches the live scenario you observed. For the original issue, I chose a conservative default to preserve expected fills when switching to OneTickSlippageFillModel, but I agree this could be configurable. I can add a boolean option, for example fallback_on_empty_simulation=true by default, and when set to false, if the simulated book yields no fills the order remains unfilled until liquidity appears with no fallback to the standard path. That would cover the edge case without breaking existing behavior. I will work on that later this week. But for now this fix will work well.

faysou:
@cjdsellers what do you think ? I think the current change is enough.

cjdsellers:
Hi @Hamish-Leahy 

Thank you for your detailed write up and pragmatic fix.

I agree that it's a reasonable expectation for market orders to be fulfilled by default. I also agree with @faysou that it should at least be possible - with some configuration - to simulate a market differently / more accurately, for example in the case of very illiquid instruments. Venues handle this in various ways; often an order will be rejected if there’s no resting liquidity, or available liquidity will be swept and any remainder canceled.

This PR is certainly an improvement on the current behavior, and additional config options would be a nice enhancement in the future.

idobz:
Thanks for this fix @Hamish-Leahy ,
Although I wonder at the utility of the new fill model generalization design, at least when using bars data.
My understanding is that it will only fill a market order if there 'top-of-book' liquidity is available - but in that case, what do I need a slippage model for?

faysou:
https://github.com/nautechsystems/nautilus_trader/issues/2194

See here

PR #3020, opened 2025-10-02
Remove partial bars

faysou:
# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [ ] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

Partial bars add complexity to the code and were used nowhere correctly in the code.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [x] Breaking change (impacts existing behavior)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

- [ ] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)

## Release notes

- [ ] I added a concise entry to `RELEASES.md` that follows the existing conventions (when applicable)

## Testing

**Ensure new or changed logic is covered by tests.**

- [ ] Affected code paths are already covered by the test suite
- [x] I added/updated tests to cover new or changed logic

PR #3040, opened 2025-10-06
Fix IBKR adapter bars request

sunlei:
## Summary

Fix IBKR adapter bars request `TypeError: _handle_bars() takes exactly 6 positional arguments (7 given)`

## Related Issues/PRs

#3020

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/data.py
+++ b/nautilus_trader/adapters/interactive_brokers/data.py
@@ -534,7 +534,6 @@ class InteractiveBrokersDataClient(LiveMarketDataClient):
             self._handle_bars(
                 request.bar_type,
                 bars,
-                bars[0],
                 request.id,
                 request.start,
                 request.end,


cjdsellers:
Hi @sunlei 

We actually removed the concept of partial bars (see [release notes](https://github.com/nautechsystems/nautilus_trader/blob/develop/RELEASES.md#breaking-changes)). These no longer need to be handled, since we’re now either subscribing to external (fully formed) bars, or aggregating them internally from higher granularity bars or ticks.

Perhaps you need a recompile?

faysou:
I think my last IB PR was merged without having been rebased on develop that container the removal of partial which is causing this issue. If so, the PR shouldn't be closed but merged.

cjdsellers:
Thanks @faysou, good catch, merged!

Also thanks for the fix @sunlei :pray:

PR #3051, opened 2025-10-08
Modify _submit_trailing_stop_market to include optional parameter

Bucanero06:
Added an optional placeholder parameter to the _submit_trailing_stop_market method. 

# Pull Request

## Summary

In Bybit's Execution Engine, the `_submit_order` passes `is_leverage` on each call regardless of method; however, the `_submit_trailing_stop_market` method only takes in arg `order: TrailingStopMarketOrder` which leads to an unexpected number of input args error, a simple if statement in the _submit_order or a placeholder arg in _submit_trailing_stop_market successfully allowed me to place the trailing stops. I chose the latter.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/nautilus_trader/adapters/bybit/execution.py
+++ b/nautilus_trader/adapters/bybit/execution.py
@@ -1390,7 +1390,7 @@ class BybitExecutionClient(LiveExecutionClient):
             is_leverage=is_leverage,
         )
 
-    async def _submit_trailing_stop_market(self, order: TrailingStopMarketOrder) -> None:
+    async def _submit_trailing_stop_market(self, order: TrailingStopMarketOrder, _: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


Bucanero06:
I Closed it because I am seeing other errors before and after the changes. The trailing stop is active on the exchange and is found  by the self.cache.orders(...) but eventually less than a minute later I see `[WARN] TRADER-001.ExecClient-BYBIT: Cannot query with client order ID for trailing stops` plus I also want to include changes to allow use of activePrice since right now immidiate activation is hard coded into BybitAccountHttpAPI.set_trading_stop and param TrailingStopMarketOrder.trigger_price is ignored/not-passed

Issue #3054, opened 2025-10-09
[Interactive Brokers, Reconciliation] Flat positions not closed entirely


idobz:
# Bug Report

## Confirmation

**Before opening a bug report, please confirm:**

- [ x] I’ve re-read the relevant sections of the documentation.
- [ x] I’ve searched existing issues and discussions to avoid duplicates.
- [ x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [ x] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

### Expected Behavior

A previous bug (https://github.com/nautechsystems/nautilus_trader/issues/3023) fixed the issue of account-closed positions not being reconciled with the cache. At current such discrepancies are dealt with, but improperly.

The expectation is that such positions existing in the cache only will be closed in the cache too.

### Actual Behavior

The system recognizes these cases and generates an inferred fill to close them, but instead of their being closed, the inferred fill generates a new position and we get two contrary positions in a NETTING account

End of the run example (showing the pairs of co-existing positions):
```
2025-10-08T14:28:27.213825095Z [WARN] IB_LIVE-001.Cache: Residual Position(SHORT 4 AGI.NYSE, id=AGI.NYSE-AGI_spread-010)
2025-10-08T14:28:27.213832922Z [WARN] IB_LIVE-001.Cache: Residual Position(LONG 4 AGI.NYSE, id=AGI.NYSE-EXTERNAL)
2025-10-08T14:28:27.213838694Z [WARN] IB_LIVE-001.Cache: Residual Position(SHORT 13 ASX.NYSE, id=ASX.NYSE-ASX_spread-004)
2025-10-08T14:28:27.213843482Z [WARN] IB_LIVE-001.Cache: Residual Position(LONG 13 ASX.NYSE, id=ASX.NYSE-EXTERNAL)
2025-10-08T14:28:27.213848104Z [WARN] IB_LIVE-001.Cache: Residual Position(LONG 5 B.NYSE, id=B.NYSE-EXTERNAL)
2025-10-08T14:28:27.213852904Z [WARN] IB_LIVE-001.Cache: Residual Position(SHORT 5 B.NYSE, id=B.NYSE-INTERNAL-DIFF)
[...]

Issue #3042, opened 2025-10-07
[Reconciliation] Fails for routing clients when instrument venue differs from client venue


cjdsellers:
Position reconciliation is currently broken for routing brokers like Interactive Brokers where the client venue (e.g., `INTERACTIVE_BROKERS`) differs from the actual instrument venues (e.g., `NYSE`, `NASDAQ`).

  In `execution_engine.py:1300`, reconciliation filters positions by the client's venue:

  ```python
  for position in self._cache.positions_open(venue):
```

For IB, venue is `IB_VENUE` but positions have venues like NYSE, so `positions_open(IB_VENUE)` returns nothing and reconciliation gets skipped entirely.

## Impact

  - Affects any routing broker (IB, potentially others).
  - Phantom positions can't be reconciled if they were closed externally.
  - Related to #3023 but broader scope.

## Potential Solution

The code already uses `venue=None` for individual position reconciliation at line 1867. We likely need similar handling earlier in the reconciliation flow, or better cache filtering that understands client-to-venue relationships.

## Related

Discovered during testing of #3023 fix by @idobz

cjdsellers:
Note there is a `reconciliation` branch which needs to be merged to `develop` (after next release) prior to any work on this.

cjdsellers:
There's been significant refactoring and improvements to reconciliation since this was reported.

The reconciliation paths now correctly use `venue=None` filtering for multi-venue brokers, with `account_id` used instead where needed. There are also specific test cases covering this logic.

Closing for now, please report any further issues you experience with recent development wheels.

Issue #1789, opened 2024-07-17
Reconciliation not working with trading node, interactive brokers and option on future


faysou:
# Bug Report

### Expected Behavior
Existing future option positions should be reconciled without errors.

Another related question I have is how would the reconciliation work if I used databento market data, for example best bid offer, so the definition of a contract in databento and interactive brokers are compatible (IB uses CME as venue, and databento uses GLBX). For example I would want to use databento as market data provider and interactive brokers for execution.

### Actual Behavior

I get the following errors

```
[...]

Issue #1154, opened 2023-06-27
Binance: reconciling the execution state fails due to a TypeError(an integer is required)


davidblom603:
# Bug Report

### Expected Behavior
The execution state is successfully reconciled without any errors.

### Actual Behavior
During initialization, the Binance execution client fails and the script is crashed. There is a TypeError exception thrown with the message that an integer is required.
```
2023-06-27T11:21:09.338621500Z [INF] TESTER-001.ExecClient-BINANCE: Nautilus clock time 1687864869338 UNIX (ms).
2023-06-27T11:21:09.338636917Z [INF] TESTER-001.ExecClient-BINANCE: Listen key 78FpuD6MkI2HcH7W8rIG73vzVTDg8xAHUAWwSM1GUpltwUMCaAlWUoKVTvw3
2023-06-27T11:21:09.338660958Z [DBG] TESTER-001.ExecClient-BINANCE: Creating task _ping_listen_keys.
2023-06-27T11:21:09.338858792Z [INF] TESTER-001.BinanceWebSocketClient: Connecting to wss://testnet.binance.vision/stream?streams=78FpuD6MkI2HcH7W8rIG73vzVTDg8xAHUAWwSM1GUpltwUMCaAlWUoKVTvw3
2023-06-27T11:21:09.339619583Z [DBG] TESTER-001.ExecClient-BINANCE: Scheduled `ping_listen_keys` to run in 300s.
2023-06-27T11:21:10.806248459Z [DBG] TESTER-001.ExecEngine: Checking in-flight orders status...
2023-06-27T11:21:10.806354584Z [DBG] TESTER-001.ExecEngine: Found 0 orders in-flight.
2023-06-27T11:21:11.094676168Z [INF] TESTER-001.BinanceWebSocketClient: Connected.
2023-06-27T11:21:11.095416626Z [INF] TESTER-001.ExecClient-BINANCE: Connected
2023-06-27T11:21:11.095809043Z [INF] TESTER-001.TradingNode: Awaiting execution state reconciliation (10.0s timeout)...
2023-06-27T11:21:11.097265668Z [INF] TESTER-001.ExecClient-BINANCE: Generating ExecutionMassStatus for BINANCE...
2023-06-27T11:21:11.099038084Z [INF] TESTER-001.ExecClient-BINANCE: Generating OrderStatusReports for BINANCE...
[...]

PR #3072, opened 2025-10-12
Fix handling of asyncio.CancelledError in execution reconciliation

dinana:
## Description

Fixes a crash during execution reconciliation startup when `asyncio.CancelledError` is returned from `generate_mass_status()`. The crash prevents the trading node from starting when reconciliation is enabled.

## Problem

The execution reconciliation crashes with:
```
AttributeError: 'CancelledError' object has no attribute 'client_id'
```

This occurs in `LiveExecEngine.reconcile_execution_state()` at line 1292-1295.

## Root Cause

In Python 3.8+, `asyncio.CancelledError` was changed to inherit from `BaseException` instead of `Exception` ([PEP 3151](https://peps.python.org/pep-3151/), [Python 3.8 release notes](https://docs.python.org/3/whatsnew/3.8.html#asyncio)).

The exception hierarchy:
- `BaseException`
  - `asyncio.CancelledError` ← NOT caught by `isinstance(x, Exception)`
  - `Exception`
    - `RuntimeError`, `ValueError`, etc.

The exception check at line 1292 only catches `Exception`:
```python
if isinstance(mass_status_or_exception, Exception):
```

When `generate_mass_status()` is cancelled, the `CancelledError` passes through this check and the code attempts to access `mass_status.client_id`, causing the crash.

## Solution

Changed the isinstance check from `Exception` to `BaseException` to catch all exception types that can be returned by `asyncio.gather(..., return_exceptions=True)`:

```python
if isinstance(mass_status_or_exception, BaseException):
```

This is more future-proof and handles any exception type that might be raised during mass status generation.

## Testing

Verified the fix resolves the issue with:
- ✅ Reconciliation enabled with 7 existing positions from IB Gateway
- ✅ StreamingConfig enabled (which previously triggered the crash)
- ✅ All engines connected successfully
- ✅ All actors reached RUNNING state
- ✅ No CancelledError crash occurred
- ✅ Reconciliation completed: "Initialized 7 open positions"

### Reproduction Environment
- **OS:** Linux (WSL2, Ubuntu 24.04)
- **Python:** 3.11.13
- **NautilusTrader:** 1.221.0 (develop branch, commit 806f1f8fc)
- **Adapter:** Interactive Brokers
- **Configuration:** Reconciliation + StreamingConfig both enabled

## Impact

- **Severity:** High - prevents node startup when reconciliation is enabled
- **Affects:** All adapters that use async operations in `generate_mass_status()`
- **Workaround (before fix):** Disable reconciliation, losing important safety features
- **Fix:** One-line change, fully backward compatible

## Related Issues

This appears to be a new issue not previously reported. Similar reconciliation issues have been filed (#1789, #1154, #3054, #3042) but none address the `CancelledError` inheritance problem.

## Checklist

- [x] Code follows project style guidelines
- [x] Changes have been tested with live adapter (Interactive Brokers)
- [x] Fix is minimal and focused on the specific issue
- [x] Commit message follows conventional commits format
- [x] Change is backward compatible

## References

- [PEP 3151 - Reworking the OS and IO exception hierarchy](https://peps.python.org/pep-3151/)
- [Python 3.8 What's New - asyncio.CancelledError](https://docs.python.org/3/whatsnow/3.8.html#asyncio)
- [asyncio.gather() documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather)

### DIFF
--- a/nautilus_trader/live/execution_engine.py
+++ b/nautilus_trader/live/execution_engine.py
@@ -1289,7 +1289,7 @@ class LiveExecutionEngine(ExecutionEngine):
 
             # Reconcile each mass status with the execution engine
             for mass_status_or_exception in mass_status_all:
-                if isinstance(mass_status_or_exception, Exception):
+                if isinstance(mass_status_or_exception, BaseException):
                     self._log.error(f"Failed to generate mass status: {mass_status_or_exception}")
                     results.append(False)
                     continue


dinana:
Closing in favor of #3073 which has the correct commit author information.

PR #3098, opened 2025-10-19
Improve and optimize AMM pool profiling

filipmacek:
# Pull Request

## Summary

- Simplify `UniswapV3PoolContract` and use native `PoolState`
- Improve core logic when bootstrapping the AMM pool profiler with and without snapshots, integrate a new method where we first stream all the liquidity events(mint/burn) from hypersync to construct to partial pool profiler state. Then, with the provided ticks and positions keys, we query the RPC and hydrate the profiler with the latest data from the blockchain to construct a pool snapshot
- Added a defensive check in the pool profiler to skip already processed events

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/crates/model/src/defi/pool_analysis/tests.rs
+++ b/crates/model/src/defi/pool_analysis/tests.rs
@@ -16,7 +16,10 @@
 use std::{
     ops::{Div, Mul},
     str::FromStr,
-    sync::Arc,
+    sync::{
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
👌 Looks good.

PR #3140, opened 2025-11-02
Optimize execution reconciliation to avoid quadratic complexity

DeirhX:
# Pull Request

## Summary

The current implementation of partial reconciliation is of O(N*N) complexity, which prevents having thousands of instruments, as then, the initialisation doesn't complete in a reasonable time. I think it can be improved upon, as it iterates `self._position_reports` inside a loop that iterates the same list.

I chose to fix the problem by moving the outer loop into `adjust_fills_for_partial_window` so it is not iterated recursively. I hope it doesn't break other functionality, as it entails changing the function signature. With this fix, 10K+ instruments are not a problem, although it needs other PRs to effectively initialize them (coming up next).

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

This is an internal method signature change, it should not be called from user code. Internal references were updated, but please double-check for me. 

## Testing

- [X] Affected code paths are already covered by the test suite

I tested manually to see if the newly generated object contains the same data. There are also tests that cover this functionality, I updated the method they are calling.

### DIFF
--- a/nautilus_trader/live/reconciliation.py
+++ b/nautilus_trader/live/reconciliation.py
@@ -563,7 +563,8 @@ def adjust_fills_for_partial_window(
     pyo3_instruments = [transform_instrument_to_pyo3(instrument) for instrument in instruments]
     results: dict[InstrumentId, tuple[dict[VenueOrderId, OrderStatusReport], dict[VenueOrderId, list[FillReport]]]] = {}
 
-    for pyo3_instrument in pyo3_instruments:
+    for instrument, pyo3_instrument in zip(instruments, pyo3_instruments):
+        assert instrument.id.value == pyo3_instrument.id.value
         pyo3_orders, pyo3_fills = nautilus_pyo3.adjust_fills_for_partial_window(
             pyo3_mass_status,
             pyo3_instrument,
@@ -588,8 +589,8 @@ def adjust_fills_for_partial_window(
             fills[venue_order_id] = reports
 
             if logger:
-                logger.debug(f"Adjusted fills for {pyo3_instrument.id}: {len(orders)} orders, {len(fills)} fills")
+                logger.debug(f"Adjusted fills for {instrument.id}: {len(orders)} orders, {len(fills)} fills")
 
-        results[pyo3_instrument.id] = (orders, fills)
+        results[instrument.id] = (orders, fills)
 
     return results


odobias:
Related PRs that make 10000+ instruments feasible to use: #3141, #3142

nautilus_trader/live/execution_engine.py
```diff
@@ -1842,43 +1843,37 @@ def _reconcile_execution_mass_status(
                     f"not in `reconciliation_instrument_ids` include list",
                 )
                 continue
+            reconciliation_instruments.append(self._cache.instrument(instrument_id))
```
cjdsellers, line 1846: I suggest we handle the `None` case here, something like:

```python
instrument = self._cache.instrument(instrument_id)
if not instrument:
    continue
```

nautilus_trader/live/execution_engine.py
```diff
@@ -1842,43 +1843,37 @@ def _reconcile_execution_mass_status(
...
-                for venue_order_id in list(final_orders.keys()):
-                    order = final_orders[venue_order_id]
-                    if order.instrument_id == instrument_id:
-                        del final_orders[venue_order_id]
-
-                for venue_order_id in list(final_fills.keys()):
-                    fills = final_fills[venue_order_id]
-                    if fills and fills[0].instrument_id == instrument_id:
-                        del final_fills[venue_order_id]
+        self._log.debug(
```
cjdsellers, line 1848: I think we should retain the original log levels here and add a config to reduce them in a follow-up.

nautilus_trader/live/execution_engine.py
```diff
@@ -1842,43 +1843,37 @@ def _reconcile_execution_mass_status(
...
+            f"Attempting to adjust fills for {len(reconciliation_instruments)} instruments",
+            LogColor.BLUE,
+        )
+        adjusted_results = adjust_fills_for_partial_window(
+            mass_status,
+            reconciliation_instruments,
+            self._log,
+        )
+        self._log.debug(
+            f"Updating adjusted fills for {len(reconciliation_instruments)} instruments",
```
cjdsellers, line 1858: Re log level, as above.

nautilus_trader/live/execution_engine.py
odobias: Sorry about that, leftover from my debugging. At first, I though it may be the log output to console that was slowing it down 🙂

DeirhX: ^ Logged from my other account, both are me 🙂

DeirhX:
Thanks for the comments, definitely feel free to address it in develop as it's well past midnight here 🙂

PR #3141, opened 2025-11-02
Add Polymarket Gamma API support for instrument loading

DeirhX:
# Pull Request

## Summary

Watching a large number (1000+) instruments currently gives us two options for initialisation: 
a) fetch every instrument separately (1000+ API calls, too slow)
b) sequentially iterate over all historic markets (faster, however, there are currently 160K of them and growing)
None of these options are particularly efficient and future-proof. This PR introduces a third option - use the Gamma API to fetch the market list, which supports server-side filtering, and we can query only for active markets.

Note, however, that the Gamma API is different and has some caveats:
- while it provides generally more info, there is also some info missing compared to CLOB API (see https://github.com/nautechsystems/nautilus_trader/compare/develop...DeirhX:nautilus_trader:gamma-markets-fetch?expand=1#diff-f3e522a1717d998a4f7bf47355991413d6189a2263720f3eeeb0ecd3d42d7ed5R243). So far, I haven't see any use of the missing info
- it provides the data in a different format. Therefore, we have to normalize the data. 
- since the raw output is different, referencing the raw output and expecting CLOB JSON will fail. So far, I haven't seen any use of that.
Given that, this API gives us the opportunity to perform powerful server-side filtering without having to iterate through an increasingly larger list of (mostly dead) markets. However, since it's not without caveats, **I've implemented it as an opt-in functionality** and haven't changed the default one, which will still use the CLOB API.

## Related Issues/PRs

Together with https://github.com/nautechsystems/nautilus_trader/pull/3140 and (coming soon) enables Nautilus to work on thousands of polymarket subscriptions

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

None

## Testing

I verified that the output is parsed into the fields returned by the CLOB API if we have them. This alternative is activated using `use_gamma_markets = True`, so it should not affect existing behaviour. I leave it up to you to change the defaults if this proves to be both useful and safe (one disadvantage I know of - Gamma API doesn't report on fees, however, we know they are zero for now. Unless they become market-dependent, we can get away with updating a constant here).

### DIFF
--- a/nautilus_trader/adapters/polymarket/providers.py
+++ b/nautilus_trader/adapters/polymarket/providers.py
@@ -65,7 +65,7 @@ class PolymarketInstrumentProvider(InstrumentProvider):
     async def load_all_async(self, filters: dict | None = None) -> None:
         await self._load_markets([], filters)
 
-    def _load_ids_using_gamma_markets(
+    async def _load_ids_using_gamma_markets(
         self,
         instrument_ids: list[InstrumentId],
         filters: dict | None = None,
@@ -75,8 +75,8 @@ class PolymarketInstrumentProvider(InstrumentProvider):
         """
         condition_ids = [get_polymarket_condition_id(inst_id) for inst_id in instrument_ids]
 
-        if filters is None:
-            filters = {}
+        # Create a copy to avoid mutating the caller's filters
+        filters = filters.copy() if filters is not None else {}
 
         if len(condition_ids) <= 100:  # We can filter directly by condition_id, but there is an API limit of max 100 condition_ids in the query string
             self._log.info(f"Loading {len(condition_ids)} instruments, using direct condition_id filtering")
@@ -84,7 +84,8 @@ class PolymarketInstrumentProvider(InstrumentProvider):
         else:
             self._log.info(f"Loading {len(condition_ids)} instruments, using bulk load of all markets")
 
-        markets = list_markets(filters=filters)  # Usually, you would use filters={"is_active": True} to skip archived markets
+        # Wrap synchronous blocking call with asyncio.to_thread
+        markets = await asyncio.to_thread(list_markets, filters=filters)
         self._log.info(f"Loaded {len(markets)} markets using Gamma API")
         for market in markets:
             condition_id = market.get("conditionId")
@@ -137,7 +138,7 @@ class PolymarketInstrumentProvider(InstrumentProvider):
             )
 
         if self._config.use_gamma_markets:
-            self._load_ids_using_gamma_markets(instrument_ids, filters)
+            await self._load_ids_using_gamma_markets(instrument_ids, filters)
         else:
             await self._load_ids_using_clob_api(instrument_ids, filters)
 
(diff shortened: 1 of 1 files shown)


odobias:
I see the assertion comparison is probably not by value, will fix it tomorrow.

nautilus_trader/adapters/polymarket/providers.py
```diff
@@ -63,6 +65,59 @@ def __init__(
     async def load_all_async(self, filters: dict | None = None) -> None:
         await self._load_markets([], filters)
 
+    def _load_ids_using_gamma_markets(
```
cjdsellers: Could we make this `async`?

nautilus_trader/adapters/polymarket/providers.py
```diff
@@ -63,6 +65,59 @@ def __init__(
...
+        if filters is None:
+            filters = {}
+
+        if len(condition_ids) <= 100:  # We can filter directly by condition_id, but there is an API limit of max 100 condition_ids in the query string
+            self._log.info(f"Loading {len(condition_ids)} instruments, using direct condition_id filtering")
+            filters["condition_ids"] = condition_ids
+        else:
+            self._log.info(f"Loading {len(condition_ids)} instruments, using bulk load of all markets")
+
+        markets = list_markets(filters=filters)  # Usually, you would use filters={"is_active": True} to skip archived markets
```
cjdsellers: Since the Polymarket clob client is synch blocking, the established pattern is to wrap with `await asyncio.to_thread(...)`

nautilus_trader/adapters/polymarket/providers.py
```diff
@@ -63,6 +65,59 @@ def __init__(
...
+        instrument_ids: list[InstrumentId],
+        filters: dict | None = None,
+    ) -> None:
+        """
+        Load instruments using Gamma API markets.
+        """
+        condition_ids = [get_polymarket_condition_id(inst_id) for inst_id in instrument_ids]
+
+        if filters is None:
+            filters = {}
```
cjdsellers: else, should we work on a copy of `filters` rather than mutate?

cjdsellers (commented):
Thanks for your work on this @DeirhX 🙏

nautilus_trader/adapters/polymarket/providers.py
DeirhX: Agreed, good idea

nautilus_trader/adapters/polymarket/providers.py
DeirhX: Thanks, applied

DeirhX:
Thanks for the suggestions, hopefully I addressed them.

cjdsellers:
@DeirhX I just followed up with [this](https://github.com/nautechsystems/nautilus_trader/commit/70e2e5d79b368fe77f4a245ae410759e49a4cd7e) to filter specific tokens.

Let me know if it causes and issues for you and I'll fix it up.

PR #3142, opened 2025-11-02
Improve Polymarket position querying using Gamma API

odobias:
# Pull Request

## Summary

When using many instruments (1000+), it is no longer feasible to query for position_allowance separately for each instrument, as that quickly runs into rate limits. Even when rate limits are addressed, such an operation takes tens of minutes to complete, making it unsuitable for large-scale order book subscriptions.

I've addressed this by using the Gamma API instead, which can give multiple balances per a single query (up to 100), notably speeding up the process. As far as I've tested, changes in balances appear around 10 seconds after submitting a fillable order using the web, which should be sufficient, as this code path is used only when initialising the engine (correct me if I'm wrong).

The Gamma API gives much more information than just balances (notably initial and current values), but we don't need that for the moment unless it's decided to make this the standard way of querying (so far, I've gated it behind `use_gamma_api` config setting). Regretfully, it doesn't reply with `allowances`, but so far I've not seen it used. In any case, it's an opt-in behaviour, suited for high-volume use.

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/pull/3140 and https://github.com/nautechsystems/nautilus_trader/pull/3141

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

Tested on ~16000 instruments and basic order submission.

### DIFF
--- a/nautilus_trader/adapters/polymarket/execution.py
+++ b/nautilus_trader/adapters/polymarket/execution.py
@@ -179,7 +179,7 @@ class PolymarketExecutionClient(LiveExecutionClient):
 
         # Get the user address (funder) - this is the address that holds positions
         # For proxy wallets, this differs from the signer address
-        user_address = http_client.builder.funder or config.funder or wallet_address
+        user_address = http_client.builder.funder if hasattr(http_client, "builder") else config.funder or wallet_address
         validate_ethereum_address(user_address)
         self._user_address = user_address
 


DeirhX:
Corrected the API name - it's Data API (https://data-api.polymarket.com/positions)

nautilus_trader/adapters/polymarket/execution.py
```diff
@@ -172,11 +173,20 @@ def __init__(
...
+        validate_ethereum_address(user_address)
+        self._user_address = user_address
+
         self._api_key = http_client.creds.api_key
         self._log.info(f"{wallet_address=}", LogColor.BLUE)
+        self._log.info(f"{user_address=}", LogColor.BLUE)
 
         # HTTP API
         self._http_client = http_client
+        self._http_session = requests.Session()
```
cjdsellers: Could we use the `HttpClient` so we can make async calls?

nautilus_trader/adapters/polymarket/execution.py
```diff
@@ -296,6 +306,54 @@ async def _update_account_state(self) -> None:
...
+            params = {
+                "user": self._user_address,
+                "limit": str(limit),
+                "offset": str(offset),
+                "sizeThreshold": str(size_threshold),
+                "sortBy": "TOKENS",
+                "sortDirection": "DESC",
+            }
+            headers = {
+                "User-Agent": (
```
cjdsellers: This is interesting, do we really need this?

nautilus_trader/adapters/polymarket/execution.py
```diff
@@ -172,11 +173,20 @@ def __init__(
 
         validate_ethereum_address(wallet_address)
         self._wallet_address = wallet_address
+
+        # Get the user address (funder) - this is the address that holds positions
+        # For proxy wallets, this differs from the signer address
+        user_address = http_client.builder.funder if hasattr(http_client, "builder") else wallet_address
```
cjdsellers: Is there better logic we could use than `hasattr`?

nautilus_trader/adapters/polymarket/execution.py
```diff
@@ -1023,6 +1146,10 @@ async def _post_signed_order(self, order: Order, signed_order) -> None:
                 )
             else:
                 venue_order_id = VenueOrderId(response["orderID"])
+                self._log.info(
```
cjdsellers: Debug logging?

nautilus_trader/adapters/polymarket/execution.py
DeirhX: I remember not getting a valid reponse before adding this, but now everything is fine, removing :)

DeirhX:
Addressed the comments, tested and corrected the parsing of the condition+tokenId from Data API

nautilus_trader/adapters/polymarket/execution.py
DeirhX: I wasn't able to remove it as tests mock the http_client and lack the builder attribute. Would you know of a better approach here? We could use try/except, but I assume it's not much better.

cjdsellers:
> Addressed the comments, tested and corrected the parsing of the condition+tokenId from Data API

Thanks for all the follow-ups @DeirhX :pray:

PR #3197, opened 2025-11-19
chore: CI improvements, type safety, and execution safety hardening

myevolve:
See description in previous attempt; retargeted to develop.


cjdsellers:
Hi @myevolve 

Thanks for the PR, I do not see a previous one from you though.

It seems to be based on an older version of the codebase around 1.90.0, and also covers many different aspects at once - which makes it tough to review. This should be broken down into at least 4-5 separate PRs.

The REST API service is also [out of scope](https://github.com/nautechsystems/nautilus_trader/blob/develop/ROADMAP.md#open-source-scope).

PR #3222, opened 2025-11-25
Implement dYdX v4 gRPC order execution

v9ai:
# Pull Request

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/dydx/src/grpc/order.rs
+++ b/crates/adapters/dydx/src/grpc/order.rs
@@ -209,6 +209,7 @@ impl OrderBuilder {
     /// Set as Market order.
     ///
     /// An instruction to immediately buy or sell an asset at the best available price when the order is placed.
+    #[must_use]
     pub fn market(mut self, side: OrderSide, size: Decimal) -> Self {
         self.order_type = Some(OrderType::Market);
         self.side = Some(side);
@@ -220,6 +221,7 @@ impl OrderBuilder {
     ///
     /// With a limit order, a trader specifies the price at which they're willing to buy or sell an asset.
     /// Unlike market orders, limit orders don't go into effect until the market price hits a trader's "limit price."
+    #[must_use]
     pub fn limit(mut self, side: OrderSide, price: Decimal, size: Decimal) -> Self {
         self.order_type = Some(OrderType::Limit);
         self.price = Some(price);
@@ -231,6 +233,7 @@ impl OrderBuilder {
     /// Set as Stop Limit order.
     ///
     /// Stop-limit orders use a stop `trigger_price` and a limit `price` to give investors greater control over their trades.
+    #[must_use]
     pub fn stop_limit(
         mut self,
         side: OrderSide,
@@ -249,6 +252,7 @@ impl OrderBuilder {
     /// Set as Stop Market order.
     ///
     /// When using a stop order, the trader sets a `trigger_price` to trigger a buy or sell order on their exchange.
+    #[must_use]
     pub fn stop_market(mut self, side: OrderSide, trigger_price: Decimal, size: Decimal) -> Self {
         self.order_type = Some(OrderType::StopMarket);
         self.trigger_price = Some(trigger_price);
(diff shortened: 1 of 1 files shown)


crates/adapters/dydx/src/data/mod.rs
```diff
@@ -123,6 +123,18 @@ pub struct DydxDataClient {
...
 
 impl DydxDataClient {
+    /// Derives the dYdX candle topic ticker from a Nautilus InstrumentId.
+    ///
+    /// dYdX candle topics use the exchange symbol without the `-PERP` suffix.
+    /// For example, `BTC-USD-PERP.DYDX` → `BTC-USD`.
+    ///
+    /// This must match the topic format used in WebSocket candle subscription
+    /// and incoming candle messages.
+    fn dydx_candle_ticker_from_instrument(instrument_id: &InstrumentId) -> String {
```
cjdsellers: This should be a more general parsing function under `common/parse.rs`, and not specific to candles. I think other adapters might take the symbol only as well?

crates/adapters/dydx/src/execution/submitter.rs
```diff
@@ -280,87 +395,161 @@ impl OrderSubmitter {
...
+    fn convert_order_side(&self, side: OrderSide) -> ProtoOrderSide {
+        match side {
+            OrderSide::Buy => ProtoOrderSide::Buy,
+            OrderSide::Sell => ProtoOrderSide::Sell,
+            _ => ProtoOrderSide::Unspecified, // Should never happen after validation
+        }
+    }
+
+    /// Convert Nautilus TimeInForce to proto TimeInForce.
+    fn convert_time_in_force(&self, tif: TimeInForce) -> ProtoTimeInForce {
```
cjdsellers: This should be a function under `common/parse.rs` probably. Or, if a 1-to-1 mapping could implement `From` on the enum.

v9ai: > if a 1-to-1 mapping could implement From on the enum.

Why it doesn't work:

1. Orphan Rule - Can't implement `From` because both the trait and 
   `ProtoTimeInForce` are from external crates (not ours)

2. Not 1-to-1 - Multiple Nautilus values map to same proto value:
   
   TimeInForce::Gtc => ProtoTimeInForce::Unspecified
   TimeInForce::Gtd => ProtoTimeInForce::Unspecified  // Same!
   TimeInForce::Day => ProtoTimeInForce::Unspecified  // Same!

v9ai: Moved to `common/parse.rs`, thx.

crates/adapters/dydx/src/data/mod.rs
v9ai: Moved Bybit's `extract_raw_symbol` into `common` crate and reused in both places, if you don't mind. Thanks.

crates/adapters/bybit/src/common/parse.rs
```diff
@@ -114,12 +115,6 @@ where
     s.parse::<u8>().map_err(serde::de::Error::custom)
 }
 
-/// Extracts the raw symbol from a Bybit symbol by removing the product type suffix.
-#[must_use]
-pub fn extract_raw_symbol(symbol: &str) -> &str {
```
cjdsellers, line 119: I don't think we should move this to the common crate.

Symbols won't always normalize in the same way across adapters - these two just happen to be the same. Because the naming is very general, it's then ambiguous whether to use the function from common (can also be hard to discover) or if the raw symbol extraction function was forgotten etc.

For the sake of a single line, I think best to have a standard pattern with a function in the adapter crates `common/parse.rs`.

crates/adapters/dydx/src/data/mod.rs
cjdsellers: I don't think we should move this to the common crate.

Symbols won't always normalize in the same way across adapters - these two just happen to be the same. Because the naming is very general, it's then ambiguous whether to use the function from common (can also be hard to discover) or if the raw symbol extraction function was forgotten etc.

For the sake of a single line, I think best to have a standard pattern with a function in the adapter crates common/parse.rs.

v9ai: Done, thanks.

crates/adapters/bybit/src/common/parse.rs
v9ai: Done, thanks.

PR #3321, opened 2025-12-20
Unify tokio runtime selection in Rust adapters

filipmacek:
# Pull Request

## Summary

Standardize all adapter code to use `get_runtime().spawn()` instead of `tokio::spawn()` for consistent runtime selection across Python and Rust contexts.

### The Error

From time to time I would encounter this panic when running adapters from Python:

```
thread '<unnamed>' panicked at crates/adapters/deribit/src/websocket/client.rs:299:17:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

### The Problem

The difference is how Tokio finds the runtime to spawn on:

**`tokio::spawn()`** - Uses thread-local storage to find the "current" runtime. It looks for a runtime context on the calling thread. When called from Python's thread (which has no Tokio context), it panics with "there is no reactor running".

**`get_runtime().spawn()`** - Gets an explicit reference to a global static runtime stored in `OnceLock<Runtime>`. You're directly calling `.spawn()` on a concrete Runtime instance, so it doesn't need thread-local context - it knows exactly which runtime to use.

```
┌─────────────────────────────────────────────────────────────┐
│  Python Thread (no Tokio context)                           │
│                                                             │
│  tokio::spawn() ───► looks for thread-local runtime ───► PANIC
│                                                             │
│  get_runtime().spawn() ───► uses global static ─────────► WORKS
│                      │                                      │
│                      ▼                                      │
│             OnceLock<Runtime> (initialized once,            │
│             accessible from any thread)                     │
└─────────────────────────────────────────────────────────────┘
```

This is exactly why Nautilus uses the `get_runtime()` pattern - it allows Rust async code to be spawned from Python callbacks, which run on Python's event loop thread, not inside a Tokio runtime context.

## Why This Is Safe

| Aspect | Analysis |
|--------|----------|
| **Semantic equivalence** | Both submit tasks to a Tokio runtime's work-stealing scheduler |
| **Thread safety** | `OnceLock` guarantees safe initialization across threads |
| **Task lifecycle** | `JoinHandle` behavior unchanged - fire-and-forget pattern preserved |
| **Cancellation** | Existing `CancellationToken` patterns continue to work |
| **Test isolation** | Test files excluded - `#[tokio::test]` keeps its own runtime |

### Test Fixes

Some tests failed after the runtime migration due to **implicit timing assumptions** that were valid with single-threaded `#[tokio::test]` but break with the multi-threaded global runtime.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

### DIFF
--- a/crates/adapters/bitmex/src/execution/submitter.rs
+++ b/crates/adapters/bitmex/src/execution/submitter.rs
@@ -1207,14 +1207,10 @@ mod tests {
 
     #[tokio::test]
     async fn test_client_stats_collection() {
-        let report = create_test_report("ORDER-1");
-        let report_clone = report.clone();
-
+        // Both clients fail so broadcast waits for all of them (no early abort on success).
+        // This ensures both clients execute and record their stats before the function returns.
         let transports = vec![
-            create_stub_transport("client-0", move || {
-                let report = report_clone.clone();
-                async move { Ok(report) }
-            }),
+            create_stub_transport("client-0", || async { anyhow::bail!("Timeout error") }),
             create_stub_transport("client-1", || async { anyhow::bail!("Connection error") }),
         ];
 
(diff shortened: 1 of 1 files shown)


cjdsellers:
Thanks @filipmacek 

Good catch and standardization, I agree with enforcing this in the adapter layer while we have Python FFI access (I'll also add a lint rule for this, as it would be easy to regress).

Once that's gone we can move back to the more idiomatic `tokio::spawn`.

PR #3356, opened 2025-12-29
Fix balance recalculation to use raw fixed-point

kirill-gr1:
## Summary

Nautilus `CashAccount._recalculate_balance` and `MarginAccount._recalculate_balance` previously recomputed `locked`/`free` using `Decimal` math and then rebuilt `Money(...)` from those `Decimal` values. Because `Money.__init__` converts inputs via `float(value)`, this can introduce a tiny rounding change in the underlying fixed‑point integer (`raw`). That can cause `AccountBalance` construction to fail its strict invariant check:

- **Invariant**: `total.raw - locked.raw == free.raw` (exact integer equality)

This manifests as:

- `ValueError("`total` (...) - `locked` (...) != `free` (...)")`

even when the values are “effectively” consistent at the decimal level, due to the float round‑trip altering `raw`.

This PR fixes the issue by recomputing balances directly in raw integer space and constructing `Money` from raw, guaranteeing the invariant holds.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

N/A

## Documentation

## Release notes

## Testing

**How tested:**
- `make build-debug`
- `make pytest`
- `make cargo-test` (cargo-nextest)
- Re-ran a backtest that previously failed with an `AccountBalance` raw invariant error; it now completes successfully.

**New regression coverage:**
- Adds large-raw balance regression tests for both cash and margin accounts, asserting `total.raw - locked.raw == free.raw` in normal and clamp scenarios.

### DIFF
--- a/tests/unit_tests/accounting/test_margin.py
+++ b/tests/unit_tests/accounting/test_margin.py
@@ -186,6 +186,46 @@ class TestMarginAccount:
         assert balance.locked.raw == raw_total
         assert balance.free.raw == 0
 
+    def test_recalculate_balance_uses_raw_and_clamps_with_maintenance_margin(self):
+        # Arrange
+        raw_total = 5_000_000_000_000  # large raw value to guard against float drift
+        total_money = Money.from_raw(raw_total, USD)
+        event = AccountState(
+            account_id=AccountId("RAW-MARGIN-MAINT"),
+            account_type=AccountType.MARGIN,
+            base_currency=USD,
+            reported=True,
+            balances=[
+                AccountBalance(
+                    total_money,
+                    Money.from_raw(0, USD),
+                    Money.from_raw(raw_total, USD),
+                ),
+            ],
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @kirill-gr1 

Thanks for the contribution!

Ideally (as optional follow-ups) it would be good to also have: 
- A test exercising the maintenance margin path
- Some "float-unfriendly" values for test cases to demonstrate the fix

We'll also need the CLA signing then this can be merged!

kirill-gr1:
Hi @cjdsellers 
1) Done
2) The tests I added will actually fail on non-patched version. Is it sufficient?
3) Fixed CLA as well - there was a mismatch in acct and commit emails

cjdsellers:
Hey @kirill-gr1 

Thanks for the follow-ups. That sounds like an ideal test case - cheers!

PR #3364, opened 2025-12-31
Improve backtest data iteration by sorting data using ts_event instead of ts_init

marcus-sa:
I've added a `HasTsEvent` trait to the `Data` enum and updated the backtest engine to sort data by `ts_event` instead of `ts_init`.

### Rationale
- `ts_event` represents when the event occurred at the exchange/venue
- `ts_init` represents when the data was ingested by Nautilus
- For backtesting, we want to process events in the order they occurred at the exchange, not when they were ingested
- This ensures proper chronological ordering for realistic backtesting

# Pull Request

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/backtest/src/data_iterator.rs
+++ b/crates/backtest/src/data_iterator.rs
@@ -20,9 +20,9 @@ use std::collections::BinaryHeap;
 
 use ahash::AHashMap;
 use nautilus_core::UnixNanos;
-use nautilus_model::data::{Data, HasTsInit};
+use nautilus_model::data::{Data, HasTsEvent};
 
-/// Internal convenience struct to keep heap entries ordered by `(ts_init, priority)`.
+/// Internal convenience struct to keep heap entries ordered by `(ts_event, priority)`.
 #[derive(Debug, Eq, PartialEq)]
 struct HeapEntry {
     ts: UnixNanos,
@@ -81,8 +81,8 @@ impl BacktestDataIterator {
             return;
         }
 
-        // Ensure sorted by ts_init
-        data.sort_by_key(HasTsInit::ts_init);
+        // Ensure sorted by ts_event (when the event occurred at the exchange)
+        data.sort_by_key(HasTsEvent::ts_event);
 
         let priority = if let Some(p) = self.priorities.get(name) {
             // Replace existing stream – remove previous traces then re-insert below.
@@ -157,7 +157,7 @@ impl BacktestDataIterator {
         self.indices.insert(entry.priority, next_index);
         if next_index < stream_vec.len() {
             self.heap.push(HeapEntry {
-                ts: stream_vec[next_index].ts_init(),
+                ts: stream_vec[next_index].ts_event(),
                 priority: entry.priority,
                 index: next_index,
             });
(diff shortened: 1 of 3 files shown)


faysou:
For now the convention of nautilus is to sort backtest data by ts_init, mixing ts_event and ts_init leads to bugs. The live system uses ts_init and the backtest system reuses parts of the live system. The benefit of using ts_event in backtest is small compared to the potential for bugs.

marcus-sa:
@faysou I ran backtests using Tardis **[book_snapshot_25](https://docs.tardis.dev/downloadable-csv-files#book_snapshot_25)** data, which I converted into Nautilus `OrderBookDepth10` by taking the top 10 levels. While doing so, I encountered the following error:

```text
Timestamp should not go backwards: old=1761955203712000000, new=1761955203711000000
```

The issue was resolved by using `ts_event` instead of `ts_init`.
In my case, `ts_init` was inferred from Tardis’ `local_snapshot` field, whereas `ts_event` was derived from Tardis’ `timestamp`, which preserved monotonic ordering.

For clarity, the relevant Tardis fields are defined as follows:
- `timestamp`: Timestamp provided by the exchange, in microseconds since epoch. If the exchange does not provide one, the `local_timestamp` value is used as a fallback.
- `local_timestamp`: Message arrival timestamp, in microseconds since epoch.

faysou:
A problem is that ts_event can point to start of bar for some adapters. It's easier to copy ts_event to ts_init while saving your data than changing the whole system. Switching ts_init  to ts_event is a major change, I wouldn't want to do this during the migration.

marcus-sa:
@faysou that makes sense. Thanks for the feedback!

faysou:
You're welcome, happy you understand.

marcus-sa:
@faysou hmm, that doesn't seem to work. Order book is also using `ts_event` instead of `ts_init` [here](https://github.com/nautechsystems/nautilus_trader/blame/51856348410ea8e0f1c04863c965d9af07b6d5d2/crates/model/src/orderbook/book.rs#L329C9-L329C39)

faysou:
I'll defer to @cjdsellers to sort this out, I would rather use ts_init everywhere to avoid issues. There were issues before in the order matching engine because ts_event was used for bars, and ts_event can be at the start of a time period.

faysou:
Another point in favour of ts_init, is that catalog requests are made with respect to ts_init. So just ensuring that ts_init is correct everywhere decreases the possibility for bugs instead of juggling between ts_event and ts_init.

cjdsellers:
Hey @marcus-sa

Thanks for the PR.

The current use of `ts_init` is intentional. In backtesting, `ts_init` represents when Nautilus "observed" the data, which is critical for:

- **Matching engine consistency** - the simulated matching engine and other components rely on a single consistent timestamp for ordering. Using `ts_event` could cause mis-ordering between data and internal events.
- **Realistic simulation** - in live trading, your system processes data when it arrives, not when it occurred at the exchange, so backtesting should mirror this.
- **Simplicity** - one timestamp for all internal ordering avoids subtle timing bugs.

For simulating network/exchange latency, we have the `LatencyModel` which could be expanded later to model more realistic delays between `ts_event` and when data is "seen" by the system.

If you're experiencing ordering issues with Tardis data specifically, as faysou suggests - that should be addressed at the adapter layer during ingestion, not by changing the backtest engine's fundamental ordering semantics.

Closing for now as this would break the intended design.

marcus-sa:
@cjdsellers @faysou appreciate the detailed explanations! I'll make sure to fix the bug on my side.

PR #3379, opened 2026-01-02
Fix trailing stops default price type

KaulSe:
## Problem

Trailing-stop calculation treated `TriggerType.DEFAULT` as if it required a **last trade price** (`LAST_PRICE` / `MARK_PRICE` path). In quote-only setups (Quote Ticks, no Trade Ticks), `last` is `None`, which triggered:

```
RuntimeError(cannot process trailing stop, no LAST price for XRPUSD.SIM (add trades or use bars))
```

However, the docs state that `DEFAULT` should behave like `BID_ASK` (quote-based):
`https://nautilustrader.io/docs/latest/concepts/orders/#trigger-type`

> `DEFAULT`: which is the same as `BID_ASK`.

This appears to be a long-standing mismatch that became obvious after upgrading from **1.221.0 → 1.222.0**, where error handling changes made the issue surface more clearly (instead of being effectively ignored / silently failing).

## Root Cause

There was an inconsistency in how `DEFAULT` was interpreted across the stack:

- **Docs**: `DEFAULT` = `BID_ASK` (quote-based)
- **Emulator subscriptions**: `DEFAULT` subscribes to quotes (same as `BID_ASK`)
- **Trailing-stop calculation**: `DEFAULT` was grouped with `LAST_PRICE` / `MARK_PRICE` (trade/last-based), requiring `last`

## Solution

Align trailing-stop calculation with the documented behavior by treating `TriggerType.DEFAULT` the same as `BID_ASK`:

- Python: `nautilus_trader/execution/trailing.pyx`
- Rust: `crates/execution/src/trailing.rs`

## Impact

- Quote-only environments can use trailing stops with `TriggerType.DEFAULT` without requiring Trade Ticks.
- Behavior now matches both the documentation and emulator subscription logic.

### DIFF
--- a/tests/unit_tests/backtest/test_exchange_trailing_stops.py
+++ b/tests/unit_tests/backtest/test_exchange_trailing_stops.py
@@ -185,6 +185,24 @@ class TestSimulatedExchange:
         with pytest.raises(RuntimeError):
             self.exchange.process(0)
 
+    def test_trailing_stop_market_order_default_when_no_quote_ticks_raises_runtime_error(
+        self,
+    ) -> None:
+        # Arrange: Prepare market
+        trailing_stop = self.strategy.order_factory.trailing_stop_market(
+            instrument_id=USDJPY_SIM.id,
+            order_side=OrderSide.BUY,
+            quantity=Quantity.from_int(200_000),
+            trailing_offset_type=TrailingOffsetType.PRICE,
+            trailing_offset=Decimal("1.0"),
+            trigger_type=TriggerType.DEFAULT,
+        )
+        self.strategy.submit_order(trailing_stop)
+
+        # Assert
+        with pytest.raises(RuntimeError):
+            self.exchange.process(0)
+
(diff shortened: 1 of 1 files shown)


KaulSe:
Alternative would be to update the documentation. I have the impression this issue is long-standing, and changing the behavior could be a significant breaking change.

KaulSe:
LMK which path makes more sense, and I will adjust the PR (failing tests), create a new one (doc update) or I am missing something ;)

cjdsellers:
Hi @KaulSe 

Thanks for the report!

I agree with the inconsistency you've identified - the trailing stop calculator should treat `DEFAULT` the same as `BID_ASK`, which would then align with both the docs and order emulator, so your fix is the right approach.

Ideally we'd have a couple of tests covering `TriggerType::Default` with bid/ask data, so looks like it's just the tests that need updating now?

KaulSe:
Alright, thank you.

I will resolve the test issues and then mark the PR as ready for review.

cjdsellers:
Hey @KaulSe 

Thanks for the following up with the tests - merging!

Issue #3391, opened 2026-01-05
Pontetial data loss in writers


netomenoci:
# Data Loss Risk During Service Restart: Analysis and Solutions

## Executive Summary

When writing feather files to S3 via fsspec/s3fs, data is buffered in memory and only uploaded to S3 under specific conditions. During service restart, there is a potential for data loss if in-memory buffers are not properly flushed to S3.

**Severity**: Medium-High (data loss possible but mitigated by `__del__` fallback)

## Important: Two Different Flush Mechanisms

Before diving into the problem, it's critical to understand there are **two separate flush mechanisms**:

### 1. Periodic Flush (Every ~10 seconds) - NOT the problem

**Code path**: `StreamingFeatherWriter.check_flush()` → `flush()`

```
writer.py:278   → check_flush() called after each write
writer.py:534   → check_flush() checks if flush_interval_ms elapsed
writer.py:541   → calls flush() if interval elapsed
writer.py:548-550 → flush() calls stream.flush() on each file handle
```

This runs continuously during normal operation. **We do NOT want to change this.**
Calling `close()` here would require reinitializing all writers, which is expensive.

### 2. Shutdown Flush (Once at termination) - THIS IS THE PROBLEM

**Code path**: `kernel.stop_async()` → `_flush_writer()` → `flush()`

```
kernel.py:1098  → stop_async() calls _flush_writer() at the end
kernel.py:1442-1444 → _flush_writer() only calls self._writer.flush()
```

This runs ONCE when the application is terminating. **This is where the fix should go.**
Calling `close()` here is safe because we're shutting down anyway.

### Visual Summary

```
NORMAL OPERATION (unchanged):
┌─────────────────────────────────────────────────────────────────┐
│  Data received → write() → check_flush() → flush()             │
│                              ↓                                   │
│                   Every 10 seconds, calls stream.flush()        │
│                   Buffer < 50MB? → Do nothing (stays in RAM)    │
│                   Buffer >= 50MB? → Upload chunk to S3          │
│                                                                  │
│  This loop continues indefinitely. Writers stay open.           │
└─────────────────────────────────────────────────────────────────┘

SHUTDOWN (this is what we're fixing):
┌─────────────────────────────────────────────────────────────────┐
│  SIGTERM received → stop_async() → _flush_writer()              │
│                                          ↓                       │
│  CURRENT (bug):     self._writer.flush()  → Data may be lost!   │
[...]

PR #3394, opened 2026-01-06
Refine closing of streaming writer

faysou:
# Pull Request

## Fix potential data loss in writers during shutdown

This PR fixes a potential data loss issue where in-memory buffered data in `StreamingFeatherWriter` may not be uploaded to S3 during service restarts. The fix ensures all buffered data is properly flushed and uploaded before the process exits.

### Problem

When writing feather files to S3 via fsspec/s3fs, data is buffered in memory and only uploaded to S3 when the buffer size exceeds 50MB (default `blocksize`). During shutdown, `_flush_writer()` was calling `flush()`, which does not force upload of buffers smaller than 50MB. This could result in data loss if:

- The process exits before `__del__` runs.
- There are circular references preventing garbage collection.
- The process is killed with SIGKILL (OOM, systemd timeout).

### Solution

Changed `_flush_writer()` to call `close()` instead of `flush()` during shutdown. The `close()` method ensures all remaining buffered data is uploaded to S3 regardless of buffer size by calling `flush(force=True)` internally, then properly closing all file handles.

#### Core change

- Modified `_flush_writer()` in `nautilus_trader/system/kernel.py` to call `self._writer.close()` instead of `self._writer.flush()`.

- This change only affects the shutdown path (`stop()` and `stop_async()` methods), ensuring it's safe to close writers since the application is terminating.

- The periodic flush mechanism (every ~10 seconds during normal operation) remains completely unchanged and unaffected.

#### Safety considerations

- `close()` is idempotent: safe to call multiple times if `dispose()` is later called.

- `_flush_writer()` is only called from terminal shutdown operations, so closing writers does not require reinitialization.

- The `StreamingFeatherWriter.close()` method internally calls `flush()` first, then closes all writers and file handles, ensuring all data is uploaded.

### Benefits

- Prevents data loss: all in-memory buffered data is guaranteed to be uploaded to S3 before process exit.

- Minimal code change: single line modification with low risk of introducing bugs.

- No performance impact: change only affects shutdown path, normal operation remains unchanged.

- Reliable shutdown: eliminates dependency on `__del__` destructor which is not guaranteed to run.

### Files changed

- `nautilus_trader/system/kernel.py`: changed `_flush_writer()` to call `close()` instead of `flush()`.

### References

- Fixes issue #3391: Potential data loss in writers.
- Solution A from the issue analysis: change shutdown flush to use `close()`.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/nautilus_trader/system/kernel.py
+++ b/nautilus_trader/system/kernel.py
@@ -1438,4 +1438,4 @@ class NautilusKernel:
 
     def _flush_writer(self) -> None:
         if self._writer is not None:
-            self._writer.flush()
+            self._writer.close()


cjdsellers:
Thanks for the fix @faysou :pray:

PR #3414, opened 2026-01-12
Fix databento_data to fetch definitions for full date range

Johnkhk:
## Summary
- Fixes an issue where `databento_definition_dates()` only returned a 1-day window regardless of the actual data range being requested
- This caused multi-year backfills to miss contract definitions for contracts that didn't exist on the start date (e.g., ESH1, ESH2, ESH3)
- Now definitions are fetched for the same date range as bar data

## Changes
- Added optional `end_time` parameter to `databento_definition_dates()`
- Updated `databento_data()` and `databento_cost()` to pass `end_time`
- Maintains backwards compatibility (falls back to 1-day window if no `end_time` provided)

## Test plan
- [x] Verified fix works for multi-year futures backfill
- [x] Backwards compatible - single-day requests still work
- [ ] Existing tests pass

### DIFF
--- a/nautilus_trader/adapters/databento/data_utils.py
+++ b/nautilus_trader/adapters/databento/data_utils.py
@@ -269,7 +269,8 @@ def databento_definition_dates(start_time: str, end_time: str | None = None) -> 
 
     if end_time is not None:
         # Use the provided end_time to get definitions for the full range
-        definition_end_date = end_time.split("T")[0]
+        # Apply next_day() since Databento uses half-open intervals (end exclusive)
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @Johnkhk 

Many thanks for the fix and contribution.

It would be ideal if we add some test coverage on this function. Also noting that Databento request time ranges are half-open (end _exclusive_), so `end_time` should also have `next_day()` applied to ensure definitions active on the final day are included.

cjdsellers:
Thanks for the follow-up @Johnkhk, merging!

PR #3467, opened 2026-01-23
Add request_funding_rates and arrow serialization

dxwil:
# Pull Request

## Summary

Add the following:
- `request_funding_rates` (and implement for Bybit)
- Arrow serialization for `FundingRateUpdate`

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/bybit/src/common/parse.rs
+++ b/crates/adapters/bybit/src/common/parse.rs
@@ -584,20 +584,16 @@ pub fn parse_trade_tick(
 pub fn parse_funding_rate(
     funding: &BybitFunding,
     instrument: &InstrumentAny,
-    ts_init: UnixNanos,
 ) -> anyhow::Result<FundingRateUpdate> {
-    let rate = parse_decimal(&funding.funding_rate, "funding.funding_rate")?;
-    let ts_event = parse_millis_timestamp(
-        &funding.funding_rate_timestamp,
-        "funding.funding_rate_timestamp",
-    )?;
+    let rate = parse_decimal(&funding.funding_rate, "funding.rate")?;
+    let ts_event = parse_millis_timestamp(&funding.funding_rate_timestamp, "funding.timestamp")?;
 
     Ok(FundingRateUpdate::new(
         instrument.id(),
         rate,
         None, // next_funding_ns not provided with historical funding rates
         ts_event,
-        ts_init,
+        ts_event,
     ))
 }
 
--- a/crates/adapters/bybit/src/http/client.rs
+++ b/crates/adapters/bybit/src/http/client.rs
@@ -3032,7 +3032,6 @@ impl BybitHttpClient {
     ) -> anyhow::Result<Vec<FundingRateUpdate>> {
         let instrument = self.instrument_from_cache(&instrument_id.symbol)?;
         let bybit_symbol = BybitSymbol::new(instrument_id.symbol.as_str())?;
-        let ts_init = self.generate_ts_init();
 
         let start_ms = start.map(|dt| dt.timestamp_millis());
         let mut seen_timestamps: AHashSet<i64> = AHashSet::new();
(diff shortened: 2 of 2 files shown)


dxwil:
Should now be ready @cjdsellers. A few things to note:

- I saw some FFI files (e.g. `model.h`) that define the Rust `Data` enum (to which I added `FundingRateUpate`), but I did not modify them (not that proficient with cross-language stuff and everything compiled and worked without changes there). So there might need to be some modifications made there.
- Not specifically to do with this PR, but in places where catalog is queried and filtered by start/end (e.g. [here](https://github.com/nautechsystems/nautilus_trader/blob/310f71bb4efb2835a6b46e9c341647689c3036b8/nautilus_trader/persistence/catalog/parquet.py#L1960)), it is filtered with `ts_init`, which from first glance is not the expected behavior. For example (and this applied to all data types, tried with bars as well), say I requested data for the last week (filtered by start and end) and wrote it to the catalog (`update_catalog=True`). The data client requested the data today so `ts_init` was stamped as today. Then if I rerun the same request, it sees that the catalog has data (from the file name which is based on `ts_event`), but then when it goes to get that data, it filters it for last week, but `ts_init` is today and no data is returned.


P.S. Sorry for the messy commit history, forgot to sync with upstream before starting to work on this... Also, no clue why pre-commit fails, it passes locally.

cjdsellers:
Hi @dxwil 

```
make pre-commit
```
or (if you're on Windows):
```
pre-commit run --all-files
```

```
pre-commit install
```

Also, we can't add `FundingRateUpdate` to the `Data` enum, because it's so tightly integrated with FFI and Cython, and the Rust `Decimal` is not FFI-friendly, and we don't want to bend/change the type just to satisfy Cython and C.

On the timestamp issue, it's a large topic. Generally for data you're writing that you intend to use for backtesting, and you don't need to preserve any latency information, then best to make `ts_init` the same as `ts_event` to avoid the exact issue you described. If that is not possible, then this is probably an issue requiring further discussion.

On the commit history, best to rebase and squash everything into a single commit - then force push (force push fine as you could take a backup locally if you want to preserve your incremental history, and this is a feature branch right now).

Doing a `make build-debug` will regenerate the C headers. This will also allow you the run the tests locally using `make pytest`. The other make target `make pre-flight` is another good one to run and have passing prior to pushing to the PR.

[edit] if pre-commit passes locally, check you have the latest Rust stable, and that you have built and run tests locally as well.

dxwil:
Hey @cjdsellers

Fixed the things you mentioned, serialization is now purely in Python (and so FundingRateUpdate is removed from the Data enum). For the timestamp issue, I created an issue [here](https://github.com/nautechsystems/nautilus_trader/issues/3491), I think more discussion is needed.

crates/common/src/actor/data_actor.rs
```diff
@@ -1850,6 +1870,40 @@ pub trait DataActor:
...
+    ) -> anyhow::Result<UUID4>
+    where
+        Self: 'static + Debug + Sized,
+    {
+        let actor_id = self.actor_id().inner();
+        let handler = ShareableMessageHandler::from_typed(move |resp: &FundingRatesResponse| {
+            get_actor_unchecked::<Self>(&actor_id).handle_funding_rates_response(resp);
+        });
+
+        DataActorCore::request_trades(
```
cjdsellers, line 1895: This should be `request_funding_rates(...)`

crates/common/src/actor/data_actor.rs
```diff
@@ -496,6 +497,16 @@ pub trait DataActor:
         Ok(())
     }
 
+    /// Actions to be performed when receiving historical trades.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if handling the historical trades fails.
```
cjdsellers, line 504: Small typo: should be "funding rates" instead of "trades".

PR #3492, opened 2026-01-27
[InteractiveBrokers] Fix fetching historical bar data for futures using raw symbology

davidsblom:
# Pull Request

## Summary

Fix fetching historical bar data for future chains when using raw symbology for the Interactive Brokers adapter. The instrument ID was regenerated with the wrong symbol. As a consequence, when building the Bars, the instrument could not be found in the Cache.

## Type of change

- [x] Bug fix (non-breaking)

## Documentation

## Release notes

## Testing

Manual testing.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/historical/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/historical/client.py
@@ -180,7 +180,7 @@ class HistoricInteractiveBrokersClient:
 
         return list(self._data_client.instrument_provider._instruments.values())
 
-    async def request_bars(
+    async def request_bars(  # noqa: C901
         self,
         bar_specifications: list[str],
         end_date_time: datetime.datetime,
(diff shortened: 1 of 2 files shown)


faysou:
Out of interest for what do you need raw symbology ? I would prefer to remove it at some point as this increases the maintenance complexity.

davidsblom:
I tried with simplified first, but that blew up in my face :-D

davidsblom:
This adapter is the most complex one in the library. Maybe it needs a full rewrite even.

davidsblom:
I'm trying to work with this contract

```
IBContract(
            secType="IND",
            exchange="EUREX",
            symbol="ESTX50",
            build_options_chain=True,
            build_futures_chain=True,
        )
```

faysou:
you should use CONTFUT, not IND, if you use the code of my PR it will work

faysou:
The adapter can do a lot, I thought initially that it required a full rewrite, but no actually, it's just adapted to how the IB api works and a lot of thing work already. It's just that people need to work on extending it when they need something. The adapter is already much cleaner than when I started touching it last year.

Also LLM allows to reduce the complexity and focus less on details and do bigger changes (and I always review after).

Also the adapter is complex because it covers so many products and markets. It's actually why I started working on nautilus, because it had an IB adapter and compatible with historical data with databento. (then I realised that the idea of doing nothing with the IB adapter because it's done was simplistic, I had to do a lot actually to make it work reliably, and still do, it's nice now though that more people are starting to contribute like @shzhng )

davidsblom:
Thanks for the tips. I'll try that. Closing this PR for now.

shzhng:
> Also the adapter is complex because it covers so many products and markets. It's actually why I started working on nautilus, because it had an IB adapter and compatible with historical data with databento. (then I realised that the idea of doing nothing with the IB adapter because it's done was simplistic, I had to do a lot actually to make it work reliably, and still do, it's nice now though that more people are starting to contribute like @shzhng )

Yeah, sadly lots of the complexity just comes from the huge coverage of products/markets, underlying API and also defensive programming against weird edge cases that IB's trading platforms (TWS, IB-GW) experiences with disconnects/reconnects, order handling, etc.

PR #3495, opened 2026-01-27
Add synthetic book support for binary markets

Javdu10:
# Pull Request

## Summary

This is to start a discussion about OwnOrderBook and binary markets... In the polymarket adapter we have 2 IntrumentId for YES and NO side, with the current system they would have their own OwnOrderBook, but to have proper view of the market we need to somehow diff the own bids of the NO side against the YES asks

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

N/A

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/model/src/orderbook/own.rs
+++ b/crates/model/src/orderbook/own.rs
@@ -445,10 +445,6 @@ impl OwnOrderBook {
     ///
     /// Returns [`BinaryMarketBookViewError::BookAndOwnSyntheticBookMustBeDifferentInstrumentId`]
     /// if `self` and `opposite` have the same instrument ID.
-    ///
-    /// # Panics
-    ///
-    /// Panics if parity transformed prices cannot be represented as [`Price`].
     pub fn combined_with_opposite(
         &self,
         opposite: &Self,
@@ -527,7 +523,11 @@ fn log_audit_error(client_order_id: &ClientOrderId) {
         "Audit error - {client_order_id} cached order already closed, deleting from own book"
     );
 }
-
+/// Transforms an order using parity price for binary markets
+///
+/// # Panics
+///
+/// Panics if  [`Price::from_decimal`] panics.
 fn transform_opposite_order(order: OwnBookOrder, side: OrderSideSpecified) -> OwnBookOrder {
     let parity_price = Price::from_decimal(Decimal::ONE - order.price.as_decimal())
         .expect("Invalid parity transformed price for OwnOrderBook::combined_with_opposite");


cjdsellers:
Hi @Javdu10 

Thanks for the PR, I think this is a good use case to support binary/prediction markets - but this needs some further design work for better integration.

Current approach here:
  - Adds a 6th optional parameter to core `OrderBook` methods
  - Embeds binary market-specific logic (1 - price) into a general-purpose component
  - Creates API inconsistency (grouped functions don't have this parameter)

I'm wondering if this functionality would be better as a standalone component rather than adding parameters to `OrderBook`?

Some alternatives to consider:
  - A utility function that combines/transforms `OwnOrderBook`s before passing to existing methods
  - A `BinaryMarketBookView` (or similarly named) wrapper that handles the transformation externally

Let me know your thoughts?

Javdu10:
Hey @cjdsellers 

From my perspective as futur user, I would also be interested in an object onto which I can call other utilities functions like "get_avg_px_for_quantity"

So I would lean for `BinaryMarketBookView`

cjdsellers:
> Hey @cjdsellers
> 
> From my perspective as futur user, I would also be interested in an object onto which I can call other utilities functions like "get_avg_px_for_quantity"
> 
> So I would lean for `BinaryMarketBookView`

Thanks @Javdu10 

Based on this work, I have a good idea about providing some valuable functionality for: 
- Combined own books for binary markets (e.g., `OwnOrderBook::combined_with_opposite()`)
- `OrderBook`s filtered for own orders (generic `filtered_view()` method)
- A couple of binary market extension methods for filtering `OrderBook`

That would then give you all of the standard book analysis methods like the one you mentioned, and avoid the expensive compute of constructing this binary market view every time it's called.

Should be able to complete this over the next couple of days. Fine to leave the PR open for now, many thanks :pray:

Javdu10:
Hello! 

I completely agree with the computation concern. The methods and API mentioned looks good to me. 

I'll keep the PR open for now, and in case i'll need new use cases within the same scope I'll commit the changes 👍 

Thanks again! Looking forward :)

cjdsellers:
Hey @Javdu10

Thanks for the update!

After reviewing, I think we should go back to the original two-piece design from this issue rather than baking the binary market logic directly into `filtered_view()`.

The current approach drops `combined_with_opposite()` and always forces a full `OrderBook` reconstruction -- even when you just need the filtered maps. With the original design:

   1. **`OwnOrderBook::combined_with_opposite()`** — merges the two own books with price transformation (cheap)
   2. **`OrderBook::filtered_view()`** — only needed when you want the full book with utility methods

The combined own book can be passed to the *existing* `bids_filtered_as_map()` / `asks_filtered_as_map()` methods -- no reconstruction needed for the common case. And `filtered_view()` stays general purpose (5 params, no binary-market-specific parameter).

Could you update the PR to restore `combined_with_opposite()` on `OwnOrderBook` and remove the `own_synthetic_book` parameter from `filtered_view()`?

Javdu10:
Hello @cjdsellers , no problems we will go back to pass around the "combined" own book.

Just to make sure, from python is this the API you are looking for ? (can just react thumb up)

```py
yes_own_book: OwnOrderBook | None = self.cache.own_order_book(yes_instrument_id)
no_own_book: OwnOrderBook | None = self.cache.own_order_book(no_instrument_id)

if yes_own_book is None:
    yes_own_book = OwnOrderBook(yes_instrument_id)

combined: OwnOrderBook = yes_own_book.combined_with_opposite(no_own_book)
filtered: OrderBook = self.orderbook.filtered_view(combined)
```

cjdsellers:
Thanks @Javdu10 this is great :+1: .

I appreciate the rework after we iterated on the design. Will merge this now and follow up with a couple of minor things. How has it been working for you in production?

Javdu10:
Thank you @cjdsellers ! The Makefile, thorough tests suite and of course the overall design make it all possible! 🙏 

It has worked wonders on the olympic games 😄. It also reduced my strategy code by 2/3 maybe.

Initially I was still figuring out how it prevents self trades, but combining it with `get_avg_px_for_qty` makes it then obvious!

In the next few days i'll also contribute a small helper onto `OrderBook`, maybe `get_target_px_for_qty`?

It is really usefull for sparse orderbooks, where even ROUND_UP `get_avg_px_for_qty`, does not give the effective next level

EDIT: just saw "get_all_crossed_levels" in the new docs, maybe that's what I need 🤔

Issue #3491, opened 2026-01-27
[RFC] Filter catalog by ts_event instead of ts_init


dxwil:
# [RFC] Filter catalog by ts_event instead of ts_init

When querying the catalog with a start / end time range, currently the filtering is done using `ts_init`. This can cause the following undesirable situation (applies to all data types): 

Steps: 
- Request data for the last week and write it to the catalog (`update_catalog=True`).
- The data client fetches the data today, so `ts_init` is stamped as today.
- Re-run the same request for the last week.
- The catalog correctly detects that data exists (based on filenames using `ts_event`).
- However, when loading the data, it is filtered by start / end using `ts_init`.
- Since `ts_init` is today (outside the requested range), no data is returned, even though the data actually exists.

This is not so much of a problem when importing external data (running it through a wrangler) and writing to the catalog, since then it is possible to tag `ts_init` the same as `ts_event`. But when requesting data from a live data client, the `ts_init` is (rightfully so) tagged as the current time (to preserve latency info) for all the data, and if `update_catalog=True` is used, that is written to the catalog.

Is there any reason why the catalog uses `ts_init` for filtering rather than `ts_event`, because currently I can't think of anything, except the problem that it causes described above?

https://github.com/nautechsystems/nautilus_trader/blob/310f71bb4efb2835a6b46e9c341647689c3036b8/nautilus_trader/persistence/catalog/parquet.py#L1960

faysou:
ts_init should be equal to ts_event for historical data, not to the received time. This is something to fix in the adapter you use. I thought I had reviewed adapters at some point so this is the case. Which adapter do you use ?

dxwil:
Oh that's good to know, thanks! I'm using the Bybit adapter. Let me know if you have time to do a fix.

faysou:
I'm already busy on a lot of stuff and I don't usually touch adapters other than IB and databento, so you will need to do a PR for this or wait that someone else does it.

dxwil:
No worries @faysou, I'll take care of it then.

PR #3502, opened 2026-01-28
Fix Bybit historical requests to use ts_event as ts_init

dxwil:
# Pull Request

## Summary

According to the discussion in #3491, historical data should be tagged with ts_event for ts_init in the adapters. This PR does so for the Bybit adapter. What I noticed though is that all adapters that I briefly looked at (bitmex, deribit, kraken, okx) all generate ts_init for historical data, instead of tagging with ts_event (so someone will need to take care of those as well). 

## Related Issues/PRs

Partial fix for #3491

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/bybit/src/common/parse.rs
+++ b/crates/adapters/bybit/src/common/parse.rs
@@ -557,7 +557,7 @@ pub fn parse_option_instrument(
 pub fn parse_trade_tick(
     trade: &BybitTrade,
     instrument: &InstrumentAny,
-    ts_init: UnixNanos,
+    ts_init: Option<UnixNanos>,
 ) -> anyhow::Result<TradeTick> {
     let price =
         parse_price_with_precision(&trade.price, instrument.price_precision(), "trade.price")?;
@@ -567,6 +567,7 @@ pub fn parse_trade_tick(
     let trade_id = TradeId::new_checked(trade.exec_id.as_str())
         .context("invalid exec_id in Bybit trade payload")?;
     let ts_event = parse_millis_timestamp(&trade.time, "trade.time")?;
+    let ts_init = ts_init.unwrap_or(ts_event);
 
     TradeTick::new_checked(
         instrument.id(),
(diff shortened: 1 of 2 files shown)


faysou:
I think that ts_event should be used for ts_init only if ts_init is not passed as argument which would be the case for requests. So ts_init should be an Option. You can see this in decode.rs in databento. This then allows to have a different ts_init than ts_event in a live setting which is useful for live data to have some latency information.

If you write this comment to an agent it should do the correction I guess.

dxwil:
The parse functions that I modified are only ever used by requests (websocket has it's own parse functions), so ts_init would never be passed to them anyway. So I guess I don't see a clear reason to have it as an option. But of course, if you think that that would just be better API/architecure, I don't mind changing it.

faysou:
Ok I agree with you. We'll see what @cjdsellers thinks when he works on merging this.

cjdsellers:
Hi @dxwil

Thanks for the contribution!

Agree with @faysou here - let's make `ts_init` an `Option<UnixNanos>` rather than removing it entirely. This follows the pattern we use in the Databento adapter, something like:

```rust
  pub fn parse_trade_tick(
      trade: &BybitTrade,
      instrument: &InstrumentAny,
      ts_init: Option<UnixNanos>,
  ) -> anyhow::Result<TradeTick> {
      // ...
      let ts_event = parse_millis_timestamp(&trade.time, "trade.time")?;
      let ts_init = ts_init.unwrap_or(ts_event);
      // ...
  }
```

Then pass `None` for the HTTP client calls.

This keeps the API flexible if these functions are ever used in a different context, and stays consistent with how Databento handles it ([see crates/adapters/databento/src/decode.rs](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/adapters/databento/src/decode.rs)).

cjdsellers:
Thanks for the follow-ups @dxwil :ok_hand:

Issue #3413, opened 2026-01-11
[Interactive Brokers] Subscribing to index does not produce expected price updates


Murph24:
### Expected Behavior

When subscribing to live index ticks in a strategy we should see price updates come through a quoteticks in a strategy.

### Actual Behavior
No quote ticks are produced when changes can be seen occurring in TWS 

### Steps to Reproduce the Problem

1. In a strategy, subsribe to an index:  self.subscribe_quote_ticks(instrument_id=InstrumentId.from_str("^SPX.CBOE"), params={"batch_quotes": True})
2. See that no QuoteTicks with price updates are  produced

It appears the adapter code is looking for bid\ask updates when its receives a tick from the IB api.  This works for most instruments, however index price ticks seem to use "last price" ticks (tickType 4)

OS platform: WSL: Ubuntu-24.04
Python version: 3.13.8
nautilus_trader version: pulled from latest develop branch

faysou:
Last price wouldn't be a quote. Maybe subscribe to trade ticks

Murph24:
@faysou  yeah I tried that as well.  

With: 
`self.subscribe_trade_ticks(instrument_id=InstrumentId.from_str("^SPX.CBOE"))
`

I get: 

2026-01-11T21:07:45.565290359Z [DEBUG] TESTER-001.InteractiveBrokersClient-001: TWS API prepared request: function=reqTickByTickData data={'reqId': 10001, 'contract': IBContract(secType='IND', conId=416904, exchange='CBOE', symbol='SPX', localSymbol='SPX', currency='USD', strike=0.0, comboLegs=[]), 'tickType': 'AllLast', 'numberOfTicks': 0, 'ignoreSize': False}

2026-01-11T21:07:45.565137199Z [ERROR] TESTER-001.InteractiveBrokersClient-001: Failed to request tick-by-tick data.AllLast tick-by-tick requests are not supported for SPX. (code: 10189, req_id=10001)

Murph24:
I think the issue is that most index quote need to be subscribed to with _eclient.reqMktData (only some CME indexes can be subscribed to with tickbytickdata per the IB docs).  The subscription works but the IB adapter but the adapter drops tick type 4, which i think is what you want.  A possible solution would be to implement the IndexPriceUpdate functionality in the IB adapter:

https://github.com/nautechsystems/nautilus_trader/commit/e0dc70cfa18fd5048091f06868dca7c474c9dabb

PR #3514, opened 2026-01-30
Add Interactive Brokers subscribe index price functionality

Murph24:
# Pull Request

## Summary

To subscribe to tick updates for index instruments such as SPX.CBOE, we need to call eclient.reqMktData.  For an index we want the last price (tickType 4)  and not the typical bid\ask tick types.  Implementing the subscribe_index_prices functionality allows for a way to differentiate between typical bid\ask ticks and index price ticks in process_tick_price() of market_data.py

## Related Issues/PRs

Addresses https://github.com/nautechsystems/nautilus_trader/issues/3413

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

None

## Documentation

## Release notes

## Testing
Possible this needs new test coverage?

### DIFF
--- a/tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py
+++ b/tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py
@@ -1173,7 +1173,7 @@ async def test_process_tick_size_negative_size_ignored(ib_client):
     )
 
     # Assert - Negative size should be ignored
-    assert 0 not in ib_client._subscription_tick_data[1]
+    assert TickTypeEnum.BID_SIZE not in ib_client._subscription_tick_data[1]
     ib_client._handle_data.assert_not_called()
 
 
@@ -1202,7 +1202,7 @@ async def test_process_tick_size_extremely_large_size_ignored(ib_client):
     )
 
     # Assert - Extremely large size should be ignored
-    assert 0 not in ib_client._subscription_tick_data[1]
+    assert TickTypeEnum.BID_SIZE not in ib_client._subscription_tick_data[1]
     ib_client._handle_data.assert_not_called()
 
 


faysou:
If only one value is produced would TradeTick be more appropriate?

nautilus_trader/adapters/interactive_brokers/client/market_data.py
```diff
@@ -881,6 +932,38 @@ async def _try_create_quote_tick_from_market_data(
...
+        if req_id not in self._subscription_tick_data:
+            return
+
+        tick_data = self._subscription_tick_data[req_id]
+
+        price = tick_data.get(4)  # IB tick type 4 = LAST_PRICE for index market data
+
+        if price is not None:
+            instrument_id = InstrumentId.from_str(subscription.name[0])
+            instrument = self._cache.instrument(instrument_id)
```
cjdsellers, line 949: We should check for `None` here, log an error and return (common pattern in other adapters). This is clearer than an attribute error on a `NoneType` later.

cjdsellers (commented):
Hi @Murph24 

Thanks for the contribution!

Ideally we need some test coverage on this. In the past this was follow up work maintainers could take on, however with the recent increase in PRs, we have to insist on complete test coverage for new functionality.

One additional comment:
```
 Tick data not cleared after creating update:

  Compare to _try_create_quote_tick_from_market_data:
  # Quote tick clears after use:
  self._subscription_tick_data[req_id].clear()

  But _try_create_index_price_tick_from_market_data doesn't clear. This means every process_tick_price callback will emit an IndexPriceUpdate even if
  the price hasn't changed (the old value stays in tick_data[4]).

```

Murph24:
> If only one value is produced would TradeTick be more appropriate?

@faysou Can you clarify?  Are you saying why not just subscribe to the index with: self.subscribe_trade_ticks(instrument_id=InstrumentId.from_str("^SPX.CBOE")) ?

This does not work with the adapter as-is

-or-

 are you suggesting to implement the SubscribeIndexPrice, but have it emit a TradeTick?   That wouldn't seem correct to me, but you guys know this adapter much better than I do.

Murph24:
> Hi @Murph24
> 
> Thanks for the contribution!
> 
> Ideally we need some test coverage on this. In the past this was follow up work maintainers could take on, however with the recent increase in PRs, we have to insist on complete test coverage for new functionality.
> 
> One additional comment:
> 
> ```
>  Tick data not cleared after creating update:
> 
>   Compare to _try_create_quote_tick_from_market_data:
>   # Quote tick clears after use:
>   self._subscription_tick_data[req_id].clear()
> 
>   But _try_create_index_price_tick_from_market_data doesn't clear. This means every process_tick_price callback will emit an IndexPriceUpdate even if
>   the price hasn't changed (the old value stays in tick_data[4]).
> ```

Thanks for taking a look!  Yes this would make sense, however I don't see this anywhere else in the adapter? 
>  Compare to _try_create_quote_tick_from_market_data:
>   Quote tick clears after use:
>   self._subscription_tick_data[req_id].clear()

faysou:
I don't know things much better than you, I'm just a contributor like you. I'm just saying that a quote is associated to a bid ask and last prices to trade ticks so it would be better to subscribe to trade ticks in this case and then there would maybe be nothing to change.

Can you test if subscribing to trade ticks for the index works ?

Murph24:
@faysou  I did try just subscribing to trade ticks, but it does not work.  (See linked bug for error output).  The _subscribe_trade_ticks function calls reqTickByTickData.  The IB api docs seem to confirm reqTickByTickData does not work for indices (with CME indices being an exception): https://ibkrcampus.com/campus/ibkr-api-page/twsapi-doc/#tick-by-tick - "Tick-by-tick data for indices is only provided for indices which are on CME."

So we could shoehorn this special case into the _subscribe_trade_ticks function and call reqMktData, but I felt it was cleaner to just implement what I've done here.

faysou:
I prefer what you just said as last ticks are associated with trade ticks in nautilus.

You could check that a symbol of an instrument id starts with ^ for an index. Or actually better, just check access the instrument at subscribe time and check if it's an index.

shzhng:
Ran into this issue myself. I would actually prefer the way @Murph24 has done it here, @faysou. SPX (and VIX, NDX, RUT, etc) itself is not a tradable instrument, but merely an index that index options are resolved around, and therefore wouldn't have any trade price/ticks - there wouldn't be any size/qty associated with it, though we could make it hacky with 0 size

It seems that Index Prices was a construct that was added to NT for this kind of use case no? I see that some of the other data adapters (specifically the crypto ones) also now implement `_subscribe_index_prices`, which I presume are also to track untradable aggregate measures without trade ticks?

@cjdsellers let us know if you have a preferred direction on this

nautilus_trader/adapters/interactive_brokers/client/market_data.py
Murph24: Added

Murph24:
After thinking about this more, I agree with @shzhng's commments.  This seems best to implement using NT's existing IndexPriceUpdate.

Thanks for all comments.  I committed some changes based on feedback:

- Address comments
- Add tests and fixed existing to accommodate this change
- Added SPX test instrument
- Updated the SubscribeStrategy to add IndexPriceUpdate

Couple thing that have not been addressed 

- Per IBs docs, CME indices can be subscribed to using reqTickbyTickData, which is likely more "real" tick flow.  This code calls reqMktData which will work for CME indies,  however this method returns tick data that is aggregated over a small time period (as I understand it).  We could handle this special case if its of interest.

nautilus_trader/adapters/interactive_brokers/client/market_data.py
```diff
@@ -881,6 +932,42 @@ async def _try_create_quote_tick_from_market_data(
...
+        self,
+        subscription: Subscription,
+        req_id: int,
+    ) -> None:
+        if req_id not in self._subscription_tick_data:
+            return
+
+        tick_data = self._subscription_tick_data[req_id]
+
+        price = tick_data.get(4)  # IB tick type 4 = LAST_PRICE for index market data
```
shzhng: is there an enum value that can be used here instead of the hardcoded `4`?

shzhng: same comment applies elsewhere such as in tests for hardcodes `1` and `4`

Murph24: yes there is an existing enum in the ibapi.  Fixed now

cjdsellers:
Thanks everyone :pray:

PR #3526, opened 2026-02-02
Refine memory management for PyCapsule

faysou:
# Pull Request

### Summary

Fixes the destructor for the `PyCapsule` created in `data_to_pycapsule` so that the Rust `Data` value is explicitly dropped when Python releases the capsule, preventing memory leaks when passing data from adapters (including databento) into the Python runtime.

### Changes

- **`crates/model/src/python/data/mod.rs`**
  - Updated the destructor passed to `PyCapsule::new_with_destructor` from `|_, _| {}` to `|data, _| drop(data)`.
  - The previous destructor ignored its arguments and did nothing, so the encapsulated `Data` was never freed when the capsule was released.
  - The new destructor receives ownership of the `Data` value and calls `drop(data)`, ensuring correct cleanup.
  - Adjusted the inline comment to state that the destructor receives ownership of the value and drops it when the capsule is released by Python.

### Rationale

`data_to_pycapsule` is used by multiple adapters (databento, binance, bybit, deribit, kraken, hyperliquid, okx, tardis, dydx, coinbase_intx, bitmex) to pass Rust `Data` into Python. When Python no longer holds a reference to the capsule, the destructor is invoked. Without explicitly dropping the `Data` in that destructor, the memory was not reclaimed. This change aligns behaviour with Rust ownership and ensures no leak for each capsule created.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/model/src/python/data/mod.rs
+++ b/crates/model/src/python/data/mod.rs
@@ -89,9 +89,9 @@ impl DataType {
 /// Rust data structure.
 #[must_use]
 pub fn data_to_pycapsule(py: Python, data: Data) -> Py<PyAny> {
-    // Register a destructor which simply drops the `Data` value once the
-    // capsule is released by Python.
-    let capsule = PyCapsule::new_with_destructor(py, data, None, |_, _| {})
+    // Register a destructor which drops the `Data` value when the capsule is
+    // released by Python. The destructor receives ownership of the value.
+    let capsule = PyCapsule::new_with_destructor(py, data, None, |data, _| drop(data))
         .expect("Error creating `PyCapsule`");
     capsule.into_any().unbind()
 }


shzhng:
Will test this during market hours tomorrow and report back

shzhng:
Unfortunately didn't fix my issue and memory continues to grow unbounded

shzhng:
i think this probably should still be kept though, because this is semantically correct vs the previous

faysou:
I think it would help to know what you are doing to observe this growth in memory.

faysou:
The increase in memory could also simply be the caches filling up form market data.

cjdsellers:
The `capsule_destructor` function in pyo3 does this:
```
  let CapsuleContents::<T, F> {
      value, destructor, ..
  } = *unsafe { Box::from_raw(ptr.cast()) };
```

Then it calls `destructor(value, ctx)`;

The value is moved into the destructor by ownership (the destructor is `FnOnce(T, ...)`). With `|_, _| {}`:
- The call takes ownership of the value, `_` just doesn’t bind it
- Since `_` doesn’t bind, the value is dropped immediately when the closure returns

pyo3 itself uses `|_, _| {}` as the default destructor. If that didn’t drop, pyo3 would leak in its own API.

Since the semantics are unchanged, this is a no-op - more of a clarity improvement than a correctness fix. Could still be worth merging though, as `_` drop behavior is subtle and the explicit drop() is clearer...

faysou:
I think there's no point in merging it it changes nothing.

PR #3562, opened 2026-02-09
Fix parsing options for Stoxx50 with Interactive Brokers

davidsblom:
# Pull Request

## Summary

Fix parsing options for Stoxx50 with Interactive Brokers

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

Added test case and manual testing.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
+++ b/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
@@ -270,7 +270,7 @@ RE_EUREX_OPT = re.compile(
     r"(?P<tradingClass>[A-Z0-9]{3,6})\s+"
     r"(?P<expiry>\d{8})\s+"
     r"(?P<strike>\d+(?:\.\d+)?)"
-    r"(?:\s+(?P<style>[A-Z]))?$"
+    r"(?:\s+(?P<style>[A-Z]))?$",
 )
 
 


cjdsellers:
Thank you @davidsblom :pray:

nautilus_trader/adapters/interactive_brokers/providers.py
```diff
@@ -562,7 +562,7 @@ async def get_contract_details(
         max_expiry = utc_now + pd.Timedelta(days=max_expiry_days)
 
         if (
-            contract.secType == "CONTFUT"
+            contract.secType in {"CONTFUT", "IND"}
```
shzhng, line 565: this will be a problem, since for US indexes we request instruments such as this:

```python
                    {
                        "secType": "IND",
                        "symbol": "SPX",
                        "exchange": "CBOE",
                        "build_options_chain": True,
                        "min_expiry_days": 0,
                        "max_expiry_days": 4,
                    }
```

and SPX isn't a future, so I wouldn't expect or want the following block to be called?

what's the contract specification that you use?

faysou: I agree with @shzhng , @davidsblom had put this instruction before as well and I removed it. Contfut should be used to query futures instruments, not Ind.

davidsblom: Hi both, is there a way to test from Europe? I'm running into errors related to my location when testing SPY.

faysou:
elif exchange in VENUES_EUREX_OPT and (m := RE_EUREX_OPT.match(instrument_id.symbol.value)):
        return IBContract(
            secType="OPT",
            exchange=exchange,
            symbol=m["tradingClass"],
            currency="EUR",
            tradingClass=m["tradingClass"],
            right=m["right"],
            strike=m["strike"],
            lastTradeDateOrContractMonth=m["expiry"],
        )
        
I don't think it makes sense to change the code like this for just one particular case. We should find a more general way, we can't have a particular case just for one exchange as this likely happens for other symbols as well.

faysou:
I've just done a follow-up PR

PR #3570, opened 2026-02-11
architect_ax: reject cancel when venue order id is unknown

andrew-cho-architect:
## Summary
- remove cancel fallback that sent client order IDs as venue `oid`
- return deterministic client-side error when `venue_order_id` is missing
- keep cancel requests wire-safe by sending only known venue order IDs

## Architect Docs
- Orders WebSocket API (cancel request expects venue order id `oid`): https://docs.architect.exchange/api-reference/order-management/orders-ws

## Files
- `crates/adapters/architect_ax/src/websocket/orders/client.rs`

## Testing
- Added tests for missing-venue-id rejection and positive cancel command path.
- Ran: `cargo test -p nautilus-architect-ax --locked`
- Result: PASS on branch `acho/ax-error-03-cancel-unknown-oid`

### DIFF
--- a/crates/adapters/architect_ax/src/websocket/orders/client.rs
+++ b/crates/adapters/architect_ax/src/websocket/orders/client.rs
@@ -653,7 +653,7 @@ impl AxOrdersWebSocketClient {
 
     /// Cancels an order via WebSocket.
     ///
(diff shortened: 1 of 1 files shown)


cjdsellers:
Please consolidate into a single PR.

PR #3571, opened 2026-02-11
architect_ax: clean order tracking maps on successful terminal flows

andrew-cho-architect:
## Summary
- remove pending request tracking entry when place-order response succeeds
- clean terminal tracking maps after filled events
- ensure rejected path also clears venue-to-client mapping
- add regression tests for pending cleanup and tracking map cleanup behavior

## Architect Docs
- Orders WebSocket API (place/cancel responses and order lifecycle events): https://docs.architect.exchange/api-reference/order-management/orders-ws

## Files
- `crates/adapters/architect_ax/src/websocket/orders/handler.rs`

## Concrete Example (Why This Matters)
1. A place-order command is sent and the handler stores `pending_orders[rid]`.
2. Venue returns successful place response for that `rid`.
3. Before this PR, the success response path did not remove `pending_orders[rid]`.
4. The order later reaches a terminal state (for example, filled).
5. Before this PR, terminal fill flow emitted `OrderFilled` but did not clear all tracking maps (`orders_metadata`, `venue_to_client_id`, `cid_to_client_order_id`).
6. Repeating this normal lifecycle over many orders caused stale in-memory tracking state to accumulate over long-running sessions.

This PR makes the success and terminal cleanup paths consistent so tracking state is released when order lifecycles complete.

## Testing
- Added handler-level regression tests.
- Ran: `cargo test -p nautilus-architect-ax --locked`
- Result: PASS on branch `acho/ax-error-04-order-tracking-cleanup`

### DIFF
--- a/crates/adapters/architect_ax/src/websocket/orders/handler.rs
+++ b/crates/adapters/architect_ax/src/websocket/orders/handler.rs
@@ -435,6 +435,7 @@ impl FeedHandler {
         match resp {
             AxWsOrderResponse::PlaceOrder(msg) => {
                 log::debug!("Place order response: rid={} oid={}", msg.rid, msg.res.oid);
+                self.pending_orders.remove(&msg.rid);
                 Some(vec![AxOrdersWsMessage::PlaceOrderResponse(msg)])
             }
             AxWsOrderResponse::CancelOrder(msg) => {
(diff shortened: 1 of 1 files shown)


cjdsellers:
Please consolidate into a single PR.

PR #3595, opened 2026-02-14
Migrate IB adapter to ibapi 10.43.2

faysou:
# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [ ] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Migrate Interactive Brokers adapter to ibapi 10.43

Upgrades the IB adapter dependency from `nautilus-ibapi` 10.37.2 to 10.43.2, bringing protobuf message support, new contract/order fields, and compatibility with the latest TWS API.

### Changes

- **Dependency upgrade**: Bump `nautilus-ibapi` from 10.37.2 to 10.43.2 and add `protobuf>=5.29.5` as a required dependency for the `ib` extra, since ibapi 10.43's decoder imports protobuf-generated modules at the module level.

- **EClient attribute rename**: Update `_initialize_connection_params()` in `client/connection.py` to use `self._eclient.host` / `self._eclient.port` (public attributes) instead of the previous `self._eclient._host` / `self._eclient._port` (private attributes), matching the attribute naming change in ibapi 10.43's `EClient`.

- **New IBContractDetails fields**: Add `minAlgoSize` (Decimal), `eventContract1`, `eventContractDescription1`, and `eventContractDescription2` (str) to `IBContractDetails` in `common.py` to match the fields added to ibapi 10.43's `ContractDetails`. Also include `minAlgoSize` in the decimal deserialization list within `dict_to_contract_details`.

- **Test fixture fix**: Set `serverVersion_ = 0` on the mocked `EClient` instances in the `data_client` and `exec_client` test fixtures. ibapi 10.43 added `useProtoBuf()` guard checks to nearly all `EClient` methods, which compare `serverVersion_` against protobuf version thresholds. When `serverVersion_` is `None` (the default for an unconnected EClient), this comparison raises `TypeError`. Setting it to `0` ensures `useProtoBuf()` returns `False` in tests, preserving the legacy code path.

### Non-breaking additions available in ibapi 10.43 (for future use)

- Full protobuf message support (transparent via decoder).
- New EWrapper callbacks: `userInfo()`, `currentTimeInMillis()`.
- New Order fields: `slOrderId`, `slOrderType`, `ptOrderId`, `ptOrderType`, `whatIfType`, `submitter`, `manualOrderIndicator`.
- New Execution field: `optExerciseOrLapseType` with `OptionExerciseType` enum.
[...]

PR #3481, opened 2026-01-25
Refine Interactive Brokers query of option chains

faysou:
# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [ ] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Refine option chain queries and execution consistency for Interactive Brokers

This PR refines the Interactive Brokers (IB) adapter, specifically focusing on the reliability of option chain queries for Index Options and improving the consistency of order tracking during execution.

### Core IB adapter improvements

* Updated `InteractiveBrokersInstrumentProvider` to handle option chain building more robustly, specifically preventing recursive lookups when encountering `CONTFUT` contracts within the chain construction process.

* Standardized the use of `venue_order_id` in `InteractiveBrokersExecutionClient` by utilizing the ID stored on the `nautilus_order` instead of re-generating it from execution details, ensuring consistent tracking across orders and combo legs.

* Modified `IBContract` to allow `strike` to be a `float` or a `str` (defaulting to an empty string), providing more flexibility for various contract definitions.

* Enhanced regex patterns in `instruments.py` for parsing Future Options (`FOP`), supporting longer strike price strings and varied decimal precision.

* Improved filled quantity validation in the execution client by explicitly using `Decimal` for precision.

### Client and logging

* Added placeholder (commented-out) logic in `InteractiveBrokersClient` for enhanced error tracing of asynchronous tasks, facilitating future debugging of complex message handling failures.

### Examples and documentation

* Updated the `with_databento_instrument_id_example.py` notebook to demonstrate Index Option chain queries and refined the strategy's bar request logic and logging configuration.

## Rationale

The primary goal was to address issues where Index Option chains were not being correctly loaded or filtered. By refining the `CONTFUT` handling and expanding regex support, the adapter can now handle a wider range of IB contract specifications. The standardization of `venue_order_id` ensures that reconciliation and trade reporting remain consistent even for complex order types like spreads.

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/3468

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Improvement (non-breaking)
- [ ] Breaking change (impacts existing behavior)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

- [ ] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)

## Release notes

[...]

PR #3599, opened 2026-02-17
Fix Interactive Brokers ibapi 10.43 protobuf compatibility

shzhng:
# Pull Request

## Summary

`ibapi 10.43` (#3595) introduced a dual protocol (legacy string-based and protobuf). This was actually already the case in `10.37` I believe. TWS/Gateway now responds in protobuf format regardless of request format, which introduces two bugs in the ibapi protobuf decoder that break NautilusTrader:

1. **`IBContract.strike` default**: The protobuf decoder decodes unset double fields as `0.0` rather than `UNSET_DOUBLE` (`sys.float_info.max`). When `strike=0.0` leaks into a subsequent `reqContractDetails` call, IB silently ignores the request (zero bytes back). Fixed by changing the `strike` field type from `float | str = ""` to `float = UNSET_DOUBLE` and adding `__post_init__` to normalize `0.0` → `UNSET_DOUBLE` on all construction paths. @faysou previously fixed [here](https://github.com/nautechsystems/nautilus_trader/pull/3481/changes#diff-c5dcd205aaeee17fb74e5e2bf57a7fea63d7a268cb4205f578c81bb1a0909180R142), but it broke again with `10.43`. I made a proper fix.

2. **`ContractDetails.underConId` typo**: The protobuf decoder (`decoder.py:decodeContractDetails`) writes `contractDetails.underConid` (lowercase 'i') instead of the canonical `contractDetails.underConId` (uppercase 'I'). This causes option chain building to silently return 0 contracts (the `underConId` filter always evaluates `0 == <conId>` → False), and `IBContractDetails(**details.__dict__)` to fail with `TypeError: Unexpected keyword argument 'underConid'`. Fixed via a `from_contract_details` classmethod that normalizes the dict key before construction, and a `getattr` fallback in the option chain filter for raw ibapi objects.

## Type of change

- [x] Bug fix (non-breaking)

N/A — `IBContract.strike` type changed from `float | str = ""` to `float = UNSET_DOUBLE`, but all existing callers already pass numeric values or omit the field. The one regex-parsed string caller is now wrapped with `float()`.

## Documentation

## Release notes

## Testing

Tested manually against IB Gateway with ibapi 10.43 — confirmed option chains load correctly with protobuf responses.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
+++ b/nautilus_trader/adapters/interactive_brokers/parsing/instruments.py
@@ -1064,7 +1064,8 @@ def ib_contract_to_instrument_id_simplified_symbology(  # noqa: C901 (too comple
         if contract.localSymbol:
             symbol = contract.localSymbol
         else:
-            symbol = f"{contract.right} {contract.tradingClass} {contract.lastTradeDateOrContractMonth} {contract.strike}"
+            strike_str = f"{contract.strike:g}"
+            symbol = f"{contract.right} {contract.tradingClass} {contract.lastTradeDateOrContractMonth} {strike_str}"
     elif security_type == "CONTFUT":
         symbol = contract.symbol
     elif security_type == "FUT":
--- a/tests/integration_tests/adapters/interactive_brokers/test_parsing.py
+++ b/tests/integration_tests/adapters/interactive_brokers/test_parsing.py
@@ -107,7 +107,7 @@ simplified_symbology_params = [
             symbol="OESX",
             tradingClass="OESX",
             lastTradeDateOrContractMonth="20260213",
-            strike=4775.0,
+            strike=4775,
             right="C",
         ),
         "C OESX 20260213 4775.EUREX",


shzhng (commented):
Note on the `strike` type change (`float | str = ""` → `float = UNSET_DOUBLE`):

Upstream had `float | str = ""` but ibapi 10.43 protobuf always sends numeric strike values. Using `UNSET_DOUBLE` matches ibapi's own `Contract.strike` default (`UNSET_DOUBLE` in `ibapi/contract.py`). The one caller that passed a string (regex-parsed strike in `instruments.py:1247`) is now wrapped with `float()`.

nautilus_trader/adapters/interactive_brokers/common.py
```diff
@@ -278,6 +291,32 @@ class IBContractDetails(NautilusConfig, frozen=True, repr_omit_defaults=True):
...
+        the constructor sees it — hence this is a classmethod rather than __post_init__.
+
+        """
+        contract_details.contract = IBContract(**contract_details.contract.__dict__)
+        d = contract_details.__dict__.copy()
+        if "underConid" in d:
+            if not d.get("underConId"):
+                d["underConId"] = d.pop("underConid")
+            else:
+                d.pop("underConid")
```
shzhng, line 317: https://github.com/nautechsystems/nautilus_ibapi/blob/6023da4597082b8e3e762a40cb056982b79ea434/ibapi/decoder_utils.py#L478

this is just embarrassing on IBKR's part.

i'll submit a PR to fix once I get access to the official upstream ibapi repo.

This is just defensive programming that should still work once we upgrade with a fix, but can be removed later on

nautilus_trader/adapters/interactive_brokers/common.py
```diff
@@ -124,7 +126,7 @@ class IBContract(NautilusConfig, frozen=True, repr_omit_defaults=True):
     multiplier: str = ""
 
     # options
-    strike: float | str = ""
+    strike: float = UNSET_DOUBLE
```
shzhng, line 129: ibapi 10.43 introduced some input validation that made `""` invalid

@faysou this is related to your change here https://github.com/nautechsystems/nautilus_trader/pull/3481/changes#r2737486167

shzhng:
@faysou @pandashark @cjdsellers

cjdsellers:
Thank you @shzhng

PR #3613, opened 2026-02-21
Add catalog deduplication functionality

ms32035:
# Pull Request

## Summary

The PR adds deduplication to the Rust parquet catalog to bring feature parity with the Python implementation.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

- [X] I added/updated tests to cover new or changed logic

### DIFF
--- a/crates/persistence/src/backend/catalog.rs
+++ b/crates/persistence/src/backend/catalog.rs
@@ -725,17 +725,12 @@ impl ParquetDataCatalog {
                     // Extract instrument_id from directory path (last component)
                     let path_parts: Vec<&str> = dir_path.split('/').collect();
                     if let Some(instrument_id_dir) = path_parts.last() {
-                        // Decode percent-encoded dir name from object store
-                        // (e.g. "%5ESPX.CBOE" → "^SPX.CBOE")
-                        let decoded_dir = urlencoding::decode(instrument_id_dir)
-                            .unwrap_or(Cow::Borrowed(instrument_id_dir));
-
                         // Apply filter if provided
                         if let Some(ref ids) = instrument_ids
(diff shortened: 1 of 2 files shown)


ms32035:
```
    pub fn list_parquet_files(&self, directory: &str) -> anyhow::Result<Vec<String>> {
        self.execute_async(async {
            let prefix = ObjectPath::from(format!("{directory}/"));
            let mut stream = self.object_store.list(Some(&prefix));
            let mut files = Vec::new();

            while let Some(object) = stream.next().await {
                let object = object?;
                if object.location.as_ref().ends_with(".parquet") {
                    files.push(object.location.to_string());
                }
            }
            Ok::<Vec<String>, anyhow::Error>(files)
        })
    }
```
https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/persistence/src/backend/catalog.rs#L1057
@faysou looks like the above function has the same issue with double `%` encoding and it returns nothing in such cases, but I can't figure out what your original fix was. Can you assist?

faysou:
If you look for % in catalog.rs you will find where it talks about it.

ms32035:
@faysou 
I did a bit more digging and looks like this comes out in a few more places. Maybe an easier and safer solution would be to undo your change and do something like this instead:
```
pub(crate) fn urisafe_instrument_id(instrument_id: &str) -> String {
    instrument_id.replace('/', "").replace("^","_")
}
```

faysou:
You are welcome to modify and test it, I'm fine with it.

cjdsellers:
Thank you @ms32035

PR #3647, opened 2026-02-27
Include .cargo config in sdist for correct platform builds

qinhanlei:
.cargo/config.toml contains platform-specific Cargo rustflags for Linux (link optimizations), macOS (dynamic_lookup), and Windows (crt-static). Poetry-core does not auto-include dot-directories, so the file was missing from the sdist, causing builds from source to use Cargo defaults instead of the intended flags.


cjdsellers:
Hi @qinhanlei 

Thank you for the fix. It's great that the build can work on macOS Intel machines after this.

PR #3648, opened 2026-02-27
Fix macOS source build missing RUSTFLAGS

qinhanlei:
The macOS linker (ld64) rejects undefined symbols by default. PyO3 extension modules reference Python C API symbols that are only resolved at runtime when Python loads the shared library, requiring the -undefined dynamic_lookup linker flag.

build.py already sets RUSTFLAGS for Linux ARM64 builds; this adds the equivalent for all macOS targets. The flag is appended to any existing RUSTFLAGS to preserve values set by the caller. CI was unaffected because it explicitly exports the flag in the workflow action.

This PR works in conjunction with #3647 to enable nautilus_trader to run on macOS Intel machines.


cjdsellers:
Hi @qinhanlei 

Closing this one as #3647 already sets the macOS `dynamic_lookup` flags for both x86_64 and aarch64 targets -- it just wasn't being included in the sdist.

PR #3694, opened 2026-03-08
Add interval field to FundingRateUpdate

### LINKED ISSUE #3676: Funding interval
# Feature Request

## Problem Statement

The funding rate without funding interval/period doesn't give enough context to determine what the actual return from funding is.

### Proposed Solution

I suggest adding `interval` or `period` to `FundingRateUpdate`. Verified that Bybit, OKX (didn't look at others, some are fixed to 1h or 8h always) all return this data together with the funding rate.

### Example Usage

```python
val funding_rate: FundingRateUpdate
funding_rate.interval
```

### Additional Context

The exchanges I checked do not return this field with historical funding rates, but by requesting one more funding rate than the user actually wants, it is possible to determine the funding interval by subtracting funding times.

I can make a PR for all of this if the change would be accepted.

dxwil:
# Pull Request

## Summary

Add `interval` field to `FundingRateUpdate`. This allows calculating APR of the funding rate. For subscriptions, the interval is taken from WebSocket messages for exchanges that return it. For historical requests, no exchanges return the interval, so it is calculated locally as the time difference between data points.

## Related Issues/PRs

Closes #3676

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

Funding subscriptions and requests were tested manually for all changed adapters.

### DIFF
--- a/crates/adapters/bybit/src/http/client.rs
+++ b/crates/adapters/bybit/src/http/client.rs
@@ -17,6 +17,17 @@
 //!
 //! Bybit API reference <https://bybit-exchange.github.io/docs/>.
 
+use std::{
+    cmp::Reverse,
+    collections::HashMap,
+    fmt::Debug,
+    num::NonZeroU32,
+    sync::{
+        Arc, LazyLock,
+        atomic::{AtomicBool, Ordering},
+    },
+};
+
 use ahash::{AHashMap, AHashSet};
 use chrono::{DateTime, Utc};
 use dashmap::DashMap;
@@ -40,16 +51,6 @@ use nautilus_network::{
 };
 use rust_decimal::Decimal;
 use serde::{Serialize, de::DeserializeOwned};
-use std::cmp::Reverse;
-use std::{
-    collections::HashMap,
-    fmt::Debug,
-    num::NonZeroU32,
-    sync::{
-        Arc, LazyLock,
-        atomic::{AtomicBool, Ordering},
-    },
-};
 use tokio_util::sync::CancellationToken;
 use ustr::Ustr;
 


cjdsellers:
Hi @dxwil 

Thanks for the contribution!

Adding interval to `FundingRateUpdate` makes sense. A few things to sort out:

**Cap'n Proto**: The new `TimeDelta` struct exists only for this one field -- a plain `Int64` would do.

**Representation**: Consider `u16` minutes instead of `chrono::TimeDelta`. No venue uses sub-minute funding intervals, and it avoids pulling chrono into model and serialization crates.

**Sourcing**: Most venues have a fixed funding interval (8h for Binance/BitMEX/Bybit/OKX/Kraken, 1h for dYdX/Hyperliquid). A venue-level default would be simpler than computing from consecutive timestamps or gating on optional WS fields. The `windows(2)` logic in Bybit/OKX HTTP clients and the `funding_interval_hour.is_some()` guard in the Bybit WS handler could all go away. Keep it optional for venues like Deribit where there's no fixed interval.

**Equality**: `Cython __eq__` compares `self.interval == self.interval` instead of `other.interval`.

dxwil:
Hey @cjdsellers,

Thanks for the notes! The problem I'm trying to fix is specifically that most venues are moving away from fixed funding intervals. [Binance](https://www.binance.com/en/futures/funding-history/perpetual/real-time-funding-rate), [Bybit](https://www.bybit.com/en/announcement-info/fund-rate), [OKX](https://www.okx.com/docs-v5/en/#public-data-websocket-funding-rate-channel) (see note under push data parameters), BitMEX (all current futures are 8h but API returns interval field so can be variable) are now changing the funding interval according to market conditions. On Bybit for example, more pairs have 4h funding than 8h. Only dYdX and Hyperliquid clearly state that the funding interval is always 1h. That's why I'm getting the interval from WS where it is returned and building it locally for historical requests.

For the `interval` type, I can change it to `u16` minutes, no problem.

dxwil:
- Changed interval to `u16` minutes.
- Moved `interval` field to the end of `FundingRateUpdate` capnp schema (so it doesn't change existing field order).
- Other than the tests passing, live tested BitMEX subscriptions and Bybit subscription and requests for returning the correct interval.

NOTE: I did not live test the changes made to OKX subscriptions and requests, since in the EU OKX does not offer futures.

Should be ready to merge from my side, unless there are more suggestions.

crates/adapters/bybit/src/http/client.rs
```diff
@@ -2979,89 +2988,66 @@ impl BybitHttpClient {
...
+            let earliest_funding_time = match funding_rates.last() {
+                Some(last) => last
+                    .funding_rate_timestamp
+                    .parse::<i64>()
+                    .map_err(|_| anyhow::anyhow!("invalid funding_rate_timestamp"))?,
+                None => break,
+            };
 
-            if !has_new {
+            if let Some(end_time) = current_end
```
dxwil: This is only needed to pass the unit test, because the test server does not support request params and always returns the same data. In a live environment the `endTime` would change, an empty result would be returned and the loop would break by the condition 4 lines above.

cjdsellers:
Hi @dxwil 

Thanks for the updates! Getting close. A few more things:

**BitMEX `CENTURY_UNIX_NANOS`**: The epoch-subtraction approach is fragile. BitMEX returns `funding_interval` as a datetime like `"2000-01-01T08:00:00.000Z"` -- we could just read the hours/minutes directly:

```rust
let hours = msg.funding_interval.hour();
let minutes = msg.funding_interval.minute();
let interval = Some((hours * 60 + minutes) as u16);
```

**Cap'n Proto field ordering**: Since `capnp` is not stable yet, we can order fields logically rather than appending for backwards compat. `interval` belongs after `rate`, not tacked on as `@5`:

```capnp
struct FundingRateUpdate {
    instrumentId @0 :Text;
    rate @1 :Base.Decimal128;
    interval @2 :UInt16;
    nextFundingTime @3 :Base.UnixNanos;
    tsEvent @4 :Base.UnixNanos;
    tsInit @5 :Base.UnixNanos;
}
```

Which then mirrors the Rust struct field order.

**OKX interval computation**: `parse_funding_rate` takes `interval_millis: Option<u64>`, but in the HTTP client the diff `raw.funding_time - window[1].funding_time` is unsigned subtraction on `u64`. If data arrives misordered this wraps silently. Better to use `checked_sub` or `i64` for safety.

**Bybit pagination**: The refactored HTTP client drops the `seen_timestamps` dedup set. If Bybit returns overlapping data across pages you could get duplicates. Also the limit semantics changed -- no limit + no start now clamps to 200 (one page) instead of paginating. Worth another look.

**Binance**: The mark price stream does return `fundingInterval` but this PR doesn't parse it. Not a blocker but worth noting as a follow-up.

dxwil:
Thanks for the notes again @cjdsellers, made the fixes you wrote about.

I would like to add interval support for Binance before merging as well, but I don't see where it's returned by the mark price stream. The documentation for that stream is [here](https://developers.binance.com/docs/derivatives/usds-margined-futures/websocket-market-streams/Mark-Price-Stream), I also verified that the docs are correct (no undocumented field). A `fundingIntervalHours` field is returned by [this](https://developers.binance.com/docs/derivatives/usds-margined-futures/market-data/rest-api/Get-Funding-Rate-Info) http endpoint, but it's not much use since it's not websocket. Could you share where you found `fundingInterval` in a websocket stream?

P.S. I noticed a bug while testing: https://github.com/nautechsystems/nautilus_trader/issues/3707

cjdsellers:
Hi @dxwil

You're right, this looks like a bug in the Bybit WS funding path. We only use the cached fields to detect change, then parse the sparse ticker frame directly, so partial updates can drop `next_funding_ns` and `interval`.

I'll follow up by caching the full funding state before emitting `FundingRateUpdate`.

Let me know if you spot anything else around the funding path.

dxwil:
Great! So that’a pretty much it, I would say ready to merge. Unless you can give more info on the Binance `fundingInterval` you found, then I would add parsing for that too.

cjdsellers:
Thanks @dxwil 

I'll follow up on these things post merge.

Issue #3183, opened 2025-11-15
Panic when loading Nautilus Catalog generated from Databento L3 MBO Catalog


marloncalvo:
# Bug Report

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates.
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

### Expected Behavior

Generating Nautilus Catalog from Databento L3 MBO catalog, and subsequently querying Nautilus Catalog should work,

```py
from nautilus_trader.persistence.catalog import ParquetDataCatalog
from nautilus_trader.adapters.databento.loaders import DatabentoDataLoader

loader = DatabentoDataLoader()
dbn_data_loaded = loader.from_dbn_file(
    path=dbn_path,
    instrument_id=InstrumentId.from_str("PLTR.XNAS"),
    as_legacy_cython=False,
)
catalog = ParquetDataCatalog("catalog")
catalog.write_data(dbn_data_loaded)
deltas = catalog.order_book_deltas(instrument_ids=["PLTR.XNAS"])
```

### Actual Behavior

Instead, the following panic is raised from ` catalog.order_book_deltas(...)`:
```
07:23:43.467 [error] Disposing session as kernel process died ExitCode: 3221226505, Reason: 
thread 'tokio-runtime-worker' (26184) panicked at crates\\model\\src\\types\\price.rs:193:14:
Condition failed: `precision` must be 0 when `raw` is PRICE_UNDEF
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is because we use `RecordBatch.price_precision` for `PRICE_UNDEF` cases as well - which of course will lead to mismatch between `R` (reset) and other records. For example, L3 MBO typical contains `precision = 2` but we expect `precision = 0` for `PRICE_UNDEF`. If row_0 == `R`, and row_1 == `A`, it will lead to panic.

https://github.com/nautechsystems/nautilus_trader/blob/ea4754530bcc691b639042ef553bec4c54ce98af/crates/model/src/types/price.rs#L187-L194

[...]

PR #3697, opened 2026-03-08
Fix PRICE_UNDEF panic in OrderBookDelta.to_pyo3_list

zshuang15:
## Summary

- `OrderBookDelta.to_pyo3_list()` caches `price_prec` from the first non-zero delta and applies it to all subsequent deltas
- When a CLEAR (action=4) message carries `PRICE_UNDEF` (which requires `precision=0`), the cached precision (e.g. 2) is incorrectly used
- This triggers a Rust panic in `Price::from_raw`: `"precision must be 0 when raw is PRICE_UNDEF"`
- The fix in #3183 addressed the DBN decode path but missed this Cython→pyo3 conversion path, which is hit during Arrow serialization (`catalog.write_data`)

## Fix

Use the delta's own precision when it is 0 (sentinel value), otherwise use the cached precision as before. Same logic applied to both price and size.

## Reproduction

Load Databento MBO data containing CLEAR messages (e.g. NQ futures) via `DatabentoDataLoader.from_dbn_file()` with `as_legacy_cython=True`, then write to `ParquetDataCatalog`. The panic occurs in `ArrowSerializer.rust_defined_to_record_batch` → `OrderBookDelta.to_pyo3_list` → `Price.from_raw`.

## Test plan

- [ ] Load MBO `.dbn.zst` files containing CLEAR messages with `PRICE_UNDEF` via `from_dbn_file(as_legacy_cython=True, include_trades=True)`
- [ ] Write decoded `OrderBookDelta` list to `ParquetDataCatalog` — should no longer panic
- [ ] Verify normal (non-UNDEF) deltas still use the correct cached precision

### DIFF
--- a/nautilus_trader/model/data.pyx
+++ b/nautilus_trader/model/data.pyx
@@ -2953,10 +2953,17 @@ cdef class OrderBookDelta(Data):
             if size_prec == 0:
                 size_prec = delta._mem.order.size.precision
 
+            # Use per-delta precision for sentinel values (PRICE_UNDEF has precision=0)
             pyo3_book_order = nautilus_pyo3.BookOrder(
                nautilus_pyo3.OrderSide(order_side_to_str(delta._mem.order.side)),
-               nautilus_pyo3.Price.from_raw(delta._mem.order.price.raw, price_prec),
-               nautilus_pyo3.Quantity.from_raw(delta._mem.order.size.raw, size_prec),
+               nautilus_pyo3.Price.from_raw(
+                   delta._mem.order.price.raw,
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @zshuang15 

Thank you for the fix! :pray: 

It turns out your PR was caught up with an unfortunate test error that made its way onto develop branch. When you get a chance please rebase on the latest and push again and the tests should pass.

zshuang15:
I've been on a business trip recently and didn't see the message in time. Thank you for fixing and merging the changes. I really like this project.

PR #3710, opened 2026-03-11
Fix backtest analyzer to include position snapshots

### LINKED ISSUE #3709: Backtest reporting divergence Rust-Python (cache-only vs snapshot-inclusive positions)
## Summary
Backtest reporting can diverge depending on whether position history is built from:

- cache positions only, or
- cache positions plus position snapshots.

When only cache positions are used, realized history can be undercounted in netting-style position cycles. This produces materially different PnL/expectancy outputs from snapshot-inclusive reporting.

## Core Code References (Nautilus)
- Python positions report entrypoint merges positions and snapshots:
	- [nautilus_trader/trading/trader.py](nautilus_trader/trading/trader.py#L876)
	- [nautilus_trader/trading/trader.py](nautilus_trader/trading/trader.py#L889)
- Python report provider API supports snapshot input:
	- [nautilus_trader/analysis/reporter.py](nautilus_trader/analysis/reporter.py#L117)
- Rust backtest analyzer currently builds from cache positions only:
	- [crates/backtest/src/engine.rs](crates/backtest/src/engine.rs#L871)
	- [crates/backtest/src/engine.rs](crates/backtest/src/engine.rs#L934)
- Snapshot data is available in cache APIs:
	- [crates/common/src/cache/mod.rs](crates/common/src/cache/mod.rs#L2211)
	- [crates/common/src/cache/mod.rs](crates/common/src/cache/mod.rs#L2217)

## Why This Is Bad
1. Same executed fills can produce different reported realized PnL depending on report path.
2. Optimizer objective values (for example net profit, expectancy, risk-adjusted metrics) become path-dependent instead of execution-dependent.
3. Cross-language parity (Python vs Rust) becomes unreliable for validation and regression testing.
4. Users can draw incorrect conclusions from backtest quality because partial position history silently drops part of realized outcomes.

## Root Cause
The reporting inputs are not consistently constructed across paths:

- Python reporting explicitly includes snapshots in `generate_positions_report`.
- Rust backtest analyzer path in `BacktestEngine::build_analyzer` adds only cache positions.

This creates an asymmetry where snapshot history is available in cache but not consumed by one reporting path.

## Expected Behavior
Backtest reporting should be source-consistent:

- If position snapshots exist, reporting should include them (or apply an equivalent canonical reconstruction policy).
- Python and Rust reporting should use equivalent position-history semantics so equal execution produces equal realized reporting outputs.

## Suggested Fix Direction
In Rust backtest analyzer construction (`BacktestEngine::build_analyzer`):

1. Load live cache positions as today.
2. Load snapshot bytes via `position_snapshot_ids` + `position_snapshot_bytes`.
3. Decode snapshot payloads into `Position` entries.
4. Add both live positions and decoded snapshots into `PortfolioAnalyzer` before computing stats.

This matches the existing Python reporting model and removes path-dependent realized PnL discrepancies.

necofx:
Fixes #3709

### DIFF
--- a/crates/backtest/tests/backtest_engine.rs
+++ b/crates/backtest/tests/backtest_engine.rs
@@ -472,10 +472,7 @@ fn test_get_result_includes_snapshot_position_history(crypto_perpetual_ethusdt: 
 
     assert!(
         (expectancy - expected_expectancy).abs() < 1e-9,
-        "expected Expectancy={} to include snapshot history {}, got {}",
-        expected_expectancy,
-        snapshots_realized,
-        expectancy
+        "expected Expectancy={expected_expectancy} to include snapshot history {snapshots_realized}, found {expectancy}"
     );
 }
 


necofx:
Adopted Copilot suggestions in follow-up commit 401cccef0:\n\n- switched snapshot decoding to serde_json streaming deserializer\n- changed snapshot collection loop to iterate current positions by id\n- added regression test covering snapshot-inclusive history in backtest result metrics\n\nRan: cargo +stable test -p nautilus-backtest --test backtest_engine test_get_result_includes_snapshot_position_history -- --nocapture

necofx:
@copilot check latest commit

cjdsellers:
Hi @necofx

Thanks for the contribution!

```
make pre-commit
```
or (if you're on Windows):
```
pre-commit run --all-files
```

```
pre-commit install
```

**Also note: after pushing your next commit the CI pipeline run will need to be approved by a maintainer and this is your first contribution.**

necofx:
> Hi @necofx
> 
> Thanks for the contribution!

Sure! Nice product nautilus trader, but since I want to get pnl status for a simple rust strategy and I was getting PNL 0 found this may help (on python it seems it's not returning 0). Hope this PR helps.


Will do this pre-commit. Have some PR's on queue (I'm working on a WFO engine for nautilus rust-based)

> 
> Looks like the pre-commit failed, but this is easy to fix. Just run the following and push any changes to the same PR:
> 
> ```
> make pre-commit
> ```
> 
> or (if you're on Windows):
> 
> ```
> pre-commit run --all-files
> ```
> 
> Also consider running the following to install the git hook which will run the pre-commit automatically on commit:
> 
> ```
> pre-commit install
> ```
> 
> **Also note: after pushing your next commit the CI pipeline run will need to be approved by a maintainer and this is your first contribution.**

necofx:
Addressed the pre-commit failure in a follow-up commit and pushed to this PR branch.

- Updated the failing test assertion message wording to satisfy conventions (", found" phrasing)
- Switched to inlined format args to satisfy clippy (uninlined_format_args)
- Re-ran make pre-commit locally and all hooks passed

Commit: 68ad02df7

cjdsellers:
Hi @necofx

Thanks for catching this divergence and the clean fix. Including position snapshots in `build_analyzer` is the right call and brings the Rust path in line with what Python already does.

I retargeted the PR to `develop` -- that's the standard branch for all PRs.

I'll follow up with a small patch for the silent `filter_map(Result::ok)` in the snapshot decoder.

Let me know if you run into anything else!

PR #3719, opened 2026-03-13
Fix IB adapter shared historical request dedup for concurrent warmup

### LINKED ISSUE #3718: [Interactive Brokers] Concurrent strategy warmup fails due to shared historical request deduplication
### Description

When multiple strategies request historical bars for the same instrument and bar type during warmup, the IB client's request deduplication logic causes some strategies to never complete their warmup.

The `get_historical_bars()` method in `InteractiveBrokersClient` deduplicates concurrent requests with the same `(bar_type, end_date_time)` key via `_name_to_req_id()`. When a duplicate is detected, the second caller immediately returns an empty list — but the first request may not have completed yet, so the bars haven't been processed by the DataEngine.

This means:
1. The second caller returns `[]` before the first request's future resolves
2. No `DataResponse` is sent to the DataEngine for the second caller
3. The DataEngine's `TimeBarAggregator` for the second strategy never finalizes
4. The strategy hangs during warmup waiting for bars that will never arrive

### Steps to Reproduce

1. Configure 2+ strategies that subscribe to the same bar type on the same instrument (e.g., `AAPL.SMART-5-SECOND-BID-EXTERNAL`)
2. Run live with the IB adapter
3. Observe that only some strategies complete warmup; others hang indefinitely

### Proposed Fix

Two changes:

**1. `client/market_data.py` — await the shared request's future before returning**

When a duplicate request is detected, the second caller should `await request.future` to ensure the first request has completed and bars have been processed before returning empty:

```python
else:
    self._log.info(f"Request already exist for {request}, awaiting shared result")
    await request.future
    return []
```

**2. `data.py` — send empty `DataResponse` when no bars are received**

When `_request_bars` receives an empty bar list, it should still call `_handle_bars` so the DataEngine can finalize the aggregator for that request:

```python
else:
    self._log.warning(f"No bar data received for {request.bar_type}")
    self._handle_bars(
        request.bar_type,
        [],
        request.id,
        request.start,
        request.end,
        request.params,
    )
```

This ensures the DataEngine receives a `DataResponse` (even with empty data) and can call `_finalize_aggregated_bars_request()` which uses only the `correlation_id` to find and finalize the matching aggregator.

### Environment

- Nautilus Trader: develop branch (latest)
- IB Gateway via Docker
- Multiple strategies sharing bar subscriptions

### Notes

Validated fix in production with 5 concurrent strategies — all complete warmup successfully. I have a branch with the fix and a test ready for PR.

Johnkhk:
Closes #3718

## Summary

- When multiple strategies request historical bars for the same instrument/bar type during warmup, the IB client's request deduplication causes secondary callers to return immediately before the first request completes — leaving their DataEngine aggregators unfinalized and strategies hanging
- Secondary callers now `await request.future` before returning empty, ensuring bars have been processed
- `_request_bars` now sends an empty `DataResponse` when no bars are received, so the DataEngine can finalize aggregators for secondary strategies

## Test plan

- [x] Added integration test `test_get_historical_bars_shared_request_awaits_future` verifying secondary callers await the shared future before returning
- [x] Validated in production with 5 concurrent strategies — all complete warmup successfully

### DIFF
--- a/tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py
+++ b/tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py
@@ -334,9 +334,11 @@ async def test_get_historical_bars(ib_client):
 
 @pytest.mark.asyncio
 async def test_get_historical_bars_shared_request_awaits_future(ib_client):
-    """When a duplicate historical bars request is made, the second caller
-    should await the first request's future before returning empty, ensuring
-    bars have been processed by the DataEngine before the second caller returns."""
+    """
+    Test that a duplicate historical bars request awaits the first request's future
+    before returning empty, ensuring bars have been processed by the DataEngine before
+    the second caller returns.
+    """
     # Arrange
     ib_client._request_id_seq = 999
     bar_type = BarType.from_str("AAPL.SMART-5-SECOND-BID-EXTERNAL")
@@ -359,7 +361,6 @@ async def test_get_historical_bars_shared_request_awaits_future(ib_client):
     )
 
     async def second_caller():
-        """Simulates a second strategy requesting the same bars."""
         result = await ib_client.get_historical_bars(
             bar_type,
             contract,


faysou:
Hi @Johnkhk have you tested your fix with TWS ? I suppose yes, it's just that it becomes increasingly difficult to understand the IB adapter, so the only way to know is to test on a real/demo case with IB.

I see that you have done PRs before for nautilus, we actually interacted on one, it should be fine.

Johnkhk:
> Hi @Johnkhk have you tested your fix with TWS ? I suppose yes, it's just that it becomes increasingly difficult to understand the IB adapter, so the only way to know is to test on a real/demo case with IB.
> 
> I see that you have done PRs before for nautilus, we actually interacted on one, it should be fine.

Hey @faysou. Yes I tested with live IB Gateway, this fix only affects IB historical backfill (for warmup), no downstream effects on any other code path.

cjdsellers:
Hey @Johnkhk 

Thank you again for another great fix :pray:

PR #3734, opened 2026-03-19
Improve Polymarket execution client

filipmacek:
# Pull Request

## Summary

Fixes several interrelated issues in the Polymarket adapter's execution pipeline that caused
orders to get stuck in incorrect states, fill quantities to be rejected by the CLOB, and
race conditions between WebSocket and HTTP responses to drop fills.

### Quote-Quantity Fill Tracking in OrderCore

**Problem:** Polymarket market BUY orders use `is_quote_quantity=true` — the order quantity
is denominated in USDC (e.g. "buy 5 USDC worth"), but fills arrive in base units (shares).
The existing `OrderCore::filled()` compared raw base-unit fill quantities directly against
the quote-unit order quantity. This meant:
- `leaves_qty` was computed incorrectly (subtracting shares from USDC)
- Orders transitioned to `Filled` prematurely or never reached `Filled` at all
- The overfill check in `reconciliation.rs` rejected valid fills as overfills

**Fix:** `OrderCore::filled()` (Rust) and `Order._filled()` (Cython) now branch on
`is_quote_quantity`. When true, each fill is converted to quote units via
`consumed = last_qty * last_px` using Decimal arithmetic, and `leaves_qty` is decremented
in quote terms. `calculate_overfill()` returns zero for quote-quantity orders since overfill
detection requires the fill price which is only available inside `filled()`.

The reconciliation overfill check now delegates to `calculate_overfill()` instead of doing
its own raw addition, keeping the logic in one place.

### CLOB Amount Precision and Sell-Side Granularity

**Problem:** Polymarket's CLOB enforces strict precision constraints on order amounts:
- Direct amounts (share quantities): max 2 decimal places (`LOT_SIZE_SCALE`)
- Computed amounts (qty × price): max `tick_decimals + LOT_SIZE_SCALE` decimal places

The previous implementation used hardcoded `MAX_PRECISION_MAKER=5` / `MAX_PRECISION_TAKER=2`
constants and a granularity-based truncation approach. This had two issues:
1. Sell-side market orders applied the wrong granularity direction (USDC precision to shares
   and vice versa)
2. The precision was not instrument-aware — different markets have different `tick_decimals`
   (1, 2, or 3), so a single hardcoded value produced amounts the CLOB rejected

**Fix:** Refactored `compute_maker_taker_amounts` and `compute_market_maker_taker_amounts`
from `parse.rs` into `order_builder.rs` with a new `tick_decimals` parameter threaded from
the instrument through `submitter.rs`. Amounts are now truncated with
`trunc_with_scale(tick_decimals + LOT_SIZE_SCALE)` before conversion to on-chain base units
via `to_fixed_decimal()`. The old `MAX_PRECISION_MAKER`/`MAX_PRECISION_TAKER` constants are
replaced by `USDC_DECIMALS` and `LOT_SIZE_SCALE`.

### Market Order FOK Order Type

**Problem:** Market orders were converting `time_in_force` to a Polymarket order type via
`convert_tif_to_polymarket_order_type()`, which could produce `GTC` for certain TIF values.
Polymarket market orders must always be `FOK` (fill-or-kill) — anything else is rejected
or behaves unexpectedly.

**Fix:** Market orders now hardcode `PolyOrderType.FOK` and pass it through via an
`order_type_override` parameter on `_post_signed_order()`, bypassing TIF conversion entirely.

### WS→HTTP Race Condition Buffering

**Problem:** When an order is submitted, the WebSocket stream can deliver fill and status
reports _before_ the HTTP response returns the `venue_order_id` and triggers the `Accepted`
event. These early messages were dropped because the order wasn't yet marked as accepted.

**Fix:** Introduced `pending_fills` and `pending_order_reports` buffers
(`FifoCacheMap<VenueOrderId, Vec<...>>`) in both Rust and Python execution clients. When a
WS message arrives for an unrecognized `venue_order_id`, it is buffered. Once the HTTP
response returns and the order is accepted, all buffered fills and reports are drained and
emitted in order.

### Per-Order Fill Tracking and Dust Detection

**Problem:** Polymarket's CLOB truncates fill quantities to 2 decimal places (`LOT_SIZE_SCALE`).
When an order is submitted for e.g. 23.696681 shares, the CLOB fills 23.69 shares and
reports the order as `MATCHED` (fully filled). The 0.006681 residual ("dust") means the
NautilusTrader engine never sees enough fill quantity to transition the order to `Filled`,
leaving it stuck in `PartiallyFilled`.

**Fix:** New `OrderFillTrackerMap` (Rust) and `OrderFillTracker` (Python) track cumulative
fills per venue order ID. Two mechanisms handle dust:

1. **Snap**: When a single fill's quantity differs from the submitted quantity by less than
   `DUST_SNAP_THRESHOLD` (0.01), the fill quantity is snapped up to the submitted quantity.
   This handles the common case of a single-fill order.

2. **Synthetic dust fill**: For multi-fill orders, when the CLOB sends a `MATCHED` status
   and the cumulative filled quantity has a sub-threshold residual, a synthetic `FillReport`
   is emitted for the dust amount at the last fill price with zero commission and
   `NoLiquiditySide`. The tracker entry is removed after settlement to prevent duplicate
   synthetic fills from repeated `MATCHED` events.

The tracker also replaces the old `accepted_orders: FifoCache<VenueOrderId>` — it subsumes
the "is accepted" check via `contains()` while adding fill state tracking. Capacity is
capped at 10,000 entries with FIFO eviction.

### LiveExecEngineConfig: convert_quote_qty_to_base

Added `convert_quote_qty_to_base: bool` (default `false`) to `LiveExecEngineConfig`. This
controls whether the execution engine converts quote-denominated quantities to base units
before submission. Defaults to false since adapters (like Polymarket) now handle quote
quantity semantics directly.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/polymarket/src/execution/mod.rs
+++ b/crates/adapters/polymarket/src/execution/mod.rs
@@ -599,7 +599,7 @@ impl PolymarketExecutionClient {
                         log::info!("User WebSocket reconnected");
                     }
                     None => {
-                        log::warn!("User WebSocket stream ended");
+                        log::debug!("User WebSocket stream ended");
                         break;
                     }
                 }


cjdsellers:
Thanks @filipmacek, great progress and improvements here.

Merging this, but we'll follow up by reverting the core model changes. The adapter should convert quote to base quantity at submission time using the crossing price (matching Polymarket's own SDK approach), and the dust tracker already handles the precision gap.

I'll also remove the deprecated `convert_quote_qty_to_base` from the execution engine to improve UX and remove a foot gun.

PR #3743, opened 2026-03-21
Add redis cluster configuration support

### LINKED ISSUE #3713: redis configuration improve
nautilus_trader/cache/config.py 

redis configuration does not support cluster mode and specifying a database.

twitu:
# Pull Request

## Summary

- **Add Redis cluster connection support**
- **Fix cluster INFO response parsing and docker-compose command**

## Related Issues/PRs

Closes #3713

## Type of change

## Documentation

## Release notes

## Testing

Added a docker compose file to launch redis cluster and a test to check connectivity with it

### DIFF
--- a/crates/infrastructure/src/redis/cache.rs
+++ b/crates/infrastructure/src/redis/cache.rs
@@ -631,7 +631,6 @@ async fn drain_buffer(
     buffer: &mut VecDeque<DatabaseCommand>,
 ) {
     let mut pipe = redis::pipe();
-    pipe.atomic();
 
     for msg in buffer.drain(..) {
         let key = if let Some(key) = msg.key {
--- a/crates/infrastructure/src/redis/mod.rs
+++ b/crates/infrastructure/src/redis/mod.rs
@@ -43,6 +43,14 @@ pub enum RedisConnection {
     Cluster(ClusterConnection),
 }
 
+impl RedisConnection {
+    /// Returns `true` if this connection is in cluster mode.
+    #[must_use]
+    pub fn is_cluster(&self) -> bool {
+        matches!(self, Self::Cluster(_))
+    }
+}
+
 impl Debug for RedisConnection {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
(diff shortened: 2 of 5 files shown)


twitu:
## Fixes for cluster compatibility

### 1. Remove `pipe.atomic()` (cache.rs, msgbus.rs)

**Why it existed:** `atomic()` wraps pipeline commands in `MULTI/EXEC`, giving all-or-nothing execution for a batch of writes.

**Why it doesn't work in cluster mode:** `MULTI/EXEC` requires all keys in the transaction to hash to the same slot. The cache buffer mixes keys across collections (instruments, orders, positions) and the msgbus uses different stream keys per topic — these will span multiple hash slots, causing `CROSSSLOT` errors.

**Why removing it is safe in both modes:** Each command in the pipeline operates on an independent key. There are no inter-command dependencies (no read-then-write, no conditional logic). A partial batch failure is already possible if the process crashes mid-flush. The pipeline still batches commands for network efficiency without `atomic()` — the only loss is the all-or-nothing guarantee, which provides no correctness benefit here.

### 2. Use `KEYS` instead of `SCAN` in cluster mode (queries.rs)

**Why SCAN existed:** `SCAN` is the recommended way to iterate keys without blocking the server, using a cursor to paginate results incrementally.

**Why it doesn't work in cluster mode:** `SCAN` is routed to a single primary node by the cluster driver. Keys stored on other shards are silently missed, breaking `load_all`, `load_currencies`, `load_instruments`, etc.

**Fix:** In cluster mode, use `KEYS` instead. The `redis` crate's cluster driver routes `KEYS` to `AllPrimaries` and merges results from every shard. `KEYS` is blocking on each node, but this is acceptable for the load/startup paths where `scan_keys` is called. Single-node mode retains `SCAN` to avoid blocking.

### 3. Apply retry/timeout config to cluster connection (mod.rs)

**What was missing:** The single-node path applied all six config fields (`connection_timeout`, `response_timeout`, `number_of_retries`, `exponent_base`, `max_delay`, `factor`) via `ConnectionManagerConfig`. The cluster path called `ClusterClient::new(urls)` with none of them.

**Fix:** Use `ClusterClientBuilder` to apply `connection_timeout`, `response_timeout`, `retries`, `retry_wait_formula`, and `max_retry_wait` so config behaves uniformly across both modes.

### 4. Add `cluster_mode`/`cluster_nodes` to `DatabaseConfig.__repr__` (config.py)

The Rust `Debug` impl was updated but the Python `__repr__` still showed only the original fields, making the new cluster fields invisible during debugging.

### 5. Improve `get_redis_version` error handling for cluster (mod.rs)

The cluster INFO path used `unwrap_or_default()` which would silently produce an empty string on unrecognized response formats, leading to a confusing parse error downstream. Now it returns explicit error messages for both "no node returned INFO" and "unexpected response type".

### Not addressed (non-issues)

- **Database != 0 validation:** The config struct has no `database` field and the URL builder doesn't append a database number — there's no way to configure a non-zero database, so no validation is needed.
- **MGET across hash slots:** The `redis` crate's cluster driver automatically splits `MGET` across slots and merges results — no code change required.
- **FLUSHDB fan-out:** The `redis` crate routes `FLUSHDB` to `AllPrimaries` in cluster mode — no code change required.
- **XTRIM routing:** `XTRIM` takes a single stream key and is routed to the correct shard by the cluster driver — no issue.

cjdsellers:
Thanks for the work on this @twitu . The `RedisConnection` enum with delegated `ConnectionLike` is a clean pattern and the cluster INFO parsing handles the edge cases well.

Now that we've explored a full implementation, I don't think cluster support is the right direction for the database layer. Removing `pipe.atomic()` drops transactional guarantees for all existing single-node users, and supporting both modes adds branching complexity throughout the pipeline and query code. We'll also be revisiting configuration and plugin ergonomics for v2, and the additional config surface area would complicate that refactoring.

Going to close this for now. Let's focus on the other Arrow/Parquet items.

PR #3783, opened 2026-04-01
Add Hyperliquid product type config for live clients

lisiyuan656:
## Summary
- add `product_types` to Hyperliquid live data and exec client configs
- thread configured product types through the live client factories into the
  instrument provider
- document how to enable HIP-3 products in live `TradingNode` configs

## Why
HIP-3 instruments can be discovered through the provider, but live
  Hyperliquid clients need an explicit way to request `PERP_HIP3` when loading instruments through the factory/config path.

## Testing
- updated Hyperliquid factory
  config tests
- updated Hyperliquid integration docs for HIP-3 live client configuration


cjdsellers:
Hi @lisiyuan656 

Many thanks for the PR. I'll follow up on a couple of minor things post-merge.

PR #3789, opened 2026-04-02
fix: increase Binance kline API rate limit to 1200 per minute

### LINKED ISSUE #2982: Implement dynamic rate limiting for API clients
# Feature Request

## Problem Statement

Requesting historical bars for multiple instruments on Binance API might cause rate limit bans, since there is no control mechanism between API calls.
I found that the actual API calls are made after all the request_bars() methods in the loop have been executed, rather than between each method call. This seems to be because it utilizes the message bus.

```python
def on_start(self) -> None:
    for bar_type in self.config.bar_types:
        self.request_bars(
            bar_type,
            start=self._clock.utc_now() - pd.Timedelta(days=60),
        )
        # expect: API call executed here...
    
    # actual: API calls executed at once here (causes rate limit ban when # of instrument is big)

```

### Proposed Solution

If I don't miss something, it would be good to have rate limiting parameter or else to mitigate the issue.

XXXXhhhx:
## Summary
This PR addresses issue #2982, which reports that batch requests for Binance kline data are prone to rate limiting and temporary bans.

## Root Cause
The original rate limit quota for Binance kline endpoints was set to 600 requests per minute, which is too low for high-frequency batch data fetching in HFT strategies.

## Solution
- Increased the rate limit for Binance futures kline endpoint (`/fapi/v1/klines`) from 600 to 1200 requests per minute
- This change aligns with Binance's official latest rate limit rules for kline data endpoints

## Related Issue
Closes #2982

### DIFF
--- a/nautilus_trader/adapters/binance/factories.py
+++ b/nautilus_trader/adapters/binance/factories.py
@@ -133,7 +133,7 @@ def get_cached_binance_http_client(
             (global_key, global_quota),
             ("binance:api/v3/order", Quota.rate_per_minute(3000)),
             ("binance:api/v3/allOrders", Quota.rate_per_minute(int(3000 / 20))),
-            ("binance:api/v3/klines", Quota.rate_per_minute(600)),
+            ("binance:api/v3/klines", Quota.rate_per_minute(1200)),
         ]
     else:
         # Futures


cjdsellers:
Hi @XXXXhhhx

Thanks for the contribution. A couple of things here though:

The PR description references `/fapi/v1/klines` (futures) but the code change is on the spot endpoint `api/v3/klines` at line 136. The futures quota at line 147 is unchanged.

The current quota of 600 req/min was set deliberately in `23e0e0f` to be conservative. With the spot klines weight of 2, that's 1,200 of the 6,000 global weight budget (20%). Doubling it to 1,200 requests would consume 40% of the global budget, leaving less headroom for order operations and increasing the risk of IP bans for users running real strategies.

I don't think this closes #2982 either. That issue is about request bursting and dynamic throttling, not the static quota size.
  
Going to close this one, but appreciate you looking into the rate limiting area.

PR #3832, opened 2026-04-10
Add PyO3 v2 live config wiring

BurnOutTrader:
# Pull Request

## Summary

This fixes the PyO3 v2 live config path so supported live engine config fields are actually propagated into the Rust v2 engines instead of being dropped back to defaults. It also narrows the exposed PyO3 v2 surface to config that currently produces real behavior, makes reserved unsupported fields fail fast on node build instead of silently no-oping, and corrects `reconciliation_startup_delay_secs` so it delays startup reconciliation itself rather than only offsetting the periodic reconciliation timers.

This branch also hardens the PyO3 v2 live-config validation path so malformed supported risk/exec values fail cleanly before any downstream `expect()` / infallible identifier conversion can be reached.

The typed PyO3 surface is now restored through native stub generation as well: `LiveNodeConfig` and `PortfolioConfig` are generated from the Rust-side PyO3 metadata again, and the old manual `nautilus_trader.live` stub patch has been removed.

## Related Issues/PRs

N/A

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

On the PyO3 v2 live path, unsupported args that were previously exposed but not implemented are no longer accepted on the Python constructors (`qsize` on data/risk/exec, plus exec `snapshot_orders`, `snapshot_positions`, `purge_from_database`, and `graceful_shutdown_on_error`). The shared Rust live config still retains these fields as reserved for future work, but `LiveNode::build(...)` and the fluent builder now reject them explicitly on the current v2 live node path. The legacy live path is unchanged.

## Root cause

The PyO3 v2 live layer had drifted into a partial/placeholder state:

- `LiveDataEngineConfig` on the v2 path only exposed `qsize`, while supported data-engine fields were not surfaced in the binding.
- The shared live config conversion into `DataEngineConfig` fell back to `DataEngineConfig::default()` instead of carrying through v2 values.
- The PyO3 `LiveNode` surface did not provide a clean `build(..., config=...)` path plus engine-config setters for the fluent builder.
- `reconciliation_startup_delay_secs` only shifted the periodic reconciliation timers, while startup reconciliation still ran immediately.
- The PyO3 config types are `from_py_object`, so constructor-only validation was not sufficient to protect the v2 build path from malformed foreign objects.
- `LiveNodeConfig` was opaque in generated stubs because native stub coverage for this class graph was incomplete: `LiveNodeConfig` did not expose stub pymethod metadata, `PortfolioConfig` lacked stub coverage, and the `live` module did not declare the `PortfolioConfig` re-export for stub generation.

## What changed

- wired supported live data/risk/exec config fields through the PyO3 v2 live path into the Rust v2 engine configs
- exposed `LiveNode.build(name, config=...)` and engine-config builder setters for the PyO3 v2 path
- removed dead-end PyO3 v2 constructor args from the Python bindings
- added build-time validation so reserved-but-unsupported live config fields error explicitly instead of silently no-oping
- added validation for supported risk/exec fields at the shared live-config boundary so malformed values now return normal errors on the current v2 path instead of risking panic paths downstream
- corrected startup reconciliation delay semantics so `reconciliation_startup_delay_secs` delays startup reconciliation itself
- added native stub coverage for `PortfolioConfig` and `LiveNodeConfig`
- declared the `PortfolioConfig` live-module re-export for stub generation
- removed the manual live-module post-processing from `python/generate_stubs.py`
- corrected a stale PyO3 v2 test expectation for `LiveNodeConfig.save_state` to match the current runtime default (`false`)
- regenerated the affected package stubs and updated Rust/Python-facing tests around the v2 live config and stub surface

## Stub generation

The previous manual live-module stub patch is no longer needed.

This PR now uses the native PyO3 stub-generation path again by:

- adding stub metadata to `PortfolioConfig`
- adding stub metadata to `LiveNodeConfig`
- declaring the `PortfolioConfig` re-export for the `nautilus_trader.live` stub module

That keeps the typed surface aligned with the PyO3 v2 architecture instead of maintaining a module-specific Python-side workaround.

## Intentionally not exposed on the current v2 live node path

These were intentionally not carried onto the PyO3 v2 surface because they do not currently lead to honest end-to-end behavior:

- `emit_quotes_from_book` / `emit_quotes_from_book_depths`
  Reason: there is no Rust v2 `DataEngineConfig` destination for these fields.
- `time_bars_origins`
  Reason: Rust has the field, but there is not yet a clean PyO3 v2 origin/timedelta representation in this layer.
- `load_cache`
  Reason: `ExecutionEngine::load_cache()` exists, but the live kernel does not currently wire a cache database adapter on the v2 path.
- `snapshot_orders`, `snapshot_positions`, `snapshot_positions_interval_secs`
  Reason: snapshot persistence/timer wiring is not complete on the current live path.
- `purge_from_database`
  Reason: purge flows currently operate on the in-memory cache only.
- `graceful_shutdown_on_error`
  Reason: there is no active live/runtime consumer for the flag yet.
- `qsize`
  Reason: the current live v2 path has no downstream engine consumer for it.

## What remains after the underlying v2 support exists

Once the lower layers are ready, the follow-up work should be straightforward:

- wire the live kernel to construct cache database adapters on the v2 path
- implement the execution-engine snapshot timer behavior end-to-end
- re-expose snapshot/purge/load-cache config on the PyO3 v2 surface once those paths are functional
- add `time_bars_origins` once there is a clean PyO3 v2 representation for the field
- if v2 live state persistence semantics are intentionally expanded or aligned differently later, revisit whether `save_state` should default back to `true`
- decide whether any currently reserved fields should become supported v2 API, instead of remaining internal/reserved only

## Documentation

No docs changed in this patch.

## Release notes

Not added in this patch.

## Testing

Validated with:

- `.venv/bin/python python/generate_stubs.py`
- `cargo test -p nautilus-live --features python --lib`
- `cargo test -p nautilus-portfolio --features python --lib`
- `PATH=/tmp/nautilus-uv-0.11.3:$PATH make pytest-v2`
- `PATH=/tmp/nautilus-uv-0.11.3:$PATH prek run --all-files`
- `PATH=/tmp/nautilus-uv-0.11.3:$PATH make check-capnp-schemas`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6074,6 +6074,7 @@ dependencies = [
  "nautilus-core",
  "nautilus-model",
  "pyo3",
+ "pyo3-stub-gen",
  "rstest",
  "rust_decimal",
  "rust_decimal_macros",
--- a/crates/live/src/builder.rs
+++ b/crates/live/src/builder.rs
@@ -28,7 +28,7 @@ use nautilus_system::{
 };
 
 use crate::{
-    config::LiveNodeConfig,
+    config::{LiveDataEngineConfig, LiveExecEngineConfig, LiveNodeConfig, LiveRiskEngineConfig},
     manager::{ExecutionManager, ExecutionManagerConfig},
     node::LiveNode,
     runner::AsyncRunner,
@@ -179,6 +179,27 @@ impl LiveNodeBuilder {
         self
     }
 
+    /// Set the live data engine configuration.
+    #[must_use]
+    pub fn with_data_engine_config(mut self, config: LiveDataEngineConfig) -> Self {
+        self.config.data_engine = config;
+        self
+    }
+
+    /// Set the live risk engine configuration.
+    #[must_use]
+    pub fn with_risk_engine_config(mut self, config: LiveRiskEngineConfig) -> Self {
+        self.config.risk_engine = config;
+        self
+    }
+
+    /// Set the live execution engine configuration.
+    #[must_use]
+    pub fn with_exec_engine_config(mut self, config: LiveExecEngineConfig) -> Self {
+        self.config.exec_engine = config;
+        self
+    }
+
     /// Adds a data client factory with configuration.
     ///
     /// # Errors
@@ -239,6 +260,8 @@ impl LiveNodeBuilder {
             self.data_client_factories.len(),
             self.exec_client_factories.len()
         );
+
+        self.config.validate_live_path()?;
 
         let runner = AsyncRunner::new();
         runner.bind_senders();
--- a/crates/live/src/config.rs
+++ b/crates/live/src/config.rs
@@ -15,19 +15,23 @@
 
 //! Configuration types for live Nautilus system nodes.
 
-use std::{collections::HashMap, time::Duration};
+use std::{collections::HashMap, str::FromStr, time::Duration};
 
 use nautilus_common::{
     cache::CacheConfig, enums::Environment, logging::logger::LoggerConfig,
-    msgbus::database::MessageBusConfig,
+    msgbus::database::MessageBusConfig, throttler::RateLimit,
 };
 use nautilus_core::UUID4;
 use nautilus_data::engine::config::DataEngineConfig;
 use nautilus_execution::engine::config::ExecutionEngineConfig;
-use nautilus_model::identifiers::TraderId;
+use nautilus_model::{
+    enums::BarIntervalType,
+    identifiers::{ClientId, TraderId},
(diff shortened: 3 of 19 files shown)


cjdsellers:
Thanks @BurnOutTrader 

I also completed some further wiring in https://github.com/nautechsystems/nautilus_trader/commit/ae59390c7f52cdb630295087bf7bdd2befbca7e4 so this should all be addressed now :+1:

Issue #3839, opened 2026-04-10
[Feature] VWAP execution algorithm


JKDasondee:
## Summary

Add a VWAP (Volume-Weighted Average Price) execution algorithm alongside the existing TWAP implementation.

The Rust algorithm module comments already mention VWAP as an intended algorithm (line 19 of `crates/core/src/nanos.rs` docs reference "algorithms like TWAP and VWAP"), and `ExecAlgorithmId("VWAP")` appears in serialization tests, but no implementation exists.

## Proposed implementation

A `VWAPExecAlgorithm` in `nautilus_trader/examples/algorithms/vwap.py` following the TWAP pattern:

- **Static volume profile**: accepts a `volume_weights` list (default: U-shaped intraday curve) that determines the proportion of total quantity to execute in each interval
- **Same timer-based execution** as TWAP: `horizon_secs` + `interval_secs` parameters, timer callback spawns market orders
- **Volume-weighted slicing**: instead of equal splits, each interval's quantity is proportional to its weight in the profile

### Parameters (via `exec_algorithm_params`)

| Parameter | Type | Description |
|-----------|------|-------------|
| `horizon_secs` | float | Total execution time window |
| `interval_secs` | float | Time between child order submissions |
| `volume_weights` | list[float] (optional) | Per-interval volume weights. Normalized internally. Default: U-shaped curve. |

### Example usage

```python
order = self.order_factory.market(
    instrument_id=BTCUSDT,
    order_side=OrderSide.BUY,
    quantity=Quantity.from_str("1.000"),
    exec_algorithm_id=ExecAlgorithmId("VWAP"),
    exec_algorithm_params={
        "horizon_secs": 60.0,
        "interval_secs": 10.0,
        "volume_weights": [3, 2, 1, 1, 2, 3],  # U-shaped
    },
)
```

## Motivation

VWAP is the most widely used execution benchmark in institutional trading. A static-profile implementation provides immediate value while a future enhancement could add live market volume tracking (participation-rate mode).

PR #3840, opened 2026-04-10
feat: add VWAP execution algorithm example

JKDasondee:
## Summary

Refs #3839. Adds a VWAP execution algorithm in `nautilus_trader/examples/algorithms/vwap.py`, following the existing TWAP pattern.

## Motivation

TWAP is the only built-in execution algorithm. VWAP is referenced in the codebase docs and serialization tests (`ExecAlgorithmId("VWAP")`) but has no implementation. This adds the most commonly used institutional execution benchmark.

## Design

Implements the standard static-profile VWAP scheduler (Konishi 2002, *J. Financial Markets* 5(2):197-221): each time bucket's target quantity is proportional to its weight in the volume profile.

```
target_qty[t] = total_qty * (weight[t] / sum(weights))
```

Scheduling uses `Decimal` arithmetic with `ROUND_DOWN` per interval. The rounding remainder is accumulated into a final cleanup slice, ensuring `sum(scheduled_sizes) == order.quantity` exactly.

### Parameters (via `exec_algorithm_params`)

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `horizon_secs` | float | required | Total execution window |
| `interval_secs` | float | required | Time between child orders |
| `volume_weights` | list[float] | U-shaped | Per-interval weights, normalized internally |

### Default volume profile

When `volume_weights` is omitted, a symmetric U-shaped curve is generated: `weight[i] = 1.0 + |i - midpoint|`. This reflects the typical intraday volume pattern (higher at open/close, lower midday). Users are expected to supply their own historical volume profile for production use.

### Example

```python
order = self.order_factory.market(
    instrument_id=BTCUSDT,
    order_side=OrderSide.BUY,
    quantity=Quantity.from_str("1.000"),
    exec_algorithm_id=ExecAlgorithmId("VWAP"),
    exec_algorithm_params={
        "horizon_secs": 60.0,
        "interval_secs": 10.0,
        "volume_weights": [3, 2, 1, 1, 2, 3],
    },
)
```

For `total_qty=100, weights=[3,2,1,1,2,3]`:
```
Sizes: [25.000, 16.666, 8.333, 8.333, 16.666, 25.000, 0.002]
                                                        └─ remainder
```

## Relationship to TWAP

Structurally identical to `twap.py` — same `ExecAlgorithm` base class, same timer-based execution loop, same `spawn_market` / `submit_order` flow, same lifecycle methods. The only difference is the quantity scheduling: TWAP divides equally, VWAP divides by weight.

If `volume_weights` are all equal, VWAP produces the same schedule as TWAP.

## Test plan

- [x] Standalone scheduling logic: verified Decimal arithmetic, rounding, remainder accumulation, U-shaped profile symmetry, equal-weight equivalence to TWAP, zero-weight bucket handling (5 test cases)
- [x] Integration: `pytest tests/unit_tests/execution/` (requires Rust/pyo3 build — will verify after CI triggers)

## Future work

This is a static-profile VWAP. A participation-rate variant (live market volume tracking via `subscribe_trade_ticks`) could follow as an enhancement, but the static version provides immediate value and matches how most institutional VWAP algos start.

### DIFF
--- a/nautilus_trader/examples/algorithms/vwap.py
+++ b/nautilus_trader/examples/algorithms/vwap.py
@@ -1,0 +1,284 @@
+# -------------------------------------------------------------------------------------------------
+#  Copyright (C) 2015-2026 Nautech Systems Pty Ltd. All rights reserved.
+#  https://nautechsystems.io
+#
+#  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+#  You may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+# -------------------------------------------------------------------------------------------------
+
+import math
+from datetime import timedelta
+from decimal import ROUND_DOWN
+from decimal import Decimal
+
+from nautilus_trader.common.enums import LogColor
+from nautilus_trader.common.events import TimeEvent
+from nautilus_trader.config import ExecAlgorithmConfig
+from nautilus_trader.core.correctness import PyCondition
+from nautilus_trader.execution.algorithm import ExecAlgorithm
+from nautilus_trader.model.enums import OrderType
+from nautilus_trader.model.identifiers import ClientOrderId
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @JKDasondee 

Closing this one. #3839 was only opened a few hours ago and has had no discussion.

Did you run this live, and if so on which adapter? we'd also need a reasonable level of automated test coverage.

Also, we'd prefer contributions to target Rust, with optional Python bindings via PyO3, as this is the direction the project is heading (see the ROADMAP, plus recent issues and PRs).

PR #3866, opened 2026-04-14
fix: consolidate_data_by_period pairwise merging in fragment-per-flush catalogs (fixes #3857)

### LINKED ISSUE #3857: Catalog consolidate_data_by_period(period=1d) produces incorrect daily-spanning fragments
### Discussed in https://github.com/nautechsystems/nautilus_trader/discussions/3824

<div type='discussions-op-text'>

<sup>Originally posted by **stefanholi** April  9, 2026</sup>
I'm trying to use `ParquetDataCatalog.consolidate_data_by_period` to produce one file per UTC day from a live data collector's fragment-per-flush on-disk state. The docs describe this method as:

  > Split data files into fixed time periods for standardized file organization.

I expected "1-day period" to produce one file per UTC day. The observed behavior is different and I want to confirm whether it's intended.

  ## Environment

  - nautilus_trader 1.223.0
  - Python 3.13.9
  - Linux x86_64, glibc 2.41

  ## Minimal reproduction

  Seed a fresh tempdir catalog with 168 synthetic `Bar` objects spanning exactly 7 UTC days (hourly bars), each written as a separate fragment via
  `catalog.write_data([bar], skip_disjoint_check=True)`. Then:

  catalog.consolidate_data_by_period(
      data_cls=Bar,
      identifier=bar_type_str,
      period=pd.Timedelta(days=1),
      ensure_contiguous_files=False,
  )

  Observed

  - Pre:  168 files, 168 bars
  - Post: 84 files, 168 bars

Bars are preserved (zero data loss), but the output file count is exactly half of the input — regardless of the period=1 day argument. It looks like pairwise adjacent-file merging rather than period-based partitioning. Running it again would presumably halve further.

For a 7-day span with period=1d, my intuition (and my reading of the "split into fixed time periods" doc) was 7 output files. Is 84 the intended behavior? If yes, could the docs clarify the cardinality relationship? If no, I'll re-file as a bug with the full minimal reproducer.

Side note — ensure_contiguous_files=True

With ensure_contiguous_files=True on the same fragment inputs:
AssertionError: Intervals are not contiguous. When ensure_contiguous_files=True, all files in the consolidation range must have contiguous timestamps.

The docs don't mention this mode for the period variants. Is ensure_contiguous_files=True documented to be incompatible with fragment-per-write collector catalogs?

I'm using NT to write a live data collector that flushes bars in small batches (NT-native filenames, one fragment per flush). Over time these fragments accumulate to thousands per instrument-timeframe. I want a periodic consolidation that produces bounded per-day files, so my rsync transfers only touch the current day's file. The base consolidate_data works but produces one ever-growing file per instrument. consolidate_data_by_period looked like the right tool — hence the question.

If the current behavior is intended and my use case is out of scope for this method, any recommendation on how to achieve per-day consolidation would be appreciated.</div>

M-Advis:
# Pull Request

## Summary

**Root cause — hardcoded +1 ns adjacency check**

`_prepare_consolidation_queries` grouped files into contiguous blocks using `prev_interval[1] + 1 == curr_interval[0]`. For fragment-per-flush catalogs where each bar is a separate file, consecutive files are spaced by the bar interval (e.g. 3,600,000,000,000 ns for hourly bars) — never 1 ns apart. Every pair was treated as a genuine data gap, producing N single-file groups instead of 1 and causing pairwise merging that halves the file count on each run and silently destroys originals.

**Intent of the original grouping**

The block was designed to honour:
```python
# Group intervals into contiguous groups to preserve holes between groups
# but allow consolidation within each contiguous group
```
The hardcoded +1 ns assumption broke this for any catalog where files are uniformly spaced but not nanosecond-adjacent.

**Fix — two-pass dominant spacing detection**

- Pass 1: collect all gaps between consecutive file start timestamps, find dominant spacing via `Counter.most_common(1)`, warn if confidence falls below 85%
- Pass 2: group files using `dominant_gap * 1.5` as threshold — only gaps larger than 1.5x the dominant spacing are treated as genuine data holes

Note: this fix introduces new grouping logic rather than a minimal patch. The 1.5x gap threshold and 85% confidence warning are reasonable defaults but are heuristics — both could be exposed as parameters in a future improvement.

## Related Issues/PRs
Closes #3857

## Type of change
- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

N/A

## Documentation

## Release notes

## Testing

Added `test_consolidate_fragment_per_flush_hourly_to_daily` to `TestConsolidateDataByPeriod` in `tests/unit_tests/persistence/test_catalog.py`. The test writes 168 hourly bars one-per-file, consolidates with `period=1d`, and asserts exactly 7 daily files with all 168 bars preserved. The test fails on the unfixed code and passes on the fix. All 17 tests in `TestConsolidateDataByPeriod` pass.

Additionally, 15 standalone scenarios were validated locally covering: minute/second/hourly/daily bar rhythms, real data gaps, intra-period missing bars, multi-instrument isolation, nanosecond jitter tolerance, weekly period boundaries, and idempotency.

### DIFF
--- a/nautilus_trader/persistence/catalog/parquet.py
+++ b/nautilus_trader/persistence/catalog/parquet.py
@@ -19,7 +19,7 @@ import itertools
 import os
 import platform
 import re
-from collections import defaultdict
+from collections import defaultdict, Counter
 from collections.abc import Generator
 from itertools import groupby
 from os import PathLike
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @M-Advis 

Thanks for the thorough diagnosis on this one - the root cause write-up is correct, and the regression test is a keeper.

I'm going to close this PR in favour of a smaller, non-heuristic fix. The proposed approach infers file spacing from the data (Counter, confidence %, 1.5x threshold), which is two layers of heuristic for a problem that already has a principled answer: the caller told us what period to bucket into, so I'll use that as the split criterion. Concretely, split groups only when `curr_start - prev_end > period_in_ns`. That handles fragment-per-flush, legacy chunked files, and genuine multi-period holes without any inference, magic numbers, or per-call logging.

I'll land that fix shortly and pull in your test case so the regression stays covered.

M-Advis:
Thanks for the rationale, makes sense. I did do the plain 1 ns check first, but hit edge cases with timestamp rounding, which tempted me toward the heuristic. 

Good to know the period itself is the cleaner anchor — glad the test and root cause write-up are useful for the final fix.

While testing this I also stumbled on two bugs in a different method (`consolidate_data_by_period` silent data destruction on repeated calls and single-file catalogs), opened a separate issue and PR to keep things clean: **#3875**.

PR #3872, opened 2026-04-15
Fix Kraken fall back to /fills in Futures generate_order_status_report

### LINKED ISSUE #3870: Kraken Futures: `generate_order_status_report` resolves fast-fill market orders as rejected
## Summary

On the Kraken Futures execution client, `generate_order_status_report` only queries the `/derivatives/api/v3/openorders` endpoint. Market orders that fully fill in under ~200ms disappear from `/openorders` before the deferred-fill reconciliation loop first polls the venue. Each of the loop's retries then receives `None`, and the order is eventually resolved as `REJECTED` even though the fill executed on-venue.

## Reproduction

1. Submit a Kraken Futures `MARKET` order on an instrument with tight spreads and sufficient taker liquidity (e.g. `PF_XBTUSD`).
2. The venue fills the order in well under the first reconciliation poll interval.
3. `ExecutionClient.generate_order_status_report` returns `None` because the order is no longer in `/openorders`.
4. After the reconciliation retries are exhausted, Nautilus emits `OrderRejected` for an order that actually filled. The fill lands on `/fills` but is never tied back to the submission.

## Evidence (redacted)

Representative log sequence with identifiers redacted as `X` (client_order_id) and `Y` (venue_order_id):

```
submit_order cli=X
venue ack   cli=X venue=Y
<venue fills in ~150 ms>
generate_order_status_report cli=X venue=Y -> None   (retry 1/5)
generate_order_status_report cli=X venue=Y -> None   (retry 2/5)
...
resolve order rejected cli=X
<later> /fills yields fill for venue=Y, instrument matches submission, no matching order in Nautilus
```

## Root cause

`generate_order_status_report` calls `request_order_status_reports` which currently only hits `/openorders`. Kraken removes fully-filled orders from that endpoint immediately on fill; it does not backfill them via the order-events history fast enough for the sub-200ms case either. The `/fills` endpoint has the execution, but the Futures execution client does not consult it from `generate_order_status_report`.

## Proposed fix

When the `/openorders` lookup returns no match, fall back to `request_fill_reports` for the last 5 minutes and synthesise a `Filled` `OrderStatusReport` from any matching fill. Match in this order:

1. `venue_order_id` (always populated in Kraken fills).
2. `client_order_id`, compared against the truncated form Kraken stores server-side (same `truncate_cl_ord_id` helper already used for `/openorders` matching).
3. `instrument_id`, scoped to the 5-minute window — unambiguous for single-order-per-rebalance strategies and a last-resort match when Kraken omits `cli_ord_id` from the fill entirely.

Return `None` only when no fill matches.

## Downstream workaround in use today

We patch this downstream in Python via a script that monkey-patches `KrakenExecutionClient.generate_order_status_report` to perform the same `/fills` fallback. Lifting the fix upstream in the Rust adapter removes the need for the patch and closes the same race for all users.

## Scope

- Affects: `KrakenFuturesExecutionClient::generate_order_status_report` only.
- Does not change: `/openorders` success path, `generate_order_status_reports` (multi), `generate_fill_reports`, any non-Futures code path.
- No public API changes.

## Would a PR be welcome?

I have a branch ready (`fix/kraken-futures-order-status-fills-fallback`) with the fix and unit tests covering all three match orders plus negative cases. Happy to open it against `develop` once this issue is triaged.

cc @cjdsellers @Stamppot82

Stamppot82:
## Summary

Fixes a race on the Kraken Futures execution client where fast-filling market orders are resolved as `REJECTED` because `generate_order_status_report` only queries `/openorders`, which no longer contains the order by the time the deferred-fill reconciliation loop polls for status.

Closes #3870

## Root cause

`KrakenFuturesExecutionClient::generate_order_status_report` currently calls `request_order_status_reports`, which hits `/derivatives/api/v3/openorders`. Kraken drops filled orders from `/openorders` immediately; a market order that fills in under ~200ms is therefore absent from every retry of the reconciliation loop, and Nautilus eventually emits `OrderRejected` for an order that actually executed. The fill lands on `/fills` but is never matched back to the submission.

## Change

- When the `/openorders` match returns `None`, fall back to `request_fill_reports` for the last 5 minutes and synthesise a `Filled` `OrderStatusReport` from any matching fill.
- Match order: `venue_order_id` first (always populated in Kraken fills), then `client_order_id` compared against the truncated form Kraken stores (`truncate_cl_ord_id`, already used on the `/openorders` path), finally `instrument_id` as a last-resort window-scoped match for the case where Kraken omits `cli_ord_id` from the fill entirely.
- Factored the match + synthesis into a pure helper `synthesize_order_status_from_fill` so the semantics are unit-testable without the full execution-client wiring.
- The existing `/openorders` success path is untouched; `generate_order_status_reports` (multi) and `generate_fill_reports` are untouched.

## Files changed

- `crates/adapters/kraken/src/execution/futures.rs` – fallback wired into `generate_order_status_report`; new `synthesize_order_status_from_fill` helper; unit tests.
- `RELEASES.md` – changelog entry under `Fixes`.

## Test plan

- [ ] `cargo test -p nautilus-kraken` – all green (adds 5 unit tests under `execution::futures::tests`).
- [ ] `cargo test -p nautilus-kraken execution::futures::tests` – the five new tests cover venue-order-id match, truncated client-order-id match, instrument-id fallback, no-match returning `None`, and empty-fills returning `None`. Verified TDD-style: all three positive-match tests fail on a stubbed helper returning `None`, pass with the implementation.
- [ ] `cargo +nightly fmt -p nautilus-kraken` – clean.
- [ ] `cargo clippy -p nautilus-kraken --all-targets -- -D warnings` – clean.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -41,6 +41,7 @@ Released on TBD (UTC).
 - Fixed Hyperliquid modify cancel-replace emitting stale `OrderCanceled` (#3827), thanks for reporting @P1YU5H-50N1
 - Fixed IB Gateway Docker image failing on ARM64 hosts (#3813), thanks for reporting @Baki-0501
 - Fixed Kraken Futures limit order `OrderUpdated` panic from wire `stop_price: 0.0` treated as trigger price
+- Fixed Kraken Futures `generate_order_status_report` resolving fast-fill market orders as rejected by falling back to `/fills` when `/openorders` misses the execution
 - Fixed Kraken Spot quote-quantity orders never reaching terminal state from base/quote size mismatch
 - Fixed Kraken trade dedup clearing the entire set at capacity instead of evicting the oldest entry
 - Fixed OKX option greeks not forwarded due to inaccessible Cython `cdef` subscription attribute
--- a/crates/adapters/kraken/src/execution/futures.rs
+++ b/crates/adapters/kraken/src/execution/futures.rs
@@ -665,7 +665,7 @@ impl ExecutionClient for KrakenFuturesExecutionClient {
 
         // Match by venue_order_id or client_order_id (comparing truncated form
         // since Kraken stores the truncated cl_ord_id for long IDs)
-        Ok(reports.into_iter().find(|r| {
+        let matched = reports.into_iter().find(|r| {
             cmd.venue_order_id
                 .is_some_and(|id| r.venue_order_id.as_str() == id.as_str())
                 || cmd.client_order_id.is_some_and(|id| {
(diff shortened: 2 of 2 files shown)


cjdsellers:
Thanks @Stamppot82, closing as fixed (see #3870).

PR #3881, opened 2026-04-16
Model Polymarket outcome tokens as base currencies

### LINKED ISSUE #3880: [Polymarket] Model Polymarket outcome tokens as base currencies
## Summary

Polymarket outcome tokens are currently treated more like off-balance positions than first-class assets.
This creates an accounting mismatch for `BinaryOption` instruments in cash accounts: cash is reduced on entry,
but the acquired outcome token is not represented in account balances as a real asset.

That gap leaks into valuation and reporting, especially for total account equity and `PnL(total)`.

## Problem

A Polymarket `BinaryOption` only carries quote/cost currency semantics in the account model.
After a buy:

- quote cash is reduced
- the open position exists
- but the acquired outcome token is not represented as a balance asset

This is different from normal spot-style accounting, where buying `BTC` with `USD` leaves the account with:

- less `USD`
- more `BTC`

Because the Polymarket outcome token is not balance-backed, account equity and PnL reporting have to infer value
from positions instead of balances. That makes the model less coherent and caused incorrect or misleading `PnL(total)`
output in Polymarket cash-account paths.

## Proposed direction

Model each Polymarket outcome token as a real base asset for `BinaryOption`.

Concretely:

1. Add optional `base_currency` support to `BinaryOption`
2. Create a token-specific currency for each Polymarket outcome token
3. Use that currency as the `base_currency` of the parsed `BinaryOption`
4. Carry that field through Python/Rust bindings and serialization
5. Let cash accounting treat fills as real `+base / -quote` or `-base / +quote` asset movements
6. Include those balances in account snapshots and portfolio valuation

This makes Polymarket outcome tokens behave more like first-class assets and removes the need to treat them as a
special off-balance case in the account model.

## Currency identity

The token currency should remain uniquely identified by `token_id`, while still being readable in metadata.

Proposed convention:

- `code = Polymarket-{token_id}`
- `name = Polymarket-{slug_or_null}-{outcome_or_null}-{token_id}`

This keeps uniqueness stable while making balances and debugging output easier to interpret.

## Scope

The work needed spans several layers:

- `BinaryOption` model:
  add `base_currency` in both Python/Cython and Rust
- Polymarket parsing/provider:
  create and register outcome-token currencies
- Serialization:
  preserve `base_currency` in Arrow/SQL/PyO3 round-trips
- Accounting:
  make cash-account fills generate real base/quote legs
- Account state reporting:
  expose token balances alongside collateral balances
- Portfolio analysis:
  value multi-asset balances correctly in reporting currency

## Expected outcome

After the change:

- Polymarket outcome tokens are represented as account balance assets
- account equity is derived from balances more naturally
- `PnL(total)` is based on portfolio value semantics instead of relying on a venue-specific workaround
- backtest/live reporting becomes more consistent with spot-like asset accounting

## Notes

- This direction is intentionally broader than a one-off `PnL(total)` patch
- The goal is to fix the asset model first, then let valuation/reporting follow that model
- This is specific to Polymarket-style binary outcome tokens and should not change normal spot cash-account semantics

fedoraiver:
# Pull Request

## Summary

This PR models Polymarket outcome tokens as first-class base assets for `BinaryOption` instead of treating them as effectively off-balance positions. It adds `base_currency` support to the instrument model, creates token-specific currencies in the Polymarket Python and Rust parsers/providers, and carries that field through serialization and bindings. Accounting and portfolio analysis are updated so token balances participate in account valuation and `PnL(total)`, and backtest footer output is limited to quote-currency totals.

## Related Issues/PRs

Closes #3880
## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

N/A.

## Documentation

## Release notes

## Testing


cjdsellers:
Hi @fedoraiver

Thanks for the write-up on #3880 proposing this change - the diagnosis is valid: outcome tokens being off-balance is why `PnL(total)` ends up inferred from positions rather than balances.

However, we can't merge this PR as is for a few reasons:

- Modeling each outcome token as a `Currency` registers it in the global static `CURRENCY_MAP`, which has no eviction. For a long-running process picking up new markets continuously, that's unbounded growth, and the xrate plumbing gets overloaded to do portfolio mark-to-market for assets that aren't really fungible mediums of exchange.
- The change reaches into shared machinery (backtest footer, `AccountsManager` single-currency routing, `Cache._xrate_symbols`, `PortfolioAnalyzer` becoming cache-dependent) in ways that alter behaviour for non-Polymarket venues. The base-currency drop in the backtest footer, for instance, affects every spot/perpetual backtest.
- Three callers of `BinaryOption::new`/`new_checked` weren't updated (`filters.rs`, `exec_client.rs`, the in-file test), so the workspace doesn't build. The Arrow column re-indexing also silently breaks existing `BinaryOption` catalog files with no migration path.
- `_update_account_state` fetching `/positions` on every account refresh to synthesize balances duplicates the fill-driven path and risks drift; that probably belongs in reconciliation, not every state emission.

The cleaner direction I'd like to explore is valuing open positions at mark in the target currency (`equity = cash + sum(position.mark_value)`), which gives the same `PnL(total)` semantics without minting a synthetic currency per market or touching xrate infra. We could convert the #3880 ticket into tracking this direction?

So closing this, but #3880 can stay open for further discussion. Thanks again for the work here, and please keep reporting if you hit other Polymarket accounting gaps.

fedoraiver:
Hi @cjdsellers,

Thanks for the detailed review and for clarifying the preferred direction.

I've updated #3880 to track the narrower mark-to-market approach for total equity / `PnL(total)` instead of the synthetic outcome-token `Currency` approach from this PR.

I'll treat #3881 as closed and use #3880 to track follow-up discussion/work on:
- valuing open Polymarket positions at mark in the target currency
- deriving `PnL(total)` from total portfolio value semantics
- avoiding broader shared-infra changes outside that scope

Thanks again.

PR #3902, opened 2026-04-20
[Polymarket] Keep token_meta in sync for request_instrument(s)

### LINKED ISSUE #3900: [Polymarket] request_instrument(s) populates instruments cache but not token_meta
## Problem Found

In the Polymarket live data client, `request_instruments()` and `request_instrument()` currently insert returned instruments into `self.instruments`, but do not update `self.token_meta`.

This can leave the client in a partially updated state:

- the instrument is already present in the live client cache
- later subscribe paths can proceed because the token ID can still be derived from `instrument.raw_symbol()`
- but incoming market WS messages still rely on `token_meta` to resolve `asset_id/token_id -> instrument_id`
- as a result, the client may drop market WS messages even though the instrument had already been returned by `request_instrument(s)`

This is easier to hit in workflows that discover markets dynamically at runtime through `request_instruments()`, rather than through bootstrap or the missing-instrument auto-load path.

Observed symptoms can include logs such as:

- `Unknown asset_id in trade: ...`
- `No instrument for token_id ...`

while the same instrument had already been fetched successfully through `request_instrument(s)`.

There is already a related improvement in `f6f2a76c37b67eeadad9a78834d6d88b93f6f5fe`, which correctly updates `token_meta` for bootstrap and missing-instrument auto-load flows. However, the explicit `request_instruments()` and `request_instrument()` paths still only update `self.instruments`.

## Question

What should the cache contract of `request_instrument(s)` be?

There seem to be two possible interpretations:

1. These methods are pure request/lookup APIs.
   In that case, they probably should not mutate the live client cache at all.

2. These methods are cache-populating discovery APIs.
   In that case, once they insert into `self.instruments`, they should also maintain the dependent routing index in `self.token_meta`.

The current implementation already behaves like option 2, because it mutates `self.instruments` and feeds returned instruments back into the live system. Because of that, only updating one side of the cache looks inconsistent.

## Proposed Solution

If the intended behavior is that `request_instrument(s)` populates the live client cache, then both methods should update `self.token_meta` whenever they insert into `self.instruments`.

Concretely:

1. clone `self.token_meta` into the async task in both `request_instruments()` and `request_instrument()`
2. after inserting each returned instrument into `self.instruments`, also insert the corresponding `TokenMeta`
3. keep this behavior aligned with the existing bootstrap and auto-load code paths

That would make all cache-populating paths consistent:

- `bootstrap_instruments()` updates both caches
- missing-instrument auto-load updates both caches
- `request_instrument(s)` should also update both caches

If instead the intended behavior is option 1, then it may be better to revisit whether `request_instrument(s)` should mutate `self.instruments` at all.

fedoraiver:
## Summary

This PR fixes the cache inconsistency described in #3900.

`request_instruments()` and `request_instrument()` were already populating the live `instruments` cache, but they were not updating the dependent `token_meta` index used by market websocket routing. That could leave the client in a partially updated state where an instrument had been fetched successfully, but subsequent WS messages for the same token were still dropped.

This change keeps those request paths aligned with the existing bootstrap and missing-instrument auto-load flows by updating both caches together.

## Changes

- clone `token_meta` into the async request paths
- insert `TokenMeta` whenever instruments are inserted into the live cache via `request_instrument(s)`
- keep cache-populating flows consistent across Polymarket data client paths

## Validation

Fixes #3900

### DIFF
--- a/crates/adapters/polymarket/src/data.rs
+++ b/crates/adapters/polymarket/src/data.rs
@@ -1104,6 +1104,7 @@ impl DataClient for PolymarketDataClient {
         let filters = self.provider.filters();
         let sender = self.data_sender.clone();
         let instruments_cache = self.instruments.clone();
+        let token_meta = self.token_meta.clone();
         let request_id = request.request_id;
         let client_id = request.client_id.unwrap_or(self.client_id);
         let venue = *POLYMARKET_VENUE;
@@ -1118,7 +1119,16 @@ impl DataClient for PolymarketDataClient {
                     log::info!("Fetched {} instruments from Gamma API", instruments.len());
 
                     for instrument in &instruments {
-                        instruments_cache.insert(instrument.id(), instrument.clone());
+                        let instrument_id = instrument.id();
+                        instruments_cache.insert(instrument_id, instrument.clone());
+                        token_meta.insert(
+                            Ustr::from(instrument.raw_symbol().as_str()),
+                            TokenMeta {
+                                instrument_id,
+                                price_precision: instrument.price_precision(),
+                                size_precision: instrument.size_precision(),
+                            },
+                        );
                     }
 
                     let response = DataResponse::Instruments(InstrumentsResponse::new(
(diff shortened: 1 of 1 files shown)


cjdsellers:
Thanks again @fedoraiver 

Closing in favor of https://github.com/nautechsystems/nautilus_trader/commit/38566beaba1bf51795551fe40783259d5529c775.

Issue #3917, opened 2026-04-23
Make fee models pluggable at runtime instead of extending `FeeModelAny`


fedoraiver:
## Summary

Issue [#3838](https://github.com/nautechsystems/nautilus_trader/issues/3838) fixed Polymarket fee calculation in the adapter/live path, but using the same venue-specific fee logic in Rust backtests still required patching core library code.

The root cause is that the backtest runtime uses a closed enum (`FeeModelAny`) rather than a pluggable runtime interface. As a result, every new fee rule currently requires adding a new enum variant and threading it through Rust and Python bindings.

I think this is an API/design limitation in the backtest layer, not just a missing Polymarket-specific model.

## Problem

Today the Rust backtest path is effectively closed for fee model extension:

- `SimulatedVenueConfig` stores `fee_model: FeeModelAny`
- `SimulatedExchange` stores `fee_model: FeeModelAny`
- `OrderMatchingEngine` stores `fee_model: FeeModelAny`
- when an order is filled, the matching engine directly calls `self.fee_model.get_commission(...)`

This means that downstream Rust users cannot implement their own `FeeModel` in an external crate and inject it into backtests.

In practice, if a venue has:

- venue-specific fee semantics
- evolving fee schedules
- strategy-specific simulation requirements
- temporary experimental fee logic

then the only way to test it in backtests is to modify Nautilus core and add another `FeeModelAny` variant.

## Why this surfaced with Polymarket

After #3838, the Polymarket adapter correctly uses:

```text
fee = qty * feeRate * price * (1 - price)
```

However, the Rust backtest path still went through the generic maker/taker model. To replay the same logic in backtests, I had to add a new `PolymarketFeeModel` variant in core and wire it through:

- `crates/execution/src/models/fee.rs`
- `crates/backtest/src/python/engine.rs`
- `crates/execution/src/python/fee.rs`
- `crates/execution/src/python/mod.rs`

That works, but it does not scale well. The next venue-specific fee rule would require the same kind of core patch again.

## Why the current design feels inconsistent

The runtime backtest API is already trait-object based in other places:

- `modules: Vec<Box<dyn SimulationModule>>`
- `latency_model: Option<Box<dyn LatencyModel>>`

But `fee_model` is still a closed enum.

So right now:

- latency behavior is extensible at runtime
- simulation module behavior is extensible at runtime
- fee behavior is not extensible at runtime

That asymmetry is surprising for downstream Rust users.

## Proposed direction

[...]

PR #3919, opened 2026-04-23
Make fee models explicitly runtime-pluggable

fedoraiver:
## Summary
- store runtime backtest fee models as `Arc<dyn FeeModel>` instead of routing everything through `FeeModelAny`
- remove the implicit `FeeModelAny -> Arc<dyn FeeModel>` conversion so runtime boundaries require explicit wrapping
- update backtest and execution call sites/tests to pass explicit shared fee models and cover custom runtime fee model injection

## Why
Issue #3917 is about making backtest fee models pluggable at runtime without extending the built-in enum each time a venue-specific fee rule is added.

The backtest runtime now accepts a trait object directly, while config-layer code can still decide how to construct that object before handing it to the simulated venue.

## Validation
- `cargo test --manifest-path Cargo.toml -p nautilus-backtest --lib`
- `cargo test --manifest-path Cargo.toml -p nautilus-execution --lib`
- `cargo clippy --manifest-path Cargo.toml -p nautilus-backtest -p nautilus-execution --tests -- -D warnings`


cjdsellers:
Hi @fedoraiver 

See comment on #3917. Please hold off on further work on this for now.

PR #3924, opened 2026-04-24
Fix custom data Parquet schema registration

### LINKED ISSUE #3923: Custom data dynamic Parquet queries can miss registered schema during DataFusion planning
## Problem

`ParquetDataCatalog::query_custom_data_dynamic` can fail or silently under-read when querying registered custom data from explicit Parquet files.

Two related issues show up in the current custom-data path:

1. Custom data schema registration happens too late for DataFusion planning. `DataBackendSession::add_file` registers Parquet with `file_sort_order = ts_init`, but custom type identity is only injected later before batch decode. If `CustomDataDecoder::get_schema` does not see `type_name` during registration, it can fall back to the dummy schema, and DataFusion reports:

```text
Schema error: No field named ts_init
```

2. `query_custom_data_dynamic` registers every explicit custom data file as the same `custom_data_table`. `DataBackendSession` intentionally skips duplicate table registration, so querying multiple explicit files may register only the first file.

## Expected behavior

For a Rust-registered custom data type, `query_custom_data_dynamic(type_name, files=...)` should register each Parquet file with the real custom Arrow schema before DataFusion plans the scan, including `ts_init` and the persisted `data_type` column. Multiple explicit files should all be registered and returned.

## Impact

This affects custom data stored in Parquet and queried dynamically by type name, particularly local backtest/catalog workflows that pass explicit custom data files instead of relying only on broad directory discovery.

## Proposed fix

- Allow `DataBackendSession::add_file` callers to provide an explicit Arrow schema.
- Have `query_custom_data_dynamic` resolve the registered custom schema from `type_name` before Parquet registration.
- Validate that the schema contains `ts_init`.
- Include the persisted `data_type` column in the schema passed to DataFusion.
- Use a unique table name per explicit custom data file.
- Add a regression test for querying two explicit Rust custom data Parquet files.

fedoraiver:
## Summary

Fixes dynamic custom-data Parquet queries so DataFusion receives the registered custom Arrow schema before planning the scan.

`query_custom_data_dynamic` now resolves the schema from `type_name`, verifies it contains `ts_init`, augments it with the persisted `data_type` column, and registers each explicit custom-data file with a unique table name.

## Root Cause

`DataBackendSession::add_file` previously registered Parquet files without an explicit schema. For custom data, the `type_name` was only injected later before batch decoding, which meant DataFusion planning could see the `CustomDataDecoder` dummy fallback schema rather than the real registered custom schema. Since Parquet registration also declares `file_sort_order` on `ts_init`, this can fail with `Schema error: No field named ts_init`.

The dynamic custom-data query path also reused the same `custom_data_table` name for every explicit file, so duplicate-table skipping could cause multi-file queries to under-read.

## Changes

- Added `DataBackendSession::add_file_with_schema` while preserving the existing `add_file` API.
- Updated `ParquetDataCatalog::query_custom_data_dynamic` to pass the registered custom schema into DataFusion registration.
- Added a `ts_init` schema validation error for unregistered or invalid custom data schemas.
- Added unique table names for explicit custom data files.
- Updated the Python `add_custom_file` binding to use the same explicit-schema registration path.
- Added a regression test for querying two explicit Rust custom-data Parquet files.

## Validation

- `cargo fmt --manifest-path nautilus_trader/Cargo.toml --package nautilus-persistence`
- `CARGO_TARGET_DIR=/tmp/nautilus-target-codex-custom-schema cargo test --manifest-path nautilus_trader/Cargo.toml -p nautilus-persistence --test test_catalog test_query_custom_data_dynamic_with_explicit_files -- --nocapture`
- Commit hooks: `cargo fmt`, `cargo clippy`, and `cargo doc` for `nautilus-persistence` passed during commit.

Closes #3923

### DIFF
--- a/crates/model/src/python/data/mod.rs
+++ b/crates/model/src/python/data/mod.rs
@@ -466,6 +466,17 @@ fn py_encode_custom_data_to_record_batch(
     })
 }
 
+#[allow(unsafe_code)]
+#[cfg(feature = "python")]
+fn pyarrow_schema_to_arrow_schema(
+    py_schema: &pyo3::Bound<'_, pyo3::PyAny>,
+) -> PyResult<arrow::datatypes::Schema> {
+    let mut ffi_schema = arrow::ffi::FFI_ArrowSchema::empty();
+    py_schema.call_method1("_export_to_c", ((&raw mut ffi_schema as usize),))?;
+    arrow::datatypes::Schema::try_from(&ffi_schema)
+        .map_err(|e| to_pyvalue_err(format!("Failed to import PyArrow schema: {e}")))
+}
+
 /// Decodes `RecordBatch` to `CustomData` via Python `decode_record_batch_py`.
 #[allow(unsafe_code)]
 #[cfg(feature = "python")]
@@ -606,7 +617,13 @@ pub fn register_custom_data_class(data_class: &Bound<'_, PyAny>) -> PyResult<()>
         ))
     })?;
 
-    let schema = Arc::new(arrow::datatypes::Schema::empty());
+    let schema = if let Ok(py_schema) = data_class.getattr("_schema") {
+        Arc::new(pyarrow_schema_to_arrow_schema(&py_schema)?)
+    } else if let Some(schema) = registry::get_arrow_schema(&type_name) {
+        schema
+    } else {
+        Arc::new(arrow::datatypes::Schema::empty())
+    };
 
     let encoder = Box::new(
         move |items: &[Arc<dyn crate::data::CustomDataTrait>]| -> Result<
--- a/crates/persistence/src/backend/catalog.rs
+++ b/crates/persistence/src/backend/catalog.rs
@@ -64,7 +64,7 @@
 
 use std::{
     borrow::Cow,
-    collections::{BTreeMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     fmt::Debug,
     io::Cursor,
     ops::Bound as RangeBound,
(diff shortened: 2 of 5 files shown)


faysou:
Is the PR ready for review ? It's still in draft mode.

fedoraiver:
> Is the PR ready for review ? It's still in draft mode.

Yes, it's ready for review now. I've marked it as ready—thank you for checking!

cjdsellers:
Hi @fedoraiver 

Looks like this is hanging CI. Are you able to test locally?

fedoraiver:
Hi @cjdsellers, I tested this locally.

On the previous head (3d1fa8c), the Rust regression test passed, but the targeted Python/PyO3 custom data test exposed a real issue: the two Python-only `customdataclass_pyo3` roundtrip cases failed because `register_custom_data_class` registered an empty Arrow schema, so query later could not find `ts_init`.

I pushed a fix in ad0a8bb: `register_custom_data_class` now imports the class `_schema` from PyArrow into the Rust Arrow registry when available, while keeping the existing Rust schema fallback.

Local checks on ad0a8bb:
- `cargo fmt --check --all`: passed (only stable rustfmt warnings for nightly-only config options)
- `pytest -n0 -vv -s tests/unit_tests/persistence/test_custom_data_roundtrip_pyo3.py`: 6 passed
- `pytest -n logical --dist=loadgroup --reruns 2 --reruns-delay 1 tests/unit_tests/persistence/test_custom_data_roundtrip_pyo3.py`: 6 passed
- `cargo nextest run -p nautilus-persistence --test test_catalog test_query_custom_data_dynamic_with_explicit_files --features "ffi,python,high-precision" --profile ci --cargo-profile nextest --no-fail-fast`: 1 passed

I did not reproduce the xdist hang in the targeted test after the fix. The new CI run for ad0a8bb is queued now.

faysou:
Tests are not passing.

faysou:
Closing this PR, I finished it in another PR

cjdsellers:
Thanks again @fedoraiver we're aware of the issue now, @faysou can continue working on this.

PR #3930, opened 2026-04-26
Fix remote catalog object paths under URI prefix

### LINKED ISSUE #3929: Remote `ParquetDataCatalog` drops the URI path prefix
## Problem

When a `ParquetDataCatalog` is created from a remote URI with a path prefix, for example:

```text
s3://my-bucket/catalog
```

catalog data is written under the bucket root:

```text
s3://my-bucket/data/...
```

instead of under the configured catalog prefix:

```text
s3://my-bucket/catalog/data/...
```

So the S3 bucket ends up with a top-level `data/` directory even though the catalog URI includes `catalog`.

## Cause

`create_object_store_from_path("s3://my-bucket/catalog", ...)` creates the S3 object store using only the bucket name:

```rust
object_store::aws::AmazonS3Builder::new().with_bucket_name(&bucket)
```

The URI path segment, `catalog`, is returned separately as `base_path`.

For local catalogs this is fine because `LocalFileSystem::new_with_prefix(...)` scopes the object store to the catalog directory, so stripping `base_path` before object-store operations makes sense.

For remote catalogs, the object store root is still the bucket/container root. The `base_path` is not built into the S3/GCS/Azure object store itself, so it must remain part of the `ObjectPath`.

However, `ParquetDataCatalog::to_object_path(...)` currently applies the same stripping behavior to remote catalogs:

```text
catalog/data/quotes/file.parquet
```

becomes:

```text
data/quotes/file.parquet
```

That stripped object path is then passed to object-store operations such as `put`, `get`, `list`, or `rename`, causing the files to be addressed relative to the bucket root instead of the catalog prefix.

fedoraiver:
Closes https://github.com/nautechsystems/nautilus_trader/issues/3929

## Summary

- Preserve or prepend the remote catalog URI path prefix when converting catalog paths to `ObjectPath`.
- Avoid duplicating the prefix when the input path already includes the remote catalog base path.
- Keep existing local catalog behavior, where the local filesystem store is already scoped to the catalog directory.
- Reconstruct remote full URIs with the catalog path prefix included.

## Root Cause

Remote object stores are created at the bucket/container root, while the URI path is returned separately as `base_path`. The previous `to_object_path` behavior stripped `base_path` for both local and remote catalogs, so paths like `catalog/data/...` became `data/...` and were addressed at the bucket root.

## Tests

- `cargo test -p nautilus-persistence --features cloud test_remote_to_object_path_preserves_base_path`
- `cargo test -p nautilus-persistence --features cloud test_reconstruct_full_uri_moved`
- `cargo check -p nautilus-persistence --features cloud`
- `cargo test -p nautilus-persistence --features cloud --test test_catalog`
- `make pre-commit`

### DIFF
--- a/crates/persistence/src/backend/catalog.rs
+++ b/crates/persistence/src/backend/catalog.rs
@@ -1350,13 +1350,18 @@ impl ParquetDataCatalog {
     /// Helper method to reconstruct full URI for remote object store paths
     #[must_use]
     pub fn reconstruct_full_uri(&self, path_str: &str) -> String {
+        if path_str.contains("://") {
+            return path_str.to_string();
+        }
+
         // Check if this is a remote URI scheme that needs reconstruction
         if self.is_remote_uri() {
             // Extract the base URL (scheme + host) from the original URI
             if let Ok(url) = url::Url::parse(&self.original_uri)
                 && let Some(host) = url.host_str()
             {
-                return format!("{}://{}/{}", url.scheme(), host, path_str);
+                let path = self.path_under_base(path_str);
+                return format!("{}://{}/{}", url.scheme(), host, path);
             }
         }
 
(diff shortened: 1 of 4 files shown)


cjdsellers:
Thanks @fedoraiver, this looks in good shape. Is the draft status intentional, or ready for review?

Before merging, it'd be good to extend `test_remote_to_object_path_preserves_base_path` to also cover `to_object_path_parsed` with an already-prefixed input, so both entry points are exercised against the round-trip case.

fedoraiver:
Thanks @cjdsellers. This is ready for review now. I pushed an update covering `to_object_path_parsed` with an already-prefixed input in `test_remote_to_object_path_preserves_base_path`, and tightened the remote catalog path handling so prefixes and URI authority are preserved. Targeted catalog tests and `cargo fmt --check` pass locally.

cjdsellers:
Thanks again @fedoraiver 

As a follow-up I'll tighten the cross-bucket full-URI path in `object_store_path` (currently falls through silently when the URI's root doesn't match the catalog's) and add a release note flagging the behavior change for prefixed remote catalogs.

PR #3933, opened 2026-04-26
Use setup-uv cache auto mode

sunlei:
# Pull Request

## Summary

Currently, the `uv` cache stored on self-hosted runners can reach 20-30GB. After each job completes, this cache file needs to be uploaded to GitHub.
As stated in the `uv` [documentation](https://docs.astral.sh/uv/guides/integration/github/#caching): `When using non-ephemeral, self-hosted runners the default cache directory can grow unbounded.`

Therefore, I'm adopting the recommended approach from the `uv` documentation to optimize these settings and enabling `prune-cache` so it can automatically execute `uv cache prune --ci`.

- Add a uv-cache-enabled input for common setup
- Default uv cache persistence to setup-uv auto mode
- Remove the custom actions/cache uv cache step
- Rely on setup-uv default dependency globs and prune-cache support

## Type of change

### DIFF
--- a/.github/actions/common-setup/action.yml
+++ b/.github/actions/common-setup/action.yml
@@ -35,6 +35,12 @@ inputs:
       runners using a persistent on-disk CARGO_TARGET_DIR).
     required: false
     default: "true"
+  uv-cache-enabled:
+    description: >-
+      Whether to persist the uv cache with GitHub Actions cache: true, false, or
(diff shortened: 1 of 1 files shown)


cjdsellers:
Thanks @sunlei 

I'll follow-up to apply the same auto pattern to the prek cache and to add an unconditional `uv cache prune --ci` so the local cache stays bounded on self-hosted runners.

PR #3935, opened 2026-04-26
Optimize performance for `ClientOrderId` generation

sunlei:
# Pull Request

## Summary

- Keep the fixed prefix in the output buffer and only truncate the previous count on same-second generation
- Use itoa to format the count and reduce format!/replace overhead on the hot path
- Preserve existing UUID behavior and ClientOrderId output formats
- Add a Criterion benchmark comparing the legacy generator logic with the optimized implementation

`cargo bench -p nautilus-common --bench client_order_id`

| Scenario | Legacy median | Optimized median | Improvement |
|---|---:|---:|---:|
| Same second, hyphenated ID | 551.39 ns/op | 248.23 ns/op | 2.22x faster, 55.0% less time |
| Same second, no hyphens | 637.82 ns/op | 272.36 ns/op | 2.34x faster, 57.3% less time |
| Cross-second, hyphenated ID | 680.84 ns/op | 480.06 ns/op | 1.42x faster, 29.5% less time |

## Type of change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5600,6 +5600,7 @@ dependencies = [
  "futures",
  "iai",
  "indexmap 2.14.0",
+ "itoa",
  "log",
  "madsim",
  "nautilus-core",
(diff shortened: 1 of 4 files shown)


cjdsellers:
Thanks @sunlei, great optimizations.

I'll apply these patterns to the other ID generators.

PR #3947, opened 2026-04-28
Add opt-in pruning for inactive Polymarket instrument state

### LINKED ISSUE #3945: [Polymarket] Add opt-in pruning for inactive Polymarket instrument state
## Summary

Add an opt-in `prune_inactive_instruments` setting to `PolymarketDataClientConfig`.

When enabled, the Polymarket data client prunes per-instrument market data state after the last active market data subscription for an instrument has been removed.

The intended use case is long-running processes which dynamically subscribe to many short-lived Polymarket markets, for example recurring prediction markets where instruments are opened, streamed briefly, then closed and never used again.

## Problem

`PolymarketDataClient` keeps several per-instrument caches while streaming live market data:

- `order_books`
- `last_quotes`
- `instruments`
- `token_meta`

For static instrument universes this is fine. For dynamic Polymarket workloads, especially short-duration markets, the client can discover and stream a large number of instruments over time. After market data subscriptions are removed, the instrument may no longer be useful to the process, but its cached state remains in memory.

In a 24/7 service, this means memory can grow with the number of historical instruments seen by the client instead of the number of currently active instruments.

## Proposal

Add a new config field:

```rust
pub struct PolymarketDataClientConfig {
    // ...
    /// Whether to prune per-instrument market data caches after all market data
    /// subscriptions are removed.
    #[builder(default)]
    pub prune_inactive_instruments: bool,
    // ...
}
```

Default: `false`.

This preserves current behavior unless the user explicitly opts in.

Expose the same option through the Python config constructor:

```python
PolymarketDataClientConfig(
    # ...
    prune_inactive_instruments=True,
)
```

## Behavior

After each market data unsubscribe path:

- `unsubscribe_book_deltas`
- `unsubscribe_quotes`
- `unsubscribe_trades`

the client checks whether the instrument still has any active market data subscriptions:

```rust
active_quote_subs.contains(&instrument_id)
    || active_delta_subs.contains(&instrument_id)
    || active_trade_subs.contains(&instrument_id)
```

If `prune_inactive_instruments` is `false`, cleanup is skipped entirely.

If the option is enabled and no active market data subscription remains, the client removes:

```rust
order_books.remove(&instrument_id);
last_quotes.remove(&instrument_id);
instruments.remove(&instrument_id);
token_meta.remove(&token_id);
```

The implementation resolves the token id from the live instrument cache when possible, and falls back to scanning `token_meta` by `instrument_id` when needed.

`ws_open_tokens` is intentionally not pruned by this cleanup path. WebSocket subscribe/unsubscribe state is reconciled separately by the existing websocket subscription logic.

fedoraiver:
Closes #3945

## Summary

- Add opt-in `prune_inactive_instruments` to `PolymarketDataClientConfig`.
- Expose the option through the Python config constructor.
- Prune per-instrument Polymarket market data state after the last quote, book-delta, or trade subscription is removed.
- Add unit coverage for disabled pruning, inactive pruning, active subscription retention, and `token_meta` fallback lookup.

## Notes

The option defaults to `false`, so existing cache retention behavior is preserved unless explicitly enabled.

`ws_open_tokens` is intentionally left under the existing websocket subscription reconciliation logic.

## Tests

```bash
cargo test -p nautilus-polymarket cleanup_inactive_instrument --lib
cargo test -p nautilus-polymarket token_id_for_instrument_falls_back_to_token_meta --lib
cargo check -p nautilus-polymarket --all-targets
```

### DIFF
--- a/crates/adapters/polymarket/src/config.rs
+++ b/crates/adapters/polymarket/src/config.rs
@@ -65,6 +65,10 @@ pub struct PolymarketDataClientConfig {
     /// The window (milliseconds) over which concurrent auto-load requests are batched.
     #[builder(default = 100)]
     pub auto_load_debounce_ms: u64,
+    /// Whether to prune per-instrument market data caches after all market data
+    /// subscriptions are removed.
+    #[builder(default)]
+    pub prune_inactive_instruments: bool,
     /// Instrument filters applied to all instruments during loading and discovery.
     #[builder(default)]
     pub filters: Vec<Arc<dyn InstrumentFilter>>,
--- a/crates/adapters/polymarket/src/data.rs
+++ b/crates/adapters/polymarket/src/data.rs
@@ -338,6 +338,44 @@ impl PolymarketDataClient {
         pending.remove(&instrument_id);
     }
 
+    fn has_active_subscriptions(&self, instrument_id: InstrumentId) -> bool {
+        self.active_quote_subs.contains(&instrument_id)
+            || self.active_delta_subs.contains(&instrument_id)
+            || self.active_trade_subs.contains(&instrument_id)
+    }
+
+    fn token_id_for_instrument(&self, instrument_id: InstrumentId) -> Option<Ustr> {
+        if let Some(instrument) = self.instruments.load().get(&instrument_id) {
+            return Some(Ustr::from(instrument.raw_symbol().as_str()));
+        }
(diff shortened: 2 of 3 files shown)


cjdsellers:
Thanks @fedoraiver, closing per #3945.

Issue #3790, opened 2026-04-02
Sandbox execution applies stale TradeTick updates to the L1 book while ts_last remains pinned


linimin:
# Bug Report

Use this template only for issues that fit the **Bug** definition.

| Term                          | Definition |
|-------------------------------|------------|
| **Bug**                       | Behavior that contradicts the platform's documented or intended design as per code, docs, or specs. (i.e., the implementation is incorrect.) |
| **Expectation mismatch**      | Behavior that follows the platform's documented or intended design but differs from what you expected. (i.e., the design/spec might be the problem.) |
| **Enhancement request**       | A request for new functionality or behavior that is not implied by existing design. (i.e., *"It would be great if the platform could..."*) |

**Note:**

- Submitting this issue automatically applies the `bug` label.
- `bug`-labeled issues are triaged with higher priority because they require corrective implementation work.
- **Expectation mismatches** and design-level concerns should be opened as [Discussions](https://github.com/nautechsystems/nautilus_trader/discussions), or RFCs instead, where they can be validated and discussed to consensus before any work is scheduled.
- The absence of a feature is typically not an expectation mismatch, and should be filed as an enhancement request.

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I've tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

Note: I reproduced this against the latest `develop` git commit `8f57525a6dfd51efb83290fac6e93588d1f10c96` in a Rust integration host and then reduced it to a deterministic sandbox adapter integration test. I have not yet repeated the repro against a prebuilt development wheel.

### Expected Behavior

When sandbox execution receives a `TradeTick` whose `ts_event` is older than the current execution-book high-water timestamp, that stale trade should not mutate the internal L1 book before stale-event handling runs.

If quote/trade events can arrive out of order, the sandbox execution path should either:

1. ignore or reorder stale events before mutating the internal book, or
2. otherwise preserve a book state whose best bid/ask and `ts_last` remain consistent with the same event ordering.

At minimum, an event already known to be older than `book.ts_last` should not change the execution-side top of book.

### Actual Behavior

[...]

PR #3992, opened 2026-05-04
Fix Cython OrderMatchingEngine.reset leaking OrderBook.ts_last on repeated runs

YeeTsai:
## Bug

`OrderMatchingEngine.reset` (Cython) calls `self._book.clear(0, 0)`, which routes through `OrderBook.increment` whose `ts_last`/`sequence` are high-water marks:

```rust
// crates/model/src/orderbook/book.rs:1024-1026
self.sequence = sequence.max(self.sequence);
self.ts_last = ts_event.max(self.ts_last);
```

So `clear(0, 0)` preserves the prior run's last timestamp.

The L1 stale-event guard added in #3790 then drops every subsequent quote/trade whose `ts_event < self.ts_last`:

```rust
// crates/model/src/orderbook/book.rs:1092-1098
if trade.ts_event < self.ts_last {
    log::warn!("Skipping stale trade: ...");
    return Ok(());  // book is not mutated
}
```

In repeated-run scenarios (sweep / parameter-search where one `BacktestEngine` instance is reset and re-run with the same data), this leaves the matching core's bid/ask uninitialised, and every market order on subsequent runs is rejected with reason `"no market for {instrument_id}"`.

The Rust matching engine reset path (`crates/execution/src/matching_engine/engine.rs:237`) already calls `book.reset()` (which zeroes `ts_last`); the Cython path was the outlier.

## Fix

One-line Cython change in `nautilus_trader/backtest/engine.pyx::OrderMatchingEngine.reset`:

```diff
-self._book.clear(0, 0)
+self._book.reset()
```

`OrderBook.reset()` clears bids/asks **and** zeroes `sequence`, `ts_last`, `update_count` — exactly the reset semantics needed so repeated runs behave like a fresh engine, matching the Rust path.

## Regression test

Added `tests/unit_tests/backtest/test_matching_engine.py::TestOrderMatchingEngine::test_reset_zeroes_book_ts_last`:

1. Seed L1_MBP book with a quote at `ts_event=100` → assert `ts_last == 100`
2. Reset matching engine
3. Assert `ts_last == 0` (the leak)
4. Replay a quote at earlier `ts_event=50` and verify it is accepted (best bid/ask updated)

Pre-fix: assertion 3 fails (`ts_last == 100`); also the replay quote in 4 is silently dropped.
Post-fix: all assertions pass.

## Validation

End-to-end reproducer (running 2 trials back-to-back through one `BacktestSweepEngine` against 2 years of BTCUSDT 1m kline data):

| | Pre-fix | Post-fix |
|---|---|---|
| Trial 1 positions | 2 | 2 |
| Trial 2 positions | **0** | **2** |
| Trial 2 OrderRejected | 266 | 0 |

Test suites: `tests/unit_tests/backtest/` 765/765 PASS (1 unrelated skip).

## Scope

- Affects: any code path that calls `BacktestEngine.reset()` (or per-venue `SimulatedExchange.reset()`) and re-runs with data whose `ts_event` is `<` the prior run's last — i.e. classic repeated-runs / sweep / parameter-search
- Unaffected: single-run backtests (each call builds a fresh engine, never hits this reset path), live trading, Rust matching engine path

### DIFF
--- a/nautilus_trader/backtest/engine.pyx
+++ b/nautilus_trader/backtest/engine.pyx
@@ -3992,7 +3992,16 @@ cdef class OrderMatchingEngine:
     cpdef void reset(self):
         self._log.debug(f"Resetting OrderMatchingEngine {self.instrument.id}")
 
-        self._book.clear(0, 0)
+        # Use `reset` (not `clear`) so `book.ts_last` is zeroed. `clear(0, 0)`
+        # routes through `OrderBook.increment` whose ts_last/sequence are
+        # high-water marks: `ts_event.max(self.ts_last)` keeps the stale
+        # timestamp, and the L1 stale-event guard added in #3790 then drops
+        # every subsequent quote/trade tick whose ts_event < that residual.
+        # Repeated runs (sweep / param-search) need the engine to accept the
+        # next run's bars even though they predate the previous run's last ts.
+        # The Rust matching engine reset path (`crates/execution/src/
+        # matching_engine/engine.rs::reset`) already calls `book.reset()`.
+        self._book.reset()
         self._account_ids.clear()
         self._execution_bar_types.clear()
         self._execution_bar_deltas.clear()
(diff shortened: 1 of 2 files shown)


cjdsellers:
Thanks for the fix @YeeTsai :pray:

Issue #3942, opened 2026-04-27
Handle tick-size changes atomically across book and matching


graceyangfan:
## Problem Statement

  NautilusTrader has an entry point for dynamic tick-size (instrument updates), but the support is incomplete end-to-end. A
  tick_size_change updates the Instrument definition, yet the transition is not treated as an atomic market-state change across:

  - adapter local book reconstruction / publication, and
  - sandbox/backtest consumption and matching (including MatchingCore).

  This can lead to either:

  1. Invalid market data emission (old-epoch book levels mixed into new tick grid), or
  2. Fatal precision validation errors in the sandbox matching engine (strict equality checks), or
  3. Stale tick-size inside MatchingCore (instrument updates do not propagate price_increment into the core), risking incorrect
     price-protection / grid-dependent logic.

  ### Current limitations (root causes)

  A. Polymarket adapter mixes epochs
  On tick_size_change, the adapter currently:

  1. reads old local book,
  2. creates new Instrument (new price_increment / price_precision),
  3. emits instrument update,
  4. rebuilds a new book from the old book,
  5. emits snapshot/deltas immediately.

  This is sometimes “okay” when tick becomes finer (e.g., 0.01 -> 0.001), but unsafe when tick becomes coarser (e.g., 0.001 ->
  0.01) because old book may contain prices like 0.505 (3 decimals) that are not valid on the new 0.01 grid. Rebuilding/quantizing
  can produce invalid or misordered data.

  B. Sandbox matching does not fully apply instrument updates
  OrderMatchingEngine.update_instrument() updates:

  - self.instrument
  - self._price_prec
  - self._size_prec

  But MatchingCore captures price_increment at initialization and keeps it stale thereafter. Additionally, strict price precision
  equality checks in market data processing (process_order_book_delta(s), process_quote_tick, process_trade_tick, etc.) can raise
  fatal errors during tick-size transitions.

  ## Proposed Solution

  Implement a two-layer fix (adapter + core), with an optional third layer (tick-compatible validation) to improve resilience
  without masking real epoch bugs.

  ———

  ### 1) Adapter fix (Polymarket): Treat tick-size change as an order book epoch transition (resync)

  In PolymarketDataClient handling of tick_size_change:

  1. Update instrument cache/provider with the new Instrument (new price_increment, price_precision) and publish the instrument
     update event.
  2. Clear local book state and last quote for that instrument immediately.
  3. Mark the instrument as pending_snapshot_after_tick_change.
[...]

PR #3994, opened 2026-05-04
Fix sandbox race on tick-size precision changes

graceyangfan:
## Summary
This PR fixes a race where a stale `QuoteTick` created just before a `tick_size_change` could be dequeued later and crash sandbox matching with a precision mismatch panic.

Handle tick-size changes atomically across book and matching #3942

## Root Cause
- Adapter updates instrument precision and resets local book around `tick_size_change`.
- But a pre-change quote can already be in the async queue.
- `OrderMatchingEngine::process_quote_tick` previously `unwrap()`ed precision checks against current instrument precision.
- On mismatch, this panicked, propagated through PyO3, and could terminate the process.

### 1) Matching engine hardening (`crates/execution/src/matching_engine/engine.rs`)
- Add `update_instrument(instrument)` to update instrument metadata in-place.
- Detect precision/grid changes using the minimal set:
  - `price_increment`
  - `price_precision`
  - `size_precision`
- When changed, atomically reset dependent market/matching state (book + queue/consumption baselines) and update core increment.
- Replace precision-check `unwrap()` in `process_quote_tick` / `process_trade_tick` with guarded early-return + warning logs.
- Add a small consecutive mismatch streak logger to avoid noisy repeated warnings.

### 2) Sandbox execution integration (`crates/adapters/sandbox/src/execution.rs`)
- Implement `on_instrument` to push instrument updates into existing matching engines via `update_instrument`.
- Add a precision pre-guard in sandbox quote handling to drop mismatched stale quotes before entering matching.

### 3) Tests
- `crates/execution/tests/matching_engine.rs`
  - `test_update_instrument_resets_market_state`
  - `test_update_instrument_without_precision_change_keeps_market_state`
- `crates/adapters/sandbox/tests/execution.rs`
  - `test_process_quote_tick_drops_precision_mismatch`
  - `test_on_instrument_updates_engine_precision`

## Why This Shape
- Prevents crash first (no panic on stale precision mismatch).
- Keeps matching engine and instrument precision/grid aligned over time.
- Keeps behavior safe in non-`tick_size_change` markets by only resetting on true precision/grid change.

## Validation
Ran locally:
- `cargo test -p nautilus-execution --test matching_engine -- --nocapture`
- `cargo test -p nautilus-sandbox --test execution -- --nocapture`
- `cargo fmt --check`
- `pre-commit run --files crates/adapters/sandbox/src/execution.rs crates/adapters/sandbox/tests/execution.rs crates/execution/src/matching_engine/engine.rs crates/execution/tests/matching_engine.rs`

## Known Limitation / Follow-up
In live probe logs, post-change recovery (change -> resumed book) can still be slow in some cases (~10s observed) because recovery depends on passive snapshot/stream progression.

Recommended follow-up: after `tick_size_change`, proactively request a fresh snapshot for that market immediately, instead of waiting passively.

### DIFF
--- a/crates/adapters/sandbox/src/execution.rs
+++ b/crates/adapters/sandbox/src/execution.rs
@@ -77,6 +77,16 @@ struct SandboxInner {
     event_handler: Option<Rc<dyn Fn(OrderEventAny)>>,
 }
 
+fn quote_matches_instrument_precision(quote: &QuoteTick, instrument: &InstrumentAny) -> bool {
+    let price_precision = instrument.price_precision();
+    let size_precision = instrument.size_precision();
+
+    quote.bid_price.precision == price_precision
+        && quote.ask_price.precision == price_precision
+        && quote.bid_size.precision == size_precision
+        && quote.ask_size.precision == size_precision
+}
+
 impl SandboxInner {
     /// Ensures a matching engine exists for the given instrument.
     fn ensure_matching_engine(&mut self, instrument: &InstrumentAny) {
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @graceyangfan 

Many thanks for the fix :pray:

PR #4041, opened 2026-05-08
Add Kraken Spot L3 order book support via WebSocket v2

### LINKED ISSUE #4001: [Kraken] L3 order book support via WebSocket v2 level3 channel
# Feature Request

## Description

The Kraken WebSocket v2 API exposes a `level3` channel providing individual order-level book data: real venue order IDs, per-order quantities, and true incremental events (`add`, `modify`, `delete`). The Nautilus framework fully supports `BookType::L3_MBO` — `OrderBookDelta`, `BookOrder`, `BookAction` are all in place. The Kraken adapter currently rejects L3 subscriptions and only handles L2.

L3 data unlocks strategies that L2 cannot support:

- **Queue-aware market making** — know exactly how many orders sit ahead of yours at a price level
- **Iceberg detection** — track individual order IDs refreshing at the same level
- **Order flow analysis** — add/cancel rates and size distributions per order, not per level
- **Toxicity filtering** — distinguish one large passive order from many small ones at the same price

## Possible Implementation

- Message structs for `level3` snapshot and incremental update (per-order `order_id`, `limit_price`, `order_qty`)
- Stable hash of venue order ID string → `u64` for `BookOrder.order_id`
- Remove the `L2_MBP`-only guard in `data/spot.rs` and route `L3_MBO` subscriptions to the `level3` channel
- Subscription could reuse the existing WS auth token (same connection as `executions`)
- Parse snapshot: `Clear` + `Add` per order; incremental: single `BookAction::{Add,Update,Delete}` delta per event
- Optional: CRC32 checksum validation (Kraken provides this for the top 10 levels)

## References

- Kraken WS v2 `level3`: https://docs.kraken.com/api/docs/websocket-v2/level3/
- Existing L3 reference implementation: `crates/adapters/architect_ax/src/websocket/data/parse.rs`
- Framework L3 types: `crates/model/src/data/delta.rs`, `crates/model/src/data/order.rs`

mcgrj:
Closes #4001.

Wires the Kraken Spot adapter into the WS v2 `level3` channel, so strategies can subscribe to `BookType.L3_MBO` and get true per-order data, real venue order IDs, individual sizes, and `Add` / `Update` / `Delete` events.

## What's in here

- New L3 WebSocket client and parser sitting alongside the existing v2 client.
- A `KrakenSpotL3DataSource` (pyo3) that owns the L3 connection, per-symbol book state, and pushes `OrderBookDeltas` straight into the Python data engine.
- Python `KrakenDataClient` now routes `L3_MBO` subscriptions to that source instead of warning and dropping them. Spot-only.
- One new config knob: `validate_l3_checksum` (default on), mirrored on both Rust and Python configs.

## Design decisions worth flagging

- **Separate L3 socket.** Kraken's `level3` channel runs on its own auth'd endpoint, so it gets its own client and lifecycle rather than being grafted onto the existing public socket.
- **Stable `u64` for `BookOrder.order_id`.** Kraken uses alphanumeric venue IDs (`O6ZQNQ-BXL4E-5WGINO`); Nautilus wants `u64`. Mapped through a fixed-seed `ahash` so the same venue ID always lands on the same number — important for replay and backtest determinism.
- **Checksum-driven resync.** On CRC32 mismatch we clear local state and resubscribe, surfacing a fresh snapshot rather than letting a divergent book drift silently. Toggleable for debug runs.
- **Snapshot semantics.** Snapshot emits `Clear` + one `Add` per resting order; incrementals map 1:1 to `BookAction::{Add,Update,Delete}`. Matches what the framework expects for `L3_MBO`.

## Reconciliation

Side-by-side with Kraken Desktop on `BTC/USD`, depth 1000 — bucketed L3 snapshot from the new source matches the venue book level-for-level (totals, best bid/ask, spread, midpoint):

Kraken L3 reconciliation vs Kraken Desktop


cjdsellers:
Hi @mcgrj,

Thanks for the careful work on this. The CRC32 + RawValue handling is well thought through. A few things I'd like addressed before merge:

**Connect/subscribe race in `KrakenSpotDataClient::subscribe_l3_book`**

The handler/connect work and the per-symbol subscribe are spawned as two separate `get_runtime().spawn` futures contending on the same `tokio::Mutex`. If a second `subscribe_l3_book` call lands while the first task is still queued, the second can win the lock and run `subscribe()` before `connect()` has fetched the token. That surfaces as "Auth token not available; call connect() before subscribe()" on the second symbol of a multi-symbol subscribe. Easiest fix: acquire the mutex once and do `connect` + `subscribe` inside the same locked section, or gate on an `is_connected` flag set under the lock.

**Duplication between the Rust client and the pyo3 wrapper**

`python/l3_data_source.rs::handle_l3_msg` and `data/spot.rs::handle_l3_ws_message` are near-identical: same snapshot-checksum, update-checksum, clear-emit, and resync-request branches, with `L3State` and `L3ResyncRequest` defined in both files. Please extract the shared L3 state machine into `websocket/spot_v2/level_3/` (e.g. a `runtime.rs`) and have both the Rust DataClient and the pyo3 source call into it. As written there are two copies of the protocol logic to keep in sync on every future change.

**Modify-path same-price detection**

In `parse_l3_update`, the same-price vs price-change branch compares cached and new prices after rounding to `price_precision`. That's effectively `f64 -> Price -> ==` rather than a check on the underlying values. Won't bite in normal venue flow, but a sub-tick discrepancy would silently take the Update branch and leave the cached `seq` in place. Comparing the raw `value`s (or the `raw` strings) is more defensive and costs nothing.

**Auth token refresh in `resync_symbol`**

`refresh_auth_token` only runs in `connect()` and `resubscribe_all()`. If a long-lived connection survives the token TTL and a checksum mismatch then triggers `resync_symbol`, the unsub/sub uses a stale token. Worth re-fetching the token at the top of `resync_symbol` (or any time we resubscribe outside the reconnect path).

The duplicated docs section and a few minor style items I'll flag in line comments and we can easily clean those up post merge.

Let me know your thoughts once you've had a look.

mcgrj:
Thanks for another thorough review, @cjdsellers, again agreed on all points:

- **Race**: `subscribe_l3_book` now acquires the mutex once and gates on a new `KrakenSpotL3WebSocketClient::is_connected()` check inside the locked section. Multi-symbol subscribes can no longer reorder connect vs subscribe
- **Duplication**: extracted the L3 state machine into `websocket/spot_v2/level_3/runtime.rs` behind an `L3Sink` trait. `data/spot.rs` uses `DataEventSink`, `python/l3_data_source.rs` uses `PyCallbackSink` — single source of truth for snapshot/update/checksum/clear/resync
- **Modify same-price**: branch now compares `cached.price_raw == event.limit_price.raw` (wire decimal) instead of the rounded `Price`. New rstest locks in that `42001.0` vs `42001.00` triggers Delete+Add and resets FIFO `seq`
- **Token refresh**: `resync_symbol` calls `refresh_auth_token` at the top before unsub/sub, doc cites the 15-min `GetWebSocketsToken` window

Manual smoke tests, etc. still all green in addition to packaged tests.

```
BTC/USD.KRAKEN L3_MBO depth=1000 batches=12806 updates=12805 ids=7377 bucket=50 view=desktop
best_bid=80986.5 best_ask=80986.6 spread=0.09999999999126885 midpoint=80986.54999999999 book_updates=30151

bid_total      bid_amount       bid_price | ask_price       ask_amount      ask_total
 17.86242398     17.86242398           80950 |     81000       2.44214783     2.44214783
 46.12247039     28.26004641           80900 |     81050      25.58647043    28.02861826
  73.5764808     27.45401041           80850 |     81100      26.58670937    54.61532763
103.91024742     30.33376662           80800 |     81150       20.2994481    74.91477573
121.54994394     17.63969652           80750 |     81200      19.75542697     94.6702027
166.26300486     44.71306092           80700 |     81250        6.4065079    101.0767106
174.84683893      8.58383407           80650 |     81300      13.79356282   114.87027342
 175.0562213      0.20938237           80600 |     81350       8.48112725   123.35140067
209.49123039     34.43500909           80550 |     81400      10.53435009   133.88575076
226.77973355     17.28850316           80500 |     81450      14.41075068   148.29650144
227.16080319      0.38106964           80450 |     81500       7.93831777   156.23481921
229.49393414      2.33313095           80400 |     81550      26.36694761   182.60176682
233.86282157      4.36888743           80350 |     81600       2.23403397   184.83580079
242.03274029      8.16991872           80300 |     81650       0.13786831    184.9736691
242.28344886      0.25070857           80250 |     81700       7.13244639   192.10611549
242.58251729      0.29906843           80200 |     81750        0.0657199   192.17183539
```

Ready for your re-review when you get a sec.

cjdsellers:
Hi @mcgrj,

Apologies, I should have flagged this in the first review and didn't. Looking again, the structural concern is bigger than the line-level findings I raised previously, and I don't want you to put more work in before we align.

The new `KrakenSpotL3WebSocketClient` reimplements connection lifecycle, auth token caching, reconnect/heartbeat config, and subscription tracking that all already live in `KrakenSpotWebSocketClient`. The promotion of `refresh_auth_token` to `pub(crate)` is the giveaway. The result is two parallel WS-client implementations in the same adapter, two reconnect loops to keep in sync, and three subscription trackers (`subscribed_symbols`, `KrakenSpotDataClient.l3_subscriptions`, Python `_l3_subscriptions`).

That's a maintenance burden we don't need to create, and several round-one findings (`SubscriptionState` gap, silent subscribe-ack handling, tracker drift) all trace back to it.

What's genuinely L3-specific is narrow:

- New wire types (`KrakenL3Snapshot`, `KrakenL3UpdateData`) and a `Level3` channel variant plus subscription params shape, added to `SpotFeedHandler`'s parsing
- A second `KrakenSpotWebSocketClient` instance in `KrakenSpotDataClient` with the URL pointing at `wss://ws-l3.kraken.com/v2`
- The per-symbol stateful book reconstruction with CRC32, which `runtime.rs` already isolates correctly and stays as the L3 state layer driven from the existing handler

That removes `client.rs` (~600 lines) and the `pub(crate)` leak, lets `KrakenSpotL3DataSource` go (the third pyo3 class only exists because the new client doesn't compose with anything else), and inherits the `SubscriptionState` ref-counting and pending/confirmed lifecycle for free. The CRC32 + parser + book-reconstruction work is the genuinely novel contribution here and it stays.

Let me know how you'd like to approach this.

mcgrj:
Agreed, @cjdsellers, the current approach was pragmatic localised duplication in some areas to de-risk the new path, which made sense perhaps when developing, but not for long term maintainability. I should have revisited before pushing for review, thanks for flagging.

Its not a large change to align as you describe, happy to pick that up :building_construction:

mcgrj:
Ready for your re-review when youve got time, @cjdsellers

cjdsellers (approved):
Hi @mcgrj 

Thanks for all the follow-ups 🙏, this is looking solid now. I'll make some very minor refinements and update the integration guide post-merge.

mcgrj:
Thanks @cjdsellers :pray:

PR #4047, opened 2026-05-10
Suppress Hyperliquid cancel/modify rejects on transport failure

M-Advis:
## Summary

In the Hyperliquid adapter, `_cancel_order`, `_cancel_all_orders`, and `_batch_cancel_orders` all caught every exception with a bare `except Exception` and unconditionally emitted `OrderCancelRejected`. This is incorrect for transport-level failures (`TimeoutError`, `OSError`) where the HTTP request may have already reached the exchange before the connection dropped — the cancel outcome is undefined, not confirmed-failed. Emitting `OrderCancelRejected` in that case misleads the strategy into believing the order is still open, causing state desync when WS subsequently delivers `OrderCanceled`. The fix adds an `isinstance` guard matching the existing pattern in `_submit_order`: transport errors log a warning and return, leaving WS reconciliation as the source of truth.

**Failure sequence**

1. Strategy sends `CancelOrder` command
2. `_cancel_order` builds the HTTP request and fires it via `await self._client.cancel_order(...)`
3. Request leaves the machine and hits Hyperliquid ✓
4. Connection drops (timeout/network blip) before response arrives
5. `TimeoutError` or `OSError` is raised on our side
6. `except Exception` catches it → `OrderCancelRejected` emitted → strategy believes cancel failed, order still open
7. Exchange did process the cancel → WS delivers `OrderCanceled`
8. NT receives `OrderCanceled` for an order it believes is still open → state desync, strategy may retry cancel or act on a position that no longer exists

## Type of change

- [x] Bug fix (non-breaking)

## Testing

The fix is a small, self-contained guard clause — the same pattern already present in `_submit_order`. No tests added; the correctness is apparent from the diff and the edge case (live transport failure during cancel) is not meaningfully unit-testable without a live node. The goal is to prevent state miscommunication on timeout, not to change any happy-path behavior.

> Note: it's possible this edge case is already handled at a higher level in the reconciliation or execution engine logic that I'm not fully across — happy to defer or close if so.

### DIFF
--- a/nautilus_trader/adapters/hyperliquid/execution.py
+++ b/nautilus_trader/adapters/hyperliquid/execution.py
@@ -855,6 +855,12 @@ class HyperliquidExecutionClient(LiveExecutionClient):
             )
 
         except Exception as e:
+            if isinstance(e, (TimeoutError, OSError)):
+                self._log.warning(
+                    f"Modify request timed out for {command.client_order_id} "
+                    f"— cancel-replace may have landed, awaiting WS reconciliation",
+                )
+                return
             self._pending_modify_keys.pop(command.client_order_id.value, None)
             self._pending_modify_target_qty.pop(command.client_order_id.value, None)
             self.generate_order_modify_rejected(
--- a/nautilus_trader/adapters/hyperliquid/execution.py
+++ b/nautilus_trader/adapters/hyperliquid/execution.py
@@ -890,7 +890,14 @@ class HyperliquidExecutionClient(LiveExecutionClient):
                 venue_order_id=pyo3_venue_order_id,
             )
             self._log.info(f"Order cancellation requested for {command.client_order_id}")
+
         except Exception as e:
+            if isinstance(e, (TimeoutError, OSError)):
+                self._log.warning(
+                    f"Cancel request timed out for {command.client_order_id} "
+                    f"— order may be canceled on exchange, awaiting WS reconciliation",
+                )
+                return
             self.generate_order_cancel_rejected(
                 strategy_id=command.strategy_id,
                 instrument_id=command.instrument_id,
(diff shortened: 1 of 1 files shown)


M-Advis:
Also included a fix for the same transport ambiguity in `_modify_order`.

Fix is the same pattern — transport errors return early without touching the markers, leaving WS reconciliation as the source of truth.

cjdsellers:
Hi @M-Advis,

Thanks for catching this and the clear writeup. Emitting `OrderCancelRejected` on a transport failure when the request may have already landed at the venue is a real state-desync, and worth fixing.

I've landed a change directly on `develop` covering both adapters, with a few changes https://github.com/nautechsystems/nautilus_trader/commit/5c639564ca3aadefbc738d5c236cc188fea992ae:

- The PyO3 client converts Rust HTTP errors through `to_pyvalue_err`, so they arrive in Python as `ValueError` carrying the Rust `Display` text, not `TimeoutError`/`OSError`. I added a `_is_transport_error()` predicate that matches the native exceptions plus the Rust prefixes (`transport error:`, `IO error:`, exact `timeout`), so the guard actually fires on the production pyo3 path.
- Kept `_pending_modify_keys` and `_pending_modify_target_qty` populated on modify transport error. `_handle_order_updated` needs them to resolve `target_qty` if the modify landed (`report.quantity` is remaining-only).
- Mirrored the change in Rust across `cancel_order`, `cancel_all_orders`, `batch_cancel_orders`, and `modify_order` via a new `Error::is_transport_error()` matching `Transport | Timeout | Io`. Also fixed `modify_order` clearing `pending_modify` unconditionally on `Err(e)`.
- While narrowing the predicate, re-categorised five sites in `signing/signers.rs` that were mis-using `Error::transport` for local signing or hex-decode failures. They now use `Error::auth` / `Error::bad_request`, which also corrects `is_retryable()` for those cases.
- Extended the same treatment to `_submit_order` and `_submit_order_list` in Python (Rust submit already deferred).
- Added coverage for the predicate and each branch (Rust unit + Python integration).

Credited you in `RELEASES.md`. Closing this out, thanks again, and please keep the reports coming.

M-Advis:
Thx for the detailed explanation covering all affected areas, helps to understand the project better 🔎.  I also felt this one leads to wider fix and coverage.

PR #4052, opened 2026-05-11
Fix Kraken Spot WS dispatch dropping delta-only exec frames

### LINKED ISSUE #4051: Kraken Spot WS dispatch: handle delta-only executions frames lacking `symbol`
Per [Kraken's executions docs](https://docs.kraken.com/api/docs/websocket-v2/executions/), `pending_new` carries full order data, then `new` carries only changed fields — `symbol` is not repeated:

```json
// new (delta only)
{"order_id": "OK4GJX-KSTLS-7DZZO5", "exec_type": "new",
 "order_status": "new", "timestamp": "..."}
```

`crates/adapters/kraken/src/websocket/dispatch/spot.rs:60-69` currently returns early when `exec.symbol.is_none()`, so the `Accepted` transition for resting orders isn't propagated. Affects `stop-loss`, `take-profit`, `trailing-stop` and their limit variants, and likely resting `LIMIT` orders.

Downstream effect on the Python adapter: the order stays local-`Submitted` while open at the venue, the inflight-check falls back to `request_order_status_reports`, and after a few cycles the order resolves locally as `OrderRejected(reason='UNKNOWN')` even though it's still live. Bracket SL/TP legs go missing without the strategy seeing it.

### Suggested fix

Mirror the existing `order_qty_cache` in the same file: cache `order_id → symbol` from frames that carry it, consult on `exec.symbol.is_none()`, evict on terminal `exec_type`. Key by `order_id` since delta frames don't always include `cl_ord_id`.

mcgrj:
## Summary

Orders submitted via the Kraken Spot WS that are not instantly executed drop the `symbol` field from the follow-up `new` frame (per Kraken's [executions docs](https://docs.kraken.com/api/docs/websocket-v2/executions/)). The Spot WS dispatch returned early on `exec.symbol.is_none()`, so the order never received `Accepted`. Local state stayed `Submitted`, then the live exec engine resolved it as `OrderRejected(reason=UNKNOWN)` while still open at the venue.

Closes #4051.

## Fix

Cache `order_id → symbol` from frames that carry it, consult on `exec.symbol.is_none()` before bailing, evict on terminal `exec_type`. Mirrors the existing `order_qty_cache` pattern.

### DIFF
--- a/crates/adapters/kraken/src/websocket/dispatch/mod.rs
+++ b/crates/adapters/kraken/src/websocket/dispatch/mod.rs
@@ -139,6 +139,11 @@ pub struct WsDispatchState {
     /// here from the first frame and look it up when later frames omit it.
     /// Keyed by venue `order_id` because delta frames often lack `cl_ord_id`
     /// as well.
+    ///
+    /// Steady-state eviction happens on terminal exec types
+    /// (`Canceled`/`Filled`/`Expired`). Bounded by `DEDUP_CAPACITY` as a
+    /// safety net so missed terminal frames (reconnects, partial replays)
+    /// cannot leak entries indefinitely.
     pub order_symbol_cache: DashMap<String, String>,
     /// Last snapshot of qty / filled / price / trigger_price seen on a
     /// tracked `OpenOrdersDelta`.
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @mcgrj,

Thanks for the careful diagnosis and the fix.

The shape is right. Two notes before merge:

`is_terminal_exec_type` already exists at `crates/adapters/kraken/src/websocket/dispatch/spot.rs:384` with the same `Filled | Canceled | Expired` set. The new eviction block in `execution()` should call it rather than re-inline the `matches!`, so the terminal definition stays in one place.

The cache only repopulates from `pending_new`, so a reconnect that lands between `pending_new` and the first `new` for an order would still drop the `new` frame (same failure mode as before, narrower window). I believe the executions `snap_orders` replay carries `symbol` and reseeds the cache, but the docstring on `order_symbol_cache` should state that assumption explicitly.

Tests look good. If you can parametrize the symbolless path across `Amended` / `Restated` / `Status` as well as `New`, that pins the docstring claim that all four delta types resolve via cache.

Let me know once you've pushed, happy to take another pass.

mcgrj:
Hi @cjdsellers, thanks for your review - 

Pushed two follow-ups addressing your findings, completely agreed with all:

- Eviction now calls `is_terminal_exec_type` instead of re-inlining `matches`
- `order_symbol_cache` docstring states the write rule and the cold-start limitation while the live spot client subscribes with `snap_orders=false` (state recovered via `request_order_status_reports`)
- Symbolless-delta test parametrized over `New` / `Amended` / `Restated` / `Status`

While addressing the above I noticed the symbol cache alone didn't fully close the issue, Kraken's `new` delta omits `cl_ord_id` too, so tracked identity still fell through to the untracked report path. Added an `order_client_id_cache` (mirrors the futures `venue_client_map` and wider pattern) and and restructured `execution()` so terminal cache eviction runs after the dispatch body regardless of which early-return path it hits.

Ready for your re-review when you get a sec, thanks :pray:

cjdsellers:
Hi @mcgrj 

Thanks again for the follow-ups, this one good to go 👍

Issue #3972, opened 2026-05-01
[Hyperliquid] Modify order race condition


AlphaTraderK:
# Feature Request

## Problem Statement

The current Hyperliquid adapter in `nautilus_trader` appears to handle Hyperliquid cancel-replace / modify semantics in the **order status report path**, but not consistently in the **fill report path**.

On Hyperliquid, a modify operation behaves like cancel-replace:
- The same Hyperliquid `cloid` / Nautilus `client_order_id` is retained
- The previous exchange `oid` / Nautilus `venue_order_id` is replaced
- The replacement order may have a new quantity and/or price

The adapter already appears to handle this in the **order status report path**, where replacement `ACCEPTED` reports can advance the local order state via:

`OrderUpdated(..., venue_order_id_modified=True)`

However, the **fill report path appears able to emit/process fills before the corresponding replacement update has advanced the local order state**.

This creates a race condition when:
- A fill for the replacement order arrives before the corresponding `OrderUpdated`
- The local order state still references the previous `venue_order_id`, quantity, or price

This can lead to false rejections inside the execution engine:

- **Venue order ID mismatch**
  - Fill references the replacement `venue_order_id`
  - Local order still has the previous `venue_order_id`

- **Overfill rejection**
  - Fill belongs to the replacement order with updated quantity
  - Local order still has the previous quantity

### Proposed Solution

For Hyperliquid modify/cancel-replace handling, the adapter should maintain pending replacement state and ensure fills are emitted only after the corresponding local order state has been advanced.

Suggested behavior:

1. When a modify request is submitted, record a pending replacement state keyed by `client_order_id` / Hyperliquid `cloid`, including the intended quantity, price, side, and any known old/new venue order identifiers.

2. When a replacement status report is received, apply the same logic currently used in the order status report path and emit:

   `OrderUpdated(..., venue_order_id_modified=True)`

   This should update the local order state to the replacement state.

3. If a fill report arrives before the replacement `OrderUpdated` has been processed, and the fill belongs to a pending replacement for the same `client_order_id` / `cloid`, the adapter should buffer the fill instead of immediately emitting `OrderFilled`.

4. After the replacement `OrderUpdated` has advanced the local order state, replay the buffered fill reports in the correct order.

5. Add a reconciliation fallback for missing/out-of-order replacement status reports, such as querying latest order status or using a bounded timeout before resolving buffered fills.

This would avoid processing fills from a replacement order against stale local order state.

[...]

PR #4076, opened 2026-05-14
Fix Hyperliquid pending fills buffer

M-Advis:
# Pull Request

## Summary
`_handle_fill_report_pyo3` silently drops fills when the order is not yet in cache at the time the WS fill message arrives. The existing comment acknowledged this (`# Don't mark as processed - order may arrive later`) but the buffering was never implemented. This PR adds `_pending_fills` — mirroring the existing `_buffered_fills` pattern (GH-3972) — to buffer the raw pyo3 fill report and drain it once the order is confirmed accepted.

## Related Issues/PRs
Related to GH-3972

## Type of change
- [x] Bug fix (non-breaking)

N/A

## Documentation

## Release notes

## Testing

No unit tests added — the logic mirrors the proven `_buffered_fills` pattern already in the codebase. Available to help build tests if needed.

### What was already there
The existing code already handled this partially — the comment `# Don't mark as processed - order may arrive later` lead to conclusion the order might arrive later and intentionally avoided marking the fill as processed to prevent dedup lock-out. 

However the buffering to actually hold the fill until that moment was never implemented, leaving the fill permanently dropped.

### What changed

**Buffer** — when `order is None` in `_handle_fill_report_pyo3`, instead of dropping:
```python
# before: fill permanently lost
self._log.error(f"Cannot process fill report - order for {client_order_id!r} not found")
return

# after: buffered until order arrives
self._pending_fills.setdefault(client_order_id.value, []).append(pyo3_report)
return
```

**Drain** — added at all three points where NT first learns an order is accepted, covering all possible WS message orderings:

1. `_handle_order_accepted_pyo3` — WS delivers a dedicated `OrderAccepted` message
2. ACCEPTED branch of `_handle_order_status_report_pyo3` — WS delivers `OrderStatusReport(status=ACCEPTED)`
3. Inside `_handle_fill_report_pyo3` when auto-generating `OrderAccepted` inline — fill arrives but order was never accepted in NT state machine yet

Missing any one of these three drain points leaves a gap depending on which message type the exchange delivers first.

**Cleanup** — `_pending_fills` cleared in `_cleanup_cloid_mapping` alongside `_buffered_fills` to prevent stranded entries on order close.

Note: this adds a small amount of complexity to `_handle_fill_report_pyo3`, requiring a `# noqa: C901` suppression.

### DIFF
--- a/nautilus_trader/adapters/hyperliquid/execution.py
+++ b/nautilus_trader/adapters/hyperliquid/execution.py
@@ -165,6 +165,10 @@ class HyperliquidExecutionClient(LiveExecutionClient):
         # from the cancel-replace ACCEPTED branch. See GH-3972.
         self._buffered_fills: dict[str, list[nautilus_pyo3.FillReport]] = {}
 
+        # FillReports buffered when fill arrives before order is in cache,
+        # drained on OrderAccepted.
+        self._pending_fills: dict[str, list[nautilus_pyo3.FillReport]] = {}
+
         self._fee_refresh_task: asyncio.Task | None = None
 
         # Get user address from HTTP client for WebSocket subscriptions.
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Thanks @M-Advis 

I'll follow up with some refinements.

PR #4077, opened 2026-05-14
Implement backtest deterministic liquidation engine

### LINKED ISSUE #3788: Liquidation engine
# Feature Request
Liquidation engine
## Problem Statement
In futures trading, positions can always be liquidated, so I think the backtesting system should also include this feature to simulate real market conditions more accurately. Thank you

abhishektang:
## Summary

Closes #3788

Implements a deterministic, configurable margin-liquidation engine for the `SimulatedExchange` backtest venue. When a position's unrealised loss causes margin utilisation to exceed a configurable threshold, the exchange automatically closes all open positions for that instrument — mirroring what real-world perpetual/derivatives venues do on margin breach.

### Configuration (`BacktestVenueConfig`)

Three new optional fields (all off by default so existing backtests are unaffected):

| Field | Type | Default | Description |
|---|---|---|---|
| `liquidation_enabled` | `bool` | `False` | Enable automatic margin liquidation |
| `liquidation_trigger_ratio` | `Decimal` | `0.90` | Fraction of margin consumed that triggers liquidation |
| `liquidation_cancel_open_orders` | `bool` | `True` | Cancel resting orders before closing positions |

### Core implementation

- **`nautilus_trader/backtest/engine.pyx`** — `SimulatedExchange._process_margin_liquidations()` checks margin utilisation on every `process()` tick and calls `liquidate_open_positions()` when the threshold is breached.
- **`crates/execution/src/matching_engine/engine.rs`** — `OrderMatchingEngine::liquidate_open_positions()` cancels open orders (optional) then synthetically closes every open position at best bid/ask, emitting `OrderAccepted` + `OrderFilled` events.
- **`crates/backtest/src/config.rs / engine.rs / exchange.rs / node.rs`** — Rust-side config propagation and plumbing.
- **`crates/backtest/src/python/config.rs / engine.rs`** — PyO3 bindings exposing the new config fields.

### Tests

- **`tests/unit_tests/backtest/test_liquidation_engine.py`** — 13 tests covering: config defaults, trigger logic, order cancellation, position closure, multiple positions, already-liquidated guard, and end-to-end market simulation.
- **`tests/unit_tests/backtest/test_config.py`** — 3 additional config tests for the new `BacktestVenueConfig` fields.

### Example

- **`examples/liquidation_demo.py`** — Standalone runnable demo: BUY 10 M XBTUSD @ $40 k, crash price to $20 k, observe liquidation firing automatically.

## Testing

All 16 new tests pass locally:

```
tests/unit_tests/backtest/test_liquidation_engine.py  13 passed
tests/unit_tests/backtest/test_config.py               3 passed (liquidation subset)
```

No existing tests were modified or broken.


abhishektang:
## Work Done

Hi maintainers! Here's a detailed breakdown of everything implemented in this PR, along with an honest self-assessment of gaps and areas I'm happy to iterate on.

---

### What was implemented

#### 1. `BacktestVenueConfig` — three new opt-in fields

```python
BacktestVenueConfig(
    venue="BITMEX",
    ...
    liquidation_enabled=True,          # default False — zero impact on existing backtests
    liquidation_trigger_ratio=Decimal("0.90"),  # liquidate when 90 % of margin is consumed
    liquidation_cancel_open_orders=True,        # cancel resting orders first
)
```

All three default to their "off" values so every existing backtest runs identically.

#### 2. `SimulatedExchange._process_margin_liquidations()` (Cython, `engine.pyx`)

Called inside `process()` on every data tick. Iterates open positions, checks unrealised PnL against available margin, and calls into the Rust matching engine when the trigger ratio is breached.

#### 3. `OrderMatchingEngine::liquidate_open_positions()` (Rust, `matching_engine/engine.rs`)

Optionally cancels all resting orders for the instrument, then synthetically closes every open position by emitting `OrderAccepted` + `OrderFilled` events at best bid/ask (falling back to settlement price), so downstream portfolio/risk accounting is updated correctly.

#### 4. Rust config plumbing

`config.rs`, `engine.rs`, `exchange.rs`, `node.rs` — propagates the three new fields from `BacktestVenueConfig` all the way down to the `OrderMatchingEngine`.

#### 5. PyO3 bindings

`python/config.rs`, `python/engine.rs` — exposes the new fields so they are visible from Python.

#### 6. Tests (16 total, all green locally)

- `test_liquidation_engine.py` — 13 tests: default config, trigger fired/not-fired at threshold boundaries, open-order cancellation, multiple concurrent positions, guard against double-liquidation, and a full tick-by-tick market simulation.
- `test_config.py` — 3 tests for the new config fields.

#### 7. Example

`examples/liquidation_demo.py` — self-contained script: buys 10 M XBTUSD at $40 k, feeds price down to $20 k tick-by-tick, prints a structured JSON result showing `liquidation_triggered: true` and zero open positions afterward.

---

### Honest self-assessment — known gaps

| Area | Current state | What a production liquidation engine would do |
|---|---|---|
| **Trigger metric** | Ratio of unrealised loss to available margin | Mark-price-based maintenance margin rate per instrument |
| **Liquidation price** | Fires at configurable ratio; fills at best bid/ask | Exchanges publish a specific liquidation price formula |
| **Partial liquidation** | Full position closed in one step | Many venues reduce size incrementally before full close |
| **Insurance fund / ADL** | Not modelled | Real venues absorb losses via insurance fund; ADL as last resort |
| **Liquidation fee** | Not charged | Venues typically charge a fee separate from taker commission |
| **Mark price source** | Uses last trade/quote price | Should use index/mark price if available in `DataEngine` |
| **Documentation** | No `docs/` pages added | A `how_to/liquidation.md` guide would help users |
| **Rust compile in CI** | Passes local debug build; CI runs full release + clippy | Waiting on CI results — happy to fix any issues surfaced |

---

### Am I meeting the issue expectations?

The original issue #3788 asks for a liquidation engine to make futures backtesting "simulate real market conditions more accurately." The core mechanic — automatically closing positions when margin is exhausted — is implemented and working. However, the gaps above (especially mark price, partial liquidation, and liquidation price formula) mean this is a solid v1 rather than a full production replica.

**I am actively available to**:

- Replace the simple ratio trigger with a proper maintenance-margin-rate formula if the maintainers can point to the preferred margin model in the codebase.
- Add partial liquidation (reduce size in steps).
- Add a liquidation fee as a separate `Money` deduction.
- Write `docs/` pages.
- Fix any Rust/Cython compile errors the CI surfaces.
- Address any API/design feedback before merge.

Happy to iterate — please leave review comments with whatever direction you'd like and I'll get on it promptly. Thank you for your time!

cjdsellers:
Hi @abhishektang

Many thanks for the contribution, this is a valuable feature.
  
The PR looks mostly good on my initial pass. I'll circle back with detailed comments after the imminent release.

cjdsellers:
Hi @abhishektang

```
make pre-commit
```
or (if you're on Windows):
```
pre-commit run --all-files
```

```
pre-commit install
```

abhishektang:
Hi @cjdsellers 

Thanks for the review feedback! I've pushed a follow-up commit (`491b73c`) that resolves all the CI pre-commit failures:

- Replaced `.cloned()` on borrowed `AccountRef<'_>` and `OrderRef<'_>` with `cache.account_for_venue_owned()` and `cache.order_owned()` respectively (fixes `clippy::E0599`)
- Added `# Panics` doc section to `liquidate_open_positions` (fixes `clippy::missing_panics_doc`)
- Inlined format arg in `log::warn!` (fixes `clippy::uninlined_format_args`)
- Added blank lines above control-flow statements (fixes `check-formatting-rs`)
- Expanded `else` blocks and reformatted long method chains (fixes `cargo fmt`)
- Fixed Python formatting: trailing commas, ruff, docformatter

All 55 `prek` hooks now pass locally. Request you to review my PR and happy to work on any changes required!

cjdsellers:
Hi @abhishektang,

Thanks for the patience here! and for the clear self-assessment in your follow-up comment. It made review easier. The feature is a solid v1 and the v1-vs-production gaps you listed (mark price, partial liquidation, fee, insurance fund) are fine to track as follow-ups rather than block merge.

A meta point before the line items: the project is migrating execution off Cython onto the Rust core, so new venue features should land Rust-only. Please drop the Cython path (`SimulatedExchange._process_margin_liquidations` and `OrderMatchingEngine.liquidate_open_positions` in `engine.pyx`, plus the new fields in `engine.pxd`), and keep only the Rust implementation. The Python `BacktestVenueConfig` fields stay so users can still configure it; they just flow through to the Rust `SimulatedExchange` via the existing PyO3 plumbing you've already wired up. Tests then need to drive the Rust path, either as Rust integration tests under `crates/backtest/tests/` or via `BacktestEngine` / `BacktestNode` from Python (which runs through the Rust core).

That collapses the rest of the review considerably. With the Cython path gone, two merge blockers remain:

**1. The `liquidated_accounts` dedupe**

In `SimulatedExchange::process_liquidations`, the `(account_id, currency)` set is filled on first breach and only cleared on `reset()`. So if a strategy re-opens positions in that currency after a liquidation and breaches again, the engine silently skips. The `maintenance > 0` check just above already prevents same-tick re-firing, so the set is unnecessary. Drop it and add a regression test that opens, liquidates, re-opens, and liquidates again.

**2. Synthetic close order has a panic path and skips `OrderSubmitted`**

In `OrderMatchingEngine::liquidate_open_positions`, the synthesised `MarketOrder` is built with no `account_id`. `generate_order_accepted` then does `self.account_ids.get(&order.trader_id()).unwrap()`. `account_ids` is populated only inside `process_order`, so any flow where a position exists in cache but no order was previously routed through this matching engine for that trader panics. Either pass the account id explicitly when building the order (look it up via `cache.account_for_venue_owned(&self.venue)`), or have the generators fall back to the cache when `account_ids` is empty.

Same place: the state transition goes `Initialized -> Accepted` directly, and the fill uses `apply_fills` with a manually picked `best_bid_price()` / `best_ask_price()`, bypassing book walk and slippage. The existing internal-synthesised path in `_process_option_expiry` is the right pattern to match for both the submitted/accepted/filled sequence and the fill mechanism. Also reuse `OrderCore::closing_side(pos.side)` (already used by `Strategy::close_position`) instead of the manual `match`, and drop the `OrderSide::NoOrderSide` fallback since `positions_open` never returns flat positions.

**Convention cleanup (also blocking)**

- The new test file and example use unicode arrows (`→`), em-dashes (`—`), box-drawing characters (`──`, `═`), and checkmarks (`✓` / `✗`). Repo policy disallows all of these. The `# ── Section ──` and `# ------` banner comments also need to go.
- Tests use `class TestLiquidationEngine` style. Repo convention is pytest-style free functions with fixtures, no test classes.
- The `liquidation_trigger_ratio` type is inconsistent across surfaces: `f64` in `SimulatedVenueConfig`, `Decimal` in `BacktestVenueConfig`, `float` in Python, with `.to_f64().unwrap_or(1.0)` glue silently swallowing out-of-range values. Pick one (`f64` is acceptable here since it's a continuous ratio) and align.
- `BacktestEngine.add_venue` docstring stops at `settlement_prices`; the three new `liquidation_*` params aren't documented.
- `examples/liquidation_demo.py` lives at top-level; backtest examples belong under `examples/backtest/`. It also wires `SimulatedExchange` by hand instead of going through `BacktestEngine.add_venue(..., liquidation_enabled=True)` like every other example.

A few smaller nits to fold in while you're touching the files:

- `crates/backtest/src/exchange.rs:26`: combine into `use ahash::{AHashMap, AHashSet};`.
- `engine.rs:2495`: synthesised market order is built with `TimeInForce::Gtc`; `Ioc` reads more correctly for an immediate fill.
- `process_liquidations` reads `positions_open(...)` inside the per-currency loop; hoist it out and bucket by `settlement_currency` once.
- Document the cross-margin assumption: a breach on currency C currently liquidates positions on every matching engine, not just those settled in C. That's right for cross-margin venues, less right for isolated; at minimum it needs a docstring note.

Happy to look at the next push as soon as it's up.

abhishektang:
Thanks @cjdsellers for the thorough and constructive review — the detailed breakdown made it very clear what needed to change and why. All merge blockers and convention issues have been addressed in this follow-up push.

---

### Summary of changes

**Architecture (blocking)**
- Removed `SimulatedExchange._process_margin_liquidations` and `OrderMatchingEngine.liquidate_open_positions` from `engine.pyx` / `engine.pxd`. Liquidation is now Rust-only, driven by `SimulatedExchange::process_liquidations` and `OrderMatchingEngine::liquidate_open_positions`.

**Merge blockers resolved**

1. **`liquidated_accounts` dedupe set** — Dropped entirely from `SimulatedExchange`. The `maintenance > 0` guard already prevents same-tick re-firing; removing the set lets the engine correctly re-trigger after a strategy re-opens a position.

2. **Synthetic close order — state sequencing and account panic**
   - Order state now follows the full `Initialized → Submitted → Accepted → Filled` sequence. Added `generate_order_submitted` in `OrderMatchingEngine` (mirrors `generate_order_accepted`); it is called with the `account_id` taken directly from the position tuple before `generate_order_accepted`.
   - `account_ids` is pre-populated from position data before either generate call, preventing the `unwrap` panic for positions not previously routed through `process_order`.
   - Closing side now uses `OrderCore::closing_side(pos.side)`.
   - Removed `OrderSide::NoOrderSide` fallback.
   - Synthesised order uses `TimeInForce::Ioc`.

**Convention fixes (blocking)**
- Removed all Unicode arrows, box-drawing characters, checkmarks, and banner comments (`# -- Section --`, `# ------`) from test file and example.
- Converted `class TestLiquidationEngine` to pytest-style free functions with a shared `_make_engine` / `_add_buy_strategy` fixture pattern.
- `liquidation_trigger_ratio` is now `f64` / `float` end-to-end — no `Decimal` in `BacktestVenueConfig`, no `.to_f64().unwrap_or` silent swallowing.
- `BacktestEngine.add_venue` docstring documents all three new `liquidation_*` parameters.
- Example moved to `examples/backtest/liquidation_demo.py` and now uses `BacktestEngine.add_venue(..., liquidation_enabled=True)` like other backtest examples.

**Smaller nits**
- `use ahash::{AHashMap, AHashSet};` combined in `exchange.rs`.
- `process_liquidations` now pre-buckets open positions by `settlement_currency` into an `AHashMap<Currency, Vec<usize>>` before the per-currency loop, eliminating the repeated full-vector scan.
- Cross-margin assumption documented on `process_liquidations`.

---

### Test scenarios covered

The 13 tests in `tests/unit_tests/backtest/test_liquidation_engine.py` drive the Rust path end-to-end via `BacktestEngine` (PyO3):

| Scenario | What it checks |
|---|---|
| `liquidation_disabled_does_not_close_position` | Feature opt-in: position survives a 50 % crash when `liquidation_enabled=False` |
| `liquidation_triggered_closes_position` | Core trigger: 1 BTC balance + 10 M-contract long + 50 % price crash forces a close |
| `no_liquidation_when_equity_above_threshold` | Healthy equity: 100 BTC balance + small position survives |
| `liquidation_cancel_open_orders_flag_true/false` | Both values of `liquidation_cancel_open_orders` close the position correctly |
| `gradual_decline_no_premature_liquidation` | No false positive: 10-step decline with ample margin leaves position open |
| `gradual_decline_liquidation_fires_at_threshold` | Threshold crossing mid-decline: liquidation fires somewhere in the 20-step path |
| `price_recovery_prevents_liquidation` | Dip-and-recover: position survives if equity recovers before breach |
| `higher_trigger_ratio_liquidates_at_least_as_early` | Monotonicity: `trigger_ratio=2.0` is at least as aggressive as `1.0` at the same price |
| `liquidation_can_retrigger_after_reopen` | **Regression for the removed dedupe set**: open → liquidate → reset → open → liquidate again — no permanent suppression |

The retrigger test directly validates the dedupe-set removal requested in blocker #1.

cjdsellers:
Hi @abhishektang,

Thanks for the follow-up push. The Rust-only path and config cleanup look much better now. I still see two issues that should be fixed before merge:

1. `SimulatedExchange::process_liquidations` triggers on one settlement currency, but then calls `liquidate_open_positions` on every matching engine for the venue. That means a breach in one currency can force-close healthy positions settled in another currency. Please either filter liquidation to instruments settled in the breached currency, or make the trigger explicitly venue-wide.

2. `OrderMatchingEngine::liquidate_open_positions` still bypasses the normal market-order fill path. It picks best bid/ask and calls `apply_fills` directly, so fill model behavior, book depth, liquidity consumption, and price protection do not apply. I think this should go through the normal market fill machinery, or reuse the same price/volume helpers, so liquidation closes behave like other taker market orders.

One test note: `test_liquidation_can_retrigger_after_reopen` resets the engine before the second run, so it would also pass with the old dedupe-set behavior because reset cleared the guard. A same-run open, liquidate, reopen, liquidate case would cover the intended regression more directly.

Once those are addressed, I think this is ready!

faysou:
Regarding pre-commit I recommend you to have a testing procedure in your agents.md where one of the checks is to run and fix make pre-commit.

cjdsellers:
Hey @abhishektang,

```
make pre-commit
```
or (if you're on Windows):
```
pre-commit run --all-files
```

```
pre-commit install
```

If we can get CI passing when/if you get the chance - then would be willing to carry on with the follow ups on my end post-merge.

abhishektang:
Hi @cjdsellers and @faysou,

Thanks for the continued feedback — very helpful in getting this over the line and sorry for the delay.

**Re May 21 review (@cjdsellers)** — both remaining blockers are now addressed in the latest push (`0c25a209a6`):

1. **Cross-currency scope** — `process_liquidations` now pre-buckets open positions by `settlement_currency` before the equity check. `liquidate_open_positions` is only called on matching engines whose instrument settles in the breached currency; positions in other currencies are left untouched.

2. **Fill mechanics** — The synthetic close no longer calls `apply_fills` directly. It now constructs a `MarketOrder` with `TimeInForce::Ioc` and `reduce_only=true`, then routes it through `fill_market_order`, so the full fill model, book-depth consumption, and slippage assumptions apply identically to a normal taker close.

3. **Retrigger test** — Replaced `test_liquidation_can_retrigger_after_reopen` (which used `engine.reset()` and would have passed with the old dedupe-set) with `test_liquidation_retriggers_within_same_run`. The new test runs a single `engine.run()`, the strategy re-opens the position immediately after the first liquidation close (`rebuy_after_close=True`), and asserts `>= 2` FILLED liquidation orders — no `reset()` involved.

**Re May 23 (@faysou)** — good call on the `agents.md` pre-commit check; noted and added that to the workflow.

**Re May 26 (@cjdsellers)** — all 55 pre-commit hooks pass locally on this push. CI should be green.

Happy to address any further review comments!

faysou:
Pre-commit is failing. Also for PRs it's much better to rebase on develop than merge from develop. Fix pre-commit is one of my favourite order to agents although I do it much less as it's written in my agents.md.

cjdsellers (approved):
Thank you for all the follow-ups and for sticking with it @abhishektang 🙏

abhishektang:
Looking forward to contribute more! 🙂

Issue #4066, opened 2026-05-12
[RFC] Polymarket resolution-aware expiration gating and settlement (no on-chain redeem integration)


graceyangfan:
# RFC: [Polymarket] Resolution-aware expiration gating and settlement (no on-chain redeem integration)

## Context

This RFC proposes a venue-specific lifecycle improvement for Polymarket `BinaryOption` instruments in NautilusTrader.

Important scope boundary:

- Polymarket already provides automatic redeem behavior at the venue level.
- This RFC does **not** propose integrating CTF/redeem transaction execution inside Nautilus adapters.
- The goal is to make Nautilus order/position lifecycle deterministic and safe around expiry/resolution.

In short: stop trading at expiry, wait for official resolution, then settle via `InstrumentClose`.

## Problem

Today, the framework has all building blocks (`InstrumentStatus`, `InstrumentClose`, matching-engine close flow), but Polymarket lifecycle handling is incomplete/inconsistent across components.

### Current gaps in code paths

1. Python Polymarket market WS schema does not include `market_resolved`.

- `nautilus_trader/adapters/polymarket/websocket/types.py`
- `nautilus_trader/adapters/polymarket/schemas/book.py`

2. Python Polymarket market subscribe payload does not set `custom_feature_enabled`.

- `nautilus_trader/adapters/polymarket/websocket/client.py`

3. Rust Polymarket adapter already parses `market_resolved`, but currently emits `InstrumentStatus(Close)` only (no `InstrumentClose` settlement event).

- `crates/adapters/polymarket/src/websocket/messages.rs`
- `crates/adapters/polymarket/src/data.rs`

4. `BinaryOption` is not treated as an engine-auto-expiring instrument class (`has_expiration=false` semantics in model tests), so expiry alone does not produce settlement.

- `nautilus_trader/model/instruments/base.pyx`
- `crates/model/src/instruments/mod.rs`

5. For Polymarket specifically, forcing settlement purely by time is unsafe because official resolution can lag expiry.

## Design goals

1. Keep venue semantics correct: resolution-driven settlement (0/1), not mark-to-last settlement.
2. Keep strategy code clean: adapter/execution should handle lifecycle internally.
3. Keep implementation minimal: no new complex user-facing config surface.
4. Keep non-Polymarket venues unaffected.

## Non-goals

1. No on-chain CTF redeem execution in Nautilus adapters.
2. No default last-price forced settlement.
3. No global framework-wide expiration redesign.

[...]

PR #4088, opened 2026-05-18
Add pending-resolution settlement and resolve polling for Polymarket

graceyangfan:
## Summary

This draft implements a Polymarket resolve pipeline for binary options that does **not** depend on `market_resolved` WebSocket delivery.

The main changes are:

- keep expired binary-option instruments in a `pending_resolution` state instead of forcing immediate expiry settlement
- drive resolve polling from open exposure via a dedicated watchlist
- use Gamma `GET /markets` results as the primary resolve path
- emit synthetic `market_resolved` messages from resolved GET responses so the normal `InstrumentClose` pipeline can settle positions
- expose sandbox simulated execution factories/configs through PyO3 so the paper-trading path can be exercised from Python

Refs #4066.

## Problem

For Polymarket binary options, waiting for a reliable `market_resolved` WebSocket event is not sufficient in practice.

When an instrument expires while orders or positions are still open, we need the following behavior:

- open orders should be canceled
- new orders should be rejected
- positions should remain open while waiting for a resolution result
- once a winner is known, the normal `InstrumentClose -> settlement -> PositionClosed` path should complete the lifecycle

The current upstream issue assumes a resolver-driven close path, but for Polymarket the practical constraint is that WebSocket resolve delivery cannot be the only source of truth.

### 1. Pending-resolution settlement semantics in the matching engine

Binary options now enter `pending_resolution` on expiry when no `InstrumentClose` has been received yet.

This state:

- cancels resting/open orders at expiry
- blocks new orders while waiting for resolution
- allows the later `InstrumentClose` event to settle the remaining position normally

### 2. Exposure-driven resolve watchlist

Resolve polling is now driven by a dedicated watchlist keyed by `InstrumentId` but carrying the minimum metadata needed to keep polling even if the instrument cache changes:

- `condition_id`
- `expiration_ns`
- `auto_poll_active`

The watchlist is updated from position events:

- `PositionOpened`, `PositionChanged`, `PositionAdjusted` add/update entries
- `PositionClosed` removes entries

This keeps the tracked set aligned with real paper/live exposure rather than transient subscription state.

### 3. Gamma GET polling as the primary resolve path

A new optional polling path was added to `PolymarketDataClientConfig`:

- `resolve_poll_enabled`
- `resolve_poll_interval_secs`
- `resolve_poll_expiry_grace_secs`
- `resolve_poll_max_wait_secs`
- `resolve_poll_emit_compensation`

When enabled, the data client:

- waits until `expiration + grace`
- polls only watchlisted instruments that are eligible for resolve
- pauses automatic polling after `max_wait` instead of deleting the watch entry
- still allows manual resolve requests to target paused entries

### 4. GET resolve compensation into the normal close pipeline

Resolved Gamma responses are converted into synthetic `market_resolved` messages and then flow through the existing mapping logic to emit:

- `InstrumentStatus(Close)`
- `InstrumentClose(ContractExpired)`

This keeps paper/live settlement on the standard `InstrumentClose` path instead of building a parallel settlement mechanism.

### 5. Safer Gamma resolve querying

Resolve GET calls now fetch markets **per `condition_id`** instead of batching multiple ids into a single comma-separated query.

In manual testing, closed-market resolve responses were not reliably returned from batched `condition_ids=a,b` requests, while single-condition requests returned the expected resolved market data.

### 6. PyO3 paper-trading support for simulated execution

This draft also exposes sandbox simulated execution factory/config extraction through the PyO3 registry and live node builder so the Rust paper-trading path can be exercised from Python without falling back to legacy Python/Cython execution wiring.

## Logging

The final version keeps the useful resolve diagnostics but removes or downgrades the high-frequency tracing that was only needed during bring-up.

In particular:

- noisy per-order forwarding logs were removed
- `InstrumentClose` forwarding logs were downgraded to debug where appropriate
- repeated empty/"waiting" resolve poll messages were throttled
- GET compensation logs were kept as normal informational events rather than warnings

### Rust tests

Added/updated coverage includes:

- WebSocket and synthetic resolve payload mapping to `InstrumentClose`
- watchlist membership driven by position events
- shared-market and multi-market watchlist selection (`YES/NO`, `BTC/ETH`)
- `pending_resolution -> InstrumentClose -> settlement` in the matching engine
- paper-lane settlement through `DataEngine -> Sandbox -> matching engine`
- multi-instrument paper settlement for multiple binary options in one run

### Pre-commit / compile checks

Passed:

- `check-anyhow-usage`
- `check-logging-conventions`
- `check-tokio-usage`
- `check-dst-conventions`
- `check-pyo3-conventions`
- `check-testing-conventions`
- `check-nautilus-conventions`
- `check-formatting-rs`
- `check-docs-conventions`
- `check-cargo-conventions`
- `cargo fmt`
- `cargo clippy`
- `cargo doc`

### Targeted cargo tests

Passed:

- `cargo test -p nautilus-polymarket market_resolved_ -- --nocapture`
- `cargo test -p nautilus-polymarket position_events_drive_resolve_watchlist -- --nocapture`
- `cargo test -p nautilus-sandbox test_paper_binary_option_multiple_instruments_close_settlement_via_data_engine -- --nocapture`
- `cargo test -p nautilus-execution test_binary_option_pending_resolution_then_instrument_close_settles_position -- --nocapture`

### Real paper-trading acceptance

Outside the repo test suite, this branch was also validated with a PyO3 paper-trading probe against live Polymarket market data.

That acceptance run verified:

- entry orders filled in paper trading
- expiry transitions to `pending_resolution`
- new post-expiry orders are rejected
- resolve results can be obtained via GET-only compensation
- `InstrumentClose` reaches the paper execution lane
- multiple positions (`BTC/ETH`, `YES/NO`) are eventually settled and closed

## Notes

- WebSocket `market_resolved` handling remains supported, but this draft does **not** rely on WS delivery for correctness.
- Automatic resolve polling is configurable and remains bounded by `max_wait`.
- Paused watchlist entries are retained so manual resolve requests can still be used for long-waiting markets.

## Follow-ups

Before marking this ready for review, I would like to gather reviewer feedback on two design points:

- whether the new polling config should remain opt-in or move toward a stronger default for Polymarket adapters
- whether the watchlist ownership should stay in the data client or later be elevated into a more general exposure-resolution abstraction


cjdsellers:
Hi @graceyangfan 

Are you open to separating this into two PRs? One for the Polymarket-specific changes, then a follow-up for the other changes necessary for this use case. It would greatly help the review process and change management :pray:

graceyangfan:
Thanks, that makes sense. I’m going to split this up.\n\nFrom the pipeline dependency side, I think it will be cleaner to land the common pending-resolution settlement path first, then follow with the Polymarket-specific resolve source on top.\n\nI’ll close this draft and reopen as two smaller PRs so the review scope is much tighter.

PR #4089, opened 2026-05-18
Guard IBKR order submit on client readiness

honvl:
## Summary

- guard Interactive Brokers order submission on the shared IB client readiness event
- reject local submit commands before calling `place_order()` when the IB client is degraded or reconnecting
- add regression coverage for the not-ready submit path and make the IB execution test connection mock mark the shared client ready

## Root Cause

`InteractiveBrokersClient.wait_until_ready()` protects startup, but a long-running client can later degrade and clear `_is_client_ready` while the `InteractiveBrokersExecutionClient` still accepts submit commands. In that state, `_submit_order()` can call `place_order()` against a TCP-connected but not fully initialized IB client. If status queries later time out, local reconciliation can resolve the order as `UNKNOWN` even though IB may still have received and filled it.

## Validation

- `python -m py_compile nautilus_trader/adapters/interactive_brokers/execution.py tests/integration_tests/adapters/interactive_brokers/test_execution.py tests/integration_tests/adapters/interactive_brokers/conftest.py`
- Targeted pytest was attempted from a shallow source clone, but the clone was not built and failed importing compiled `nautilus_trader.core.data` before reaching the tests.

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -191,6 +191,13 @@ class InteractiveBrokersClient(
         # Start client
         self._request_id_seq: int = 10000
 
+    @property
+    def is_client_ready(self) -> bool:
+        """
+        Return whether the Interactive Brokers client is ready for requests.
+        """
+        return self._is_client_ready.is_set()
+
     def _start(self) -> None:
         """
         Start the client.
--- a/nautilus_trader/adapters/interactive_brokers/execution.py
+++ b/nautilus_trader/adapters/interactive_brokers/execution.py
@@ -981,6 +981,7 @@ class InteractiveBrokersExecutionClient(LiveExecutionClient):
         PyCondition.type(command, SubmitOrder, "command")
 
         try:
+            self._ensure_client_ready_for_order_submission()
             ib_order: IBOrder = self._transform_order_to_ib_order(
                 command.order,
                 command.params,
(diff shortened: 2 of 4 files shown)


cjdsellers:
Hi @honvl,

Thanks for this change. The readiness guard on submit paths is the right direction, and the regression test makes the failure mode clear.

A nice follow-up would be to apply the same readiness check to cancel and modify paths, then tidy up the direct `_is_client_ready` access to use the public equivalent. That would keep the execution client from reaching into private client state.

Let me know if you run into anything while doing that.

honvl:
Thanks, this makes sense. I pushed a follow-up commit (`5ff14d74`) addressing it:

- Added a public `InteractiveBrokersClient.is_client_ready` property.
- Updated the execution client to use that public readiness surface instead of reaching into `_is_client_ready` directly.
- Applied the same fail-closed readiness guard to modify and cancel paths, including cancel-all/batch cancel behavior.
- Added regressions for submit, modify, single cancel, and batch cancel when the IB client is not ready.

Local checks run:

```text
python -m pytest tests/integration_tests/adapters/interactive_brokers/test_execution.py --tb=short
# 32 passed

python -m ruff check nautilus_trader/adapters/interactive_brokers/client/client.py nautilus_trader/adapters/interactive_brokers/execution.py tests/integration_tests/adapters/interactive_brokers/test_execution.py
# All checks passed

python -m ruff format --check nautilus_trader/adapters/interactive_brokers/client/client.py nautilus_trader/adapters/interactive_brokers/execution.py tests/integration_tests/adapters/interactive_brokers/test_execution.py
# 3 files already formatted
```

faysou:
Why did you close the PR ?

honvl:
My API key got exposed. I will open a replacement PR.

PR #4090, opened 2026-05-18
Add generic structured key-value fields to LogLine 

filipmacek:
## Summary

- Add generic `fields: LogFields` to `LogLine` that captures all non-reserved kv pairs from log records via the `log` crate's existing kv API
- `LogFields = SmallVec<[(Ustr, String); 4]>` — up to 4 fields stored inline (no heap alloc), keys use `Ustr` interning so repeated keys like `"venue"` across log calls share one allocation
- Add `FieldCollector` implementing `log::kv::VisitSource` for single-pass extraction of all key-value pairs
- Fields render as `key=value` in plain/colored text output and as top-level keys in JSON output
- Reserved JSON keys (`timestamp`, `trader_id`, `level`, `color`, `component`, `message`) are protected from override by user-supplied fields
- Zero macro changes needed — `log::info!(key = "val"; "msg")` syntax works out of the box

## Performance

Net improvement. The single-pass `VisitSource` collector replaces multiple `key_values.get()` calls (each of which iterated all kv pairs). `SmallVec::new()` has zero heap allocation, and up to 4 fields are stored inline on the stack. `Ustr` key interning eliminates per-call string allocation for repeated field names.

## Usage

```rust
log::info!(
    component = "DataEngine",
    venue = "BINANCE",
    order_id = "O-12345";
    "Order filled"
);
```

Plain output:

```
2026-05-19T10:00:00Z [INFO] TRADER-001.DataEngine: Order filled venue=BINANCE order_id=O-12345
```

JSON output:

```json
{"timestamp":"...","trader_id":"TRADER-001","level":"INFO","component":"DataEngine","message":"Order filled","venue":"BINANCE","order_id":"O-12345"}
```

## Test plan

- [x] `cargo check --all-features` passes
- [x] All 131 logging tests pass
- [x] New integration tests: `test_logging_to_file_with_kv_fields`, `test_logging_to_file_json_with_kv_fields`
- [x] New unit tests: fields in plain text, JSON, Display, and reserved key protection
- [x] `make pre-commit` all green

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -253,7 +253,7 @@ serde_json = { version = "1.0.149", default-features = false, features = [
   "std",
 ] }
 serde_urlencoded = "0.7.1"
-smallvec = "1.15.1"
+smallvec = { version = "1.15.1", features = ["serde"] }
 sockudo-ws = { version = "1.7.4", default-features = false, features = [
   "simd",
   "fastrand",
--- a/crates/common/src/logging/logger.rs
+++ b/crates/common/src/logging/logger.rs
@@ -32,6 +32,7 @@ use nautilus_core::{
 };
 use nautilus_model::identifiers::TraderId;
 use serde::{Deserialize, Serialize, Serializer};
+use smallvec::SmallVec;
 use ustr::Ustr;
 
 pub use super::config::LoggerConfig;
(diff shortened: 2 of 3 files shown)


cjdsellers (approved):
👌 thanks @filipmacek

PR #4100, opened 2026-05-20
Guard IBKR order submit on client readiness

honvl:
Replaces #4089 after recreating the fork path to remove cached sensitive-data exposure.\n\nThis PR contains only the IBKR client readiness guard changes from the original PR.


cjdsellers:
Hey @honvl

Thanks again for persevering with it.

Two small things worth addressing before merge:
- Property name `client.is_client_ready()` reads a bit redundant (`is_ready()` would be cleaner)
- Add a test for the `_cancel_all_orders` case

cjdsellers (approved):
Thanks for the follow-ups @honvl 👌

PR #4101, opened 2026-05-20
Add pending-resolution settlement pipeline for binary options

graceyangfan:
## Summary

This PR adds the common pending-resolution settlement pipeline for binary options.

When a binary option expires without an immediate resolved close price, the matching engine now moves it into a `pending_resolution` state instead of forcing immediate settlement. In that state, open resting orders are canceled, new orders are rejected, and any remaining position exposure stays pending until a later `InstrumentClose` arrives.

This PR also wires the sandbox simulated execution client into the PyO3 live-node builder path so the paper-trading lane can exercise the same delayed-close settlement flow.

This is the first of two PRs for [#4066](https://github.com/nautechsystems/nautilus_trader/issues/4066).
A follow-up PR will add the Polymarket-specific resolve source, watchlist ownership, and GET-polling integration.

## What Changed

- add `pending_resolution` handling to the matching engine for binary options
- cancel open resting orders when an expired binary option enters pending resolution
- reject new orders while pending resolution is active
- settle remaining exposure only when a later `InstrumentClose` arrives
- route delayed `InstrumentStatus` and `InstrumentClose` messages through the sandbox paper-trading lane
- expose `SandboxExecutionClientFactory` through the PyO3 registry and `LiveNodeBuilder`
- refine settlement handling so the common pipeline does not depend on cache-based engine recreation or fee-model fallbacks

## Design Intent

This PR is intentionally venue-agnostic.

For delayed-settlement venues, instrument expiration and market resolution may not happen at the same instant. The generic engine behavior needed here is:

- expiry should stop new trading activity
- resting orders should be canceled
- open position exposure should remain pending
- final settlement should occur only when a later close signal arrives

That lifecycle belongs in the common execution and paper-trading pipeline.

The Polymarket-specific questions of:

- which expired instruments still need later resolve tracking
- how long they should remain tracked
- when to GET resolve results
- how to map those results into `InstrumentClose`

are intentionally left for the follow-up PR.

## Testing

Verified with:

- `cargo test -p nautilus-execution test_binary_option_pending_resolution_then_instrument_close_settles_position -- --nocapture`
- `cargo test -p nautilus-sandbox --tests -- --nocapture`
- `cargo test -p nautilus-sandbox --features python --lib -- --nocapture`
- `cargo test -p nautilus-system --features python --lib -- --nocapture`
- `cargo test -p nautilus-live --features python --lib -- --nocapture`

Added coverage for:

- binary option expiry entering `pending_resolution`
- resting-order cancellation at the expiry boundary
- rejection of new orders while pending resolution is active
- settlement via later `InstrumentClose`
- paper-trading settlement through sandbox directly
- paper-trading settlement through the `DataEngine` route
- multiple binary-option instruments settling independently

## Notes

This PR does not include any Polymarket-specific resolve polling, watchlist management, or Gamma API fallback logic. Those changes will follow in a separate PR.


graceyangfan:
Thanks again for the feedback on splitting the original change set.

A bit more context on the intended follow-up (`PR2`) and why it stays separate from this PR:

`PR1` is meant to land only the common pending-resolution lifecycle for binary options:
- expiry transitions the instrument into a pending-resolution state
- resting orders are canceled
- new orders are rejected
- open exposure remains pending until a later settlement signal arrives
- paper/sandbox can consume that later signal and close the position cleanly

The remaining Polymarket work is venue-specific and belongs in a separate PR.

For Polymarket, the missing piece is not the settlement pipeline itself, but the settlement *source*:
- `market_resolved` WebSocket events are documented, but in practice they have been unreliable
- the venue does not give us a Hyperliquid-style account-level `Settlement` fill that can directly close the position in the execution pipeline
- after expiry / unsubscribe / rotation, instrument cache lifetime is no longer a reliable source of truth for unresolved exposure

So the planned `PR2` is:
- maintain a small waiting-resolution watchlist keyed by unresolved expired instruments with open exposure
- after a grace period, query the venue's resolve state via HTTP/GET
- translate a resolved result into `InstrumentClose`
- feed that `InstrumentClose` into the common lifecycle introduced here

That separation is intentional:
- this PR defines *what the engine should do once settlement is pending*
- the follow-up PR defines *how Polymarket obtains the settlement result when the venue does not reliably push it*

This is also why I have tried to keep venue-specific fallback logic out of this PR.

For comparison, Hyperliquid HIP-4 outcomes are modeled differently: the venue can emit a `Settlement` fill, so its live settlement path is naturally fill-driven rather than `InstrumentClose`-driven. In other words, both venues share the same “expiry -> pending exposure” problem, but they differ in how the final settlement signal enters Nautilus.

So the intended shape is:
- common pending-resolution state machine here (`PR1`)
- Polymarket-specific resolve watchlist / GET polling / `InstrumentClose` emission in `PR2`
- Hyperliquid continues to use its venue-driven settlement fill path

I wanted to leave this note here so the split is easier to review in terms of pipeline ownership and long-term maintenance.

cjdsellers:
Hi @graceyangfan,

Thanks for the clean split and the detailed `PR1`/`PR2` context note. The pipeline-vs-source boundary reads well, and the three-layer test coverage made the intent easy to follow.

A few of things left to address are the redundant cancel pass in `cancel_open_orders_for_expiration`, the `max(clock, ts_init)` semantic shift in the expiration check, the silent-drop asymmetry between `process_instrument_close` and `process_instrument_status` (which does lazy-create), and de-duplication of the two near-identical sandbox tests.

Happy to look at the next push as soon as it's up.

graceyangfan:
Thanks for the focused follow-up. I pushed a revision that addresses each of the four points directly:

1. `cancel_open_orders_for_expiration`
- I removed the redundant two-pass cancel flow.
- The final version now builds a single de-duplicated `ClientOrderId` set before canceling.
- I did not reduce it to `cache`-only, because the pending-resolution path still needs to cover both the matching-core view and the order-ledger/cache view without issuing duplicate cancels.
- This was re-validated with the pending-resolution execution/sandbox regressions.

2. `max(clock, ts_init)` semantic shift
- I reverted the expiration gate back to engine-clock-only semantics.
- Added a regression test to lock that in: `test_binary_option_expiration_check_uses_engine_clock_not_order_ts_init`.

3. `process_instrument_close` vs `process_instrument_status`
- I kept the intentional asymmetry, but made it explicit and test-backed.
- `InstrumentStatus` still lazy-creates from cache because it belongs to active instrument lifecycle routing.
- `InstrumentClose` no longer recreates from cache and only routes to an existing engine, because delayed settlement belongs to an existing exposure lifecycle rather than cache lifetime.
- `InstrumentStatus` cache-miss is now warned instead of silently dropped.
- Added a regression test to cover the intended behavior: `test_instrument_status_lazy_creates_but_close_requires_existing_engine`.

4. Sandbox test duplication
- I kept both entry paths (`direct close publish` and `via DataEngine`) but extracted the common pending-resolution harness/setup so the coverage remains the same without carrying two near-identical test bodies.

Re-ran after the changes:
- `cargo test -p nautilus-sandbox --test execution -- --nocapture` -> `34 passed`
- `cargo test -p nautilus-execution --test matching_engine -- --nocapture` -> `154 passed, 3 ignored`
- `pre-commit run --all-files` -> passed

Please take another look when convenient.

cjdsellers (approved):

PR #4102, opened 2026-05-20
Fix LiveNode signal handling during startup connection wait

filipmacek:
# Pull Request

### Problem

When engine clients are slow to connect or fail entirely, the LiveNode enters `await_engines_connected()` which polls for the duration configured by `timeout_connection` (default 120s). During this window:

1. **Stop signals are ignored** — the polling loop only checks `check_engines_connected()`, never `should_stop()`
2. **Stop signals are erased** — after the timeout, `set_state(NodeState::Running)` clears the stop flag (by design)
3. **The node becomes uncontrollable** — any stop/restart command sent during startup is silently lost, and the operator must wait the full `timeout_connection` duration before the node begins to respond

This was documented as a known limitation via a TODO comment in `run()`.

### Fix

Four targeted changes in `nautilus_trader/crates/live/src/node.rs`:

- **`await_engines_connected()`** — check `should_stop()` each iteration of the 100ms polling loop; return `false` immediately if set
- **`run()`** — after `connect_exec_phase()` completes, check `should_stop()` and return early with `NodeState::Stopped` *before* `set_state(Running)` clears the flag
- **`start()`** — same early-exit guard after `connect_exec_clients()` for the non-event-loop startup path
- Remove the stale TODO comment that described this limitation

### Result

Stop/restart commands sent during the engine connection wait are now detected within 100ms. The node exits startup cleanly and the restart loop can rebuild immediately.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/crates/live/src/node.rs
+++ b/crates/live/src/node.rs
@@ -373,6 +373,12 @@ impl LiveNode {
 
         self.kernel.connect_exec_clients().await;
 
+        if self.handle.should_stop() {
+            log::info!("Stop signal received during startup, aborting start");
+            self.handle.set_state(NodeState::Stopped);
+            return Ok(());
+        }
+
         if !self.await_engines_connected().await {
             log::error!("Cannot start trader: engine client(s) not connected");
             self.handle.set_state(NodeState::Running);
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Thanks @filipmacek, this improves the main startup responsiveness issue.

I’ll follow up after merge on a few lifecycle gaps: preserving stop signals through the `start()` timeout path, running client/kernel cleanup before reporting `Stopped`, and covering the remaining `ctrl_c`/`ShutdownSystem` startup cases.

PR #4107, opened 2026-05-21
Fix unbounded Cache VecDeque memory leak

filipmacek:
# Pull Request

## Summary

- Add `BoundedVecDeque<T>` — a runtime-capacity newtype wrapper around `VecDeque<T>` that auto-evicts the oldest entry on `push_front()` when at capacity
- Replace all 7 unbounded `VecDeque` fields in `Cache` (quotes, trades, bars, mark_prices, index_prices, funding_rates, instrument_statuses) with `BoundedVecDeque`
- Fix `add_bars()` batch method incorrectly using `tick_capacity` instead of `bar_capacity` for pre-allocation

## Problem

`CacheConfig` defines `tick_capacity` (default 10,000) and `bar_capacity` (default 10,000), but these were only passed to `VecDeque::with_capacity()` — a pre-allocation hint, **not a size limit**. Every `add_*` method calls `push_front()` without eviction, so VecDeques grow unboundedly in long-running nodes.

Discovered in production: one trading node subscribing to hundreds of instruments grew from ~98 MB to nearly 2 GB over two days before being OOM-killed. Other trading nodes showed the same linear growth pattern at ~20 MB/hr. Both confirmed unbounded cache accumulation as the root cause.

## Approach

The codebase already has `ArrayDeque`/`FifoCache` in `cache/fifo.rs`, but these require compile-time `const N` — unusable here since capacity comes from runtime `CacheConfig`.

`BoundedVecDeque<T>` uses the **newtype pattern** to encode the capacity invariant in the type system rather than relying on manual `pop_back()` calls across 14 insertion sites. The eviction logic lives in one place (`push_front`), making it impossible to introduce new `add_*` methods that forget to enforce the bound. `Deref<Target = VecDeque<T>>` provides transparent read-only access, while mutable operations go through the bounded API.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/crates/common/src/cache/bounded.rs
+++ b/crates/common/src/cache/bounded.rs
@@ -1,0 +1,247 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2026 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
+//  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+//  You may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Thanks @filipmacek, this is a good find 👍

PR #4109, opened 2026-05-21
Add Binance futures open interest request custom data

### LINKED ISSUE #4104: Add Binance Futures open interest request custom data for PyO3 strategies
## Problem
The Binance Futures adapter already supports current open interest at the HTTP client layer, but it is not exposed through the `DataClient.request_data(...)` / PyO3 strategy path.

On current `develop`, the request pathway already exists in core (`RequestCustomData` -> `CustomDataResponse` -> `on_historical_data(...)`), but the Binance Futures adapter does not implement `request_data(...)` for open interest.

At the moment:

- `GET /fapi/v1/openInterest` and `GET /dapi/v1/openInterest` are already implemented in the Binance Futures HTTP client.
- `GET /futures/data/openInterestHist` is not implemented yet.
- `BinanceFuturesDataClient` does not expose either endpoint as requestable custom data for PyO3 strategies.

This makes it difficult to build OI-aware strategies using the standard Nautilus historical request workflow.

## Scope
This should support both Binance futures markets already handled by the adapter:

- USD-M futures
- COIN-M futures

The adapter already models both as Binance Futures product types, so this feature should follow that existing surface rather than adding a USD-M-only request path.

## Proposal
Add two Binance Futures request-only custom data types:

1. `BinanceFuturesOpenInterest`
2. `BinanceFuturesOpenInterestHist`

with a supporting point type:

- `BinanceFuturesOpenInterestHistPoint`

These should be exposed through the existing `RequestCustomData` pathway. No new core abstraction should be required.

## Request API shape
Use `DataType` metadata as the request input surface.

### Canonical request metadata
Use `instrument_id` as the canonical metadata field.

This matches the existing Binance adapter custom-data direction better than raw `symbol` metadata, and it lets the adapter derive the correct REST params internally for both USD-M and COIN-M.

Examples:

```python
DataType(
    BinanceFuturesOpenInterest,
    metadata={"instrument_id": "BTCUSDT-PERP.BINANCE"},
)

DataType(
    BinanceFuturesOpenInterestHist,
    metadata={
        "instrument_id": "BTCUSDT-PERP.BINANCE",
        "period": "5m",
    },
)
```

### Request field mapping
Map Nautilus request fields where applicable:

- `limit` -> Binance `limit`
- `start` / `end` -> Binance `startTime` / `endTime` for `openInterestHist`

The adapter should derive Binance-specific HTTP params from `instrument_id`:

- USD-M current OI -> `symbol`
- COIN-M current OI -> `symbol`
- USD-M hist OI -> `symbol`
- COIN-M hist OI -> `pair` + `contractType`

That keeps the user-facing API venue-native (`instrument_id`) while still supporting the REST differences between the two futures markets.

### `BinanceFuturesOpenInterest`
First-scope fields:

- `instrument_id`
- `open_interest`
- `ts_event`
- `ts_init`

### `BinanceFuturesOpenInterestHistPoint`
First-scope fields:

- `sum_open_interest`
- `sum_open_interest_value`
- `ts_event`

### `BinanceFuturesOpenInterestHist`
First-scope fields:

- `instrument_id`
- `period`
- `points`
- `ts_event`
- `ts_init`

I suggest not including the USD-M-only `CMCCirculatingSupply` field in the first scope. It is not present in the COIN-M history response, so excluding it keeps the initial type surface consistent across both futures products.

## Time semantics
A live check on 2026-05-21 shows the two endpoints have different timestamp behavior:

- `openInterest` returns a near-current snapshot timestamp close to request time.
- `openInterestHist` returns an ascending bucketed series whose point timestamps are aligned to the requested period.

Because of that, the timestamp model should be explicit:

- `BinanceFuturesOpenInterest.ts_event` = Binance `time`
- `BinanceFuturesOpenInterestHistPoint.ts_event` = Binance point `timestamp`
- `BinanceFuturesOpenInterestHist.ts_event` = last point timestamp
- `ts_init` = local request completion / object initialization time

The adapter should preserve Binance's returned point ordering and should not attempt to infer stronger semantics than the exchange provides for the latest bucket.

## Historical batch shape
`openInterestHist` should be modeled as a single batch custom-data object containing `points: Vec<BinanceFuturesOpenInterestHistPoint>`.

It should not be expanded into many individual point responses.

Reasoning:

- Binance returns an array-based historical result natively.
- One request should map to one logical historical response.
- This keeps the request/response semantics cleaner than emitting many point-level custom responses.

## PyO3 ergonomics
Preferred PyO3 ergonomics:

- `on_historical_data(...)` should expose the historical result as a `CustomData` wrapper whose `.data` payload is `BinanceFuturesOpenInterestHist`.
- `BinanceFuturesOpenInterestHist.points` should be exposed as `list[BinanceFuturesOpenInterestHistPoint]`, not `list[dict]`.

This is an explicit user-experience goal for this feature.

## Implementation notes
Suggested implementation shape:

- Add the new Binance Futures custom data structs in the adapter custom-data module.
- Implement `request_data(...)` in `BinanceFuturesDataClient` for these two types.
- Return responses through the existing custom-data request path using `CustomDataResponse`.
- Add `openInterestHist` support to the Binance Futures HTTP client.
- If needed, extend the HTTP path handling so the public `/futures/data/openInterestHist` endpoint can be requested cleanly for both futures products.

### Preferred implementation strategy for PyO3 exposure
For the `openInterestHist` Python surface, prefer adapter-local handwritten PyO3 exposure over broadening core abstractions or extending the custom-data macro system.

In particular:

- Keep `BinanceFuturesOpenInterestHist` as the batch wrapper type.
- Expose `BinanceFuturesOpenInterestHistPoint` as a typed PyO3 class.
- Expose `BinanceFuturesOpenInterestHist.points` in Python as a typed `list[BinanceFuturesOpenInterestHistPoint]`.

This keeps the feature aligned with the current adapter-level architecture and avoids expanding the scope into new core request abstractions or generic macro enhancements.

## Acceptance criteria
- `BinanceFuturesOpenInterest` can be requested from PyO3 strategies via `request_data(...)`.
- `BinanceFuturesOpenInterestHist` can be requested from PyO3 strategies via `request_data(...)`.
- Both work for USD-M and COIN-M futures.
- Historical OI is returned as one batch custom-data object containing ordered points.
- Timestamps follow the semantics above.
- In PyO3 strategies, `payload.points` is exposed as typed point objects and can be iterated without manual dict parsing.
- Add Rust tests covering:
  - request dispatch
  - metadata validation
  - USD-M parameter mapping
  - COIN-M parameter mapping
  - response construction for current OI and historical OI
- Add a minimal live example showing periodic OI requests from a strategy.

## Example usage
```python
from nautilus_trader.adapters.binance import BINANCE_CLIENT_ID
from nautilus_trader.adapters.binance import BinanceFuturesOpenInterest
from nautilus_trader.adapters.binance import BinanceFuturesOpenInterestHist
from nautilus_trader.model import CustomData
from nautilus_trader.model.data import DataType

self.request_data(
    DataType(
        BinanceFuturesOpenInterest,
        metadata={"instrument_id": "BTCUSDT-PERP.BINANCE"},
    ),
    BINANCE_CLIENT_ID,
)

self.request_data(
    DataType(
        BinanceFuturesOpenInterestHist,
        metadata={
            "instrument_id": "BTCUSDT-PERP.BINANCE",
            "period": "5m",
        },
    ),
    BINANCE_CLIENT_ID,
    limit=96,
)

def on_historical_data(self, data):
    if not isinstance(data, CustomData):
        return

    payload = data.data

    if isinstance(payload, BinanceFuturesOpenInterest):
        ...
    elif isinstance(payload, BinanceFuturesOpenInterestHist):
        for point in payload.points:
            ...
```

## References
Official Binance references:

- USD-M current open interest:
  - https://developers.binance.com/docs/derivatives/usds-margined-futures/market-data/rest-api/Open-Interest
- USD-M historical open interest statistics:
  - https://developers.binance.com/docs/derivatives/usds-margined-futures/market-data/rest-api/Open-Interest-Statistics
- COIN-M historical open interest statistics:
  - https://developers.binance.com/docs/derivatives/coin-margined-futures/market-data/rest-api/Open-Interest-Statistics

graceyangfan:
## Summary

This PR adds Binance Futures open interest request custom data for both USD-M and COIN-M markets.

It wires current open interest and historical open interest requests through the Rust Binance futures data client so PyO3 strategies can call `request_data(...)` and receive typed custom data payloads. Historical responses are returned as a batch `BinanceFuturesOpenInterestHist`, with `payload.points` exposed in Python as `list[BinanceFuturesOpenInterestHistPoint]`.

## What Changed

- add `BinanceFuturesOpenInterest`, `BinanceFuturesOpenInterestHist`, and `BinanceFuturesOpenInterestHistPoint` custom data types
- implement Binance futures `request_data(...)` handling for open interest requests
- support both USD-M and COIN-M request mapping from `metadata.instrument_id`
- add HTTP client/query/model support for current open interest and `openInterestHist`
- preserve the public `/futures/data/openInterestHist` path without incorrectly prefixing `/fapi` or `/dapi`
- add handwritten PyO3 bindings/getters so historical `points` are exposed as typed point objects rather than generic dicts
- export the new Binance adapter types in Python and stub files
- add Rust coverage for USD-M/COIN-M request routing, metadata validation, typed historical payloads, and the public hist URL path

## Testing

Verified locally with:

- `make pre-commit`
- `cargo test -p nautilus-binance --features python open_interest_hist_points_roundtrip_as_typed_python_list -- --nocapture`
- `cargo test -p nautilus-binance --features python --test futures -- --nocapture`

Coverage added for:

- USD-M current open interest request custom data
- USD-M historical open interest batch custom data
- COIN-M current open interest request mapping
- COIN-M historical `pair + contractType` request mapping
- missing metadata validation
- unsupported custom data requests returning `Ok(())`
- Python typed `payload.points` access for historical batches
- public `/futures/data/openInterestHist` URL routing

## Related Issues

Closes #4104


cjdsellers:
Hi @graceyangfan,

Thanks for the PR. The USD-M/COIN-M routing, the public `/futures/data/` path handling, and the typed `points` Python exposure all look good. Test coverage across both products is solid.

One item I think should be addressed before merge:

- The open interest values are stored as `String` (`open_interest`, `sum_open_interest`, `sum_open_interest_value`). Repo convention is `Decimal` for discrete domain values, and the peer type `BinanceFuturesLiquidation` already uses `Price` / `Quantity` for the same reason. The wire model can stay `String` if Binance returns it that way, but the public custom data crossing into strategy code should parse to `Decimal` so users aren't doing `Decimal(payload.sum_open_interest)` at every callsite.

Two smaller items worth folding in:

- COIN-M historical OI here is perpetual-only (the `_PERP` suffix and hardcoded `PERPETUAL` contract type). Binance also supports `CURRENT_QUARTER` and `NEXT_QUARTER` on this endpoint. Either extend the mapping, or document the restriction in the type-level docs so users know upfront.
- Errors inside the spawned task (HTTP failure, COIN-M mapping failure, unsupported product type) only log and drop. The caller of `request_data` then has a pending request it cannot reconcile. A doc-comment note that fetch errors are logged-only would at least set expectations.

A few smaller nits: the redundant `match data_type_name.as_str()` after the early-return guard, `open_interest.time as u64` silently wrapping onnegatives, and the `py_points` getter returning `PyResult<Py<PyAny>>` while using `into_py_any_unwrap` internally. Can elaborate on any of these.

graceyangfan:
Thanks for the review.

Agreed on the `Decimal` point — I'll change the public custom data surface to parse the Binance wire strings into `Decimal`, while keeping the HTTP models as raw strings.

On the COIN-M historical OI point: after re-checking the current Binance adapter more broadly, the restriction is really at the adapter/instrument layer rather than the endpoint itself. Binance's COIN-M OI endpoint supports `CURRENT_QUARTER` / `NEXT_QUARTER`, and Nautilus has `CryptoFuture` in the model, but the current Binance futures instrument parsing and symbology path are still perpetual-only on both USD-M and COIN-M. So for this PR I'm going to document the current COIN-M historical OI behavior as perpetual-only rather than make `request_data(...)` wider than the adapter's supported instrument surface.

I'll also fold in the smaller cleanups you called out: checked timestamp conversion, remove the redundant post-guard `match`, and tighten the `py_points` getter return path.

If delivery-contract support is desired for Binance futures more generally, I think that belongs in a follow-up adapter-level PR that adds proper Binance dated-future instrument parsing/symbology first, then broadens OI routing on top of that.

cjdsellers (approved):
Thank you for the follow-ups @graceyangfan 🙏

PR #4127, opened 2026-05-24
Align Polymarket PyO3 bootstrap config with instrument_config

### LINKED ISSUE #4126: Polymarket pyO3 live data client should honor instrument_config-scoped bootstrap
# Feature Request

## Problem Statement

The Polymarket Python adapter supports `PolymarketDataClientConfig.instrument_config`, which lets users scope startup instrument loading via `load_ids`, `filters`, and `event_slug_builder`.

The Rust/pyO3 live data client path does not currently honor that configuration model. In practice, the pyO3 `LiveNode.builder(...).add_data_client(...)` path expects the Rust `PolymarketDataClientConfig`, and the live Rust data client bootstraps instruments with a broad `load_all(None)` path instead of a provider-scoped bootstrap.

This creates two practical problems:

1. Python Polymarket configs do not migrate naturally to the pyO3 live path.
2. Narrow-universe strategies (for example event-scoped or rolling UpDown strategies) cannot scope startup instrument loading the same way as the Python adapter, which can lead to excessive bootstrap traffic and startup failures.

## Proposed Solution

Align the Rust/pyO3 Polymarket live data client with the Python adapter's `instrument_config` layering.

Suggested scope:

- Add nested `instrument_config` support to the Rust `PolymarketDataClientConfig`.
- Add a Rust `PolymarketInstrumentProviderConfig` with the bootstrap-scoping subset needed for live migration.
- Make the Rust data client/provider bootstrap instruments from `instrument_config` instead of unconditional broad bootstrap.
- Allow the pyO3 config extraction path to accept the higher-level Python `PolymarketDataClientConfig` object so existing Python-side config shapes can migrate more naturally.

A compatibility-oriented first step can focus on startup bootstrap scope without trying to solve every dynamic provider behavior in one PR.

## Example Usage

```python
from nautilus_trader.adapters.polymarket.config import PolymarketDataClientConfig
from nautilus_trader.adapters.polymarket.providers import PolymarketInstrumentProviderConfig

config = PolymarketDataClientConfig(
    instrument_config=PolymarketInstrumentProviderConfig(
        event_slug_builder="myproject.slug_builders:build_event_slugs",
    ),
    update_instruments_interval_mins=1,
    subscribe_new_markets=False,
)
```

The goal is for the pyO3 live path to understand this configuration model instead of falling back to broad bootstrap.

## Alternatives Considered

- Add a new Rust-only top-level config field for startup event slugs.
  This is simpler short term, but it diverges from the Python adapter model and makes migration less natural.

- Keep broad bootstrap and rely only on ad-hoc auto-loading.
  This does not solve startup behavior for narrow-universe strategies and does not match the Python adapter's provider ownership model.

## Additional Context

This is especially useful for Polymarket strategies that operate on a narrow rolling subset of markets and already depend on Python-side `instrument_config` scoping.

graceyangfan:
## Summary
- add nested `instrument_config` support to the Rust/pyO3 Polymarket data client config
- move startup bootstrap onto provider-owned scoped initialization instead of unconditional full `load_all(None)`
- make `request_instruments()` serve the already loaded scoped universe from cache instead of broad-fetching Gamma
- add periodic additive scoped refresh so pyO3/live paths can load newly rotated instruments without dropping existing ones
- preserve Python adapter-style `update_instruments_interval_mins=None` semantics through the pyO3 config extractor

## Problem
The Python Polymarket adapter already models instrument loading through `PolymarketDataClientConfig.instrument_config`, including builder-driven scoped universes and periodic refresh. The Rust/pyO3 live path still bootstrapped by unconditionally calling `load_all(None)`, and strategy-facing `request_instruments()` could fall back to a broad Gamma fetch.

That broke natural migration for narrow-universe strategies, loaded far more markets than needed, and prevented pyO3 strategies from relying on the same scoped bootstrap and rotation model as the Python adapter.

## Solution
This PR aligns the Rust/pyO3 path with the Python adapter's config layering by introducing `PolymarketInstrumentProviderConfig` on `PolymarketDataClientConfig.instrument_config` and teaching the provider/data client to honor scoped inputs such as `load_ids`, `filters`, `event_slugs`, `market_slugs`, and Python-compatible `event_slug_builder` materialization.

It also keeps ownership boundaries consistent with Nautilus Trader:
- the provider/data client own instrument-universe bootstrap and refresh
- strategies continue to own active market rotation and subscribe/unsubscribe decisions

Concretely, the PR now:
- uses provider-scoped initialization for startup bootstrap
- keeps scoped reloads additive so old instruments remain available for open positions and resolution flows
- refreshes scoped instruments periodically when configured, then republishes them into the live cache/data engine
- keeps `request_instruments()` cache-scoped so strategy requests do not reopen the broad-universe fetch path
- accepts higher-level Python adapter-style config objects in the pyO3 extractor and preserves explicit `None` for refresh disable semantics on that path

Closes #4126.

## Testing
- `cargo fmt --check`
- `cargo clippy -p nautilus-polymarket --lib --tests --features python -- -D warnings`
- `cargo test -p nautilus-polymarket --features python extract_data_config -- --nocapture`
- real pyO3 rotation smoke:
  - startup loaded only the scoped `slug1` yes/no pair
  - after the configured refresh interval, the rotated `slug2` yes/no pair was present in cache without dropping the original scoped instruments


cjdsellers:
Hi @graceyangfan,

Thanks for the PR. The scoped bootstrap direction looks right to me, and the cache-backed `request_instruments()` path is the behavior I would expect for the pyO3 migration.

Two items I think should be addressed before merge:

- `PolymarketDataClientConfig.__new__` maps `update_instruments_interval_mins=None` back to the default `Some(60)`. The new refresh task only disables itself when this field is `None`, so direct pyO3 users cannot opt out with `None` even though the config type and PR summary expose that as the disable path. The custom extractor preserves explicit `None`, but the pyO3 class constructor does not.
- The generated Python stubs were not updated for the new pyO3 surface. `python/nautilus_trader/polymarket/__init__.pyi` and `python/nautilus_trader/adapters/polymarket/__init__.pyi` still omit `PolymarketInstrumentProviderConfig`, and `PolymarketDataClientConfig.__init__` does not expose `instrument_config`, so typed Python users cannot see or pass the new config cleanly.

Happy to look at the next push when you get a chance.

graceyangfan:
@cjdsellers Implemented both follow-ups on top of the latest `develop` merge.

1. Direct pyO3 `PolymarketDataClientConfig(...)` now preserves the Python adapter semantics for `update_instruments_interval_mins`:
- omitted keeps the default refresh interval
- explicit `None` disables periodic refresh
- positional direct-constructor calls also preserve explicit `None`

I kept the pyO3 surface as an explicit `#[pyo3(signature = (...))]` constructor rather than a kwargs-only path, so this stays closer to the existing Nautilus adapter config style.

2. The Polymarket stubs are now aligned with the current Rust/pyO3 surface:
- added `PolymarketInstrumentProviderConfig`
- added `instrument_config` on `PolymarketDataClientConfig`
- updated the current data/exec config signatures in both Polymarket stub modules

Validation run:
- `cargo fmt --check`
- `cargo clippy -p nautilus-polymarket --lib --tests --features python -- -D warnings`
- `cargo test -p nautilus-polymarket --features python extract_data_config -- --nocapture`

Also added direct pyO3 regression coverage for:
- explicit `None`
- omitted default
- positional `None` compatibility

cjdsellers (approved):
Thanks again for all the follow-ups @graceyangfan :pray:

Issue #3505, opened 2026-01-29
Is it possible to provide a method to override the `_order_id_generator` field of the `OrderFactory` class


sh7ning:
Is it possible to provide a method to override the `_order_id_generator` field of the `OrderFactory` class, without using the implementation of the `ClientOrderIdGenerator` class? Many exchanges have their own rules, such as gates need start with "t-", and some exchanges requiring numbers. I don't want to directly set the `client_order_id` parameter to the order, because I want to rely on the framework to check the uniqueness of `client_order_id` through caching.

xin676231-cmyk:
I’ve prepared a fix for this by allowing `Strategy` to provide a custom client order ID generator to `OrderFactory`.

The change adds:
- optional custom generator injection in `OrderFactory`
- interface validation for custom generators
- cache collision retries for custom-generated IDs
- a strategy hook for providing the generator
- unit coverage for both `OrderFactory` and `Strategy`

I’m preparing the PR now.

skyharbor:
I opened PR #4143 with a small `OrderFactory.set_client_order_id_generator(...)` API for this use case.

The approach allows a venue-specific generator object to provide `generate() -> ClientOrderId` while keeping the factory's existing cache-based uniqueness checks for non-UUID IDs. I also added unit coverage with a prefixed generator example (`t-1`, etc.).

cjdsellers:
Thanks all,

Closing this for now per #4143.

PR #4143, opened 2026-05-27
Allow custom client order ID generators

skyharbor:
## Summary
- Adds `OrderFactory.set_client_order_id_generator(...)` so adapters and strategies can supply venue-specific `ClientOrderId` generation logic.
- Keeps the existing cache-based uniqueness loop in `OrderFactory.generate_client_order_id()` for non-UUID generators.
- Adds unit coverage for a custom prefixed generator, covering the use case described in #3505.

## Issue
Addresses #3505.

## Test plan
- `uvx ruff@0.15.13 check tests/unit_tests/common/test_factories.py`
- `uvx cython==3.2.4 -3 nautilus_trader/common/factories.pyx --working /tmp/nautilus-cython`

Full pytest was not run locally because this machine does not have `rustc`, which is required to build Nautilus Trader from source.

### DIFF
--- a/nautilus_trader/common/factories.pxd
+++ b/nautilus_trader/common/factories.pxd
@@ -19,7 +19,6 @@ from cpython.datetime cimport datetime
 
 from nautilus_trader.cache.base cimport CacheFacade
 from nautilus_trader.common.component cimport Clock
-from nautilus_trader.common.generators cimport ClientOrderIdGenerator
 from nautilus_trader.common.generators cimport OrderListIdGenerator
 from nautilus_trader.core.rust.model cimport ContingencyType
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @skyharbor,

Thanks for the PR. I'm going to hold off on merging this one. The factory lives in the Cython layer, whose remaining service life is short, and I'd rather not churn it for a non-urgent change this close to the end.

There's also a contract gap worth noting: the framework calls `set_count()` and `reset()` on the generator and reads `.count` during the strategy lifecycle, so a custom generator needs more than just `generate()`.

We'll revisit venue-specific client order ID generation once the v2 port lands, where it fits more cleanly.

Thanks again for the contribution.

PR #4153, opened 2026-05-27
Refine live cancel/update reconciliation

### LINKED ISSUE #4152: [RFC] Refine live reconciliation for pending cancel/update orders
## Context

This is a policy refinement rather than a bug fix.

The current live reconciliation policy for inflight orders treats `SUBMITTED` orders and `PENDING_CANCEL` / `PENDING_UPDATE` orders the same once the retry budget is exhausted in the runtime check path. That works for never-acknowledged submits, but it is too aggressive for ambiguous cancel/update outcomes, where the command may have timed out, never reached the venue, or reached the venue while the follow-up status query failed.

The specific failure mode is a network outage or prolonged connectivity problem that lasts long enough for `inflight_check_retries` to be exhausted. In that case, a live `PENDING_CANCEL` or `PENDING_UPDATE` order can be marked locally as `CANCELED` even though the venue never confirmed the cancel/modify request. Once that synthetic terminal state is applied, later venue recovery and reconciliation can no longer move the local order to the real final state, because the order has already been closed out incorrectly.

For those pending states, retry exhaustion alone does not prove the venue canceled or modified the order. Forcing a synthetic terminal event at that point can create a local state transition that is more certain than the actual venue outcome.

## Proposed Change

Keep the existing `SUBMITTED` timeout rejection behavior.

For `PENDING_CANCEL` and `PENDING_UPDATE`, log a warning and leave the order unresolved so that venue reconciliation can determine the final state.

## Why this is preferable

- It avoids a false terminal transition when the venue outcome is still ambiguous.
- It prevents a connectivity outage from hard-closing orders locally before the venue has actually confirmed the outcome.
- It keeps reconciliation as the source of truth for pending cancel/update outcomes.
- It keeps Rust and Python live execution behavior aligned.

## Considerations

- Unresolved pending orders may remain tracked longer, which is safer than resolving them incorrectly.
- We may want a future follow-up for stale unresolved-order metrics or alerting, but that should be separate from the state transition policy itself.
- The change is intentionally narrow and does not alter the existing `SUBMITTED` timeout behavior.

## Alternatives considered

1. Keep synthesizing `CANCELED` for pending cancel/update orders after retry exhaustion. Simple, but it can misrepresent the real venue state.
2. Introduce a new local terminal state for unresolved outcomes. That adds complexity without solving the core reconciliation problem.
3. Leave pending cancel/update orders unresolved until venue reconciliation resolves them. Preferred.

## Related implementation notes

- Rust: `crates/live/src/manager.rs`
- Python: `nautilus_trader/live/execution_engine.py`
- Tests: `crates/live/tests/manager.rs`, `tests/unit_tests/live/test_execution_engine.py`, `tests/unit_tests/live/test_execution_order_checks.py`
- Docs: `docs/concepts/live.md`, `docs/how_to/configure_live_trading.md`

JiajunWan:
# Pull Request

## Summary

Refine live cancel/update reconciliation so ambiguous `PENDING_CANCEL` and `PENDING_UPDATE` outcomes are no longer converted into synthetic `OrderCanceled` events after the inflight retry budget is exhausted. The Rust live manager and Python live execution engine now leave those orders unresolved and warn for venue reconciliation instead, while preserving the existing `SUBMITTED` timeout rejection path.

## Related Issues/PRs

Closes #4152

## Type of change

- [ ] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

Not expected to be breaking in the API or schema sense. The behavioral change is that unresolved pending cancel/update orders are no longer force-closed locally when retry exhaustion alone does not prove the venue outcome.

## Documentation

## Release notes

Suggested release note:

```markdown
- Refined live cancel/update reconciliation so pending orders remain unresolved until venue reconciliation, instead of being synthesized as canceled on retry exhaustion.
```

## Testing

Planned validation / examples:

```bash
cargo test -p nautilus-live --test manager inflight_pending -- --nocapture
cargo test -p nautilus-live --test manager inflight -- --nocapture
cargo test -p nautilus-live --test manager -- --nocapture
uv run --group test pytest tests/unit_tests/live/test_execution_engine.py -q
uv run --group test pytest tests/unit_tests/live/test_execution_order_checks.py -q
cargo fmt --check -p nautilus-live
git diff --check
```

### DIFF
--- a/crates/live/src/manager.rs
+++ b/crates/live/src/manager.rs
@@ -55,7 +55,7 @@ use nautilus_execution::{
 };
 use nautilus_model::{
     enums::{OrderSide, OrderStatus, OrderType, TimeInForce},
-    events::{OrderCanceled, OrderEventAny, OrderFilled, OrderInitialized},
+    events::{OrderEventAny, OrderFilled, OrderInitialized},
     identifiers::{
         AccountId, ClientOrderId, InstrumentId, PositionId, StrategyId, TradeId, TraderId,
         VenueOrderId,
@@ -882,20 +882,12 @@ impl ExecutionManager {
                                 }
                             }
                             OrderStatus::PendingUpdate | OrderStatus::PendingCancel => {
-                                // Generate cancellation for orders stuck in pending modify/cancel
-                                let event = OrderEventAny::Canceled(OrderCanceled::new(
-                                    order.trader_id(),
-                                    order.strategy_id(),
-                                    order.instrument_id(),
+                                log::warn!(
+                                    "Order {} exceeded max inflight retries while {}, leaving unresolved for venue reconciliation",
                                     order.client_order_id(),
-                                    UUID4::new(),
-                                    ts_now,
-                                    ts_now,
-                                    true, // reconciliation
-                                    order.venue_order_id(),
-                                    order.account_id(),
-                                ));
-                                result.events.push(event);
+                                    order.status()
+                                );
+                                continue;
                             }
                             _ => {
                                 // Order already resolved, just clear tracking
--- a/crates/live/tests/manager.rs
+++ b/crates/live/tests/manager.rs
@@ -1767,7 +1767,7 @@ fn test_inflight_increments_retry_count_before_max() {
 }
 
 #[rstest]
-fn test_inflight_pending_update_generates_canceled() {
+fn test_inflight_pending_update_stays_unresolved_at_max_retries() {
     let config = ExecutionManagerConfig {
         inflight_threshold_ms: 100,
         inflight_max_retries: 1,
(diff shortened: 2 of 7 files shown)


cjdsellers:
Hi @JiajunWan,

Thanks for the PR. The direction is useful, and I agree with the main policy point: **retry exhaustion alone should not synthesize terminal cancel/update outcomes when the venue result is still unknown**.

I've now tightened the policy/docs toward that in `docs/concepts/live.md` and kept `docs/how_to/configure_live_trading.md` focused on configuration with a link back to the runtime policy https://github.com/nautechsystems/nautilus_trader/commit/f720640a62f3a1a31bee94c2aa93902bf593423e.

So at this point we're not going to take this PR as-is. The implementation changes touch the inflight timeout path directly in both `ExecutionManager` and `LiveExecutionEngine`, but this needs to land with the broader command outcome standardization work rather than as a one-off behavior change. The docs changes also left some dense policy in the how-to guide, which I moved into the concept guide on our side.

I’m going to close this and follow up with the implementation from our side as the related standardization threads settle. Thanks again for pushing on this!

PR #4160, opened 2026-05-28
Add support for trigger children statuses in Hyperliquid order responses

sonnymai:
# Pull Request

## Summary

Fixes two HL order-submission bugs that silently orphaned positions on the venue:

- `normalTpsl` brackets were rejected wholesale because HL returns the SL/TP children as bare strings (`"waitingForFill"` / `"waitingForTrigger"`) that the response enum couldn't deserialize.
- MARKET orders that filled atomically never surfaced as `OrderFilled` — the WS submit path threw the parsed response away, the inflight check eventually timed out, and the order resolved as `OrderRejected(reason='UNKNOWN')` while the position sat open on HL.

Both transports (HTTP and WS, singleton and batch) now parse the response through a shared helper and feed each `OrderStatusReport` into the same handler the WS userEvents stream uses, so orders transition `SUBMITTED → ACCEPTED|FILLED|REJECTED` at submit time. Trigger children of a bracket carry a `pending-cloid:` placeholder venue id until the real oid arrives via WS.

## Type of change

- [x] Bug fix (non-breaking)

`HyperliquidWebSocketClient::submit_order` / `submit_orders` (and their PyO3 wrappers) now return the parsed report(s) instead of `()`. Callers that ignored the return are unaffected.

## Documentation

## Release notes

## Testing

Added a deserialization test covering the exact `normalTpsl` payload from the bug report. Existing single/batch submit tests still pass. `cargo check -p nautilus-hyperliquid` is clean. Manually reproduced on HL mainnet — MARKET fills and bracket submits now propagate end-to-end.

### DIFF
--- a/crates/adapters/hyperliquid/src/http/client.rs
+++ b/crates/adapters/hyperliquid/src/http/client.rs
@@ -3089,95 +3089,202 @@ impl HyperliquidHttpClient {
         let mut reports = Vec::with_capacity(order_response.statuses.len());
 
         for (order, order_status) in orders.iter().zip(order_response.statuses.iter()) {
-            let instrument_id = order.instrument_id();
-            let symbol = instrument_id.symbol.as_str();
-            let product_type = HyperliquidProductType::from_symbol(symbol).ok();
+            reports.push(self.build_status_report(
+                order.instrument_id(),
+                order.client_order_id(),
+                order.order_side(),
+                order.order_type(),
+                order.quantity(),
+                order.time_in_force(),
+                order.price(),
+                order.trigger_price(),
+                order_status,
+                account_id,
+                ts_init,
+            )?);
+        }
 
-            // Mirror the alias `cache_instrument` stored (token form for
-            // outcomes, leading segment for perps / spots).
-            let asset = cache_alias_for_symbol(symbol).unwrap_or_else(|| symbol.to_string());
-            let instrument = self
-                .get_or_create_instrument(&Ustr::from(asset.as_str()), product_type)
-                .ok_or_else(|| Error::bad_request(format!("Instrument not found for {asset}")))?;
+        Ok(reports)
+    }
 
-            let report = match order_status {
-                HyperliquidExecOrderStatus::Resting { resting } => self.create_order_status_report(
-                    order.instrument_id(),
-                    Some(order.client_order_id()),
-                    VenueOrderId::new(resting.oid.to_string()),
-                    order.order_side(),
-                    order.order_type(),
-                    order.quantity(),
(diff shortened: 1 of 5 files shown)


cjdsellers:
Hi @sonnymai,

Thanks for the PR and the clear write-up. The bare `waitingForFill` / `waitingForTrigger` parsing looks like the right direction, but I found a couple of lifecycle blockers before this can land.

The main one is the immediate-fill path. `_submit_order` now passes a `FILLED` submit response into `_handle_order_status_report_pyo3`, but that handler does not emit `OrderAccepted` or `OrderFilled` for `OrderStatus.FILLED`; it only marks `_terminal_orders` / `_pending_filled` and waits for a later `FillReport`. If that fill report is delayed or missing, the order still remains `SUBMITTED` and can resolve as `OrderRejected(reason="UNKNOWN")`, which is the failure mode this PR is meant to fix.

The second one is the deferred trigger-child path. `Tag` statuses are now elided, so those orders remain `SUBMITTED` until a later user-events `OrderAccepted` with a real oid. For `normalTpsl` children, that can be a normal long-lived state, but the live engine treats stale `SUBMITTED` orders as delayed in-flight orders and eventually rejects them as `UNKNOWN`. I think this needs either a local non-inflight accepted/deferred state for these children, or another way to stop the in-flight resolver from rejecting expected deferred tags.

One smaller follow-up: `HyperliquidHttpClient::submit_order` still rejects `Tag` statuses, while the new model docs say standalone trigger orders can also return them. If that venue behavior is possible for singleton submits, the HTTP path still has the old bug.

Could you add coverage around the live execution behavior, not just deserialization? The useful cases are a submit response with `filled` and no immediate `FillReport`, and a `normalTpsl` response where the child tags stay deferred past the in-flight threshold.

Thanks again for working through this.

sonnymai:
Thanks for feedback, I will address comments over the week - have found another issue with vaults too, will add it together there

cjdsellers:
Hi @sonnymai,

Thanks again for the PR.

I've picked this up and landed it on `develop`, adapting it to the recent change that standardized Hyperliquid's wire fields on `Decimal`. Both the HTTP and WebSocket submit paths (single and batch) now run through one shared `OrderStatusReport` builder, so the `normalTpsl` trigger children parse and the report surfaces at submit time. As your later revisions found, I kept the deferred children elided rather than carrying a `pending-cloid:` placeholder, since that placeholder was getting deduped against the real accept. I also verified it end-to-end on testnet with atomic market fills and a bracket, and credited you in the release notes.

Please let me know if you run into anything else!

PR #4161, opened 2026-05-28
Optimize datetime and UUID formatting

sunlei:
# Pull Request

- [x] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

This PR optimizes two fixed-format core formatting paths without changing public APIs or wire/FFI layout. It replaces chrono RFC3339 formatting in `unix_nanos_to_iso8601[_millis]` with fixed-width UTC integer formatting, and replaces `UUID4::new` runtime `write!` formatting with fixed-position lowercase hex table writes.

The retained changes are behavior-equivalent: the datetime functions still return `String`, out-of-range timestamps still fall back to the raw nanosecond string, and `UUID4` remains `#[repr(C)]` with the same `[u8; 37]` NUL-terminated layout.

## Performance summary

Benchmarks compare the baseline on `develop` (`63966996dc`) against this branch. Values are Criterion mean ranges from a local macOS/aarch64 workstation; latency reduction and speedup are approximate midpoint calculations unless noted.

| Area | Benchmark | Before | After | Latency reduction | Speedup |
| --- | --- | ---: | ---: | ---: | ---: |
| Datetime nanos | `datetime/unix_nanos_to_iso8601/epoch` | 66.079-66.317 ns | 47.189-47.289 ns | ~28.6% | ~1.40x |
| Datetime nanos | `datetime/unix_nanos_to_iso8601/current_sample` | 53.335-53.440 ns | 46.990-47.210 ns | ~11.8% | ~1.13x |
| Datetime nanos | `datetime/unix_nanos_to_iso8601/max_chrono` | 54.516-60.261 ns | 47.470-47.721 ns | ~17.1% midpoint, noisy | ~1.14-1.21x |
| Datetime fallback | `datetime/unix_nanos_to_iso8601/raw_fallback` | 94.683-94.911 ns | 30.190-30.355 ns | ~68.1% | ~3.13x |
| Datetime millis | `datetime/unix_nanos_to_iso8601_millis/epoch` | 54.429-54.536 ns | 43.995-44.185 ns | ~19.1% | ~1.24x |
| Datetime millis | `datetime/unix_nanos_to_iso8601_millis/millis_sample` | 51.505-51.604 ns | 43.668-43.784 ns | ~15.2% | ~1.18x |
| Datetime fallback | `datetime/unix_nanos_to_iso8601_millis/raw_fallback` | 94.667-94.946 ns | 29.667-29.819 ns | ~68.6% | ~3.18x |
[...]

PR #4162, opened 2026-05-28
Optimize AtomicMap snapshot borrowing

sunlei:
# Pull Request

## Summary

This PR updates hot adapter paths to reuse `AtomicMap::load()` snapshots inside message and batch handlers, and to borrow cached values from those snapshots instead of cloning them where possible. The main target is repeated `InstrumentAny` lookup in live WebSocket parsing and order-building paths, where the benchmarked `u64` fixture already shows a repeated-read win and likely understates the benefit for large enum payloads.

No new `AtomicMap` API is added. The change uses the existing `load()` snapshot API and keeps cold/control paths unchanged when the benchmarked win does not apply.

## Performance summary

Benchmarks compare three read patterns over 256 lookups using `AtomicMap<String, u64>`. The production migration applies the winning pattern to repeated reads and to hot instrument lookups that previously cloned larger `InstrumentAny` values.

| Optimization | Benchmark / pattern | Before | After | Latency reduction | Speedup / throughput |
| --- | --- | ---: | ---: | ---: | ---: |
| Reuse one snapshot for repeated reads | `get_cloned_each_lookup` -> `load_once_guard` over 256 reads | 2.946-2.962 us | 2.180-2.189 us | ~26.0% | ~1.35x |
| Avoid loading a guard per lookup | `load_each_lookup` -> `load_once_guard` over 256 reads | 2.952-2.958 us | 2.180-2.189 us | ~26.1% | ~1.35x |
| Improve read throughput | 256-read batch throughput | ~86.6-86.7 Melem/s | ~117.2 Melem/s | n/a | ~35% higher throughput |
| Batch writes, measured only | 16 repeated `insert` clone/swap -> one `rcu` batch | 490.02-492.76 us | 47.729-48.312 us | ~90.2% | ~10.2x |

The batch-write result is included as measured context but is not a production change in this PR. The audited candidate call sites were bootstrap/refresh/control paths rather than steady-state message or order hot paths.

## Detailed benchmark results

| Pattern | Result | Throughput | Decision |
| --- | ---: | ---: | --- |
| `get_cloned_each_lookup` over 256 reads | 2.946-2.962 us | ~86.7 Melem/s | Baseline |
| `load_each_lookup` over 256 reads | 2.952-2.958 us | ~86.6 Melem/s | Baseline |
| `load_once_guard` over 256 reads | 2.180-2.189 us | ~117.2 Melem/s | Use this pattern for repeated reads |
| 16 repeated `insert` clone-and-swap writes | 490.02-492.76 us | ~32.6 Kelem/s | Measured, not migrated here |
| Single `rcu` batch for 16 writes | 47.729-48.312 us | ~333 Kelem/s | Measured, no retained hot write-path migration |

### ASM evidence

There is no useful single-function `cargo asm` diff for this PR. The change is not an instruction-level rewrite of `AtomicMap`; it changes how adapter call sites use the existing `ArcSwap<AHashMap<_, _>>` API:

- before: repeatedly call `get_cloned()` or repeatedly create guards inside loops;
- after: call `load()` once per message/batch scope, then borrow from the snapshot with `get()`;
- after: avoid cloning large cached values such as `InstrumentAny` when parser/builders only need borrowed access.

The performance evidence is therefore the Criterion read-pattern benchmark plus the production call-site audit below.

## Production paths changed

| Adapter | Path | Why it is hot enough | Optimization applied |
| --- | --- | --- | --- |
| Bybit | Python WebSocket handlers for orderbook, trade, kline, ticker, order, execution, position | Runs inside live WebSocket message handling; several handlers process records per inbound message | Load one snapshot per handler and borrow instruments instead of cloning `InstrumentAny` |
| Coinbase | WebSocket L2, trade, ticker, user, candle handlers | Live WebSocket message loop; handlers can process multiple events per inbound message | Load one alias snapshot per message handler instead of loading the alias map for each product/trade |
| Hyperliquid | Order submit, modify, cancel, and batch order builders | Trading-path order APIs repeatedly resolve asset index and price precision before posting | Use already-interned `InstrumentId` symbols for cache lookups, avoiding `String` allocation and repeated `Ustr::from` on order paths |
| OKX | WebSocket submit and batch submit/modify/cancel order builders | Trading-path order APIs; batch APIs loop over user orders | Borrow instrument metadata from short-lived snapshots and load `inst_id_code_cache` once per batch |
| Polymarket | WebSocket order/trade dispatch and batch submit order builder | User-channel dispatch is a live execution path; batch submit loops over orders | Borrow token instruments in WS dispatch and load `neg_risk_index` once per batch submit |
| Kraken | Rust/Python Spot/Futures data and execution dispatch, including L3 runtime | Live ticker/trade/book/fill/order WebSocket paths; several messages carry repeated records | Borrow instrument snapshots instead of cloning `InstrumentAny`; fill-delta and batch arms load once per message |

## Scope deliberately not changed

| Candidate | Reason left unchanged |
| --- | --- |
| Cold helper methods, request-instrument paths, subscription-map insert/remove, ref-count lookups | The measured win is from repeated reads or avoiding large clones, not one-off control-plane operations |
| Bootstrap/reconnect owned snapshots such as `values().cloned().collect()` | Owned values are needed for setup/reconnect flows and are not steady-state message hot-path work |
| Single `Copy` lookups such as bar-type maps or boolean flags | Guard hoisting may add churn without evidence that a single guard dominates |
| OKX mass cancel owned lookup | Treated as control path, not tight trading-loop work |
| Polymarket reconciliation report builders | Background reconciliation path, not live WebSocket dispatch |
| Batch `rcu` write migrations | Strong microbench result, but audited candidates were bootstrap/refresh paths rather than hot batch writers |

## Related Issues/PRs

Related to #4161, which contains the separate core datetime/UUID formatter PR from the same performance investigation.

## Type of change

N/A. This PR does not add or remove public APIs and does not change `AtomicMap` semantics. It changes adapter call sites to use existing snapshot borrowing patterns.

## Testing

Validation run on this branch:

```bash
cargo check -p nautilus-core --benches
cargo check -p nautilus-kraken --features python
cargo check -p nautilus-bybit --features python
cargo check -p nautilus-coinbase -p nautilus-hyperliquid -p nautilus-okx -p nautilus-polymarket
pre-commit run check-formatting-rs --files crates/core/benches/concurrent_map.rs crates/adapters/kraken/src/data/spot.rs crates/adapters/kraken/src/python/websocket_spot.rs crates/adapters/okx/src/websocket/client.rs
```

All commands above passed locally.

### DIFF
--- a/crates/adapters/bybit/src/python/websocket.rs
+++ b/crates/adapters/bybit/src/python/websocket.rs
@@ -304,11 +304,6 @@ impl BybitWebSocketClient {
                 let mut funding_cache: AHashMap<Ustr, (Option<String>, Option<String>)> =
                     AHashMap::new();
                 let _client = client;
-                let _resolve = |raw_symbol: &Ustr| -> Option<InstrumentAny> {
-                    let key =
-                        product_type.map_or(*raw_symbol, |pt| make_bybit_symbol(raw_symbol, pt));
-                    instruments.get_cloned(&key)
-                };
 
                 tokio::pin!(stream);
 
@@ -1365,13 +1360,13 @@ fn register_batch_pending(
     }
 }
 
-fn resolve_instrument(
+fn resolve_instrument_from_snapshot<'a>(
     raw_symbol: &Ustr,
     product_type: Option<BybitProductType>,
-    instruments: &AtomicMap<Ustr, InstrumentAny>,
-) -> Option<InstrumentAny> {
+    instruments: &'a AHashMap<Ustr, InstrumentAny>,
+) -> Option<&'a InstrumentAny> {
     let key = product_type.map_or(*raw_symbol, |pt| make_bybit_symbol(raw_symbol, pt));
-    instruments.get_cloned(&key)
+    instruments.get(&key)
 }
 
 fn send_data_to_python(data: Data, call_soon: &Py<PyAny>, callback: &Py<PyAny>) {
@@ -1402,12 +1397,15 @@ fn handle_orderbook(
     call_soon: &Py<PyAny>,
     callback: &Py<PyAny>,
 ) {
-    let Some(instrument) = resolve_instrument(&msg.data.s, product_type, instruments) else {
+    let instruments_snapshot = instruments.load();
+    let Some(instrument) =
+        resolve_instrument_from_snapshot(&msg.data.s, product_type, &instruments_snapshot)
+    else {
         return;
     };
     let ts_init = clock.get_time_ns();
 
-    match parse_orderbook_deltas(msg, &instrument, ts_init) {
+    match parse_orderbook_deltas(msg, instrument, ts_init) {
         Ok(deltas) => {
             send_data_to_python(
                 Data::Deltas(OrderBookDeltas_API::new(deltas)),
@@ -1421,7 +1419,7 @@ fn handle_orderbook(
     let instrument_id = instrument.id();
     let last_quote = quote_cache.get(&instrument_id);
 
-    match parse_orderbook_quote(msg, &instrument, last_quote, ts_init) {
+    match parse_orderbook_quote(msg, instrument, last_quote, ts_init) {
         Ok(quote) => {
             quote_cache.insert(instrument_id, quote);
             send_data_to_python(Data::Quote(quote), call_soon, callback);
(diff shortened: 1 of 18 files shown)


cjdsellers (approved):
Thank you @sunlei, more great work 📈

PR #4164, opened 2026-05-29
Add Polymarket condition-level resolution for expired markets

graceyangfan:
## Summary
This PR delivers **PR2 for Polymarket pending-resolution** on top of PR1:
- condition-level resolve tracking owned by `PositionEvent`
- resolve reconciliation via both:
  - automatic post-expiry polling, and
  - manual `request_data(PolymarketResolveRequest)`

Core goal: settle expired unresolved exposure reliably without speculative winner inference.

Refs #4066

---

## Data-Driven Design (Gamma real payloads)
We validated the settle rules against live Gamma payloads and then encoded those samples as fixtures in this PR.

Added real fixtures:
- `gamma_market_closed_binary_accepting_false.json`
- `gamma_market_closed_binary_accepting_true.json`
- `gamma_market_closed_zero_zero_legacy.json`
- `gamma_market_closed_nonbinary_legacy.json`

### Observed categories and settlement rules
1. `closed=true`, binary `outcomePrices` (1/0), `acceptingOrders=false`  
Action: resolve and emit grouped YES/NO closes.

2. `closed=true`, binary `outcomePrices` (1/0), `acceptingOrders=true` (rare outlier)  
Action: resolve (do **not** block on `acceptingOrders=false`).

3. `closed=true`, `outcomePrices=["0","0"]`  
Action: skip (no winner determinable).

4. `closed=true`, non-binary `outcomePrices` (e.g. 0.58/0.42 scalar-like)  
Action: skip (not binary settlement shape for this adapter path).

### Observed prevalence (for reviewer context)
From live Gamma sampling used in this PR:

- Recent window (closer to current production): `N=10,100` closed markets
  - Case 1 (`closed + binary + acceptingOrders=false`): `9,939` (`98.406%`)
  - Case 2 (`closed + binary + acceptingOrders=true`): `2` (`0.020%`)
  - Case 3 (`closed + 0/0`): `0` (`0.000%`)
  - Case 4 (`closed + non-binary`): `159` (`1.574%`)

- Legacy window: `N=1,826` closed markets
  - Case 1: `0` (`0.000%`)
  - Case 2: `0` (`0.000%`)
  - Case 3: `41` (`2.245%`)
  - Case 4: `100` (`5.476%`)
  - Note: many legacy records omit `acceptingOrders` (`null`), so case1/case2 labeling is not meaningful there.

### Why this rule set
- `closed` is the finality gate.
- `outcomePrices` is used only for winner derivation when it is strictly binary.
- `acceptingOrders` is **not** a hard gate anymore (real closed markets can still report `true`).

---

### 1) Condition-level watch ownership
- watch entries keyed by `condition_id`
- YES/NO legs under one condition are settled together in one apply path

### 2) Unified reconcile engine
- auto poll and manual `request_data` share one fetch/parse/apply helper
- avoids path drift and duplicate logic

### 3) Manual selector semantics hardened
- explicit selectors: `condition_id`, `condition_ids`, `instrument_ids`
- non-Polymarket `instrument_ids` are ignored
- explicit-but-invalid selector returns summary error (no silent watchlist fallback)

### 4) Manual-only mode usability
When `resolve_poll_enabled=false`, no-param manual request includes expired eligible entries (not only paused/timed-out), so operator-driven reconcile remains usable.

### 5) Batch reconcile observability
- dedup + chunked condition fetch
- summary includes requested/emitted/failed IDs and error field

---

## Config Surface (Rust / PyO3 / Python / .pyi aligned)
- `resolve_poll_enabled` (default `true`)
- `resolve_poll_interval_secs` (default `30`)
- `resolve_poll_grace_secs` (default `10`)
- `resolve_poll_max_wait_secs` (default `1800`)
- `subscribe_new_markets` parity on Python config side

---

### New/updated tests
- strict resolve parser accepts/rejects categories from real Gamma fixtures
- grouped close emission and watchlist cleanup closure
- duplicate resolve no-op after watch removal
- manual fallback behavior with auto-poll on/off
- explicit invalid selector no-fallback behavior
- non-Polymarket `instrument_ids` ignored
- explicit multi-condition reconcile

### Validation

Both passing on this branch.

---

## Review Guide
1. `crates/adapters/polymarket/src/data.rs`
   - strict settle classification
   - condition-level watchlist ownership
   - shared reconcile path
   - manual request selector semantics
2. `crates/adapters/polymarket/test_data/*.json`
   - real payload categories backing settle rules
3. Config/binding parity:
   - `crates/adapters/polymarket/src/config.rs`
   - `crates/adapters/polymarket/src/python/config.rs`
   - `crates/adapters/polymarket/src/python/mod.rs`
   - `nautilus_trader/adapters/polymarket/config.py`
   - `python/.../__init__.pyi`

---

## Out of Scope
- changing PR1 pending-resolution state model
- soft winner fallback heuristics
- introducing non-binary payout settlement semantics in this PR


cjdsellers:
Hi @graceyangfan,

Thanks for the PR and the detailed Gamma notes. The condition-level resolve path looks like the right direction, but I found a few things I think should be fixed before this lands.

The main one is watch ownership. `update_resolve_watchlist_from_position_event` removes a token from the resolve watchlist on any `PositionClosed`, but Nautilus can still have other open `PositionId`s for the same instrument under hedging or split ownership. That means one close can drop the token while exposure remains, and the later resolve emits no `InstrumentClose` for that token. I think the watch entry needs to track open position IDs, or at least a refcount per token.

Two smaller points:

- `build_strict_resolved_market` should reject anything that is not exactly two token IDs, two outcomes, and two prices. Right now a malformed or multi-outcome payload with one `1` and the rest `0` can still pass the strict path.
- `apply_condition_resolution` uses `Price::new(f64, ...)`, and the tests compare with `.as_f64()`. Please switch that to Decimal-backed price construction and `.as_decimal()` assertions to match the adapter policy.

graceyangfan:
Thanks for the detailed review — we addressed the requested points and tightened the fallback path further.

### What was fixed

1. **Watch ownership safety (PositionClosed handling)**
- Watch entries continue to track per-token ownership via `open_position_ids`.
- Token watch state is only removed when the last open position for that token closes.
- This prevents premature watch removal under hedging/split ownership.

2. **Strict resolved-market shape constraints**
- Gamma strict path remains exact-cardinality only (binary shape): exactly two token IDs, two outcomes, and two prices.
- Malformed/multi-outcome/non-binary payloads are rejected from strict resolution.

3. **Decimal price policy in close emission**
- Resolution close prices are emitted with Decimal-backed construction (`Price::from_decimal_dp(ONE/ZERO, ...)`) and decimal assertions in tests.

### Additional hardening included in this update

4. **A->B resolve fallback semantics (condition_id ownership unchanged)**
- Keep `condition_id` as the ownership key.
- Resolve flow is now:
  - **A:** Gamma strict resolution first.
  - **B:** only unresolved condition IDs then query CLOB `GET /markets/{condition_id}` and use `tokens[].winner`.
- This improves real-world resolve hit-rate without weakening strictness.

5. **Real fixture coverage for CLOB market resolution**
- Added real CLOB fixtures (captured from production API) for both:
  - `closed=true`, binary, `accepting_orders=false`
  - `closed=true`, binary, `accepting_orders=true`
- Added parsing + resolution tests against those fixtures.

6. **Resolve summary accuracy fix**
- Adjusted stats behavior so if Gamma fails for a chunk but CLOB fallback resolves successfully, the condition is **not** incorrectly reported in `failed_condition_ids`.
- Added regression test to lock this behavior.

### Validation
- `cargo test -p nautilus-polymarket` passes.
- `pre-commit run --files ...` passes for the changed files and new fixtures.

If you want, I can also split the fixture-related additions into a standalone prep commit in a follow-up, but I kept them together here because they directly validate the new fallback path.

cjdsellers (approved):
Thanks again for the follow-ups @graceyangfan 🙏

cjdsellers:
@graceyangfan by the way, a small workflow improvement. It would help if you run `make format` on your final pass as there are some nightly only import sorting we've been using (hopefully they'll stabilize it one of these days).

PR #4171, opened 2026-05-30
Bybit demo trigger

sonnymai:
# Pull Request

## Summary

Bybit demo previously denied any order with `take_profit` / `stop_loss` / `order_iv` / `mmp` set, because the HTTP `submit_order` binding didn't accept those fields — even though the underlying `BybitBatchPlaceOrderEntry` already carries them and Bybit demo's `/v5/order/create` endpoint is identical to mainnet's. The deny gate was purely defensive against the pyo3 binding limitation. This plumbs the missing fields through so demo can use native TP/SL just like mainnet.

A new `BybitNativeTpSlParams` bundle carries the 12 TP/SL + option fields the entry struct already supports (`take_profit`, `stop_loss`, `tp/sl_trigger_by`, `tp/sl_order_type`, `tp/sl_limit_price`, `tpsl_mode`, `close_on_trigger`, `order_iv`, `mmp`). The Rust `BybitHttpClient::submit_order` takes it as a trailing `Option<&...>`, mirrors the WS path's TP/SL-implies-`Full`-mode default, and threads each populated field onto the entry builder. The PyO3 binding exposes it as a pyclass mirror with string-typed enum fields (matching `BybitWsPlaceOrderParams`), parsed at the boundary. Both the Rust `execution.rs` and Python `execution.py` demo branches drop the deny gate and pass the bundle through.

## Type of change

- [x] Bug fix (non-breaking)

`BybitHttpClient::submit_order` gains a trailing `native_tp_sl: Option<&BybitNativeTpSlParams>` arg; existing callers must pass `None` (the PyO3 binding defaults it to `None`). `tp_trigger_price` / `sl_trigger_price` are still silently ignored on the demo HTTP path — the `/v5/order/create` entry doesn't carry them. Same behavior as before the fix (where they were also denied), but worth flagging as a known limitation.

## Documentation

## Release notes

## Testing

Rewrote the stale `test_submit_order_with_tp_sl_in_demo_mode_emits_order_denied` as `..._routes_to_http` — now asserts the HTTP path is called with the `native_tp_sl` bundle populated from the `tp_sl` command params. `cargo check -p nautilus-bybit` and `cargo test -p nautilus-bybit --no-run` are clean.

### DIFF
--- a/crates/adapters/bybit/bin/flatten.rs
+++ b/crates/adapters/bybit/bin/flatten.rs
@@ -190,6 +190,7 @@ async fn close_position(
             position_idx,
             None,
             None,
+            None,
         )
         .await?;
 
--- a/crates/adapters/bybit/src/common/parse.rs
+++ b/crates/adapters/bybit/src/common/parse.rs
@@ -159,10 +159,13 @@ use crate::{
         },
         symbol::BybitSymbol,
     },
-    http::models::{
-        BybitExecution, BybitFeeRate, BybitFunding, BybitInstrumentInverse, BybitInstrumentLinear,
-        BybitInstrumentOption, BybitInstrumentSpot, BybitKline, BybitOrderbookResult,
-        BybitPosition, BybitTrade, BybitWalletBalance,
+    http::{
+        models::{
+            BybitExecution, BybitFeeRate, BybitFunding, BybitInstrumentInverse,
+            BybitInstrumentLinear, BybitInstrumentOption, BybitInstrumentSpot, BybitKline,
+            BybitOrderbookResult, BybitPosition, BybitTrade, BybitWalletBalance,
+        },
+        query::BybitNativeTpSlParams,
     },
     websocket::parse::parse_millis_i64,
 };
@@ -1547,6 +1550,26 @@ impl BybitTpSlParams {
 
     pub fn has_bbo(&self) -> bool {
         self.bbo_side_type.is_some()
+    }
+
+    /// Projects the native TP/SL + option fields onto the bundle the HTTP
+    /// submit_order entry expects. BBO + position_idx + leverage stay separate
+    /// because they're already first-class args on the submit signature.
+    pub fn to_native_tp_sl(&self) -> BybitNativeTpSlParams {
+        BybitNativeTpSlParams {
+            take_profit: self.take_profit.map(|p| p.to_string()),
+            stop_loss: self.stop_loss.map(|p| p.to_string()),
+            tp_trigger_by: self.tp_trigger_by,
+            sl_trigger_by: self.sl_trigger_by,
+            tp_order_type: self.tp_order_type,
+            sl_order_type: self.sl_order_type,
+            tp_limit_price: self.tp_limit_price.clone(),
+            sl_limit_price: self.sl_limit_price.clone(),
+            tpsl_mode: None,
+            close_on_trigger: self.close_on_trigger,
+            order_iv: self.order_iv.clone(),
+            mmp: self.mmp,
+        }
     }
 }
 
--- a/crates/adapters/bybit/src/execution.rs
+++ b/crates/adapters/bybit/src/execution.rs
@@ -1135,16 +1135,6 @@ impl ExecutionClient for BybitExecutionClient {
             return Ok(());
         }
 
-        if self.config.environment == BybitEnvironment::Demo
-            && (tp_sl.has_tp_sl() || tp_sl.order_iv.is_some() || tp_sl.mmp.is_some())
-        {
-            self.emitter.emit_order_denied(
-                &order,
-                "Native TP/SL and option params are not supported in demo mode",
-            );
-            return Ok(());
-        }
-
         log::debug!("OrderSubmitted client_order_id={}", order.client_order_id());
         self.emitter.emit_order_submitted(&order);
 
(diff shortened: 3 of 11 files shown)


cjdsellers:
Hi @sonnymai,

Thanks for the PR, and for the clear write-up on the demo deny-gate. The Bybit direction is sound: if demo's `/v5/order/create` accepts the same fields as mainnet, threading them through instead of denying is the right call.

A few things before this can land.

The bigger one is scope. Three of the four commits are Hyperliquid changes (the new `Tag` status variant, the `build_submit_orders_reports` refactor, and `submit_orders` returning `Vec<OrderStatusReport>`), but the title and description are Bybit-only. I'd prefer these split into two PRs. If they stay together, the description needs to cover the Hyperliquid work, and the `submit_orders` signature change should be called out as breaking on both the Rust and PyO3 sides, alongside the Bybit `submit_order` arg you already noted.

On testing: the new core logic isn't exercised yet. The `tests/http.rs` edits all pass `None` for `native_tp_sl`, so the field-threading block and the `Full`-mode default in `client.rs` have no coverage, and neither does the `TryFrom` enum parsing (including the invalid-enum error path). The Python `..._routes_to_http` test is a mock assertion on routing, not serialization. Could you add a Rust test that submits a populated `BybitNativeTpSlParams` against the mock server and asserts the request body, plus a unit test for the `TryFrom` error case? Worth noting CI hasn't run on this yet, so nothing here is verified beyond a compile.

**Two correctness questions:**

- For the Hyperliquid `pending-cloid:<client_order_id>` placeholder, the comment says the real oid arrives via userEvents and is reconciled through the cloid mapping, but that reconciliation isn't in this diff. Can you confirm the userEvents path actually recognizes and overwrites a `pending-cloid:` venue id? Otherwise those orders stay stuck with a synthetic id.
- `to_native_tp_sl` hardcodes `tpsl_mode: None` and `client.rs` then forces `Full` whenever TP or SL is set. If a user can set `tpsl_mode` to `Partial` on the source params, that selection gets silently dropped on demo. The Python `_build_native_tp_sl_params` omits `tpsl_mode` too. Is a partial mode reachable here?

Minor: both new structs have the derive order reversed from our convention (standard traits first, so `#[derive(Debug, Clone, Default)]`), and a `RELEASES.md` entry would be good since this changes user-facing behavior on two adapters.

Thanks again for working through this.

sonnymai:
Hi @cjdsellers, sorry about the Hyperliquid commits — rushed my branching, they'll go up in separate PR. This branch is now Bybit-only.

- `test_submit_order_with_native_tp_sl_serializes_fields` (`tests/http.rs`) — populated `BybitNativeTpSlParams` against the mock server, asserts every field in the `/v5/order/create` body including the `tpslMode=Full` default. Extended `CapturedOrder` with a `raw_body: Value` so further tests don't need struct churn.
- `test_submit_order_with_explicit_partial_tpsl_mode_is_preserved` — covers the new `tpsl_mode` invariant below.
- `test_native_tp_sl_params_try_from_{accepts_valid_enums,rejects_invalid_enum}` (`python/params.rs`) — parameterised across all five enum fields. Refactored `TryFrom`'s error type from `PyErr` to `anyhow::Error` (mapped at the pyo3 boundary) so unit tests don't need a live interpreter.

**`tpsl_mode` reachability.** Wasn't reachable from user input — plumbed it through end-to-end (added to `BybitTpSlParams` + `parse_tpsl_mode` + `_BYBIT_VALID_TPSL_MODES` + both `_apply_tp_sl_fields` and `_build_native_tp_sl_params`). `client.rs` already used `tp_sl.tpsl_mode.or_else(...)`, so a user-set `Partial` is now preserved.

Caught a bug while testing this: `BybitTpSlMode` has `#[serde(other)] Unknown`, so garbage strings were silently mapping to `Unknown`. Added an explicit reject in `TryFrom`.

**Derive order + RELEASES.md.** Fixed — both structs are `#[derive(Debug, Clone, Default)]`. Added `Fixes` and `Breaking Changes` entries.

cjdsellers:
Thanks for the follow-ups @sonnymai,

```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

cjdsellers:
Hi @sonnymai,

It looks like some of those fixes may not have made it into the current PR head: I still see `to_native_tp_sl` setting `tpsl_mode: None`, the HTTP tests only passing `None` for `native_tp_sl`, no `RELEASES.md` change, and the pre-commit check is still failing.

Could you check whether there is a local commit or generated change that was left behind, then if you push that to this branch I could take it the rest of the way?

sonnymai:
Ah apologies you're correct, I'll fix on the weekend as well as patch up the other Hyperliquid PR too

cjdsellers:
Hi @sonnymai,

Thanks again for getting this started, and for the clear write-up on the demo deny-gate. The direction was right: demo's `/v5/order/create` accepts the same TP/SL fields as mainnet.

I took it the rest of the way on `develop` in https://github.com/nautechsystems/nautilus_trader/commit/e547bb569369bd1e70d13241891ae8107468e070. Native TP/SL and the option params now route through the demo HTTP path instead of being denied, `tpsl_mode` is plumbed through so a user-set `Partial` survives, and invalid `tpsl_mode` values are rejected instead of silently mapping to `Unknown`. Custom TP/SL trigger prices stay denied on demo, since the create-order entry can't carry them. I also smoke-tested all three paths against the live demo API.

This closes the PR. Thanks again for the contribution, and please open a new one if you run into anything else!

(I'll also handle your Hyperliquid PR in a similar way to this.)

Issue #3941, opened 2026-04-27
Missing calculate_from_*() in MaxDrawdown/CAGR/CalmarRatio


a1zb2yc3z:
# Bug Report
## Confirmation

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates.
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

### Expected Behavior

MaxDrawdown/CAGR/CalmarRatio to be calculated in a backtest when they are registered with PortfolioAnalyzer; however exceptions are raised.

These statistics were added alongside the visualization tearsheets. It seems that these statistics implement `calculate_from_returns()` but not `calculate_from_realized_pnls()` and `calculate_from_positions()`, which is inconsistent with the other statistics. 

### Actual Behavior
An exception is raised, please see output with code snippet below.

### Steps to Reproduce the Problem
See code snippet below.

### Code Snippets or Logs
I used ` examples/backtest/crypto_ema_cross_ethusdt_trade_ticks.py` and added the following before `engine.run()`. Only one statistic demonstrated here but it applies to all statistics reported here.

```python
from nautilus_trader.analysis import MaxDrawdown
# ...
engine.portfolio.analyzer.register_statistic(MaxDrawdown())
engine.run()
# ...
```

The result:

```bash
Traceback (most recent call last):
  File "/workspace/dev/nautilus_trader/examples/backtest/crypto_ema_cross_ethusdt_trade_ticks.py", line 103, in <module>
    engine.run()
    ~~~~~~~~~~^^
  File "nautilus_trader/backtest/engine.pyx", line 1374, in nautilus_trader.backtest.engine.BacktestEngine.run
    stats_pnls: dict[str, dict[str, float]] = {}
  File "nautilus_trader/backtest/engine.pyx", line 1422, in nautilus_trader.backtest.engine.BacktestEngine.end
    bint missing_book_data
  File "nautilus_trader/backtest/engine.pyx", line 2148, in nautilus_trader.backtest.engine.BacktestEngine._log_post_run
    del self._data_name[data_priority]
  File "/usr/local/lib/python3.14/site-packages/nautilus_trader/analysis/analyzer.py", line 531, in get_stats_pnls_formatted
[...]

PR #4174, opened 2026-05-31
Add missing PortfolioStatistic trait methods returning None

mahimn01:
# Pull Request

## Summary

`MaxDrawdown`, `CAGR`, and `CalmarRatio` only override `calculate_from_returns()` in their `PortfolioStatistic` trait impls. `calculate_from_realized_pnls()` and `calculate_from_positions()` inherit the default panicking impls, so a Rust caller that registers one of these with `PortfolioAnalyzer` and calls `get_performance_stats_pnls()` or `get_performance_stats_general()` panics on dispatch (both iterate every registered statistic).

The PyO3 wrappers already return `None` from these methods after #3941, so Python users are insulated. This PR mirrors that fix to the underlying Rust trait impls, matching the existing stub pattern used by `SharpeRatio`, `SortinoRatio`, and `ReturnsVolatility` (which also have no per-stat tests for the stubs).


cjdsellers:
Hi @mahimn01 

Thanks again for a solid contribution :pray:.

cjdsellers (approved):

PR #4188, opened 2026-06-02
Fix raw value boundary constants in Rust

po4yka:

## Summary

`PRICE_RAW_MAX`/`MIN`, `QUANTITY_RAW_MAX`, and `MONEY_RAW_MAX`/`MIN` were computed with an `f64` multiply by `FIXED_SCALAR`, which rounded the constant below the true representable maximum (e.g. by 11,161,508,380,672 raw units in high-precision mode, and 512 in standard mode). A value built by `new` at the bound then exceeded the constant, so `from_raw` panicked and `from_raw_checked`/`checked_add` spuriously rejected valid boundary values. The fix casts both operands to the raw integer type before multiplying, so scaling stays exact in all feature modes.

## Related Issues/PRs

N/A

## Type of change

- [x] Bug fix (non-breaking)

N/A

## Documentation

## Release notes

- [x] I added a concise entry to `RELEASES.md` that follows the existing conventions

## Testing

Added boundary round-trip regression tests in `crates/model/src/types/{price,quantity,money}.rs`. Verified with `cargo test -p nautilus-model --lib types::` in both default (591 passed) and `--features high-precision` (609 passed) modes; `cargo fmt --check` and `cargo clippy` are clean on the touched files.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -91,6 +91,7 @@ Released on TBD (UTC).
 
 ### Fixes
 - Fixed `OrderMatchingEngine` panicking on `PriceType::Mark` bars during bar execution, now logs a warning and skips the bar (Rust)
+- Fixed `PRICE_RAW_MAX`/`MIN`, `QUANTITY_RAW_MAX`, and `MONEY_RAW_MAX`/`MIN` computed with lossy `f64` scaling, which rounded the bound below the representable maximum and could panic `from_raw` at the limits (Rust)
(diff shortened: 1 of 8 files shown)


cjdsellers:
Hi @po4yka,

Thanks for the PR and clear write-up. I reviewed the boundary constant changes and did not find any issues, good catch.

Please let me know if anything else turns up!

cjdsellers (approved):

PR #4191, opened 2026-06-03
Fix sandbox cached account configuration after initial AccountState

### LINKED ISSUE #4190: Rust SandboxExecutionClient leaves cached accounts with calculate_account_state=false, so fill-driven balances never update
# Bug Report

## Confirmation

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates.
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue against recent `develop` source and can still reproduce it.

### Expected Behavior

In the Rust `SandboxExecutionClient`, when `frozen_account=false`, fill-driven account balances should update in the cached account and subsequent `AccountState` / `QueryAccount` responses should reflect realized PnL and commissions.

This should match:

- existing Cython backtest/live behavior, where calculated accounts are registered before first account creation, and
- existing Rust backtest behavior, where the cached account is explicitly updated after initial account creation.

### Actual Behavior

On the Rust sandbox path, the cached account is created from the initial `AccountState` with `calculate_account_state=false`, so `Portfolio.update_order()` exits before running fill-driven balance updates.

As a result:

- fills occur normally,
- realized PnL is visible at the portfolio/position layer,
- but `AccountState` / `QueryAccount` can keep returning the starting balance (for example `20000 USDC`) even after many fills and non-zero PnL.

1. Start a Rust sandbox execution client with `frozen_account=false`, a cash or margin account, and non-zero starting balances.
2. Connect the client so it emits the initial `AccountState`.
3. Submit/fill orders which realize non-zero PnL and/or commissions.
4. Call `QueryAccount` or inspect emitted `AccountState` updates.
5. Observe that balances can remain at the starting values instead of reflecting fill-driven changes.

### Code Snippets or Logs

Relevant code paths on current `develop`:

- Missing-account creation defaults accounts to `calculate_account_state=false`:
  - `crates/model/src/accounts/any.rs`
- Cache uses that default path when first applying an `AccountState`:
  - `crates/common/src/cache/mod.rs`
- Fill-driven accounting is gated on `calculate_account_state`:
  - `crates/portfolio/src/portfolio.rs`
- Rust backtest explicitly fixes the cached account after initialization:
  - `crates/backtest/src/exchange.rs`
- Rust sandbox emits the initial `AccountState` but does not perform the same post-initialization cached-account configuration:
  - `crates/adapters/sandbox/src/execution.rs`

For comparison, the legacy/Cython path registers calculated accounts before first account creation:

- `nautilus_trader/accounting/factory.pyx`
- `nautilus_trader/backtest/execution_client.pyx`

### Why this looks like a bug instead of an enhancement

This behavior already exists in the Cython implementation and is already partially mirrored in Rust backtest. The Rust sandbox appears to have missed the same cached-account initialization step, so this looks like a parity/implementation bug rather than a design change.

### Specifications

- OS platform: macOS
- Python version: N/A for Rust-source analysis / local repo reproduction
- `nautilus_trader` version: current `develop` source (`539b0e3c03c7bcb77d687af1d560377858c455e4` at time of analysis)

graceyangfan:
## Summary
- sync the cached sandbox account configuration immediately after the initial `AccountState` is published
- enable calculated account state for non-frozen sandbox accounts and propagate configured margin leverage values into the cached account
- add focused regression tests covering the synced margin-account path and the frozen-account path

## Problem
Rust sandbox creates the cached account from the initial `AccountState`, but that cached account keeps the default `calculate_account_state=false` configuration. As a result, subsequent fills do not drive account balance updates, even though the sandbox account has already been connected.

## Solution
After publishing the initial `AccountState`, the sandbox execution client now reloads the cached account, applies the sandbox configuration (`calculate_account_state` and margin leverage settings), and writes the configured account back to the cache.

## Testing
- `pre-commit run --files crates/adapters/sandbox/src/execution.rs crates/adapters/sandbox/tests/execution.rs`
- `cargo test -p nautilus-sandbox`

Fixes #4190.

### DIFF
--- a/crates/adapters/sandbox/src/execution.rs
+++ b/crates/adapters/sandbox/src/execution.rs
@@ -579,6 +579,23 @@ impl SandboxExecutionClient {
         self.get_account_balances()
     }
 
+    fn sync_cached_account_config(&self) -> anyhow::Result<()> {
+        let Some(mut account) = self.get_account() else {
+            return Ok(());
+        };
+
+        account.set_calculate_account_state(!self.config.frozen_account);
+
+        if let AccountAny::Margin(margin_account) = &mut account {
+            margin_account.set_default_leverage(self.config.default_leverage);
+            for (instrument_id, leverage) in &self.config.leverages {
+                margin_account.set_leverage(*instrument_id, *leverage);
+            }
+        }
+
+        self.cache.borrow_mut().update_account(&account)
+    }
+
     /// Processes a quote tick through the matching engine.
     ///
     /// # Errors
@@ -776,6 +793,7 @@ impl ExecutionClient for SandboxExecutionClient {
             .generate_account_state(balances, margins, reported, ts_event, ts_init);
         let endpoint = MessagingSwitchboard::portfolio_update_account();
         msgbus::send_account_state(endpoint, &state);
+        self.sync_cached_account_config()?;
         Ok(())
     }
 
--- a/crates/adapters/sandbox/tests/execution.rs
+++ b/crates/adapters/sandbox/tests/execution.rs
@@ -26,18 +26,22 @@ use nautilus_common::{
         ExecutionEvent,
         execution::{SubmitOrder, TradingCommand},
     },
-    msgbus::{self, MessagingSwitchboard, stubs::get_typed_into_message_saving_handler},
+    msgbus::{
+        self, MessageBus, MessagingSwitchboard, TypedHandler,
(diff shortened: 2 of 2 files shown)


cjdsellers (approved):
Thanks @graceyangfan, solid changes.

PR #4200, opened 2026-06-04
Add check for internal aggregation in bar subscription

HaakonFlaaronning:
`execute_subscribe`/`execute_unsubscribe` always forward the bars command to the data client even when the bar type is internally aggregated, causing an error if the client doesn't support that bar type or forwarding a request the venue can't satisfy. This adds an early return for internally-aggregated bars after the local subscribe/unsubscribe.

Tested in my workflow, but not rigorously.

May need some input on this one @cjdsellers


faysou:
Continuing this PR on another one.

PR #4201, opened 2026-06-04
Fix quarter bar volume precision

HaakonFlaaronning:
`OrderMatchingEngine` quartered `bar.volume.raw` at the `FIXED_PRECISION` scale instead of the bar's own precision scale, producing tick sizes whose raw values weren't multiples of `10^(FIXED_PRECISION − precision)`. Downstream, `quantity_matches_precision` silently rejected every fill that hit the unaligned remainder creating loads of `Skipping fill … not compatible with size_precision=N` warnings in backtests. The fix splits at the volume's own precision scale (aligned by construction) and floors each tick to `size_increment`. Covered by unit and integration tests and tested with my workflow.

@cjdsellers Could use some input here.

### DIFF
--- a/crates/execution/src/matching_engine/engine.rs
+++ b/crates/execution/src/matching_engine/engine.rs
@@ -1132,6 +1132,20 @@ impl OrderMatchingEngine {
         quantity.raw.is_multiple_of(scale)
     }
 
+    fn quarter_bar_volume(volume: Quantity, size_increment: Quantity) -> (Quantity, Quantity) {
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @HaakonFlaaronning,

Thanks for the PR and the clear write-up. I’ll merge this now and follow up locally on a couple of edge cases: preserving total bar volume for very small bars, rounding synthetic sizes to `size_increment` multiples, and adding a regression that exercises the actual fill path through `normalize_fill_quantity`.

Please let me know if anything else turns up!

cjdsellers (approved):

PR #4207, opened 2026-06-05
Fix Hyperliquid account balances for zeroed perp summaries

### LINKED ISSUE #4206: Hyperliquid Unified/Portfolio Margin accounts can report zero USDC collateral
### Summary

Hyperliquid Unified Account / Portfolio Margin accounts can be parsed as having zero USDC collateral when the perp `crossMarginSummary` is present but zeroed and the real collateral is reported in `spotClearinghouseState`.

### Current behavior

`parse_combined_account_balances_and_margins` treats the presence of a perp `crossMarginSummary` as enough evidence that the perp leg should own USDC. It then skips spot USDC during the merge.

For Unified/PM accounts, Hyperliquid can return a zeroed perp summary such as `totalRawUsd=0` and `withdrawable=0` while the usable USDC sits in `spotClearinghouseState`. In that case, the merge drops the real spot USDC and keeps the zeroed perp USDC balance.

### Expected behavior

Spot USDC should be surfaced when the perp state does not actually carry meaningful USDC collateral. Standard perp accounts where the perp leg has non-zero collateral should continue deduplicating spot USDC to avoid double-counting.

### Suggested fix

Gate spot-USDC deduplication on whether the perp leg actually reflects USDC collateral, not merely on whether `crossMarginSummary` exists.

Regression coverage should include:

- Standard funded perp summary still deduplicates spot USDC.
- Missing perp summary still uses spot USDC.
- Zeroed Unified/PM perp summary surfaces spot USDC.
- Margin-used-only or total-raw-usd-only cases remain perp-authoritative.

jungsooyun:
## Summary
- treat Hyperliquid perp USDC as authoritative only when the perp summary reflects collateral or margin
- surface spot USDC for zeroed Unified/Portfolio Margin perp summaries
- add regression coverage for zeroed Unified/PM summaries plus each perp-authoritative signal (`totalRawUsd`, `totalMarginUsed`, `withdrawable`)

The margin-used-only case remains perp-authoritative intentionally; that preserves the existing standard cross-margin assumption that active perp margin state belongs to the perp leg rather than the spot merge path.

Fixes #4206

## Tests
- cargo +1.96.0 test -p nautilus-hyperliquid test_parse_combined --lib
- cargo +1.96.0 fmt --check (exits 0; stable rustfmt warns that repo import grouping options are nightly-only)
- cargo +1.96.0 test -p nautilus-hyperliquid --lib

### DIFF
--- a/crates/adapters/hyperliquid/src/common/parse.rs
+++ b/crates/adapters/hyperliquid/src/common/parse.rs
@@ -980,7 +980,7 @@ pub fn parse_combined_account_balances_and_margins(
         .cross_margin_summary
         .as_ref()
         .is_some_and(|summary| {
-            summary.total_raw_usd > Decimal::ZERO
+            summary.total_raw_usd != Decimal::ZERO
                 || summary.total_margin_used > Decimal::ZERO
                 || perp_state.withdrawable.unwrap_or(Decimal::ZERO) > Decimal::ZERO
         });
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @jungsooyun 

Thanks for the contribution!

```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

jungsooyun:
Hi @cjdsellers , 

  Thanks for the note and for the guidance.

  This is my first contribution to NautilusTrader, so I missed running the full pre-commit checks before submitting the PR. I’ve now run `make pre-commit` locally, fixed the reported formatting issue, and pushed the update to the same PR branch.

Thanks again!

cjdsellers:
Hi @jungsooyun,

Thanks for the PR and the quick pre-commit follow-up. The zeroed Unified/PM path looks like the right direction, but I found one edge case before this can land: `perp_reflects_usdc` only treats positive `total_raw_usd` as perp-owned USDC, while the existing parser preserves negative `totalRawUsd` as a real USDC balance. A negative perp balance with zero `withdrawable` and zero margin would now have its USDC entry removed and could be replaced by spot USDC, so I think this should check for non-zero `total_raw_usd` instead and add a regression for that case.

jungsooyun:
@cjdsellers Thanks, I checked this against the existing parser behavior and you’re right: negative `totalRawUsd` is preserved as a real USDC balance, so the combined-path guard should treat any non-zero `total_raw_usd` as perp-owned USDC.

My original change was implicitly assuming the liquidation/risk path would always keep this from surfacing as a negative perp-side balance, which was too narrow for a parser handling raw API snapshots.

I pushed a follow-up commit that changes the check from `> Decimal::ZERO` to `!= Decimal::ZERO` and adds a regression for negative `totalRawUsd` with zero withdrawable and zero margin.

I also verified the combined parser tests and `make pre-commit` locally.

Please let me know if there’s anything else I missed in the issue or in my follow-up changes. I’m really enjoying learning through nautilus_trader.

cjdsellers (approved):
Thanks again for the follow-ups @jungsooyun 👌

Issue #4193, opened 2026-06-03
[Interactive Brokers] Connection timeout socket leak and client ID conflict on reconnection


vortexua:
### Expected Behavior

Upon a connection handshake timeout (`TimeoutError` waiting for the `managedAccounts` message), the `InteractiveBrokersClient` should cleanly close the TCP socket, cancel all background message reading/processing tasks, and resume the reconnection loop in a clean state.

### Actual Behavior

A `TimeoutError` is caught in `except TimeoutError:` inside `_start_async()` but does not trigger `self._stop()` or any cleanup/disconnection. 
* The TCP socket remains open.
* The background message-reading/processing tasks (`_tws_incoming_msg_reader_task`, `_internal_msg_queue_processor_task`, and `_msg_handler_processor_task`) are left dangling.
* On the next reconnection iteration (5 seconds later), a new connection is established under the same client ID. The gateway detects a client ID conflict with the still-open dangling socket and terminates or resets the new connection, resulting in a continuous `socat` reset/reconnection storm.
* In addition, if `self._stop()` is called (e.g., from `except Exception:` on transient connection failures), `self._is_shutting_down` is set to `True` (inside `_stop_async`), which permanently blocks subsequent reconnection attempts if the loop condition variables are not reset.

---

### Introduction of the Regression in v1.227.0

This behavior became critical in **v1.227.0** due to a combination of three changes:
1. **502 added to CLIENT_ERRORS:** In `1.227.0`, error code `502` (Couldn't connect to TWS) was added to the client errors list. When a transient 502 is received during gateway startup, the error mixin immediately clears `_is_ib_connected`, causing the handshake to fail and triggering the connection `TimeoutError` or `ConnectionError`.
2. **Latent socket leak:** The handshake timeout raises a `TimeoutError`, which hits the `except TimeoutError:` block. This block has historically logged the timeout but failed to perform a `.disconnect()` or `self._stop()` cleanup, exposing the socket leak.
[...]

PR #4210, opened 2026-06-05
Harden Interactive Brokers reconnect startup handling

faysou:
# Pull Request

## Summary

- Close the raw IB socket transport during failed Python startup attempts so `EClient.disconnect()` does not synchronously trigger `connectionClosed()` and start a competing reconnect loop.
- Add capped exponential reconnect backoff using the existing `_reconnect_delay` as the base delay and a per-client jitter of 0-3 seconds.
- Preserve the configured IB Gateway client ID for normal connects and reconnects, including `ibg_client_id=0` behavior.
- Add a one-shot temporary client ID fallback only after IB error 326, when Gateway still reports the previous client ID as in use.
- Force `fetch_all_open_orders=True` when the 326 fallback is activated so recovery uses `reqAllOpenOrders()` under the temporary client ID.
- Stop reconnect startup from proceeding after shutdown or dispose is requested during reconnect backoff.

### Design decisions
- Failed startup cleanup now calls `self._eclient.conn.disconnect()` only when a raw connection exists. Normal explicit stop and disconnect paths still call `self._eclient.disconnect()`.
- `_configured_client_id` keeps the originally configured Nautilus identity stable for component naming and provider cache keys.
- Runtime `_client_id` uses `_configured_client_id` by default so `reqOpenOrders()` continues to target the configured session scope during normal recovery.
- IB error 326 sets a one-shot `_randomize_client_id_on_next_connect` flag. The next `_initialize_connection_params()` call consumes the flag, chooses a 1000-9999 client ID that differs from both the configured and current stale IDs, and then clears the flag.
- The 326 fallback also enables `_fetch_all_open_orders` because a temporary client ID cannot safely recover orders placed under the configured client ID with `reqOpenOrders()`.
- Reconnect backoff follows 5, 10, 20, 40, 80, 160, then a 300 second cap, plus fixed per-client jitter.
- After any reconnect sleep, startup checks shutdown/dispose state before opening a new socket so stop/dispose cannot reconnect later.

### Files covered
- `nautilus_trader/adapters/interactive_brokers/client/client.py`
- `nautilus_trader/adapters/interactive_brokers/client/common.py`
- `nautilus_trader/adapters/interactive_brokers/client/connection.py`
- `nautilus_trader/adapters/interactive_brokers/client/error.py`
- `nautilus_trader/adapters/interactive_brokers/factories.py`
- `tests/integration_tests/adapters/interactive_brokers/client/test_client.py`
- `tests/integration_tests/adapters/interactive_brokers/client/test_client_connection.py`
- `tests/integration_tests/adapters/interactive_brokers/client/test_client_error.py`
- `tests/integration_tests/adapters/interactive_brokers/client/test_client_order.py`

### Verification
- `uv run --no-sync pytest tests/integration_tests/adapters/interactive_brokers/client/test_client_connection.py tests/integration_tests/adapters/interactive_brokers/client/test_client.py tests/integration_tests/adapters/interactive_brokers/client/test_client_error.py tests/integration_tests/adapters/interactive_brokers/client/test_client_order.py`
- `make format`

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/4193

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -245,6 +245,7 @@ class InteractiveBrokersClient(
                         f"Attempt {self._connection_attempts}: attempting to reconnect in {reconnect_delay} seconds...",
                     )
                     await asyncio.sleep(reconnect_delay)
+
                     if self._is_shutting_down or self.state in _SHUTDOWN_STATES:
                         break
 
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -16,6 +16,7 @@
 import asyncio
 import functools
 import os
+import secrets
 import traceback
 from collections.abc import Callable
 from collections.abc import Coroutine
@@ -121,6 +122,7 @@ class InteractiveBrokersClient(
         self._cache = cache
         self._host = host
         self._port = port
+        self._configured_client_id = client_id
         self._client_id = client_id
         self._fetch_all_open_orders = fetch_all_open_orders
         self._request_timeout_secs = request_timeout_secs
@@ -166,8 +168,11 @@ class InteractiveBrokersClient(
         self._max_connection_attempts: int = int(os.getenv("IB_MAX_CONNECTION_ATTEMPTS", "0"))
         self._indefinite_reconnect: bool = not self._max_connection_attempts
         self._reconnect_delay: int = 5  # seconds
+        self._reconnect_delay_max: int = 300  # seconds
+        self._reconnect_jitter_secs: int = secrets.randbelow(4)
         self._had_ib_connection: bool = False
         self._last_disconnection_ns: int | None = None
+        self._randomize_client_id_on_next_connect: bool = False
 
         # MarketDataMixin
         self._bar_type_to_last_bar: dict[str, BarData | None] = {}
@@ -220,6 +225,10 @@ class InteractiveBrokersClient(
 
         while not self._is_ib_connected.is_set():
             try:
+                if self.state in _SHUTDOWN_STATES:
+                    break
+
+                self._is_shutting_down = False
                 self._connection_attempts += 1
 
                 if (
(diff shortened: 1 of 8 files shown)


HKOWL:
@faysou Great fix for the core socket leak — we hit this exact scenario in production during IB's nightly maintenance window and want to share what we found.

**Our experience with `eclient.disconnect()` (your line 297)**

We applied a local patch for the same bug and went through two iterations:

**v1** used `self._eclient.disconnect()` — the same call in this PR. On our first nightly maintenance cycle (2026-06-01 03:33 UTC) it caused a silent crash. The root cause: `EClient.disconnect()` calls `self.wrapper.connectionClosed()` **synchronously** before returning. That fires `_handle_disconnection()`, which schedules a new `_start_async()` coroutine via `ensure_future` — while the original `_start_async()` is still running inside the reconnect loop. During IB maintenance, each `TimeoutError` iteration spawned another concurrent `_start_async()` task, compounding into multiple racing loops.

**v2** uses `self._eclient.conn.disconnect()` — the raw transport-level close. This closes the TCP socket without invoking the `connectionClosed()` wrapper callback, so `_handle_disconnection()` is never triggered. No duplicate `_start_async()` task. This has been stable through all nightly maintenance cycles since.

Suggested change:
```python
# Instead of:
self._eclient.disconnect()
# Use:
self._eclient.conn.disconnect()
```

Your task-cancellation + `asyncio.gather` before the disconnect call may mitigate the duplicate-task issue in some cases (since the reader task is cancelled), but `connectionClosed()` fires synchronously from inside `disconnect()` itself — before the event loop yields — so the cancelled state of the reader task doesn't block it.

---

**Related: Error 326 "client ID already in use" on mid-session reconnect**

Once the socket is cleaned up correctly per this PR, a second reconnection failure surfaces: after a mid-session IB disconnect, reusing the same `_client_id` triggers IB error 326 ("client ID already in use") because the gateway still has the old session registered. The bot then loops indefinitely.

Fix: randomise `_client_id` in `_initialize_connection_params()` on every call (e.g. `random.randint(1000, 9999)`) after invoking the original method. We've run this in production for several months without issues — a fresh client ID appears in every reconnect log.

---

**Related: exponential backoff in the reconnect loop**

Even with clean socket handling, the fixed `self._reconnect_delay: int = 5` (hardcoded in `client.py`, not user-configurable) causes a problem we measured in production: during a prolonged IB maintenance window, `startApi()` queues commands faster than the gateway can drain them. We observed **953 connection attempts** in one maintenance cycle — the gateway's handshake initialization queue never fully cleared because a new attempt arrived every 5 seconds before the previous handshake completed.

Replacing the fixed delay with exponential backoff (5 → 10 → 20 → 40 → 80 → 160 → 300s cap) plus per-client jitter (0–3s) gives the gateway time to complete its startup between attempts. Since deploying this we've had clean single-attempt reconnects after maintenance. This is standard network programming practice and has no behavioural side-effects — `_reconnect_delay` is still used as the base.

faysou:
Thank you for your experience, I've updated the PR accordingly.

faysou:
@HKOWL you can maybe test this branch before it gets merged.

Issue #4224, opened 2026-06-08
[RFC] Add rolling binary-option scope subscriptions


graceyangfan:
# RFC: Rolling binary-option scope subscriptions

Polymarket Up/Down first, with a public shape that can later support HIP-4-style binary scopes.

## Context

Nautilus already models Polymarket and Hyperliquid binary markets as concrete `BinaryOption` instruments. What many live strategies actually trade, however, is a **recurring logical scope** rather than one short-lived concrete `instrument_id`.

Examples:
- Polymarket: `BTC 5m Up/Down`
- Hyperliquid HIP-4: `BTC daily priceBinary`

Today, the lifecycle of that recurring scope leaks into strategy code:
- discover the current concrete market
- subscribe quotes / trades / book streams for both legs
- detect expiry / rotation
- discover the replacement market
- unsubscribe the old legs
- subscribe the new legs
- update the execution target instrument ids

For short-dated binary products such as Polymarket `5m` / `15m` Up/Down, that is the wrong layer to own venue lifecycle logic.

## Goal

Provide a high-level rolling binary-option scope API so that strategies can subscribe to a logical scope and receive the currently active concrete tradable members as that scope rotates.

The intended public subscription family is:
- `subscribe_binary_option_scope(...)`
- `unsubscribe_binary_option_scope(...)`

The intended strategy-facing update is:
- `BinaryOptionScopeSlice`

## Core modeling distinction: active tradable market vs parent container

This RFC is about the **active tradable child market / outcome set**, not the broad parent event or question container.

### Polymarket
For Polymarket Up/Down there are three layers:
1. parent event / question container
2. active child market window, for example one `BTC 5m Up/Down` expiry
3. concrete outcome instruments inside that child market

The rotating unit is the **child market window**. Its concrete tradable legs rotate together.

### HIP-4
For HIP-4 there are also multiple layers:
1. recurring question-level scope
2. concrete outcome market
3. concrete tradable side tokens

For `priceBinary`, one active outcome market yields two concrete tradable side-token instruments.
For future `priceBucket`, multiple concrete outcome markets may belong to the broader question, and some may close earlier than others.

## Public shape

The public common layer should stay venue-neutral.

### Generic common types
The common layer should own:
- `BinaryOptionScope`
- `BinaryOptionScopeStreams`
- `BinaryOptionScopeMember`
- `BinaryOptionScopeSlice`

[...]

PR #4227, opened 2026-06-08
Add generic binary option scope subscription plumbing

graceyangfan:
## Summary
- add the venue-neutral binary-option scope domain and custom-data types: `BinaryOptionScope`, `BinaryOptionScopeStreams`, `BinaryOptionScopeMember`, and `BinaryOptionScopeSlice`
- wire binary-option-scope subscribe/unsubscribe commands through `DataActor`, `DataClient`, the message bus, and `DataEngine`
- expose the new surface to Python `DataActor` / `Strategy` users, including `on_binary_option_scope(...)` callbacks and generated `.pyi` stubs

Closes the phase-1 portion of #4224.

## What This PR Adds
This PR establishes the common plumbing needed for rolling binary-option scope subscriptions without introducing any venue-specific lifecycle logic into the abstraction layer.

Concretely, it adds:
- a generic `BinaryOptionScope { scope_id, venue }` key in `model`
- a stream-intent object (`BinaryOptionScopeStreams`) that can be carried by a logical scope subscription
- a strategy-facing aggregated update object (`BinaryOptionScopeSlice`) with concrete `BinaryOptionScopeMember` entries
- a full generic subscribe / unsubscribe path across `DataActor`, `DataClient`, `msgbus`, and `DataEngine`
- Python bindings and stubs so strategies can subscribe to scope slices and receive `on_binary_option_scope(...)` updates through a venue-neutral API

## What This PR Deliberately Does Not Add
This PR does **not** add:
- `PolymarketUpDownScope`
- slug or window parsing
- expiry-driven refresh / discovery
- automatic concrete instrument rebinding across rotations
- adapter-side manager logic

That boundary is intentional: the common layer owns the generic subscription surface, while concrete scope discovery and rollover remain adapter responsibilities.

## Design Notes
The new public surface matches the direction we want for both Polymarket and future HIP-4-style binary markets:
- strategies subscribe to a logical scope
- adapters remain free to implement venue-specific discovery and rotation internally
- strategies consume `BinaryOptionScopeSlice` updates carrying the currently active concrete `instrument_id` values

This keeps `model`, `common`, and `trading` free of venue-specific scope types such as a Polymarket Up/Down config object.

## Follow-up PR
The next PR for #4224 will build on top of this surface from inside the Polymarket adapter. It will:
- add an adapter-local `PolymarketUpDownScope` config / parsing layer
- implement expiry-bound discovery and rotation for the active Up/Down child market window
- rebind the requested stream intent (`quotes`, `trades`, `book_deltas`) to the newly active concrete instruments
- publish rotated `BinaryOptionScopeSlice` updates back to strategies through the generic path introduced here
- include the remaining adapter lifecycle hardening needed around reset / resubscribe generation ownership

## Testing
- `cargo test -p nautilus-model binary_option_scope -- --nocapture`
- `python3 python/generate_stubs.py`
- `pre-commit run --files crates/common/src/actor/data_actor.rs crates/common/src/clients/data.rs crates/common/src/messages/data/mod.rs crates/common/src/messages/data/subscribe.rs crates/common/src/messages/data/unsubscribe.rs crates/common/src/msgbus/api.rs crates/common/src/msgbus/core.rs crates/common/src/msgbus/mod.rs crates/common/src/msgbus/switchboard.rs crates/common/src/python/actor.rs crates/data/src/client.rs crates/data/src/engine/mod.rs crates/model/src/data/binary_option_scope.rs crates/model/src/data/mod.rs crates/model/src/python/mod.rs crates/trading/src/python/strategy.rs nautilus_trader/model/__init__.py python/nautilus_trader/common/__init__.pyi python/nautilus_trader/model/__init__.pyi python/nautilus_trader/trading/__init__.pyi`


graceyangfan:
Follow-up plan for PR2: keep the concrete Polymarket Up/Down scope type and expiry-driven rotation manager inside the Polymarket adapter, then build it on top of the generic subscription surface introduced here. That follow-up will cover slug/window parsing, near-expiry refresh scheduling, scope slice publication for rotated members, and the remaining adapter lifecycle hardening around reset/resubscribe generation ownership.

cjdsellers:
Hi @graceyangfan,

Thanks for the PR and the clear write-up. The generic common-layer surface looks aligned with the RFC, but I found a few things I think should be sorted before this lands.

The main one is the scope routing key. `BinaryOptionScope` and `BinaryOptionScopeSlice` both include `venue`, but the message-bus topic is only `data.binary_option_scope.{scope_id}` and both the `DataActor` subscribe/unsubscribe path and the `DataEngine` publish path use only `scope_id`. That means two venues with the same logical scope id can cross-deliver slices, duplicate subscription checks can collapse distinct scopes, and unsubscribing one venue can remove the handler for the other. I think the switchboard topic/key should include both `venue` and `scope_id`.

Two other functional points:

- `DataClientAdapter` de-dupes `SubscribeBinaryOptionScope` only by `BinaryOptionScope`, so a later subscription for the same scope with a broader `BinaryOptionScopeStreams` intent never reaches the adapter. Since the RFC says the stream intent belongs to the logical scope subscription and must be preserved across rotations, this should either merge/refcount stream masks per scope or otherwise forward stream-intent changes.
- `handle_custom_data_pipeline` now publishes `BinaryOptionScopeSlice` on the live binary-scope topic. `process_pipeline` is meant to publish only on `data.pipeline.*`, so historical/replay scope slices can leak into live `on_binary_option_scope` subscribers. This should use a pipeline scope topic or skip the typed scope publish in pipeline mode.

One stub issue too: the new PyO3 methods live in `crates/model/src/data/binary_option_scope.rs`, but `python/generate_stubs.py` only collects getter/default fixups from `crates/**/src/python/**/*.rs`. The generated stubs expose getters like `slice.members()` as methods and miss the default args for `BinaryOptionScopeStreams()`, even though the runtime bindings use `#[getter]` and `quotes=false, trades=false, book_deltas=false`.

graceyangfan:
Thanks, I pushed a follow-up that addresses the issues you called out.

- Scope routing now keys binary option scope topics on both `venue` and `scope_id`, and the `DataActor` subscribe/unsubscribe path plus the `DataEngine` live publish path all use the full `BinaryOptionScope`.
- `DataClientAdapter` now tracks `BinaryOptionScope -> BinaryOptionScopeStreams`, merges widened stream intent for an already-subscribed scope, and forwards the broadened subscribe downstream.
- `handle_custom_data_pipeline` no longer publishes `BinaryOptionScopeSlice` on the live typed topic, so pipeline/replay scope slices stay on `data.pipeline.*`.
- The Python-facing `BinaryOptionScope*` `#[pymethods]` were moved under `crates/model/src/python/...` so the stub generator picks them up correctly; the generated `.pyi` now preserves getter properties and the default args for `BinaryOptionScopeStreams()`.

I also added regression coverage for the routing key, stream-intent widening, and live-vs-pipeline separation, then reran targeted tests plus `python3 python/generate_stubs.py` and `pre-commit run --files ...` locally.

Please let me know if you want me to break any of this out further.

cjdsellers:
Hi @graceyangfan 

Taking a step back on this one, I’m not convinced this belongs in the common platform layer. The change introduces a new first-class subscription family across model, common, data, trading, msgbus, and Python APIs, but the motivating behavior appears to be Polymarket-specific scope discovery and rotation. Binary option instruments are generic; “binary option scope subscription with stream intent and rotated members” feels like adapter-managed grouping/rebinding logic.

Before merging something like this into core, I’d want to see either a second concrete venue/use case that needs the same abstraction, or a smaller design that keeps the scope manager and emitted custom data inside the adapter. The platform already has custom data and adapter-specific subscription surfaces that could support this without permanently widening core APIs.

graceyangfan:
@cjdsellers   I want to clarify the actual intent behind this PR and the follow-up implementation, because it is not just “adding one more subscription type,” and it is not about moving Polymarket-specific lifecycle logic into common/core.

  There are really two layers here.

  The first layer is what this PR adds: a minimal, venue-neutral public semantic surface for recurring binary scopes:

  - BinaryOptionScope
  - BinaryOptionScopeStreams
  - BinaryOptionScopeMember
  - BinaryOptionScopeSlice
  - subscribe_binary_option_scope(...)
  - on_binary_option_scope(...)

  The second layer is what the adapter will own in follow-up work: automatic discovery, automatic rotation, and automatic rebinding. Concretely, that means:

  - deriving current and next windows from a business-level scope definition
  - automatically discovering the active child market
  - preloading or chasing the replacement market near expiry
  - carrying the declared stream intent across rotations onto the new generation’s concrete members
  - publishing a new BinaryOptionScopeSlice so the strategy is told explicitly which instrument_ids are currently tradable

  So this design is not “just expose a generic callback and still make the strategy do the real work.” The intent is the opposite: the strategy
  subscribes to the logical scope, the adapter/runtime owns discovery + rotation + rebinding, and the strategy consumes an authoritative slice.

  That is also why I think the analogy to OptionChainManager is strong, not superficial. An option chain could also be implemented manually in strategy space by discovering legs, subscribing quotes/greeks, aggregating snapshots, and rebalancing membership over time. But the platform intentionally does not leave that ongoing aggregate lifecycle in strategies. Instead, subscribe_option_chain and OptionChainManager let DataEngine own the active  set, bootstrap, wire-level subscribe/unsubscribe churn, and slice publication.crates/data/src/option_chains/manager.rs:78 crates/data/src/engine/mod.rs:3350

  A rolling binary scope is the same category of problem. It is not just a single-instrument subscription problem. It is a logical object backed by a changing set of concrete instruments, rotating on a fixed interval, where the strategy needs a stable logical view rather than raw exposure to the underlying churn. The driver of membership change is different, but the abstraction boundary is the same: option chains change through ATM/strike/expiry dynamics, while binary scopes change through fixed-interval child-market rotation.

On the question of whether the existing APIs can already compose the desired functionality, my view is:
They can compose an approximation of it, but they cannot compose the full intended behavior cleanly and robustly.
That distinction matters.
Looking at the current Polymarket path in this codebase, the building blocks we have today are:
- bootstrap / scoped loading
    We can load concrete markets/instruments through event_slug_builder, event_slugs, and market_slugs.crates/adapters/polymarket/src/config.rs:46
    crates/adapters/polymarket/src/providers.rs:241
  - periodic refresh
    We can periodically refresh scoped instruments using update_instruments_interval_mins.crates/adapters/polymarket/src/data.rs:651
  - new-market discovery
    We can enable subscribe_new_markets, receive NewMarket, fetch by slug, and filter before accepting
    instruments.crates/adapters/polymarket/src/data.rs:1060 crates/adapters/polymarket/src/filters.rs:28
  - per-instrument market-data subscription
    Once an instrument_id is known, we can subscribe quotes / trades / book deltas, and even auto-load a missing instrument if it is referenced before
    being cached.crates/adapters/polymarket/src/data.rs:1556

  These capabilities prove one reviewer point: this is not impossible today.

  But they also show why the existing APIs are not yet a first-class fit for the problem. They provide “how to load a concrete market” and “how to subscribe a concrete instrument.” They do not provide “how to maintain a recurring logical scope across rotations.”

  In other words, what is missing is not raw access to data. What is missing is semantic guarantees:

  - no first-class identifier for a recurring logical scope
    We can identify slugs, events, markets, and instruments, but not “the continuous BTC 5m UpDown scope.”
  - no authoritative active-generation concept
    The current APIs can tell us that markets/instruments exist, but not directly “this is the one current generation for the scope right now.”
  - no native stream-intent continuity
    Today quotes/trades/book subscriptions are bound to concrete instruments, not to a logical scope that automatically preserves intent across rotations.
  - no atomic rebinding / cutover semantic
    We can discover a new market and subscribe its instruments, but there is no formal lifecycle path that says “the scope has now rotated from  generation A to generation B.”
  - no unified strategy-facing authoritative slice
    Without BinaryOptionScopeSlice, the strategy has to reconstruct truth by combining instrument events, custom data, subscription state, and cache state.

  That is why I would say the current Polymarket APIs can “assemble” this, but they cannot “perfectly assemble” it.

  This becomes especially important in the 5-minute fixed-interval case, because the problem is not static discovery. It is dynamic rotation. The hard part is not only finding the market, but doing all of the following at the right time:

  - knowing which generation is current
  - knowing which generation is next
  - preparing next early enough
  - preserving stream intent across the cutover
  - retargeting execution
  - avoiding skew from overlap, late discovery, retries, or ordering drift

  If all of that remains in strategy space, the strategy is no longer just implementing trading logic. It is implementing a mini scope manager. That splits discovery, subscription, and execution state and makes the result less robust and less reusable.

  So the more precise position I want to express is:

  1. The existing APIs are indeed sufficient to manually assemble a rolling binary scope.
  2. But what they assemble is instrument orchestration, not scope semantics.
  3. The actual goal of the follow-up implementation is adapter-local automatic discovery, automatic rotation, and automatic rebinding.
  4. The value of this PR is to first make recurring binary scope a first-class concept the platform can express and strategies can consume directly.
  5. That is structurally the same motivation behind OptionChainManager: not that strategies cannot do it, but that they should not own this class of ongoing aggregate/rotation lifecycle.

cjdsellers:
Hi @graceyangfan,

Thanks for clarifying. I agree strategies should not own discovery, rotation, and rebinding, but that does not imply the common platform needs a new public subscription family first. This PR still adds `BinaryOptionScope` types, subscribe/unsubscribe commands, msgbus routing, `DataEngine` / `DataActor` handling, Python callbacks, and stubs across 22 files for a Polymarket-first workflow, with the second venue still hypothetical. That is a wide core API change, not a minimal adapter change.

The path I’d be open to is implementing the Polymarket scope manager inside the adapter first, using adapter-local custom data or an adapter-specific surface where possible. If a second concrete venue proves the same shape, we can promote the abstraction into core with much stronger evidence.

Issue #4228, opened 2026-06-08
[Interactive Brokers] Three data-integrity bugs during degraded connections: false reconciliation trigger, None position iteration, and commission crash


HKOWL:
### Expected Behavior

During a degraded IB connection (socket error, partial disconnect, or maintenance window):
1. `get_positions()` should distinguish between "IB confirmed no open positions" and "request failed" — callers must not treat a failed request as a confirmed-empty position list.
2. `generate_position_status_reports()` should handle a failed `get_positions()` result gracefully and skip reconciliation for the cycle rather than iterating over an invalid value.
3. `_create_fill_report()` and related commission sites should handle IB's pending commission sentinels (`None` / `-1.0`) gracefully rather than crashing.

### Actual Behavior

**Bug 1 — `get_positions()` returns `[]` on socket error → false "venue has no positions"**

`get_positions()` in `account.py` returns `[]` on both a clean "no positions" response and a socket error/timeout. The reconciliation path cannot distinguish the two cases:
* A socket error during `reqPositions()` causes `all_positions` to be `None` internally, but the method silently converts it to `[]`.
* The exec engine receives `[]`, concludes the venue has no open positions, and flags all locally-tracked positions as ghost `EXTERNAL SHORT` discrepancies.
* This can trigger unwanted reconciliation orders mid-trade during IB maintenance windows.

**Bug 2 — `generate_position_status_reports()` iterates `None` → `TypeError`**

Once Bug 1 is fixed (returning `None` on socket error instead of `[]`), `generate_position_status_reports()` and `generate_order_status_reports()` in `execution_engine.py` receive `None` and attempt `for report in None` → `TypeError: 'NoneType' object is not iterable`. This exception surfaces inside `asyncio.gather(return_exceptions=True)`, where it is silently swallowed rather than logged, making the root cause difficult to diagnose.

Note: even without Bug 1's fix, if a `None` ever reaches this path by any other means, the same crash applies.

**Bug 3 — `commissionAndFees = None` or `-1.0` crashes `Money()` constructor**

During degraded connections, IB sends `commissionAndFees = None` or `commissionAndFees = -1.0` (IB's well-known pending/unknown sentinel) in commission callbacks. This value reaches `Money()` directly at multiple sites in `execution.py`:
* `Money(None, USD)` raises `TypeError`
* `Money(-1.0, USD)` raises a value error
* Both crash `_create_fill_report()`, preventing fill reports from being generated mid-trade.

---

[...]

PR #4230, opened 2026-06-09
Harden Interactive Brokers reconciliation against connection loss

HKOWL:
# Pull Request

## Summary

Fixes four bugs that combine during IB Gateway disconnects (e.g. daily CME/COMEX trading halt or nightly maintenance) to silently corrupt the execution cache and produce a startup crash loop.

- Return `None` from `get_positions()` on socket error instead of `[]`, so callers can distinguish &#34;venue confirmed flat&#34; from &#34;venue unreachable&#34;.
- Raise `ConnectionError` in `generate_position_status_reports()` and `generate_order_status_reports()` when `get_positions()` returns `None`, allowing `asyncio.gather(return_exceptions=True)` to catch it and skip reconciliation cleanly.
- Guard `commissionAndFees = None` or `-1.0` (IB pending sentinel) before constructing `Money(...)` at all fill-report sites in Python and Rust. Without this guard, `Money(None, ...)` raises during degraded connections.
- Fix `asyncio.gather(return_exceptions=True)` result loops in the live execution engine to check `isinstance(..., BaseException)` instead of `isinstance(..., Exception)`. In Python 3.9+, `asyncio.CancelledError` inherits from `BaseException` — it slipped through the `Exception` check and was iterated as a `list[PositionStatusReport]`, raising `TypeError: &#39;CancelledError&#39; object is not iterable`. Four sites are fixed; line 1716 already used `BaseException` correctly.

### Design decisions

- `get_positions()` return type is widened to `list[Position] | None`. `None` strictly means &#34;request failed — venue state unknown&#34;. An empty list continues to mean &#34;IB confirmed no open positions&#34;. This distinction is load-bearing: treating a failed request as confirmed-zero is what triggered the ghost fill described below.
- `ConnectionError` is raised (not returned) so it propagates through `asyncio.gather(return_exceptions=True)` without any change to calling signatures in the execution engine.
- Commission clamping uses `if fees is None or fees < 0: fees = 0.0` — preserves the fill event with zero commission rather than dropping it entirely, which matches the behaviour when IB later sends the real commission via `commissionReport`.
- `BaseException` is the correct catch-all for `asyncio.gather(return_exceptions=True)` results; `Exception` was always incorrect here since Python 3.8 when `CancelledError` was reparented.

### Production incident

On 2026-06-01, bugs 1–3 combined at 16:30 CT (daily CME/COMEX trading halt) to generate a ghost inferred SELL fill for an open MGC position:

1. `reqPositions()` errored with socket disconnect; default `result=[]` was treated as &#34;venue confirmed flat&#34;.
2. The commission guard was absent — `generate_fill_reports()` raised `TypeError` trying to compare `None` to `int`, but the exception was swallowed.
3. `ExecEngine` reconciled `cached_qty=1` against `venue_qty=0` and generated an inferred `OrderFilled(side=SELL)` — a ghost external SHORT position was written into the in-memory cache.
4. The bot ran normally for several hours, then shut down gracefully. The graceful stop cleanly persisted the invalid logical state to Redis — no partial write; just faithful persistence of bad data.
5. Every subsequent startup loaded the orphaned `index:order_position` entry and crashed on `cache.add_order()` integrity check. The crash loop persisted until Redis was manually flushed.

With bugs 1 + 2 + 4 fixed together: `get_positions()` returns `None` → `generate_position_status_reports()` raises `ConnectionError` → `asyncio.gather` catches it as `BaseException` → reconciliation skipped with a WARN log → no ghost fill, no corrupted state.

### Files covered

- `nautilus_trader/adapters/interactive_brokers/client/account.py`
- `nautilus_trader/adapters/interactive_brokers/execution.py`
- `nautilus_trader/live/execution_engine.py`
- `crates/adapters/interactive_brokers/src/execution/parse.rs`

### Verification

```bash
uv run --no-sync pytest tests/integration_tests/adapters/interactive_brokers/
make format
make pre-commit
```

New and updated tests:
- `tests/integration_tests/adapters/interactive_brokers/client/test_client_account.py` — `test_get_positions_returns_none_on_connection_error`
- `tests/integration_tests/adapters/interactive_brokers/test_execution_reconciliation.py` — `test_generate_position_status_reports_raises_on_disconnected`, `test_generate_order_status_reports_raises_on_disconnected`
- `tests/integration_tests/adapters/interactive_brokers/test_execution_reconciliation.py` — fixed existing `test_generate_position_status_reports_flat_when_no_positions` (`return_value=None` → `return_value=[]` to preserve its original intent after the semantic change)

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/4228

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

`get_positions()` return type widens from `list[Position]` to `list[Position] | None`. Any caller that assumed a non-None return must now handle `None`. Within this PR all call sites are updated.

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/interactive_brokers/src/execution/core_updates.rs
+++ b/crates/adapters/interactive_brokers/src/execution/core_updates.rs
@@ -245,9 +245,16 @@ impl InteractiveBrokersExecutionClient {
                     let mut cache = commission_cache
                         .lock()
                         .map_err(|_| anyhow::anyhow!("Failed to lock commission cache"))?;
+                    // IB sends -1.0 as a pending-sentinel before the real commission arrives;
+                    // clamp it to zero so downstream Money construction never receives a negative value.
+                    let commission_value = if commission.commission == -1.0_f64 {
+                        0.0_f64
+                    } else {
+                        commission.commission
+                    };
                     cache.insert(
                         commission.execution_id.clone(),
-                        (commission.commission, commission.currency.clone()),
+                        (commission_value, commission.currency.clone()),
                     );
                 }
 
--- a/crates/adapters/interactive_brokers/src/execution/core_updates.rs
+++ b/crates/adapters/interactive_brokers/src/execution/core_updates.rs
@@ -245,8 +245,8 @@ impl InteractiveBrokersExecutionClient {
                     let mut cache = commission_cache
                         .lock()
                         .map_err(|_| anyhow::anyhow!("Failed to lock commission cache"))?;
-                    // IB sends -1.0 as a pending-sentinel before the real commission arrives;
-                    // clamp it to zero so downstream Money construction never receives a negative value.
+                    // IB uses -1.0 as a pending-sentinel before the real commission arrives;
+                    // clamp only that sentinel to zero (legitimate rebates can be negative).
                     let commission_value = if commission.commission == -1.0_f64 {
                         0.0_f64
                     } else {
--- a/crates/adapters/interactive_brokers/src/execution/parse.rs
+++ b/crates/adapters/interactive_brokers/src/execution/parse.rs
@@ -108,8 +108,9 @@ pub fn parse_execution_to_fill_report(
     let last_qty = Quantity::new(execution.shares, instrument.size_precision());
     let last_px = Price::new(execution_price, instrument.price_precision());
 
-    // Create commission
-    let commission_money = Money::new(commission, Currency::from_str(commission_currency)?);
+    // Create commission — clamp IB's -1 pending sentinel to 0.0 to avoid invalid Money values
+    let commission_clamped = if commission < 0.0 { 0.0 } else { commission };
+    let commission_money = Money::new(commission_clamped, Currency::from_str(commission_currency)?);
 
     // Parse execution time
     let ts_event = parse_execution_time(&execution.time)?;
--- a/nautilus_trader/adapters/interactive_brokers/client/account.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/account.py
@@ -135,7 +135,7 @@ class InteractiveBrokersClientAccountMixin(BaseMixin):
         else:
             self._log.debug(f"Subscription doesn't exist for {name}")
 
-    async def get_positions(self, account_id: str) -> list[Position]:
+    async def get_positions(self, account_id: str) -> list[Position] | None:
         """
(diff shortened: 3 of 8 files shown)


cjdsellers:
Hi @HKOWL,

Thanks for the PR and the detailed incident write-up. The `get_positions()` distinction and the `BaseException` handling look like the right direction, but I found a few things I think should be sorted before this lands.

The main one is that `get_open_orders()` still maps request timeout or connection failure to `[]`, and `generate_order_status_reports()` treats that as a confirmed empty venue response. That leaves the same unknown-as-empty failure mode in the order path: startup reconciliation can build synthetic orders with empty `open_order_fills`, and continuous reconciliation with `open_check_open_only=False` can eventually reject or cancel cached live orders after repeated empty venue responses.

Two smaller functional points:

- The Rust live spread path still uses raw commission values in `build_pending_combo_fill()` and `generate_leg_fill()`. The PR clamps `parse_execution_to_fill_report()`, but spread combo and leg fills can still receive IB's `-1.0` pending sentinel as a negative `Money`.
- The Python guards clamp every negative `commissionAndFees` value, not only `None` or the pending sentinel. Since `CommissionAndFees` is combined commission and fees, and IBKR tiered pricing can include venue rebates, I think this should be narrowed to `None` and the exact sentinel value unless we are sure real negative net fees cannot appear here.

I did not run the suite locally. Please let me know if anything here needs more detail.

HKOWL:
Thanks for the thorough review. All three points addressed:

1. **`get_open_orders()` failure-as-empty** — now returns `None` on timeout/socket error (same pattern as `get_positions()`); both call sites in `generate_order_status_report` and `generate_order_status_reports` raise `ConnectionError` on `None` rather than treating it as a confirmed-empty response. We also applied the same fix to `get_executions()` which had the identical pattern — a timeout would silently map to no fills during fill report reconciliation.

2. **Rust combo/leg commission sentinel** — clamped at the commission cache insertion point in `core_updates.rs` so both `build_pending_combo_fill()` and `generate_leg_fill()` never receive the `-1.0` sentinel.

3. **Python commission guard scope** — narrowed from `< 0` to `== -1.0` across all four call sites to preserve legitimate negative values (venue rebates under tiered pricing).

PR #4245, opened 2026-06-11
Fix Deribit chart bar volume for inverse perpetuals

filipmacek:
# Pull Request

## Summary

On Deribit inverse perpetuals and inverse futures (e.g. `BTC-PERPETUAL`, `BTC-27DEC24`),
the chart channel's `volume` is in base currency (BTC) while the matching
`trades.{instrument}` channel reports each trade's `amount` in USD. As a result
`Bar.volume` and `TradeTick.size` were on incompatible scales. This routes those
products through the chart `cost` field (USD) — both on the live WebSocket feed
and the HTTP TradingView chart endpoint — so historical and live bars line up
with trade ticks.

Options are deliberately excluded: Deribit flags BTC/ETH options as `is_inverse`,
but per the API docs their trade `amount` is still reported in the underlying
base currency, so they keep using `volume`.

### Details

- New helper `use_cost_for_bar_volume(&InstrumentAny)` in `common/parse` centralises
  the decision: `true` only when `is_inverse() && instrument_class ∉ {Option, OptionSpread}`.
- `parse_chart_msg` (WS) and `parse_bars` (HTTP) take `use_cost_for_volume: bool`;
  `DeribitWsFeedHandler` and `DeribitHttpClient::request_bars` derive it from the
  cached instrument.
- `DeribitTradingViewChartData.cost` is no longer `#[serde(default)]` — it's a
  documented standard field of the response — and the existing length-parity
  check now covers it.
- Reference: Deribit API docs on the `amount` field — *"For perpetual and inverse
  futures the amount is in USD units. For options and linear futures it is the
  underlying base currency coin."* (https://docs.deribit.com/api-reference/market-data/public-get_last_trades_by_currency)
- Consistent with the cross-adapter invariant that `Bar.volume` and `TradeTick.size`
  share units per instrument (verified against `bitmex`, `bybit`, `binance`, `okx`,
  `hyperliquid`).

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/crates/adapters/deribit/src/common/parse.rs
+++ b/crates/adapters/deribit/src/common/parse.rs
@@ -25,12 +25,12 @@ use nautilus_core::{
 };
 use nautilus_model::{
     data::{Bar, BarType, BookOrder, TradeTick},
-    enums::{AccountType, AggressorSide, BookType, OptionKind, OrderSide},
+    enums::{AccountType, AggressorSide, BookType, InstrumentClass, OptionKind, OrderSide},
     events::AccountState,
     identifiers::{AccountId, InstrumentId, Symbol, TradeId},
     instruments::{
         CryptoFuture, CryptoFuturesSpread, CryptoOption, CryptoOptionSpread, CryptoPerpetual,
-        CurrencyPair, any::InstrumentAny,
+        CurrencyPair, Instrument, any::InstrumentAny,
     },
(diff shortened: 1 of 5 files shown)


cjdsellers (approved):
Thanks @filipmacek 👌

Issue #4253, opened 2026-06-11
[Polymarket] Coordinate pruning of inactive dynamic instruments across data client, global cache, and execution client


graceyangfan:
## Summary

Follow up to #3945.

Polymarket continuously creates new short-lived markets, so a long-running live process can discover and stream an unbounded number of venue instruments over time.

Today, an inactive Polymarket instrument can remain retained across multiple layers of the live stack:

- `PolymarketDataClient` local instrument state
- global `Cache` instrument-owned state
- `PolymarketExecutionClient` local lookup maps

This means memory growth is not only a data-client-local issue. Even if per-subscription market-data state is pruned, instrument definitions and execution-side lookup state can still accumulate over long-running sessions.

This RFC proposes a coordinated lifecycle for inactive Polymarket instruments, rather than pruning only one layer.

## Relationship to #3945

Issue #3945 proposed opt-in pruning of inactive Polymarket instrument state in the data client.

That RFC was directionally correct, but it only addressed part of the retention chain in `crates/adapters/polymarket/src/data.rs`:

- data-client-local `instruments`
- data-client-local `token_meta`
- data-client-local `order_books`
- data-client-local `last_quotes`

For Polymarket-style dynamic venues, the broader issue is that the same inactive instrument can also remain retained in:

- the global `Cache`
- execution-side lookup structures:
  - `shared_token_instruments`
  - `neg_risk_index`

So the main problem is not just "one cache is not pruned", but that the same inactive instrument can be retained by multiple layers without a coordinated removal path.

## Problem

For static instrument universes, retaining instruments indefinitely is usually acceptable.

For Polymarket, that assumption does not hold well because the venue continuously introduces new markets/instruments, many of which are streamed briefly and then never used again.

Over time, a long-running live process can accumulate inactive Polymarket instruments in several places.

### 1. Polymarket data client local instrument state

The core local retained owners are:

- `PolymarketDataClient.instruments`
- `PolymarketDataClient.token_meta`

The common write path is `cache_instrument(...)` in `crates/adapters/polymarket/src/data.rs`, which inserts into both maps.

That path is reached from multiple runtime flows:

- bootstrap via `bootstrap_instruments()`
- periodic refresh via `refresh_scoped_instruments(...)`
- `request_instruments(...)`
- `request_instrument(...)`
- `new_market` discovery in the WebSocket handler

Not all of these write paths propagate equally far through the runtime.

[...]

PR #4272, opened 2026-06-14
Improve Polymarket Rust expired market state handling

graceyangfan:
## Summary

Implements Polymarket PR1 for expired instrument retirement in the paper/live data path.

This change reduces non-essential runtime state retention after instrument expiry while preserving only the minimum metadata/watchlist state needed for later resolution handling.

Closes or advances: #4253

## What changed

- retire expired live instrument state from the Polymarket data client runtime
- keep watchlisted expired instruments only at the minimal metadata layer needed for resolution
- remove unwatched expired instruments from the local cache/runtime state
- prevent expired instruments from being reintroduced into live subscription/request paths
- stop auto-load retry loops for expired instruments
- keep an expiry sweeper active even when `resolve_poll_enabled = false`
- split common live-vs-expired test fixtures so regular data-client tests use a real active Polymarket market and expiry tests use an explicitly expired fixture

## Behavior changes

- cached expired instruments are now rejected for live subscribe/request paths
  - quotes
  - book snapshot
  - trades
- `request_instrument` still returns a response for expired instruments, but does not cache or publish `DataEvent::Instrument`
- `resolve_poll_enabled = false` still runs the expiry retirement sweep; it only disables resolution fetch/polling
- watchlisted expired instruments retain only the minimal metadata needed until resolution; other live runtime state is retired eagerly

## Why

This keeps expired markets from continuing to occupy:

- WS subscription intent/state
- local order book / quote caches
- token meta used for live routing
- auto-load retry state

while avoiding pushing cleanup pressure into the later auto-resolve / settlement path.

## Tests

Validated with:

- `cargo test -p nautilus-polymarket -- --nocapture`

Added/updated regressions cover:

- expiry sweep still runs when auto resolve polling is disabled
- unwatched expired instruments are removed from cache/runtime state
- mixed watched/unwatched batch retirement keeps only watchlist-required state
- message handler startup does not re-seed `token_meta` for retained watched expired instruments
- expired auto-load retires without retrying
- cached expired instruments are rejected across quotes/book/trades live paths
- regular data-client integration tests use an official current active Polymarket market fixture

## Scope note

This PR is intentionally adapter-scoped and keeps settlement/PositionClosed follow-up work out of scope for PR1.


cjdsellers:
Hi @graceyangfan,

Thanks for the PR and the write-up. The core retirement path, the `token_meta` seed switch to the live cache, and the auto-load interaction all look good. A few things to address before it lands, none of them correctness blockers:

- The retirement sweep re-runs a full `retire_local_instrument_state` on every watchlisted-expired instrument each `resolve_poll_interval_secs`, for the life of the watch. That's wasted work after the first pass; retiring each instrument once rather than every tick would avoid it.
- `test_data/gamma_market_live.json` hardcodes `endDate` `2028-11-07` against a real-time clock, so the request/reset tests will start failing once expiry enforcement rejects the fixture after that date. A fixed clock or a now-relative expiry would stop it recurring.
- A few cleanups worth a look: `keep_local_metadata` is read before the `sync_ws_subscription_async` await and used after it; `resolve_token_id_from` in `data_runtime.rs` duplicates the one in `subscriptions.rs`; the expiry checks are spread across the cache-population callers rather than the `cache_instrument` chokepoint; and a couple of the new tests assert or await on things that can't fail (`_tx` kept alive, the pre-insert `active_quote_subs` check).

I reviewed against the diff and didn't run the suite locally. Please let me know if any of these need more detail.

graceyangfan:
Thanks for the review. I pushed a follow-up pass to address the points you called out.

Changes in this round:
- retirement sweep now skips repeated work for watchlisted expired instruments once their live runtime state has already been torn down, while still retaining the instrument metadata needed by the watchlist/settlement path
- removed the fixed-date `gamma_market_live.json` fixture and switched the request/recheck tests to derive a test-local future expiry from the stable `gamma_market.json` fixture
- moved the `keep_local_metadata` decision in the retirement path to after the WS sync await
- removed the duplicate `resolve_token_id_from` helper from `data_runtime.rs`
- tightened the cache-population path by routing request/new-market cache inserts through the same active-instrument cache helper
- cleaned up the message-handler startup test so the receiver closure is observable rather than relying on a live sender being kept around

Validation run locally:
- `cargo test -p nautilus-polymarket retire_expired_local_instruments_retires_watchlisted_runtime_state_once -- --nocapture`
- `cargo test -p nautilus-polymarket resolve_poll_task_ -- --nocapture`
- `cargo test -p nautilus-polymarket --test data_client -- --nocapture`
- `cargo test -p nautilus-polymarket new_market_condition_empty_then_success_recheck_loads_instrument -- --nocapture`
- `pre-commit run --files crates/adapters/polymarket/src/data/dispatch.rs crates/adapters/polymarket/src/data/instruments.rs crates/adapters/polymarket/src/data/lifecycle.rs crates/adapters/polymarket/src/data/mod.rs crates/adapters/polymarket/src/data/requests.rs crates/adapters/polymarket/src/data/subscriptions.rs crates/adapters/polymarket/src/data_runtime.rs crates/adapters/polymarket/test_data/gamma_market_live.json crates/adapters/polymarket/tests/data_client.rs`

The new regression test here is `retire_expired_local_instruments_retires_watchlisted_runtime_state_once`, which exercises two retirement sweeps on the same retained expired instrument and asserts only the first sweep performs WS/runtime teardown.

cjdsellers:
@graceyangfan looks like an issue with pre-commit when you get a chance

graceyangfan:
Addressed the `pre-commit` failure.

The issue was an unfulfilled `#[expect(clippy::large_enum_variant)]` in `crates/common/src/messages/mod.rs` under the current toolchain/lint settings. I switched it to `#[allow(clippy::large_enum_variant)]`, which keeps the intended local suppression without tripping `-D unfulfilled-lint-expectations`.

Re-validated with:
- `pre-commit run cargo-clippy-network-turmoil-non-linux --all-files`

Both pass locally now.

cjdsellers (approved):
Thanks for the follow-ups @graceyangfan 👌

PR #4274, opened 2026-06-14
Fix Kraken request_bars to keep the most recent bars on count-only requests

### LINKED ISSUE #4254: [Kraken] request_bars with limit keeps the oldest bars of the OHLC page, so count-only requests return stale history
### Bug description

`KrakenSpotHttpClient::request_bars` (and the futures equivalent) iterates Kraken's OHLC response oldest-first and returns early once `limit` is reached, so the `limit` truncation keeps the **oldest** `limit` bars of the page:

```rust
bars.push(bar);

if let Some(limit_count) = limit
    && bars.len() >= limit_count as usize
{
    return Ok(bars);
}
```

Kraken's public spot `/0/public/OHLC` ignores any count parameter and always returns its full ~720-candle page (ascending). So a count-anchored request — `RequestBars { start: None, end: None, limit: Some(500) }` on a 1-minute bar type — returns bars covering `now − 12h` to `now − 3.7h`: the most recent 220 candles are silently discarded. A live strategy that warms up indicators with "the most recent 500 bars" instead gets a window ending hours in the past, with no error or warning.

Everywhere else a count-only history request means "the most recent N bars" (Binance/Bybit kline endpoints with only `limit`, and the long-standing `request_bars(..., limit)` convention in the Python API), so this is surprising and easy to ship without noticing — the response looks plausible (correct count, correct shape) and only the timestamps reveal it.

### Observed

Live run against Kraken spot, 2026-06-12, `ETH/USD.KRAKEN-1-MINUTE-LAST-EXTERNAL`:

- Request: `RequestBars { start: None, end: None, limit: Some(500) }` sent at `06:29:33Z`
- Response: 500 bars, first `ts_event = 2026-06-11T18:29:00Z`, last ≈ `2026-06-12T02:49:00Z` — ends 3h40m before the request.

### Expected

With `start: None`, `limit: N` should return the **most recent** N bars (i.e. keep the tail of the ascending page, not the head).

### Affected code

- `crates/adapters/kraken/src/http/spot/client.rs` — `request_bars`, early return inside the ascending candle loop
- `crates/adapters/kraken/src/http/futures/client.rs` — `request_bars`, identical pattern
- Present in `0.58.0` (crates.io) and unchanged at current `develop` HEAD.

### Workaround

Always pass `start` (e.g. `now − limit × interval`) so the page begins at the window start and the oldest-first truncation is harmless.

### Environment

Rust crates from crates.io `0.58.0` (no Python), `LiveNode` + `KrakenSpotDataClient` (keyless), Linux.

folknor:
## Summary

Fixes #4254.

Kraken's OHLC endpoints ignore any count parameter and always return their full page **oldest-first** (ascending). `request_bars` truncated to `limit` by returning early inside the ascending loop, so it kept the **oldest** `limit` bars of the page.

A count-only request — `RequestBars { start: None, end: None, limit: Some(N) }` — therefore returned a window ending hours in the past instead of the most recent N bars, silently (correct count and shape; only the timestamps reveal it). Everywhere else a count-only history request means "the most recent N bars" (Binance/Bybit kline endpoints, and the Python `request_bars(..., limit)` convention).

## Changes

- Add a shared `apply_bar_limit` helper in `http/mod.rs`:
  - `start` set → keep the oldest `limit` bars from that anchor (head of the page), so an explicit window is unaffected.
  - `start` is `None` → keep the **tail** of the ascending page, so a count-only request returns the most recent N bars.
- Apply it in both `http/spot/client.rs` and `http/futures/client.rs` `request_bars`, replacing the in-loop early return.

## Tests

- Added unit tests for `apply_bar_limit` covering: no limit, count-only (keeps most recent), `start`-anchored (keeps oldest), and fewer bars than the limit.
- `cargo test -p nautilus-kraken --lib http::tests` passes.

### DIFF
--- a/crates/adapters/kraken/src/http/futures/client.rs
+++ b/crates/adapters/kraken/src/http/futures/client.rs
@@ -1394,18 +1394,16 @@ impl KrakenFuturesHttpClient {
                         continue;
                     }
                     bars.push(bar);
-
-                    if let Some(limit_count) = limit
-                        && bars.len() >= limit_count as usize
-                    {
-                        return Ok(bars);
-                    }
                 }
                 Err(e) => {
                     log::warn!("Failed to parse futures bar: {e}");
                 }
             }
         }
+
+        // Kraken returns the page oldest-first; keep the most recent `limit`
+        // bars for count-only requests rather than the oldest (issue #4254).
+        crate::http::apply_bar_limit(&mut bars, start, limit);
 
         Ok(bars)
     }
--- a/crates/adapters/kraken/src/http/mod.rs
+++ b/crates/adapters/kraken/src/http/mod.rs
@@ -20,10 +20,38 @@
 //! - [`spot`]: Kraken Spot REST API
 //! - [`futures`]: Kraken Futures REST API
 
+use chrono::{DateTime, Utc};
+use nautilus_model::data::Bar;
+
 pub mod error;
 pub mod futures;
 pub mod models;
 pub mod spot;
+
+/// Applies a `limit` to OHLC `bars` returned oldest-first by Kraken.
+///
+/// Kraken's OHLC endpoints ignore any count parameter and always return their
+/// full page in ascending (oldest-first) order. When the caller anchors the
+/// window with `start`, the oldest `limit` bars from that anchor are the
+/// intended result, so the head of the page is kept. Otherwise a count-only
+/// request (`start` is `None`) means "the most recent `limit` bars", so the
+/// tail is kept rather than the head (see issue #4254).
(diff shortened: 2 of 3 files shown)


cjdsellers (approved):
Hi @folknor,

Thanks for the PR and the clear write-up. I reviewed the change and it looks correct: count-only requests now keep the most recent N bars, while `start`-anchored requests still keep the oldest N from the window. Safe to merge from my side.

`request_trades` carries the same count-only pattern in both clients, so I'll follow up separately to fix that path too.

PR #4275, opened 2026-06-14
Add Kraken spot WebSocket idle timeout to detect dead connections

### LINKED ISSUE #4255: [Kraken] acked spot WS v2 ohlc subscription can deliver no data, and the adapter cannot detect it
### Bug description

Intermittently, Kraken's spot WS v2 accepts an `ohlc` subscription (ack with `success: true`) and then never delivers any data on the connection — no ohlc updates for many minutes on a pair that is demonstrably trading. The connection stays open (no close frame, no transport error, no reconnect), so from the adapter's point of view everything is healthy: subscription confirmed, socket up — and the strategy silently receives no bars, indefinitely.

The adapter has no way to notice: `KrakenSpotWebSocketClient::connect` hardcodes `idle_timeout_ms: None` in its `WebSocketConfig`, and there is no per-subscription liveness check.

### Observed

2026-06-12, ~06:03–06:18 UTC, `ETH/USD` spot, interval 1, keyless data client, `wss://ws.kraken.com/v2`:

- Two consecutive `LiveNode` runs subscribed and got `Subscription confirmed: channel=Ohlc` — then zero ohlc messages for 10 and 4 minutes respectively (the pair traded continuously throughout; ~5–20 trades/minute confirmed via REST `Trades`).
- During that same window, a separate raw WebSocket connection from the same host sent the byte-identical subscribe request (`{"method":"subscribe","params":{"channel":"ohlc","symbol":["ETH/USD"],"interval":1,"snapshot":false}}`) and received the ack, 1/s `heartbeat` messages, and an ohlc update **within one second**.
- Runs started after ~06:25 behaved normally on the first try (ack → heartbeats → ohlc updates within seconds, verified down to raw frames with trace logging). Same binary, same params.

So the condition is per-connection and intermittent — it looks like a load-balanced backend accepting the subscription without attaching the fan-out.

### Expected

The client should be able to detect and recover from a connection that stops delivering data. Kraken sends `heartbeat` channel messages once per second whenever at least one subscription is active, so a conservative idle timeout (heartbeats are text frames and reset `last_data_time` in the read loop) would cleanly distinguish "quiet market" from "dead connection" and trigger the existing reconnect + resubscribe path.

### Suggested fix

Set (or expose on `KrakenDataClientConfig`) `idle_timeout_ms` for the spot v2 WS — e.g. 10s, an order of magnitude above the 1/s heartbeat cadence. If heartbeats can also go missing independently of channel data (we could not determine whether heartbeats flowed on the dead connections — they log at trace), a per-subscription data watchdog would be the stronger complement.

### Affected code

- `crates/adapters/kraken/src/websocket/spot_v2/client.rs` — `connect()`, `idle_timeout_ms: None`
- Present in `0.58.0`, unchanged at current `develop` HEAD.

### Environment

Rust crates from crates.io `0.58.0` (no Python), `LiveNode` + `KrakenSpotDataClient` (keyless), Linux.

folknor:
## Summary

Fixes #4255.

`KrakenSpotWebSocketClient::connect` hardcoded `idle_timeout_ms: None`. Intermittently, Kraken's spot WS v2 accepts a subscription (acks with `success: true`) and then never delivers data on that connection — the socket stays open (no close frame, no transport error, no reconnect), so nothing detects it and the strategy silently receives no data indefinitely.

## Changes

- Add `ws_idle_timeout_ms` to `KrakenDataClientConfig` (default `10_000`) and wire it into the spot v2 `WebSocketConfig`. `0` disables it.
- Thread the field through the PyO3 config constructor and `.pyi` stub.
- The authenticated execution WS opts out (`0`) to preserve its current behavior — this issue concerns the public data WS.

## Why 10s

The network read loop resets its idle timer only on **application** frames (text/binary), not ping/pong. Kraken sends a `heartbeat` text frame **once per second** while at least one subscription is active, so a live subscribed connection resets the timer well within 10s, while a silent (acked-but-no-fan-out) connection trips it and runs the existing reconnect + resubscribe path.

The default is deliberately shorter than the client's keepalive interval: the keepalive `ping` is answered with a `pong` *text* frame that also resets the timer (~every `heartbeat_interval_secs`, default 30s), so a longer timeout would be masked by the keepalive and never fire. Relying on the 1/s heartbeats assumes the connection carries at least one subscription; this trade-off is documented on the field, and a connection deliberately held open without any subscription can disable it (`0`) or raise it above `heartbeat_interval_secs`.

## Tests

- Added config unit tests (default value and `0` override).
- `cargo test -p nautilus-kraken --lib config::tests` passes; `cargo check -p nautilus-kraken --features python` is clean.

### DIFF
--- a/crates/adapters/kraken/src/config.rs
+++ b/crates/adapters/kraken/src/config.rs
@@ -59,6 +59,27 @@ pub struct KrakenDataClientConfig {
     pub timeout_secs: u64,
     #[builder(default = 30)]
     pub heartbeat_interval_secs: u64,
+    /// Idle timeout (milliseconds) for the spot v2 WebSocket.
+    ///
+    /// If no application data (any text or binary frame) is received within this
+    /// window, the connection is treated as dead and the client reconnects and
+    /// resubscribes. This recovers from a backend that acknowledges a
+    /// subscription but never attaches the data fan-out: the socket stays open
+    /// with no close frame or transport error, so nothing else detects it.
+    ///
+    /// Kraken sends a `heartbeat` text frame once per second while at least one
+    /// subscription is active, so a live subscribed connection resets this timer
+    /// well within the window. Note the client's keepalive `ping` is answered
+    /// with a `pong` *text* frame, which also resets the timer roughly every
+    /// `heartbeat_interval_secs`; the default below is therefore kept short
+    /// enough to rely on the 1/s heartbeats rather than the keepalive, which
+    /// assumes the connection carries at least one subscription. A connection
+    /// held open without any subscription should disable this (`0`) or raise it
+    /// above `heartbeat_interval_secs`.
+    ///
+    /// `0` disables the idle timeout.
+    #[builder(default = 10_000)]
+    pub ws_idle_timeout_ms: u64,
     pub max_requests_per_second: Option<u32>,
     #[builder(default)]
     pub transport_backend: TransportBackend,
@@ -281,6 +302,19 @@ validate_l3_checksum = false
         assert_eq!(config.environment, KrakenEnvironment::Live);
         assert_eq!(config.timeout_secs, 45);
         assert!(!config.validate_l3_checksum);
+    }
+
+    #[rstest]
+    fn test_data_config_ws_idle_timeout_default() {
+        let config = KrakenDataClientConfig::default();
+        assert_eq!(config.ws_idle_timeout_ms, 10_000);
+    }
+
+    #[rstest]
+    fn test_data_config_ws_idle_timeout_override() {
+        let config: KrakenDataClientConfig = toml::from_str("ws_idle_timeout_ms = 0").unwrap();
+
+        assert_eq!(config.ws_idle_timeout_ms, 0);
     }
 
     #[rstest]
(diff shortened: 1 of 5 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR, and for the clear write-up.

I confirmed the timing assumption against Kraken's v2 docs: heartbeats are generated on subscription to any channel, so a live subscribed connection resets the idle timer roughly once per second, well inside the 10s window. The approach is sound.

One thing I wanted to confirm before merging: in the stuck state from #4255, did the 1/s `heartbeat` frames stop too, or only the book/trade fan-out? Since the timer resets on any text frame, the timeout only fires if the heartbeats stop as well. Your write-up has the connection going fully silent, so I expect they do, but I'd like to be sure you observed that and not just the missing data.

If you can confirm, this is otherwise good to merge.

folknor:
Yes, the connection went completely silent in that instance. I did not perform a root-cause analysis of why, because it was difficult to reproduce.

It has not happened since - it only happened once thus far. But that's enough for me to want a guard in place.

This was not a problem with my network equipment or hardware, because other connections were active all over the place. Nothing weird in my system logs either.

Thank you

PR #4287, opened 2026-06-18
Harden Polymarket execution lookup retention

graceyangfan:
## Summary

Related to #4253.

This PR addresses the execution-client portion of the inactive-instrument retention chain described in #4253.

Specifically, it hardens Polymarket execution-side lookup retention so expired instruments are retained only while live execution flows still need them, and are pruned once both order and position state have reached terminal conditions.

In scope:
- `shared_token_instruments`
- `neg_risk_index`
- execution-client lifecycle for lookup-retention subscriptions

Out of scope:
- sandbox / matchengine behavior
- global cache pruning
- data-client retention redesign
- periodic execution sweep timers
- reconnect-time stale lookup sweeping

## Problem

As described in #4253, Polymarket continuously creates short-dated markets, so long-running sessions can accumulate inactive instrument state across multiple layers.

On the execution side, the relevant retained owners are:
- `shared_token_instruments`
- `neg_risk_index`

Before this change, those lookup maps were effectively append-only within a connection lifetime:
- `load_instruments_from_cache()` populated execution lookup state on connect
- `on_instrument()` kept inserting newly seen instruments
- expired instruments were not later re-evaluated once execution-side order/position state became terminal

That meant expired markets could remain retained in execution-local lookup state even after execution flows no longer needed them.

## What This PR Changes

This PR adds a single execution-local retention policy:
- retain active instruments
- retain expired instruments while an open order or open position still exists
- prune expired instruments once neither open orders nor open positions remain

The policy is applied through execution-side lifecycle events rather than connect-time pruning:

1. `load_instruments_from_cache()`
   - on connect / reconnect, preload execution lookup state from cache
   - do not prune here, so startup reconciliation and late venue state cannot lose required lookup entries prematurely

2. `on_instrument_update()`
   - newly seen instruments are upserted into execution lookup state
   - do not prune here, so mid-session updates do not remove lookup state while venue events may still arrive

3. terminal execution events
   - while connected, the execution client subscribes to:
     - `PositionClosed`
     - terminal order events (`Canceled`, `Expired`, `Rejected`, `Filled`)
   - each event re-evaluates lookup retention for that single instrument
   - expired lookup state is pruned only when no open order and no open position remain

This also makes the subscription lifecycle symmetric:
- install on connect
- remove on disconnect
- remove on stop
- remove on connect-failure teardown
- allow clean reinstall after disconnect cleanup

## Why This Shape

This PR intentionally keeps the PR2 scope narrow.

It does not try to solve the entire retention chain from #4253 in one patch. Instead, it closes the execution-client part of that chain without introducing broader lifecycle changes.

In particular, this PR intentionally does not:
- add a periodic execution sweep timer
- expand data-client-local ownership
- introduce sandbox / matchengine cleanup changes
- change global cache purge behavior
- promise expiry-time shrink for never-traded instruments

The intended behavior here is convergence, not immediate expiry-time shrink.

Execution lookup state is no longer permanently append-only for traded instruments:
- it is preloaded on connect / reconnect
- it converges after terminal order / position events for expired instruments

This means:
- traded expired instruments can contract once execution state is fully terminal
- expired instruments that were only loaded / discovered and never reach execution terminal events are not pruned by this PR

That is the execution-side piece of the larger inactive-instrument cleanup plan in #4253.

## Testing

- `cargo test -p nautilus-polymarket execution::lifecycle --lib`
- `pre-commit run --files crates/adapters/polymarket/src/execution/lifecycle.rs crates/adapters/polymarket/src/execution/mod.rs`

Added lifecycle coverage for:
- preloading expired lookup state during cache load without connect-time pruning
- retaining expired lookup state while an open order remains
- retaining expired lookup state while an open position remains
- pruning expired lookup state after terminal order events
- retaining expired lookup state after `Filled` when the position remains open
- pruning expired lookup state after `PositionClosed`
- ignoring other-venue position events
- reinstalling order / position event subscriptions after disconnect cleanup

### DIFF
--- a/crates/adapters/polymarket/src/execution/lifecycle.rs
+++ b/crates/adapters/polymarket/src/execution/lifecycle.rs
@@ -18,17 +18,16 @@ use std::{
     time::{Duration, Instant},
 };
 
-use ahash::{AHashMap, AHashSet};
+use ahash::AHashMap;
 use anyhow::Context;
 use nautilus_common::{
-    cache::Cache,
     live::{runner::get_exec_event_sender, runtime::get_runtime},
     msgbus::{self, TypedHandler},
 };
-use nautilus_core::{MUTEX_POISONED, UnixNanos, collections::AtomicMap, time::AtomicTime};
+use nautilus_core::{MUTEX_POISONED, collections::AtomicMap, time::AtomicTime};
 use nautilus_model::{
-    events::PositionEvent,
-    identifiers::{AccountId, InstrumentId},
+    events::{OrderEventAny, PositionEvent},
+    identifiers::InstrumentId,
     instruments::{Instrument, InstrumentAny},
 };
 use ustr::Ustr;
@@ -43,6 +42,39 @@ use crate::{
 };
 
 impl PolymarketExecutionClient {
+    fn ensure_order_event_subscription(&mut self) {
+        if self.order_event_handler.is_some() {
+            return;
+        }
+
+        let core = self.core.clone();
+        let clock = self.clock;
+        let shared_token_instruments = self.shared_token_instruments.clone();
+        let neg_risk_index = self.neg_risk_index.clone();
+        let handler = TypedHandler::from(move |event: &OrderEventAny| {
+            if !is_terminal_order_event(event) || event.instrument_id().venue != core.venue {
+                return;
+            }
+
+            sync_execution_lookup_for_instrument(
+                &core,
+                clock,
+                &shared_token_instruments,
+                &neg_risk_index,
+                event.instrument_id(),
+            );
+        });
+
+        msgbus::subscribe_order_events("events.order.*".into(), handler.clone(), Some(10));
+        self.order_event_handler = Some(handler);
+    }
+
+    fn clear_order_event_subscription(&mut self) {
+        if let Some(handler) = self.order_event_handler.take() {
+            msgbus::unsubscribe_order_events("events.order.*".into(), &handler);
+        }
+    }
+
     fn ensure_position_event_subscription(&mut self) {
         if self.position_event_handler.is_some() {
             return;
@@ -246,26 +278,6 @@ impl PolymarketExecutionClient {
         neg_risk_index.get(instrument_id).copied().unwrap_or(false)
     }
 
-    fn should_retain_execution_lookup(
-        &self,
-        cache: &Cache,
-        instrument: &InstrumentAny,
-        now_ns: UnixNanos,
-        account_id: AccountId,
-    ) -> bool {
-        if !crate::filters::is_expired(instrument, now_ns) {
-            return true;
-        }
-
-        cache.has_positions_open(
-            Some(&self.core.venue),
-            Some(&instrument.id()),
-            None,
-            Some(&account_id),
-            None,
-        )
-    }
-
     fn upsert_execution_lookup(&self, instrument: &InstrumentAny) {
         upsert_execution_lookup(
             &self.shared_token_instruments,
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @graceyangfan,

Thanks for the PR, and for keeping the PR2 scope tight. The subscription lifecycle, the `events.position.*` handler, and the symmetry with the data client all look right.

My main concern is the retention check itself. `shared_token_instruments` is the map the live user-channel dispatch uses to resolve an incoming fill's `asset_id` back to an instrument, and a miss there just logs `Unknown asset_id` and drops the message. Keying retention on `cache.has_positions_open(...)` prunes a token in two states where the WS layer still needs it: at connect, `load_instruments_from_cache()` runs before startup reconciliation, so a venue-known open position on an expired instrument isn't in the cache yet; and a resting order that hasn't filled never populates `positions_open` at all. In both cases the expired instrument gets removed, nothing re-adds it, and the closing or settlement fill is dropped. Before this change the map was append-only, so that fill always resolved.

A few smaller things in the same area:

- `on_instrument_update` now prunes on the per-update path too, so the same drop can happen mid-session the moment an instrument update arrives for an expired token that only has a resting order.
- `remove_execution_lookup` always clears `neg_risk_index`, so a later `get_neg_risk` on a pruned instrument silently defaults to `false` and would sign against the wrong exchange. Narrow trigger, but worth a guard.
- `prune_execution_lookup_not_in_retained` removes one entry at a time, and each `AtomicMap` remove clones the whole map, so the prune is quadratic on exactly the large, long-running sessions this targets. Rebuilding the retained map once and storing it in a single swap would avoid that.
- The lifecycle test asserts `position_event_handler.is_none()` right after `start_client()`, but `start_client` never installs the subscription, so that path isn't actually covered. The `PositionClosed` and venue filters in the handler aren't exercised either, since the test calls `sync_execution_lookup_for_instrument` directly rather than publishing through the bus.

I reviewed against the diff and didn't run the suite locally. The cleanest fix is probably to also retain while an open order exists, not just an open position, and to hold the connect-time prune until after reconciliation. Happy to go into more detail on any of these.

graceyangfan:
Thanks for the detailed review. I pushed a follow-up update (head is now `48b21841a3`) that addresses the main retention/lifecycle concerns.

What changed:
- `sync_execution_lookup_for_instrument(...)` now retains expired instruments while either an open order or an open position exists, instead of checking only `has_positions_open(...)`.
- `load_instruments_from_cache()` no longer prunes on connect/reconnect; it only preloads / upserts execution lookup state.
- `on_instrument_update()` also no longer prunes on the per-update path; it only upserts.
- The old bulk prune path is gone. Lookup contraction is now event-driven on a single-instrument basis via:
  - `PositionClosed`
  - terminal order events (`Canceled`, `Expired`, `Rejected`, `Filled`)
- Added an execution-side `events.order.*` subscription with symmetric install/teardown alongside the existing position-event subscription.

Testing updates:
- lifecycle tests now publish through msgbus rather than only calling the sync helper directly
- added coverage for:
  - retaining expired lookup state while an open order remains
  - pruning after terminal order events
  - pruning after `PositionClosed`
  - ignoring other-venue position events
  - retaining after `Filled` when the position is still open
  - reinstalling order / position event subscriptions after disconnect cleanup

I also updated the PR description to match the current behavior: this version no longer claims connect/reconnect-time shrink. The convergence is now event-driven rather than connect-time prune-based.

I did not try to solve the `neg_risk_index` fallback concern in this PR, since that started to push the scope beyond the execution-retention hardening we were aiming for here. If you think that needs to block this PR, I’m happy to follow up on it directly; otherwise I’d prefer to keep it as a focused follow-up from the remaining #4253 chain.

cjdsellers (approved):
Thank you for the follow-ups @graceyangfan 👌

Issue #3827, opened 2026-04-09
[Hyperliquid] Exchange sends a Cancel and new message on Order modification, which isn't handled


P1YU5H-50N1:
# Bug Report

Use this template only for issues that fit the **Bug** definition.

| Term                          | Definition |
|-------------------------------|------------|
| **Bug**                       | Exchange sends a Cancel and new message on Order modification, which isn't handled. Nautilus assumes the order is canceled while the order remains open at exchange |
| **Expectation&nbsp;mismatch** |Nautilus should keep order as open instead of cancelling it to reflect state at exchange |

**Note:**

- Submitting this issue automatically applies the `bug` label.
- `bug`-labeled issues are triaged with higher priority because they require corrective implementation work.
- **Expectation mismatches** and design-level concerns should be opened as [Discussions](https://github.com/nautechsystems/nautilus_trader/discussions), or RFCs instead, where they can be validated and discussed to consensus before any work is scheduled.
- The absence of a feature is typically not an expectation mismatch, and should be filed as an enhancement request.

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates.
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

Checking a recent development wheel can save time because the issue may already have been fixed.
You can install a development wheel by running:

```bash
pip install -U nautilus_trader --pre --index-url https://packages.nautechsystems.io/simple
```

See the [development-wheels](https://github.com/nautechsystems/nautilus_trader#development-wheels) section for more details.

### Expected Behavior

Nautilus should keep order as open instead of cancelling it to reflect state at exchange upon order modified instead of marking order as cancelled

### Actual Behavior

Exchange sends a Cancel and new message on Order modification, which isn't handled. Nautilus assumes the order is canceled while the order remains open at exchange. 

### Steps to Reproduce the Problem

1.Setup a trading node to modify order on hyperliquid

### Code Snippets or Logs

```2026-04-09T08:00:52.944988000Z [INFO] LIGHTER-NODE-001.ExecClient-HYPERLIQUID: Modify ClientOrderId('O-20260409-080047-001-000-1') VenueOrderId('375273671786')
[...]

PR #4288, opened 2026-06-18
Fix reconciliation canceling live orders on cancel-replace venues

filipmacek:
# Pull Request

### Overview

On venues that implement `modify_order` as atomic cancel-replace (Hyperliquid, dYdX), the HTTP reconciliation path can synthesize a phantom `OrderCanceled` for an order that is still resting at the venue. Strategies react by clearing tracking and replenishing — leaving two live orders on the venue for the same `client_order_id`.

Two complementary guards: stop probing confirmed-modified orders (root cause), and refuse to synthesize a terminal event while mid-command or while the report references a previously-promoted `venue_order_id` (backstop).

### Problem

Observed in production: `OrderCanceled` events delivered to strategies for orders that subsequently traded as makers on the same `venue_order_id`. Four cancel-then-fill collisions in a 12-minute window on one instrument, each causing the strategy to submit a replacement — silent double exposure.

The phantom requires three conditions to align:

1. **Cancel-replace modify venue** (single-voi venues are never affected).
2. **`client_order_id` quiet ≥ `inflight_check_threshold`** (default 5 s) after its last command.
3. **Venue HTTP snapshot returns the cancel-half of a recent modify-replace** (eventual consistency or a transient flake).

Active liquid instruments rarely hit gate 2. The phantoms concentrate on illiquid perpetuals during periods of venue HTTP stress.

### Root cause

Two parallel paths deliver venue state: the adapter's lifecycle stream (WS dispatch — source of truth, has GH-3827 modify-replace awareness) and a periodic HTTP reconciliation (safety net). The reconciliation path was written under a single-voi mental model and **does not have the equivalent guard**.

Two mirror-image failure modes both produce a phantom:

- **Case A — cache-advanced.** WS dispatch already promoted `cache.venue_order_id` to NEW; HTTP report arrives referencing OLD. Distinguishable by `report.voi != cache.voi`.
- **Case B — cache-stale.** WS lag/drop means the promotion never arrived; HTTP report arrives referencing OLD, which matches the stale cache. Distinguishable only by `order.status() == PendingUpdate`.

**Why HTTP fires at all on healthy orders:** `LiveNode` clears the inflight reconciliation tracking on terminal events but not on `Updated`, `ModifyRejected`, or `CancelRejected`. After a confirmed modify the `client_order_id` lingers in `inflight_checks`; the next quiet window past threshold triggers a `QueryOrder`. 

### Commit 1 — crates/live/src/node.rs

Add `Updated`, `ModifyRejected`, `CancelRejected` to the match arm that clears `clear_recon_tracking`. Once the lifecycle stream confirms the command outcome, periodic reconciliation probes must stop.

### Commit 2 — crates/execution/src/reconciliation/orders.rs

Two guards in the `Canceled` branch of `reconcile_order_report`:

1. **Defer** when `order.status()` is `PendingUpdate` or `PendingCancel`. The adapter's lifecycle stream must confirm the command outcome before a terminal event can be applied.
2. **Suppress** when `report.venue_order_id` differs from the cached one and is present in `order.venue_order_ids()` — i.e., a previously-promoted leg whose successor is still live. Mirrors the GH-3827 Rule 1 suppression already in the HL WS dispatch.

Function docstring extended to list both new `None`-returning conditions.

Both new log lines are `info` — these events are rare under normal venue conditions but operationally meaningful; `debug` hides them by default, `warn` is wrong (nothing is broken).

### Non-regression

Walked every realistic scenario in which a `Canceled` report reaches `reconcile_order_report`:

| scenario | result |
|---|---|
| Real cancel of a live order (post-only crossed, margin, user UI, …) | forwarded |
| `Strategy::cancel_order` → WS confirms → HTTP echoes | forwarded (engine de-dups) |
| External venue cancel of a `venue_order_id` we never tracked | forwarded |
| Single-voi venue, ordinary cancel | forwarded (history check trivially never matches) |
| Brand-new submission, no `Accepted` yet | forwarded |
| **Modify-replace race, cache advanced (Case A)** | **suppressed** |
| **Modify-replace race, cache stale (Case B)** | **deferred** |
| Multi-step modify chain, report on an old leg | suppressed |

**Single-`venue_order_id` venues unaffected.** The suppress guard's `in_history && != current` clause cannot be satisfied. The defer guard does fire briefly but introduces at most one reconciliation cycle of latency on cancels arriving inside the in-flight window.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/crates/execution/src/reconciliation/orders.rs
+++ b/crates/execution/src/reconciliation/orders.rs
@@ -116,6 +116,14 @@ pub fn generate_reconciliation_order_events(
 /// Returns `None` for pending venue states (`PendingUpdate`, `PendingCancel`)
 /// regardless of local state, since an unconfirmed amend or cancel must not
 /// drive any local mutation until the venue surfaces a confirmed status.
+///
+/// Returns `None` for `Canceled` reports when:
+/// - the local order is mid-command (`PendingUpdate`, `PendingCancel`) — the
+///   adapter's primary lifecycle stream (WebSocket or HTTP) must confirm the
+///   outcome before a terminal event can be applied;
+/// - the report references a previously-promoted `venue_order_id` whose
+///   successor is still live in the cache (the cancel-half of a cancel-replace
+///   modify on venues like Hyperliquid).
 #[must_use]
 pub fn reconcile_order_report(
     order: &OrderAny,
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hi @filipmacek,

Thanks for this, good findings and the reasoning holds up 👌. I traced both guards and the `node.rs` tracking change and they do what you describe.

I'll follow up after it lands with some tests for the two new branches, and take a closer look at the cache-stale path, where the inflight backstop could still synthesize a cancel if the lifecycle stream never confirms.

PR #4289, opened 2026-06-18
Fix Binance futures leverage initialization aborting execution client connect

YeeTsai:

## Summary

`BinanceFuturesExecutionClient._update_account_state` could abort `_connect` during leverage initialization in two ways. Both are now best-effort, so a leverage-init issue never prevents the execution client from connecting. This surfaced when connecting a futures testnet/demo account whose `symbolConfig` lists every venue symbol with `leverage=0` for the untraded ones.

**1. A single `set_leverage` rejection tears down the whole `TaskGroup`.** The per-symbol `set_leverage` calls run inside a `TaskGroup`; any failure propagates as an unhandled `ExceptionGroup`. A symbol whose venue leverage cap is below the requested value (for example `-4028 leverage too large`) aborts `_connect` and leaves the client unable to connect. Each `set_leverage` is now caught per symbol (logged as a warning) and the connect proceeds.

**2. `symbolConfig` with `leverage=0` raises in `MarginAccount.set_leverage`.** The leverage mirror calls `account.set_leverage(instrument_id, 0)` for symbols the account has not configured. Binance testnet/demo returns `leverage=0` for every untraded symbol (hundreds of them), and `MarginAccount.set_leverage` asserts `leverage >= 1`, raising `ValueError`. The `except KeyError` was dead code because `_get_cached_instrument_id` fabricates instrument IDs rather than raising. Entries with `leverage < 1` are now skipped before mirroring.

## Related Issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)

N/A.

## Documentation

- [x] N/A — no documentation changes.

## Release notes

- [ ] Happy to add a `RELEASES.md` entry if preferred.

## Testing

Added two regression tests in `tests/integration_tests/adapters/binance/test_execution_futures.py` exercising `_update_account_state`:
- a `set_leverage` `BinanceError` is non-fatal (warns and continues) — fails before the fix via `TaskGroup` propagation;
- `symbolConfig` `leverage < 1` entries are skipped (not mirrored into account leverage state) — fails before the fix when `set_leverage(id, 0)` hits the `leverage >= 1` assertion.

### DIFF
--- a/nautilus_trader/adapters/binance/futures/execution.py
+++ b/nautilus_trader/adapters/binance/futures/execution.py
@@ -25,6 +25,7 @@ from nautilus_trader.adapters.binance.common.enums import BinanceAccountType
 from nautilus_trader.adapters.binance.common.enums import BinanceEnvironment
 from nautilus_trader.adapters.binance.common.enums import BinanceErrorCode
 from nautilus_trader.adapters.binance.common.enums import BinanceExecutionType
+from nautilus_trader.adapters.binance.common.symbol import BinanceSymbol
 from nautilus_trader.adapters.binance.config import BinanceExecClientConfig
 from nautilus_trader.adapters.binance.execution import BinanceCommonExecutionClient
 from nautilus_trader.adapters.binance.futures.enums import BinanceFuturesEnumParser
@@ -167,6 +168,9 @@ class BinanceFuturesExecutionClient(BinanceCommonExecutionClient):
             self._log.info("TRADE_LITE events will be used", LogColor.BLUE)
 
         self._leverages = config.futures_leverages
+        # Symbols already warned about a failed best-effort leverage set, so the
+        # warning is emitted at most once (this runs on every account query).
+        self._set_leverage_warned: set[str] = set()
         self._margin_types = config.futures_margin_types
 
         self._decoder_futures_user_msg = msgspec.json.Decoder(BinanceFuturesUserMsgData)
@@ -175,7 +179,7 @@ class BinanceFuturesExecutionClient(BinanceCommonExecutionClient):
         self._decoder_futures_trade_lite = msgspec.json.Decoder(BinanceFuturesTradeLiteMsg)
         self._decoder_futures_algo_update = msgspec.json.Decoder(BinanceFuturesAlgoUpdateMsg)
 
-    async def _update_account_state(self) -> None:
+    async def _update_account_state(self) -> None:  # noqa: C901 (too complex)
         account_info: BinanceFuturesAccountInfo = (
             await self._futures_http_account.query_futures_account_info(recv_window=str(5000))
         )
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @YeeTsai,

Thanks for the PR and the clear write-up.

The fix is the right shape: catching per-symbol inside the coroutine does stop the `TaskGroup` re-raising, and skipping `leverage < 1` correctly avoids the `MarginAccount.set_leverage` assertion.

**A few things to improve before it lands:**

- The `leverage < 1` regression test doesn't exercise the failure it guards. `stub.account` is a `MagicMock`, so `account.set_leverage(id, Decimal(0))` never reaches the real `leverage >= 1` assertion. Pre-fix the test fails only because the call count is 3 rather than 2, not because of the assertion the comment cites. Asserting against a real `MarginAccount` (and correcting that comment) would make it a genuine guard.
- The banner comment and the `→` in the `leverage=0` comment will trip pre-commit. We don't use separator/banner comments, and unicode arrows aren't allowed (use `->`).
- The catch is `BinanceError` only, so the "never prevents connect" goal holds for venue rejects but not for a transient transport error from `set_leverage`, which still aborts `_connect`. Worth a deliberate call on whether to broaden it.

Two minor nits, neither blocking: `_update_account_state` also runs on every `QueryAccount`, so the new per-symbol warning becomes recurring noise for capped or untraded symbols rather than a one-off at startup; and the `except KeyError` is now dead (`_get_cached_instrument_id` fabricates the id rather than raising), so it can go along with its comment.

I reviewed against the diff and didn't run the suite locally. Please let me know if any of these need more detail.

YeeTsai:
Thanks for the thorough review @cjdsellers - all addressed in the latest push:

1. **The zero-leverage test now uses a real `MarginAccount`** (`TestExecStubs.margin_account()` with real `InstrumentId`s), so it genuinely exercises the assertion: pre-fix it fails on `MarginAccount.set_leverage(<id>, Decimal(0))` raising `leverage was not >= 1`, not on a call count. Comment corrected accordingly.
2. **Removed the banner comment and the unicode arrow.** The new test code is ASCII-only now.
3. **Kept the catch scoped to `BinanceError` deliberately.** A transport error during `_connect` means the connection itself is unhealthy, so it should surface rather than be swallowed; only a venue reject is treated as best-effort. Added a comment making that explicit.
4. **The per-symbol warning is now emitted at most once** (via `_set_leverage_warned`), so the recurring `QueryAccount` path no longer re-warns for capped/untraded symbols.
5. On the now-dead `except KeyError`: I left it in place because the existing `test_leverage_initialization_from_symbol_config` mocks `_get_cached_instrument_id` to raise `KeyError` for an unloaded symbol and asserts it is skipped, so removing the handler breaks that test. I'm happy to drop both the handler and that mock-`KeyError` case (it exercises a path that can no longer occur, since `_get_cached_instrument_id` fabricates the id) if you'd prefer; I just kept this PR's scope minimal. Let me know.

cjdsellers (dismissed):
Thank you for the follow-ups @YeeTsai 👌

cjdsellers:
@YeeTsai I recommend a faster feedback cycle if you could run the pre-commit locally with:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

YeeTsai:
Thanks @cjdsellers, will do - I'll run `make pre-commit` (and `prek install` for the hook) going forward. The back-and-forth was me running the hooks piecemeal rather than the full suite; pre-commit is green now.

PR #4290, opened 2026-06-19
Shard dYdX Indexer WebSocket into a per-channel connection pool

filipmacek:
# Pull Request

## Summary

- The hosted dYdX Indexer caps each WebSocket connection at **32 subscriptions per channel** for `v4_trades`, `v4_candles`, `v4_orderbook`, and `v4_markets`. Strategies that watch more than 32 markets (e.g. our 119-instrument `VolumeRankingWss` candle subscribe) silently lose every subscription past 32.
- Wrap `DydxWebSocketClient` in a lazy-growing `Vec<ConnectionSlot>` pool. New subscriptions route to the first slot with capacity on the requested channel; if every slot is full and the cap isn't reached, an extra connection is opened up to `max_ws_connections` (default 8 → 256-market capacity per 32-limit channel).
- Pattern follows `BinanceFuturesWebSocketClient` — including its `connect_lock` race fix — but tracks capacity **per channel kind** (`[u16; 4]`) instead of a single flat stream count, matching dYdX's per-channel-per-connection model.
- Subaccount and block-height streams stay pinned to slot 0 (their 256-per-connection cap is unreachable in practice and keeps related fill/position events in a single ordered stream).
- Two new `DydxDataClientConfig` fields, `max_ws_connections` and `per_channel_subscription_limit`, let self-hosted Indexer deployments raise the defaults.

## Notes

- `connection_mode_atomic()` now reflects slot 0 only (documented). Use `is_connected()` for the pool-wide check.
- `disconnect()` tears down slots sequentially with a 2 s timeout each — fine for typical shutdown, could be parallelized later if needed.
- The pre-existing per-connection rate-limit quota (2 msg/s) is kept. Multiple slots × 2 msg/s stays under the 100 req / 10 s IP-level budget for `max_ws_connections ≤ 8`.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

### DIFF
--- a/crates/adapters/dydx/src/config.rs
+++ b/crates/adapters/dydx/src/config.rs
@@ -165,6 +165,14 @@ fn default_data_retry_delay_max_ms() -> u64 {
     5000
 }
 
+fn default_max_ws_connections() -> usize {
+    8
+}
+
+fn default_per_channel_subscription_limit() -> usize {
+    32
+}
+
 impl DydxAdapterConfig {
     /// Creates a config with URLs and chain ID resolved for the given network.
     ///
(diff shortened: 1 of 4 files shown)


cjdsellers (approved):
Thank you @filipmacek 👍

PR #4293, opened 2026-06-20
Finalize sandbox expired engine retention

graceyangfan:
Related to #4253

## Summary
- clean up sandbox expired instruments immediately once they are expired and have no open positions
- retain expired instruments with open positions until settlement completes, then retry cleanup on `PositionClosed`
- treat delayed order-terminal cache updates as non-owning in the sandbox expiry path via a narrow `purge_instrument_skip_order_guard(...)` cache helper
- add regression coverage across sandbox, cache, and execution-engine layers for expiry cleanup and position-close ordering

## Why
PR2 narrowed Polymarket adapter retention boundaries. PR3 closes the equivalent sandbox paper-trading lifecycle gap without introducing an extra `events.order.*` retry path.

In the sandbox expiry pipeline:
- `InstrumentClose` (or local expiry fallback) cancels resting orders and blocks new orders
- order state may still be converging asynchronously in the cache
- open positions remain the real owner until settlement closes them

So this PR draws the owner boundary explicitly:
- expired + no open position: release the matching engine and purge the instrument immediately
- expired + open position: keep the instrument alive until settlement, then retry cleanup on `PositionClosed`
- order state no longer blocks expired cleanup in the no-position path

## Design Notes
- sandbox cleanup now keys off factual instrument expiry (`expiration_ns <= now`) rather than matching-engine internal expiration-processing state
- `PositionClosed` retry remains guarded by the execution-engine contract that cache position state is updated before the event is published
- cache-level `purge_instrument_skip_order_guard(...)` is intentionally narrow and still preserves the position guard

## Tests
- `cargo test -p nautilus-common test_purge_instrument_skip_order_guard_allows_orders_open -- --nocapture`
- `cargo test -p nautilus-execution --test exec_engine test_process_closing_fill_updates_cache_before_publishing_position_closed -- --nocapture`
- `cargo test -p nautilus-sandbox --test execution -- --nocapture`
- `pre-commit run --files crates/adapters/sandbox/src/execution.rs crates/adapters/sandbox/tests/execution.rs crates/common/src/cache/mod.rs crates/common/src/cache/tests.rs crates/execution/tests/exec_engine.rs`

### DIFF
--- a/crates/adapters/sandbox/src/execution.rs
+++ b/crates/adapters/sandbox/src/execution.rs
@@ -303,7 +303,7 @@ impl SandboxInner {
         // with the already-initialized matching engine.
         if let Some(engine) = self.matching_engines.get_mut(&instrument_id) {
             engine.get_engine_mut().process_instrument_close(*close);
-            self.try_finalize_expired_instrument(instrument_id);
+            self.sync_expired_cleanup(instrument_id);
         } else {
             log::warn!(
                 "Ignoring instrument close for {instrument_id}: no existing matching engine",
@@ -311,12 +311,21 @@ impl SandboxInner {
         }
     }
 
-    fn try_finalize_expired_instrument(&mut self, instrument_id: InstrumentId) {
+    fn is_expired_now(&self, instrument_id: InstrumentId) -> bool {
         let Some(engine) = self.matching_engines.get(&instrument_id) else {
-            return;
+            return false;
         };
 
-        if !engine.get_engine().is_expiration_processed() {
+        let now_ns = self.clock.borrow().timestamp_ns();
+        engine
+            .get_engine()
+            .instrument
+            .expiration_ns()
+            .is_some_and(|ns| now_ns >= ns)
+    }
+
+    fn sync_expired_cleanup(&mut self, instrument_id: InstrumentId) {
+        if !self.is_expired_now(instrument_id) {
             return;
         }
 
@@ -333,20 +342,9 @@ impl SandboxInner {
         }
 
         self.matching_engines.remove(&instrument_id);
-
-        let has_open_orders = self.cache.borrow().has_orders_open(
-            Some(&self.config.venue),
-            Some(&instrument_id),
-            None,
-            None,
-            None,
-        );
-
-        if has_open_orders {
-            return;
-        }
-
-        self.cache.borrow_mut().purge_instrument(instrument_id);
+        self.cache
+            .borrow_mut()
+            .purge_instrument_skip_order_guard(instrument_id);
     }
 }
 
@@ -561,9 +559,11 @@ impl SandboxExecutionClient {
                     && position_closed.account_id == account_id
                     && let Some(inner_rc) = inner_weak.upgrade()
                 {
+                    // ExecutionEngine updates the cached position state before publishing
+                    // PositionClosed, so this retry observes the post-settlement cache view.
                     inner_rc
                         .borrow_mut()
-                        .try_finalize_expired_instrument(position_closed.instrument_id);
+                        .sync_expired_cleanup(position_closed.instrument_id);
                 }
             })
         };
@@ -961,6 +961,7 @@ impl ExecutionClient for SandboxExecutionClient {
             engine
                 .get_engine_mut()
                 .process_order(&mut order, account_id);
+            inner.sync_expired_cleanup(instrument_id);
         }
 
         Ok(())
@@ -1022,6 +1023,7 @@ impl ExecutionClient for SandboxExecutionClient {
                     engine
                         .get_engine_mut()
                         .process_order(&mut order_clone, account_id);
+                    inner.sync_expired_cleanup(instrument_id);
                 }
             }
         }
--- a/crates/adapters/sandbox/src/execution.rs
+++ b/crates/adapters/sandbox/src/execution.rs
@@ -346,6 +346,12 @@ impl SandboxInner {
             .borrow_mut()
             .purge_instrument_skip_order_guard(instrument_id);
     }
+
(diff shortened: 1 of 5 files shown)


cjdsellers:
Hi @graceyangfan,

Thanks for the PR, and for the clear design write-up. The owner-boundary model reads well, the `purge_instrument_skip_order_guard` split keeps the position guard intact, and the lifecycle tests cover the happy paths nicely. A few things to address before it lands, the first two being the ones I'd want fixed:

- `submit_order_list` runs `sync_expired_cleanup` per leg, so the first rejected leg on an expired, position-free instrument purges the instrument mid-loop. The remaining legs then find no instrument in the cache and are skipped after their `OrderSubmitted` was already dispatched, leaving them stuck in `SUBMITTED` with no terminal event. Hoisting the cleanup out of the per-leg loop (run once after the list) would avoid it. The single-order path is fine.

- The new `PositionClosed` handler does `inner.borrow_mut()` with no re-entrancy guard. When the sandbox runs without a registered exec-event sender, matching-engine fills dispatch synchronously through `msgbus::send_order_event` into the `ExecutionEngine`, which publishes `PositionClosed` on the same stack and re-enters the handler while `submit_order` / `process_instrument_close` still hold the `inner` borrow. Under the live runner (fills routed via the channel) it won't fire, so it's latent, but the matching engine already guards against this exact synchronous-dispatch case, so it's worth covering here too: defer the cleanup, or guard the handler.

- `purge_instrument_skip_order_guard` removes the `instrument_orders` index but leaves the orders open until their cancel events apply, so `has_orders_open(venue, instrument)` reports `false` for a still-open order in that window. It's an intentional trade-off, but worth calling out in the method doc.

- A few smaller things: the doc block moved onto the now-private `purge_instrument_inner`, leaving the public `purge_instrument` undocumented (and a `///` on a private item); `is_expired_now` re-derives the engine's own expiry check rather than reading `is_expiration_processed()`; and `position_closed_event` hand-writes the full `PositionClosed` literal where `PositionClosed::create` would do.

I reviewed against the diff and didn't run the suite locally. Happy to expand on any of these.

graceyangfan:
Thanks again for the careful review. Here is a single point-by-point response mapped to each item:

1. `submit_order_list` cleaning up per leg
Addressed.

`submit_order_list` no longer runs expired cleanup inside the per-leg loop. It now:
- processes the full list first,
- records the touched instrument IDs, and
- runs expired cleanup once after the loop.

That prevents the first rejected leg on an expired, position-free instrument from purging the instrument before later legs finish their own terminal path. I also added `test_submit_order_list_keeps_processing_all_expired_legs_before_cleanup` to lock in that later legs still emit terminal rejection events rather than getting stranded after `OrderSubmitted`.

2. `PositionClosed` handler re-entrancy
Addressed.

The sandbox `PositionClosed` handler now uses `try_borrow_mut()` instead of `borrow_mut()`. On the synchronous no-runner / direct-msgbus path, if the handler is re-entered while the sandbox still holds an active borrow on the same stack, it now skips that immediate retry instead of panicking.

I chose this narrower guard rather than introducing another deferred queue because the existing cleanup path already converges through the outer owner-boundary cleanup points. In other words, the fix here is to make the retry non-panicking under re-entry, not to add a parallel cleanup mechanism.

I added `test_instrument_close_sync_cleanup_handles_synchronous_position_closed_reentry` to exercise that synchronous path directly.

3. `purge_instrument_skip_order_guard` trade-off should be documented
Addressed.

I expanded the method docs to call out the intentional trade-off explicitly: there is a brief window where order objects may still exist even though `instrument_orders` has already been removed from the cache index.

That keeps the expiration-lifecycle cleanup semantics visible at the public cache API boundary rather than only implied by the internal helper split.

4. Smaller follow-ups

- Public `purge_instrument` docs
  Addressed.

  I updated the public docs while touching the cache surface so the distinction is explicit:
  - `purge_instrument(...)` is the strict/public path: it refuses to purge when orders or positions are still non-terminal.
  - `purge_instrument_skip_order_guard(...)` is the narrow expiration-lifecycle path: it still honors the position guard, but it no longer treats delayed order-terminal cache updates as owning retention.

- `is_expired_now` vs `is_expiration_processed()`
  Left unchanged intentionally.

  For this PR I kept cleanup keyed off factual contract expiry (`expiration_ns <= now`) rather than engine-local processed state.

  The two answer slightly different questions:
  - `expiration_ns <= now`: has the instrument crossed the venue/domain expiry boundary yet?
  - `is_expiration_processed()`: has the matching engine already consumed and marked that transition internally?

  For the cleanup paths touched here, I wanted to preserve the existing PR3 policy that once the instrument is factually expired and there are no open positions, sandbox cleanup may converge even if the engine-local processed state is only observed slightly later on the stack.

  So I did not want this reviewer-fix patch to change the cleanup policy surface from “instrument lifecycle boundary” to “engine-local flag boundary”.

- `position_closed_event` helper using `PositionClosed::create`
  Left unchanged intentionally in this patch.

  I agree this is a valid style cleanup, but it is orthogonal to the two correctness fixes above. The current helper is test-only and stable, and changing it would not strengthen the regression coverage for either:
  - the order-list mid-loop cleanup bug, or
  - the synchronous `PositionClosed` re-entrancy bug.

  I therefore kept this patch narrowly scoped to:
  - the destructive-cleanup timing fix,
  - the re-entrancy guard, and
  - the cache API doc clarification tied directly to the retention semantics.

Validation run locally:
- `cargo test -p nautilus-sandbox --test execution -- --nocapture`
- `pre-commit run --files crates/adapters/sandbox/src/execution.rs crates/adapters/sandbox/tests/execution.rs crates/common/src/cache/mod.rs`

I kept the scope to the sandbox/cache changes only for this PR.

cjdsellers (approved):
Thank you for the follow-ups @graceyangfan 👍

PR #4295, opened 2026-06-20
Use slices for catalog writes

sunlei:
## Summary

Use borrowed slices for Parquet catalog write paths instead of taking owned `Vec<T>` values. This removes unused direct dependencies from persistence and the patched `pyo3-stub-gen`, and replaces those call sites with standard library equivalents.

## Type of change

- [x] Improvement (non-breaking for Python users)
- [x] Breaking change (Rust API signature change)
- [x] Maintenance / chore
- [x] Documentation update

## Breaking change details

Rust callers of `ParquetDataCatalog::write_to_parquet` and `data_to_record_batches` now pass borrowed slices instead of owned vectors:

```rust
catalog.write_to_parquet(&quotes, None, None, None)?;
```

The Python catalog API remains unchanged.

## Removed dependencies

Direct dependencies removed by this PR:

| Dependency | Removed from | Replacement |
| --- | --- | --- |
| `itertools` | Workspace root and `nautilus-persistence` | `slice::chunks`, `Vec::sort`, `Vec::join` |
| `itertools` | `patches/pyo3-stub-gen` | Standard iterator collection plus `Vec::sort` / `Vec::join` |
| `maplit` | `patches/pyo3-stub-gen` | `HashSet::from([...])` and `HashSet::new()` |
| `rustpython-parser` | `patches/pyo3-stub-gen` | Removed because the patched crate no longer uses it directly |

## Performance impact

| Area | Before | After | Expected impact |
| --- | --- | --- | --- |
| `data_to_record_batches` chunking | `itertools::chunks` plus per-chunk `collect_vec()` | Slice `chunks()` passed directly to encoders | Fewer temporary `Vec` allocations; expected improvement for large writes |
| Rust catalog write API | Takes ownership of `Vec<T>` | Borrows `&[T]` | Avoids unnecessary ownership transfer and caller clones |
| Test/catalog setup helpers | Temporary `Vec` literals | Temporary slice literals | Fewer test-only allocations |
| `pyo3-stub-gen` formatting joins | `itertools::join` | `collect::<Vec<_>>().join(",")` | Possible small regression in stub generation only; not runtime trading path |

No benchmark was run for this PR; the performance assessment is static allocation analysis.

## Documentation

- Updated the Rust backtest how-to snippet to pass borrowed catalog data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4901,15 +4901,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "itertools"
-version = "0.15.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b4baf93f58d4425749ca49a51c50ebab072c5df6994d08fed93541c331481dc"
-dependencies = [
- "either",
-]
-
-[[package]]
 name = "itoa"
 version = "1.0.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5269,12 +5260,6 @@ dependencies = [
  "quote",
  "syn 1.0.109",
 ]
-
-[[package]]
-name = "maplit"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e2e65a1a2e43cfcb47a895c4c8b10d1f4a61097f9f254f183aee60cad9c651d"
 
 [[package]]
 name = "matchers"
@@ -6529,7 +6514,6 @@ dependencies = [
  "datafusion",
  "futures",
  "indexmap 2.14.0",
- "itertools 0.15.0",
  "log",
  "nautilus-common",
  "nautilus-core",
@@ -7818,14 +7802,11 @@ dependencies = [
  "chrono",
  "indexmap 2.14.0",
  "inventory",
- "itertools 0.14.0",
  "log",
- "maplit",
  "num-complex",
  "pyo3",
  "pyo3-stub-gen-derive",
  "rust_decimal",
- "rustpython-parser",
  "serde",
  "serde_json",
  "time",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,7 +182,6 @@ http = "1.4.2"
 implied-vol = { version = "2.0.0" }
 indexmap = { version = "2.14.0", features = ["serde"] }
 itchy = "0.3"
-itertools = "0.15.0"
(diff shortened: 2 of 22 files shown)


sunlei:
Superseded by #4296.

faysou:
You don't need to close a PR and open a new one, you can just soft reset your commit modify it locally then do git push --force, or something similar for modifying commits, like rebasing as well. I have a bash function that pulls develop, rebases on it and does n soft resets in one command, it stashes uncommitted changes as well then pops them back at the end. Basically rebasing is a no brainer with it.

sunlei:
Thanks @faysou for the helpful comment.

The issue was on our side: the branch was based on `develop`, but when I created the PR I accidentally selected `master` as the target branch. That meant the normal `develop` CI checks were not created. After changing the target branch back to `develop`, GitHub still did not trigger those checks because changing the base branch does not fire the default `pull_request` workflow events.

An empty commit or force push would have triggered CI, but that felt less clean, so I closed this PR and reopened it as #4296 against `develop` directly.

Thanks again for pointing it out.

faysou:
Here's the function I was talking about for reference, it used to be called custom_rebase, but I use it all the time so I shortened the name to initials:

```bash
cr() {
    local target_branch="develop"
    local commits=0
    local current_branch=$(git rev-parse --abbrev-ref HEAD)

    # 1. Parse Arguments
    while [[ $# -gt 0 ]]; do
        case $1 in
            -b|--branch)
                target_branch="$2"
                shift 2 # Move past flag and value
                ;;
            [0-9]*)
                commits="$1"
                shift # Move past the number
                ;;
            *)
                echo "Unknown argument: $1"
                return 1
                ;;
        esac
    done

    # 2. Stash changes and capture output
    local stash_output=$(git stash push -u)

    # 3. Update logic
    if [ "$current_branch" = "$target_branch" ]; then
        echo "Updating current branch ($target_branch) from origin..."
        git pull origin "$target_branch"
    else
        echo "Fetching $target_branch and rebasing..."
        git fetch origin "$target_branch":"$target_branch"
        git rebase "$target_branch"
    fi

    # 4. Only pop if we actually pushed something
    if [[ "$stash_output" != "No local changes to save" ]]; then
        git stash pop
    fi

    # 5. Soft reset
    if [ "$commits" -gt 0 ]; then
        echo "Resetting $commits commits (soft)..."
        git reset HEAD~"$commits" --soft
    fi
}
```

PR #4305, opened 2026-06-24
Add Bybit instrument subscriptions

dxwil:
# Pull Request

## Summary

Add instrument subscription support for the Bybit adapter. Uses the same poll loop as the instrument statuses to avoid adding an extra parameter (now `insturment_poll_interval_secs`) and gets instrument definitions and statuses in one endpoint request. `on_instrument` is fired for every new instrument (if the subscription is venue-wide) and for changes to the economic fields of the instrument (e.g. fees).

Also emit instrument status on subscribe so the cache gets populated (not needed for instrument definitions since those are put into the cache on connect).

Currently economic difference is determined by a seperate`economics_differ()` function because an insturment is considered equal simply if it's id is equal (so fee changes and the like would be skipped). This helper function sits in the Bybit adapter but perhaps it would be better to move it into Instrument in the future if multiple adapters want to use it.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/bybit/src/common/instruments.rs
+++ b/crates/adapters/bybit/src/common/instruments.rs
@@ -30,10 +30,20 @@ use nautilus_model::{
 fn economics_differ(a: &InstrumentAny, b: &InstrumentAny) -> bool {
     a.maker_fee() != b.maker_fee()
         || a.taker_fee() != b.taker_fee()
+        || a.margin_init() != b.margin_init()
+        || a.margin_maint() != b.margin_maint()
+        || a.price_precision() != b.price_precision()
+        || a.size_precision() != b.size_precision()
         || a.price_increment() != b.price_increment()
         || a.size_increment() != b.size_increment()
+        || a.multiplier() != b.multiplier()
+        || a.lot_size() != b.lot_size()
         || a.min_quantity() != b.min_quantity()
+        || a.max_quantity() != b.max_quantity()
         || a.min_notional() != b.min_notional()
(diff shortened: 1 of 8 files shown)


cjdsellers:
Hi @dxwil,

Thanks for the PR, this is a pragmatic approach for Bybit instrument updates.

A few things I think need addressing before this lands:

- Clear the new `instrument_subs` and `subscribe_all_instruments` state in `reset()` and `disconnect()`, otherwise a reconnect can keep polling/emitting from stale subscriptions.
- Broaden `economics_differ()` to cover the other order-validation and pricing fields, such as max quantities/notionals, min/max price, precision, multiplier, and margins?

Small style note: the new Rust tests should use `#[rstest]` rather than bare `#[test]` to match the repo convention (otherwise pre-commit will fail, I suggest you run the pre-commit).

Given this is live Bybit behavior, I think you are in the best position to make the changes and confirm the final polling/subscription behavior against your own use case. That will give better signal than a local-only fix from my side.

Happy to go into more detail on any of these.

dxwil:
Hi @cjdsellers,

Thanks for taking a look, fixed the things you flagged. Let's keep the `economics_differ` function inside the Bybit adapter until more potential callers that need the same logic come up (then it would be smart to move it to `Instrument`).

cjdsellers (approved):
Thanks for the follow-ups @dxwil 👌

PR #4309, opened 2026-06-25
Fix Interactive Brokers crypto order sizing for SELL orders

bebop23:
## Summary

The Interactive Brokers execution client mis-sizes crypto (e.g. PAXOS) **SELL** orders, so they are rejected by the venue.

In `InteractiveBrokersExecutionClient._transform_order_to_ib_order` (`adapters/interactive_brokers/execution.py`), every `is_inverse` instrument's order is converted to a cash quantity:

```python
if self.instrument_provider.find(order.instrument_id).is_inverse:
    ib_order.cashQty = int(ib_order.totalQuantity)
    ib_order.totalQuantity = 0
```

This is incorrect for crypto in two ways:

1. **IBKR accepts a cash quantity (`cashQty`) only for BUY orders.** A SELL must be sized with the base/coin quantity (`totalQuantity`). Converting a SELL to `cashQty` makes IBKR reject the order.
2. **`int()` truncates a fractional coin quantity** (e.g. `0.031 -> 0`), so the size collapses to zero and the venue rejects with *"size value cannot be zero"*.

## Fix

Only convert **BUY** orders to `cashQty` (and without `int()` truncation, since `cashQty` accepts decimals); leave **SELL** orders' `totalQuantity` (coin quantity) untouched.

```python
if is_inverse and order.side == OrderSide.BUY:
    ib_order.cashQty = float(ib_order.totalQuantity)
    ib_order.totalQuantity = 0
```

## Verification

Verified against a **live IBKR paper account**: running a strategy that buys then sells `ETH/USD.PAXOS`, a market BUY (cash quantity) followed by a market SELL (coin quantity) now **both fill**. Before the fix, the SELL was rejected (`size value cannot be zero`). The behavior matches IBKR's documented rule that `cashQty` is valid for buy orders only.

## Notes

- I wasn't able to run the full local suite (no Rust/Cython build in this environment). Happy to add a unit test in `tests/integration_tests/adapters/interactive_brokers/test_execution_order_transform.py` — a faithful one needs an inverse instrument with **fractional** size precision (the existing inverse stubs like `xbtusd_bitmex` are whole-contract), mirroring IBKR PAXOS crypto. Glad to add it if you'd like it in this PR.
- Added a `RELEASES.md` Fixes entry.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -115,6 +115,7 @@ This release includes many breaking changes across the user-facing Rust v2 APIs.
 - Fixed HTTP client errors discarding the underlying cause from the reqwest source chain (Rust)
 - Fixed `HttpClient` rejecting invalid response header keys instead of silently dropping them (Rust)
 - Fixed `Instrument` rejecting negative `min_price`, preventing spread instruments from loading in Python
+- Fixed Interactive Brokers crypto order sizing where SELL orders were converted to `cashQty` (valid for BUY only) and fractional coin quantities were truncated to zero via `int()`, causing venue rejection ("size value cannot be zero")
 - Fixed live external order claim registration in Rust
 - Fixed live reconciliation logging below-cached fill mismatches as errors, halting `shutdown_on_error` nodes (Rust)
 - Fixed live reconciliation logging transient venue report-query failures as errors (Rust)
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -115,7 +115,7 @@ This release includes many breaking changes across the user-facing Rust v2 APIs.
 - Fixed HTTP client errors discarding the underlying cause from the reqwest source chain (Rust)
 - Fixed `HttpClient` rejecting invalid response header keys instead of silently dropping them (Rust)
 - Fixed `Instrument` rejecting negative `min_price`, preventing spread instruments from loading in Python
-- Fixed Interactive Brokers crypto order sizing where SELL orders were converted to `cashQty` (valid for BUY only) and fractional coin quantities were truncated to zero via `int()`, causing venue rejection ("size value cannot be zero")
+- Fixed Interactive Brokers crypto order sizing where inverse quote-quantity SELL orders were converted to `cashQty` (which IBKR accepts for BUY only) and fractional coin quantities were truncated to zero via `int()`, causing venue rejection ("size value cannot be zero"); the Python and Rust adapters now apply `cashQty` only for inverse quote-quantity BUYs and reject quote-quantity SELLs
 - Fixed live external order claim registration in Rust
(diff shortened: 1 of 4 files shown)


faysou:
Hi @bebop23 can you please do the equivalent change in the rust IB adapter as well as the python adapter is close to the end of its supported life ? Thanks

cjdsellers:
Hi @bebop23,

Thanks for the clear write-up and the live paper-account verification. I think this needs one predicate tweak before it lands: the Python path should still gate `cashQty` on `order.is_quote_quantity`, not only on `OrderSide.BUY`.

As written, an inverse base-quantity BUY can be sent as `cashQty`, and an inverse quote-quantity SELL can fall through as `totalQuantity`. The Rust IB transform already uses `instrument.is_inverse() && order.is_quote_quantity()`, so matching that here, plus rejecting quote-quantity SELLs if IBKR only accepts `cashQty` on BUYs, should keep both paths aligned.

Please let me know if I missed an IBKR constraint here.

bebop23:
Thanks both, good catches — pushed an update.

@cjdsellers you're right, gating only on `OrderSide.BUY` was too loose. The Python path now gates `cashQty` on `is_inverse and order.is_quote_quantity` (matching the Rust transform's `is_inverse() && is_quote_quantity()`) and raises on quote-quantity SELLs instead of letting them fall through. That closes both gaps you flagged: an inverse base-quantity BUY is no longer sent as `cashQty`, and an inverse quote-quantity SELL no longer slips through as `totalQuantity`. No IBKR constraint missed as far as I can tell — `cashQty` is BUY-only, so a quote-sized SELL can't be faithfully represented and is rejected up front.

@faysou done — same change is now in the Rust adapter (`execution/transform/policy.rs`). `apply_quantity_policy` rejects quote-quantity SELLs and only sets `cash_qty` for inverse quote-quantity BUYs; it returns a `Result` now and the call site propagates it. Non-inverse quote-quantity is still rejected one layer up in `core_orders.rs`, same as Python, so the two paths stay aligned.

Both adapters now share the same predicate and error message. `cargo check` passes and the existing IB transform tests are green; `RELEASES.md` updated.

On tests: a faithful Rust unit test needs an inverse instrument cached in the provider, which currently means embedded IB contract-info JSON (and `cache_instrument` is private), so it's awkward from the transform test module. Happy to add coverage whichever way you prefer — a small test hook on the provider, or a Python integration test with a fractional-precision inverse stub. Let me know.

cjdsellers:
Hi @bebop23 

Thank you for the follow-ups.

If you could just sign the CLA then we can merge this. I'll tidy up the release notes post-merge (as we shipped a release in the mean time).

cjdsellers (approved):

PR #4319, opened 2026-06-26
Batch Polymarket RTDS incremental subscribe snapshots

### LINKED ISSUE #4317: [Polymarket] RTDS incremental subscribe sends one wire message per symbol (reconnect batches); potential multi-symbol overwrite
## Summary

The Polymarket RTDS feed (`crates/adapters/polymarket/src/rtds.rs`) sends a **separate WebSocket `subscribe` message per symbol** on the incremental subscription path, while the reconnect/fresh-connect path correctly **batches all symbols into a single message**. This asymmetry risks silent symbol overwrite if the Polymarket RTDS server treats `subscribe` as replace-semantics (last-write-wins), and is needlessly chatty regardless.

Both `crypto_prices` and `equity_prices` are affected — they share the same `subscribe_live` path.

## Background: client-side tracking is correct

To be clear, the in-process tracking does **not** overwrite symbols:

- `track_subscribe` keys entries by `topic:symbol_lower` (e.g. `crypto_prices:btcusdt`), so distinct symbols never collide: `rtds.rs:199-223`
- Dispatch (`matching_data_types`) routes each update by exact `topic:symbol` key: `rtds.rs:729-742`
- Re-subscribing the same symbol reuses the wire (dedup): `test_track_subscribe_reuses_symbol_wire_subscription`

So this is purely a **wire-protocol / round-trip** concern, not a client-state bug.

## The asymmetry

**Reconnect / fresh connect — CORRECT (batched):**
```rust
// rtds.rs:440-444 (ensure_connected_locked)
let subscriptions = self.snapshot_wire_subscriptions(); // ALL symbols
if !subscriptions.is_empty() {
    self.send_wire_request(&ws, "subscribe", subscriptions).await?;
}
```
```rust
// rtds.rs:449-455
fn snapshot_wire_subscriptions(&self) -> Vec<RtdsWireSubscription> {
    self.inner.subscriptions.iter().map(|entry| entry.wire.clone()).collect()
}
```
Reconnect replay does the same: `rtds.rs:485-494`.

**Incremental subscribe while connected — RISKY (per-symbol):**
```rust
// rtds.rs:263-281 (subscribe_live)
self.send_wire_request(&ws, "subscribe", vec![wire]).await  // ONE message, ONE symbol
```
Each `DataClient::subscribe` (custom data) spawns its own task that calls `subscribe_live` for that one symbol: `data/mod.rs:381-410`.

So subscribing to `BTCUSDT`, `ETHUSDT`, `XRPUSDT` while already connected produces three separate wire messages, each with a single `subscriptions` entry.

## Why this matters

1. **Potential silent overwrite.** The RTDS wire shape is:
   ```json
   {"action":"subscribe","subscriptions":[{"topic":"crypto_prices","type":"update","filters":"{\"symbol\":\"BTCUSDT\"}"}]}
   ```
   Per the official RTDS client and docs ([real-time-data-client](https://github.com/Polymarket/real-time-data-client)), each entry carries its own `filters: '{"symbol":"..."}'`. The `subscriptions` array natively supports batching multiple symbols in **one** message. Whether the server **accumulates** successive `subscribe` messages (additive) or **replaces** the active set with the latest message (replace-semantics) is not documented. If it is replace-semantics, the per-symbol incremental messages mean only the **last** symbol subscribed receives updates — earlier symbols silently stop updating. This is the "overwrite" risk.

2. **Needless round-trips.** Even if the server is additive, N symbols = N wire messages instead of 1. The batched form is already implemented for the reconnect path, so the incremental path is just inconsistent.

## Contrast with the market channel

The CLOB market channel (`sync_ws_subscription_async`, `data/subscriptions.rs:45-78`) uses the same per-token incremental pattern, but the CLOB market WS is well-documented as additive (`operation: "subscribe"` adds `assets_ids`). RTDS subscribe semantics are not documented as additive, which is why RTDS is the riskier surface.

## Suggested fix

Coalesce pending incremental subscriptions into a single batched `subscribe` message, matching the reconnect path. Concretely, while holding `wire_mutex`, collect any newly-tracked subscriptions and send them in one `send_wire_request(&ws, "subscribe", vec![...])` call rather than one message per symbol.

This:
- eliminates the overwrite risk regardless of server semantics, and
- reduces N round-trips to 1.

## Affected versions / reproduction

- Branch: `develop` (HEAD `b257b8617a`)
- Subscribe to two or more RTDS custom data types (e.g. `PolymarketRtdsCryptoPrice` with `metadata['symbol']='BTCUSDT'` then `'ETHUSDT'`) on an already-connected client; observe two separate `subscribe` wire frames.

## Environment

- Adapter: `nautilus-polymarket` (Rust)
- Files: `crates/adapters/polymarket/src/rtds.rs`, `crates/adapters/polymarket/src/data/mod.rs`

## References

- Official RTDS client: https://github.com/Polymarket/real-time-data-client (`src/client.ts`, `src/model.ts`)
- RTDS topics/filters table (crypto_prices / equity_prices use `{"symbol":string}` filters)

graceyangfan:
## Summary
This PR hardens Polymarket RTDS multi-symbol subscriptions by making the adapter reconcile tracked intent into the final wire state, instead of letting each logical subscribe call become an independent wire action.

Before this change:
- RTDS tracked the logical subscription set internally
- connect and reconnect replay used tracked state
- live incremental subscribe still behaved like a direct wire action path
- repeated venue snapshots could re-emit historical RTDS points downstream

After this change:
- the adapter keeps one tracked desired RTDS subscription state
- the adapter tracks the last applied live wire state
- connect, reconnect, incremental subscribe, and incremental unsubscribe all reconcile through the same path
- live sync sends only the delta between desired state and live venue state
- inbound RTDS snapshot/update replay is guarded by per-symbol last-seen timestamp dedup

This keeps the fix tight to issue #4317 while addressing the reviewer concern about historical snapshot replay.

## Root Cause
The bug was not URL selection, topic naming, or JSON parsing.

The real issue was subscription lifecycle asymmetry:
- the adapter already knew the full logical RTDS subscription set
- reconnect replay used that tracked state
- but incremental live changes did not reconcile against the live venue state

That left the adapter vulnerable to repeated wire subscribe behavior and venue-side replay semantics.

## What Changed
- replaced the direct incremental RTDS wire path with tracked-state reconciliation
- added `live_subscriptions` state so the adapter can sync only the delta from desired state to live wire state
- coalesced repeated live sync requests through `schedule_sync()` so multiple logical subscribe calls can collapse into one final wire subscribe batch
- kept one replay model across connect, reconnect, subscribe, and unsubscribe
- added inbound per-symbol timestamp dedup so repeated RTDS snapshots do not replay historical custom data into strategies
- simplified `track_subscribe` / `track_unsubscribe` to return `changed` instead of venue wire payloads, aligning the public behavior with the new reconcile design

## Why This Fits Nautilus Style
- one logical source of truth: tracked RTDS subscription state
- one replay/reconcile model across lifecycle transitions
- no parallel channel or alternate state machine
- incremental wire sync is derived from tracked state, not caller-local behavior
- venue replay hardening lives at the adapter boundary

## Tests
Passed:
- `cargo test -p nautilus-polymarket rtds::tests:: -- --nocapture`
- `pre-commit run --files crates/adapters/polymarket/src/rtds.rs crates/adapters/polymarket/src/data/mod.rs`

Added / updated RTDS regression coverage for:
- duplicate logical subscribe sharing one tracked symbol identity
- new-symbol incremental subscribe sending only the new wire delta
- incremental unsubscribe sending only the removed wire delta
- coalescing multiple pending subscribe changes into one wire subscribe batch
- crypto snapshot replay dedup
- equity snapshot replay dedup
- baseline crypto/equity snapshot and update parsing still emitting expected custom data

## Live Verification
Outside the upstream PR, I also reran the local PyO3 + LiveNode smoke harness against the real RTDS venue using the patched worktree.

Confirmed live:
- crypto multi-symbol subscriptions (`BTCUSDT`, `ETHUSDT`, `SOLUSDT`, `XRPUSDT`) continue receiving data concurrently
- symbols do not overwrite each other and do not show blocking gaps in the harness run
- direct websocket validation showed that repeating a subscribe for an already-active RTDS symbol can re-trigger a snapshot, which is why the inbound timestamp guard remains necessary as a safety rail

Equity live verification is time-of-day dependent. The most recent rerun was on Saturday, June 27, 2026, when US equities were not actively trading, so that run does not provide fresh live equity ticks. The adapter parse and routing paths remain covered by the equity regression tests in this PR.

## Related
- Closes #4317

### DIFF
--- a/crates/adapters/polymarket/src/data/mod.rs
+++ b/crates/adapters/polymarket/src/data/mod.rs
@@ -391,7 +391,7 @@ impl DataClient for PolymarketDataClient {
             "Tracking Polymarket RTDS custom data subscription: {}",
             cmd.data_type
         );
-        let Some(wire) = self.rtds_feed.track_subscribe(cmd.data_type)? else {
+        let Some(_) = self.rtds_feed.track_subscribe(cmd.data_type)? else {
             return Ok(());
         };
 
@@ -401,7 +401,7 @@ impl DataClient for PolymarketDataClient {
 
         let feed = self.rtds_feed.clone();
         get_runtime().spawn(async move {
-            if let Err(e) = feed.subscribe_live(wire).await {
+            if let Err(e) = feed.subscribe_live().await {
                 log::error!("Failed to subscribe RTDS custom data: {e}");
             }
         });
--- a/crates/adapters/polymarket/src/data/mod.rs
+++ b/crates/adapters/polymarket/src/data/mod.rs
@@ -391,20 +391,16 @@ impl DataClient for PolymarketDataClient {
             "Tracking Polymarket RTDS custom data subscription: {}",
             cmd.data_type
         );
-        let Some(_) = self.rtds_feed.track_subscribe(cmd.data_type)? else {
+        let changed = self.rtds_feed.track_subscribe(cmd.data_type)?;
+        if !changed {
             return Ok(());
-        };
+        }
 
         if !self.is_connected() {
             return Ok(());
         }
 
-        let feed = self.rtds_feed.clone();
-        get_runtime().spawn(async move {
-            if let Err(e) = feed.subscribe_live().await {
-                log::error!("Failed to subscribe RTDS custom data: {e}");
-            }
-        });
+        self.rtds_feed.schedule_sync();
 
         Ok(())
     }
@@ -533,20 +529,16 @@ impl DataClient for PolymarketDataClient {
             "Tracking Polymarket RTDS custom data unsubscription: {}",
             cmd.data_type
         );
-        let Some(wire) = self.rtds_feed.track_unsubscribe(&cmd.data_type)? else {
+        let changed = self.rtds_feed.track_unsubscribe(&cmd.data_type)?;
+        if !changed {
             return Ok(());
-        };
+        }
 
         if !self.is_connected() {
             return Ok(());
         }
 
-        let feed = self.rtds_feed.clone();
-        get_runtime().spawn(async move {
-            if let Err(e) = feed.unsubscribe_live(wire).await {
-                log::error!("Failed to unsubscribe RTDS custom data: {e}");
-            }
-        });
+        self.rtds_feed.schedule_sync();
 
         Ok(())
     }
--- a/crates/adapters/polymarket/src/rtds.rs
+++ b/crates/adapters/polymarket/src/rtds.rs
@@ -17,7 +17,10 @@
 
 use std::{
     str::FromStr,
-    sync::{Arc, Mutex as StdMutex},
+    sync::{
+        Arc, Mutex as StdMutex,
+        atomic::{AtomicBool, AtomicU64, Ordering},
+    },
     time::Duration,
 };
 
@@ -66,9 +69,15 @@ struct PolymarketRtdsFeedInner {
     clock: &'static AtomicTime,
     data_sender: tokio::sync::mpsc::UnboundedSender<DataEvent>,
     subscriptions: dashmap::DashMap<String, TrackedSubscription>,
+    last_emitted_timestamps_ms: dashmap::DashMap<String, u64>,
+    // Tracks the last venue state we successfully pushed so incremental syncs
+    // can send only the delta from desired state to live wire state.
+    live_subscriptions: StdMutex<AHashMap<String, RtdsWireSubscription>>,
     ws_client: StdMutex<Option<Arc<WebSocketClient>>>,
     task_handle: StdMutex<Option<tokio::task::JoinHandle<()>>>,
     wire_mutex: tokio::sync::Mutex<()>,
+    sync_version: AtomicU64,
+    sync_running: AtomicBool,
 }
 
 #[derive(Clone, Debug)]
@@ -180,9 +189,13 @@ impl PolymarketRtdsFeed {
                 clock,
                 data_sender,
                 subscriptions: dashmap::DashMap::new(),
+                last_emitted_timestamps_ms: dashmap::DashMap::new(),
+                live_subscriptions: StdMutex::new(AHashMap::new()),
                 ws_client: StdMutex::new(None),
                 task_handle: StdMutex::new(None),
                 wire_mutex: tokio::sync::Mutex::new(()),
+                sync_version: AtomicU64::new(0),
+                sync_running: AtomicBool::new(false),
             }),
         }
     }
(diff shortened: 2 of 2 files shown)


cjdsellers:
Hi @graceyangfan,

Thanks for the PR, and for the clear root-cause write-up. The shape is right: one tracked subscription set, with one replay path across connect, reconnect, and incremental subscribe.

One thing I want to confirm before it lands. Since `subscribe_live` now resends the full snapshot, every incremental subscribe re-sends a `subscribe` action for symbols that are already active. The `subscribe`-type handlers (`handle_crypto_price_subscribe` / `handle_equity_price_subscribe`) replay the whole `data` array with no inbound dedup, so if the RTDS server re-sends a symbol's snapshot on re-subscribe, those historical points get re-emitted downstream each time another symbol is added.

In your live trading, have you observed what the server does when you re-subscribe an already-active symbol? Does it re-send that symbol's snapshot, or treat the repeat as a no-op? If it does re-snapshot, a per-symbol last-seen-timestamp guard in the subscribe handlers would make the full-replay safe under either semantics (and harden the existing reconnect path too).

The test asserts the outbound payload, which is the right check for the wire fix, but it doesn't exercise the inbound side, so I can't tell from the diff alone.

Happy to go into more detail. Let me know what you've seen live.

graceyangfan:
Thanks, this concern was valid.

I checked the RTDS venue behavior directly against the live websocket and confirmed that repeating a subscribe for an already-active symbol can trigger that symbol's snapshot again rather than acting as a pure no-op.

I changed the fix accordingly:
- the adapter no longer re-sends already-live symbols on incremental subscribe
- it now reconciles tracked desired state against the last applied live wire state and sends only the delta
- reconnect, subscribe, and unsubscribe all go through the same reconcile path
- I also kept an inbound per-symbol last-seen timestamp guard so replayed RTDS snapshot points are dropped before reaching strategies

So the final shape is:
- primary fix: avoid repeated wire subscribe for already-active symbols
- safety rail: suppress replayed historical points if the venue still re-sends them

I also expanded the tests to cover both sides:
- outbound: incremental subscribe sends only the new symbol
- outbound: incremental unsubscribe sends only the removed symbol
- outbound: multiple pending subscribe changes coalesce into one wire subscribe batch
- inbound: duplicate crypto subscribe snapshots do not re-emit historical points
- inbound: duplicate equity subscribe snapshots do not re-emit historical points

I reran the local PyO3 + LiveNode smoke harness against the patched worktree as well. Crypto multi-symbol delivery remained continuous across `BTCUSDT`, `ETHUSDT`, `SOLUSDT`, and `XRPUSDT` with no overwrite behavior. The latest rerun was on Saturday, June 27, 2026, so it did not provide fresh live US equity ticks, but the equity replay/dedup path is covered by the regression tests in this PR.

cjdsellers (approved):
Thanks again @graceyangfan 👍

PR #4320, opened 2026-06-26
Fix Binance TRADING_HALT ValidationError on futures

YeeTsai:

## Summary

Binance returns `status="TRADING_HALT"` for an active USDⓈ-M Futures contract that is temporarily halted. This value was missing from `BinanceFuturesContractStatus`, so the strict msgspec decode of `GET /fapi/v1/exchangeInfo` rejects the **entire** `symbols` array (`ValidationError` at `$.symbols[N].status`). That fails `BinanceFuturesInstrumentProvider._connect`, so the data/execution clients never connect and the `TradingNode` cannot start — even when none of the configured instruments is the halted symbol (one bad entry takes down the whole decode).

I hit this on the Binance USDⓈ-M Futures **testnet** (`demo-fapi.binance.com`), where a symbol's `exchangeInfo` entry reported `status="TRADING_HALT"` and prevented both the data and execution nodes from starting (`_connect` failed → engines timed out → process exited and kept restarting).

The fix adds `TRADING_HALT` to `BinanceFuturesContractStatus`, mirroring the prior `LIQUIDATION`/`ADL` enum-drift fix (34bbfd49). `_parse_instrument` is intentionally unchanged: it still skips only `PENDING_TRADING`, so a halted symbol is parsed into the cache like any other instrument (harmless — it is simply not traded).

## Related Issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

- [ ] Documentation changes follow the style guide (`docs/developer_guide/docs.md`) — N/A, no docs changes

## Release notes

- [x] I added a concise entry to `RELEASES.md` that follows the existing conventions

## Testing

Added `test_parse_futures_exchange_info_with_trading_halt_status` in `tests/integration_tests/adapters/binance/test_parsing_http.py`. It injects `status="TRADING_HALT"` into a real `exchangeInfo` fixture and decodes it via `BinanceFuturesExchangeInfo`, exercising the exact production decode path (`$.symbols[N].status`). It **fails before** the fix (reproducing `ValidationError: Invalid enum value 'TRADING_HALT' - at $.symbols[0].status`) and **passes after**.

- `pytest tests/integration_tests/adapters/binance/` — 428 passed
- `pre-commit run --files ...` on the changed files — passed (ruff, ruff-format, mypy, codespell, typos, copyright, testing-conventions, …)


cjdsellers:
Hi @YeeTsai,

Thanks for the PR, and for the clear write-up tracing it back to the strict decode.

The fix is the right shape for the Python adapter: adding `TRADING_HALT` to `BinanceFuturesContractStatus` unblocks the decode, and the regression test exercises the real `$.symbols[N].status` path.

One heads-up on direction: this only touches the v1 Python Binance adapter, which we're retiring within the next couple of weeks as the v2 cutover lands. If you're able to, I'd encourage moving over to the v2 (Rust with an optional Python API) Binance adapter, which is where ongoing work is focused from here.

Either way, fixes like this need mirroring on the Rust side. There the `BinanceContractStatus` enum already absorbs undocumented values through a `#[serde(other)] Unknown` catch-all, so it won't crash the way the Python decode did, but `TRADING_HALT` currently falls into that `Unknown` bucket (mapped to `NotAvailableForTrading`). It's worth adding as an explicit variant so the status maps deliberately rather than by fallback.

Happy to provide further details on the v2 adapter if it helps.

https://nautilustrader.io/docs/latest/how_to/

YeeTsai:
Thanks for the quick and detailed review, @cjdsellers — and for the heads-up on the v1 → v2 cutover.

I've pushed a second commit (fe6c7146d8) mirroring the fix on the Rust side: added an explicit `TradingHalt` variant to `BinanceContractStatus` and mapped it to `MarketStatusAction::Halt`, so it no longer falls into the `#[serde(other)] Unknown` bucket (`NotAvailableForTrading`) but maps deliberately. Added a unit test covering both the deserialize and the mapping.

I kept the Python-side change as well since it's still useful for v1 users until the cutover lands, but happy to drop it if you'd prefer the PR be Rust-only.

Checks run locally: `cargo test -p nautilus-binance` (incl. the new test) and `cargo clippy -p nautilus-binance` clean; pre-commit clean on both the Rust and Python files.

I'd definitely appreciate any pointers on the v2 (Rust + optional Python API) Binance adapter — we'll be moving onto it, so understanding the intended structure for status handling there would help us contribute fixes in the right place going forward.

cjdsellers:
Hey @YeeTsai 

We've been having issues with disk space on GitHub hosted runners. Could you please rebase when you get a chance which will trigger some new disk space freeing strategies.

PR #4239, opened 2026-06-10
Fix DonchianChannel period window

KaizynX:
# Pull Request

- [x] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

Fixes the Rust/PyO3 `DonchianChannel` rolling-window behavior so `upper`, `lower`, and `middle` are computed over the configured `period`, not all retained values up to `MAX_PERIOD`.

This aligns the Rust implementation with the legacy Cython indicator behavior and prevents stale highs/lows from affecting the channel after they leave the active window.

## Related Issues/PRs

Closes #4238

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Improvement (non-breaking)
- [ ] Breaking change (impacts existing behavior)
- [ ] Documentation update
- [ ] Maintenance / chore

## Breaking change details (if applicable)

N/A

## Documentation

- [ ] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)

No documentation changes were needed.

## Release notes

- [ ] I added a concise entry to `RELEASES.md` that follows the existing conventions (when applicable)

No release notes entry was added.

## Testing

**Ensure new or changed logic is covered by tests.**

- [ ] Affected code paths are already covered by the test suite
- [x] I added/updated tests to cover new or changed logic

Added a regression test for stale values leaving the Donchian window and updated the existing period-10 expectations.

Validation run locally:

```text
rustfmt --check crates/indicators/src/volatility/dc.rs
git diff --check
cargo test -p nautilus-indicators volatility::dc::tests -- --nocapture
prek run --all-files
```

The targeted Rust test run passed with 7 tests passed, 0 failed. The full local pre-commit run also passed.

PR #4326, opened 2026-06-26
Fix RateOfChange period window

Martingale42:
## Summary

Fix `RateOfChange` to respect its `period` window.

The `prices` deque is declared with a fixed `MAX_PERIOD` (1024) capacity, but the
window was never bounded to `period`. As a result `prices.front()` — the reference
price for the calculation — stayed pinned to the first-ever price instead of the
price `period` bars ago, so the indicator was correct only at the moment it first
became initialized and drifted from then on.

This mirrors the fix applied to `DonchianChannel` in #4239: evict the oldest price
once the window reaches `period` before pushing, matching the Cython reference
which uses `deque(maxlen=period)`.

## Tests

- Added `test_value_respects_period_window` (an early outlier must leave the window).
- Corrected the baked-in expected value in `test_value_with_all_higher_inputs`.

```
cargo nextest run -p nautilus-indicators roc   # 7 passed
```

### DIFF
--- a/crates/indicators/src/momentum/roc.rs
+++ b/crates/indicators/src/momentum/roc.rs
@@ -112,7 +112,7 @@ impl RateOfChange {
 
         if let Some(first) = self.prices.front() {
             if self.use_log {
-                self.value = (price / first).log10();
+                self.value = (price / first).ln();
             } else {
                 self.value = (price - first) / first;
             }
@@ -160,7 +160,7 @@ mod tests {
         }
 
         assert!(roc_10.initialized());
-        assert_eq!(roc_10.value, 0.28428852094735724);
+        assert_eq!(roc_10.value, 0.6545985104427102);
     }
 
     #[rstest]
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @Martingale42,

Thanks for the PR, and for matching the Cython `deque(maxlen=period)` reference. The window fix is correct.

One thing worth folding in while you're in here. The `use_log` branch computes `.log10()`, but the Cython reference does `from math import log` then `log(price / first)`, which is natural log. So for `use_log=true` the two diverge by a constant `ln(10)` factor, and the new expected value bakes the base-10 result in. `roc.rs` is also the only indicator in the crate using `.log10()`. Switching it to `.ln()` and recomputing that one test value would bring it back in line.

Please let me know if you'd like me to expand on any of this.

Martingale42:
Updated the use_log case to use .ln() instead of .log10() and recomputed the affected test value (test_value_with_all_higher_inputs → 0.6545985104427102). roc.rs no longer uses .log10(), so the use_log result now matches the Cython log(price / first) reference. Pushed in fce2f7e.

cjdsellers (approved):
Thanks for the follow-ups @Martingale42 👌

PR #4330, opened 2026-06-26
Replace bincode in event_store with a positional codec

### LINKED ISSUE #4316: Replace bincode in event_store — RUSTSEC-2025-0141 (unmaintained)
## Why now

bincode is unmaintained as of [RUSTSEC-2025-0141](https://rustsec.org/advisories/RUSTSEC-2025-0141) (filed 2026-01-07; development ceased permanently, no fix coming). `cargo audit` / `cargo deny` will flag it in CI. We use it for the on-disk envelope in `event_store`, so this is a durability-format decision.

Doing it now is the cheapest it will ever be: the crate is **pre-2.0 and has shipped only in beta releases** (v1.227.0–v1.229.0), so the format break carries no compat obligation — a window that closes as you approach 2.0.

## Current state

- bincode 2.x serde API (`bincode::serde::encode_to_vec` / `decode_from_slice`, `config::standard()`), ~15 call sites: `backend/redb.rs`, `markers/redb.rs`, `kernel.rs`, plus tests/benches.
- Encodes `EventStoreEntry`, `RunManifest`, `SnapshotAnchor`, and the marker family — all plain `#[derive(Serialize, Deserialize)]` data with `#[serde(with = "wire::nanos_as_u64")]` adapters. No object graph / cycles.
- **No codec-level format marker today** — records are stored as raw `&[u8]` blobs with nothing identifying the codec. A naive swap would *silently mis-decode* old stores rather than reject them.

## Requirements for any replacement

- Non-self-describing / positional (`wire.rs` relies on `deserialize_any` rejection)
- serde-compatible (keep the derives + `wire::` adapters)
- compact, deterministic
- bounds-checked decode (feeds the existing `Corrupted` path)

Self-describing was deliberately rejected for the envelope (MessagePack is already used *inside* payloads).

## Mandatory regardless of codec choice

Add a store-level format/codec marker (meta record or per-blob tag) so pre-change stores are **rejected with a clear error**, not mis-decoded. This is the migration story for a beta-only format.

## Migration: none

Shipped only in beta → no forward-compat obligation. Hard break, discard old stores, regenerate. Changelog note: *"event_store on-disk format changed; beta stores from v1.227–1.229 must be regenerated."* No migrator, no dual-read path.

## Codec options

- **wincode** — deliberate drop-in for the bincode 2.x serde API; ~15 call sites change, nothing else moves. Actively maintained. Risk: young, single-maintainer.
- **Roll our own serde backend** — our own `Serializer` / `Deserializer` over a small hardened byte layer (magic + version, length caps vs forged prefixes, bounds-checked reads, deterministic). Keeps all derives + `wire::` adapters. One-time bounded cost; permanently ends codec-churn exposure. Cost: implement the serde data model + preserve `deserialize_any` rejection.
- **Rejected:** postcard (near-unmaintained; `postcard2` fork exists), musli (stale — no recent release), bitcode (wire-format stability weak for durable data), rmp-serde (self-describing; rejected for envelope).

**Decision axis:** effort vs durability. wincode = smallest diff, re-exposed to single-maintainer risk. Own backend = bounded one-time cost, ends the churn.

**Recommendation (author leans, open to reviewers):** roll our own serde backend, to permanently end the codec-churn exposure (we've now burned through bincode and postcard in this same domain). Left to reviewers to weigh against wincode's much smaller diff.

## Validation

- Bench the winner vs current bincode (size + speed) via the existing `benches/codec.rs`.
- Round-trip tests, corrupt-input tests, and an old-format-rejection test.

folknor:
## Summary

`bincode` is unmaintained (RUSTSEC-2025-0141). This replaces it in the
`event_store` on-disk envelope with a hand-rolled, non-self-describing
positional serde codec, and adds a store-level format marker so pre-change
beta stores are rejected with a clear error rather than silently mis-decoded.

Closes #4316.

## Format break (beta only)

This is a hard, intentional on-disk format change with no migrator. Event
stores written by beta versions **v1.227–1.229** are no longer readable and
must be regenerated. Opening such a store now fails fast with a clear
`Corrupted`-class error ("...must be regenerated") instead of mis-decoding.

## What changed

Two commits:

1. **Codec module** (`crates/event_store/src/codec.rs`) — a non-self-describing,
   positional serde `Serializer`/`Deserializer` framed with a `NESC` magic +
   version header over a fixed-width little-endian byte layer. Decoding is
   bounds-checked (forged-prefix and length-overflow rejection), requires full
   buffer consumption, and rejects `deserialize_any` to preserve the `wire.rs`
   invariant. Decode failures feed the existing `EventStoreError::Corrupted`
   path.

2. **Cutover** — swap every production call site (`backend/redb.rs`,
   `markers/redb.rs`, `kernel.rs`) to the codec; add the store-level format
   marker (`format.rs`) — a native-redb `store_format` table written at store
   init and verified before any record decode; migrate the test corpus; rework
   `benches/codec.rs` to a one-time captured baseline; remove the `bincode`
   2.0.1 dependency.

## On the choice of a hand-rolled codec

We rolled our own rather than adopting another crate, since the encoded data
model is small and fully positional-safe (all enums are unit-only), which keeps
the codec bounded and adds no new dependency — ending the codec-churn exposure
(bincode → postcard → bincode) permanently.

That said, if you'd prefer the more conservative route of depending on
[`wincode`](https://crates.io/crates/wincode) instead of maintaining a
bespoke codec, we're happy to accommodate — say the word and we'll rework the
PR to use it.

## Dependency note

Only `event_store`'s direct `bincode 2.0.1` dependency (and the workspace
entry + lockfile nodes) is removed. The transitive `bincode 1.3.3` pulled via
`hypersync-client`/`madsim` is unaffected, so the existing RUSTSEC-2025-0141
ignore in `deny.toml` is retained; only the now-redundant duplicate-version
`bans` skip is dropped.

## Scope

In: the `event_store` envelope codec only. Untouched: MessagePack inside
payloads (used there deliberately), the redb storage layer, and the BLAKE3
hash framing.

## Testing

- Round-trip, corrupt-input, forged-prefix, and property tests for the codec
- Old-format and unsupported-version rejection tests for both the event and
  marker backends
- `cargo fmt`, `clippy -D warnings`, the full feature-expanded crate test
  suite, and `cargo deny check advisories bans` all pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1676,26 +1676,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bincode"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "36eaf5d7b090263e8150820482d5d93cd964a81e4019913c972f4edcc6edb740"
-dependencies = [
- "bincode_derive",
- "serde",
- "unty",
-]
-
-[[package]]
-name = "bincode_derive"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf95709a440f45e986983918d0e8a1f30a9b1df04918fc828670606804ac3c09"
-dependencies = [
- "virtue",
-]
-
-[[package]]
 name = "bip32"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4554,7 +4534,7 @@ dependencies = [
  "anyhow",
  "arrayvec",
  "arrow 57.3.1",
- "bincode 1.3.3",
+ "bincode",
  "capnp 0.23.2",
  "faster-hex",
  "fastrange-rs",
@@ -5247,7 +5227,7 @@ dependencies = [
  "async-channel",
  "async-stream",
  "async-task",
- "bincode 1.3.3",
+ "bincode",
  "bytes",
  "downcast-rs",
  "errno",
@@ -6067,7 +6047,6 @@ dependencies = [
  "ahash 0.8.12",
  "anyhow",
  "async-trait",
- "bincode 2.0.1",
  "blake3",
  "bytes",
  "criterion",
@@ -10482,12 +10461,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8ecb6da28b8a351d773b68d5825ac39017e680750f980f3a1a85cd8dd28a47c1"
 
 [[package]]
-name = "unty"
-version = "0.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d49784317cd0d1ee7ec5c716dd598ec5b4483ea832a2dced265471cc0f690ae"
-
-[[package]]
 name = "url"
 version = "2.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


cjdsellers (approved):
Thanks @folknor, this is a well-scoped replacement 👍

PR #4331, opened 2026-06-26
Add caller-supplied clock factory seam for live/sandbox nodes

### LINKED ISSUE #4304: [RFC] Injection seam for a caller-supplied Clock in live/sandbox nodes
> Code references pinned to branch `develop` at commit `c44c605d5b55385af8e8f4cc77605fc042a5ec84` (around `v1.229.0`).

### Summary

Allow a live/sandbox node to be built with a **caller-supplied clock factory**, instead of hardwiring `LiveClock::default()` from the `Environment` discriminant. The `Clock` trait is already a complete control surface (time reads *and* timer scheduling/firing), so a custom implementation can deliver alternative time semantics on its own — it just has no way in.

### Motivation

We run an integration harness: a real external venue process speaking our adapter's protocol over a real transport, with a nautilus live node trading against it — exercising the genuine live execution and data path. We want to drive that closed loop on a **non-wall clock** (e.g. accelerated faster-than-real-time but coherent, so time-in-force expiries and bar windows fire on the same axis as the data timestamps). We're prepared to implement that clock ourselves; we just need the node to *use* it.

This is the **Live** environment (real sockets, real clients). It is *not* Sandbox — that simulates execution in-process via `OrderMatchingEngine`, which is the thing our external venue replaces. But since Live and Sandbox share the same clock construction, the seam covers both.

### Why this is purely a seam (no timer/runner changes needed)

The `Clock` trait (`crates/common/src/clock.rs`) already owns timer creation: `set_timer_ns`, `set_time_alert_ns`, `register_default_handler`, `next_time_ns`. And firing is impl-internal: `LiveTimer::start` (`crates/common/src/live/timer.rs`) pushes `TimeEvent`s into a `TimeEventSender`, and the live runner (`crates/live/src/runner.rs`, the `tokio::select!` loop) just `recv()`s them with no wall-clock assumption. A custom `Clock` can therefore schedule on any axis and report any `timestamp_ns()`, pushing through the public `get_time_event_sender()`. The trait is sufficient; only construction is closed.

### Current behavior (two hardwired construction sites)

**Kernel clock** — `crates/system/src/kernel.rs`:

```rust
fn initialize_clock(environment: Environment) -> Rc<RefCell<dyn Clock>> {
    match environment {
        Environment::Backtest => { /* TestClock */ }
        Environment::Live | Environment::Sandbox => {
            Rc::new(RefCell::new(nautilus_common::live::clock::LiveClock::default()))
        }
    }
}
```

**Per-component clocks** — `crates/system/src/trader.rs`:

```rust
pub fn create_component_clock(&mut self, component_id: ComponentId) -> Rc<RefCell<dyn Clock>> {
    let clock = match self.environment {
        Environment::Backtest => Rc::new(RefCell::new(TestClock::new())),
        Environment::Live | Environment::Sandbox => Self::create_live_clock(), // LiveClock::default()
    };
    // ...
}
```

Neither the config nor the builder can override these.

### Proposed change

Let the live node builder accept an optional **clock factory** — e.g. `Fn() -> Rc<RefCell<dyn Clock>>` — used for the kernel clock and for each `create_component_clock` call in Live/Sandbox, defaulting to `LiveClock::default()` when not supplied (fully backward compatible). A factory (not a single instance) preserves the existing per-component-clock invariant noted in `create_component_clock` ("Each component gets its own clock instance so that the default time event callback registered on each clock is independent").

### Acceptance criteria

- A live/sandbox node can be built with a caller-supplied clock factory, with no fork of `kernel`/`trader`.
- When no factory is supplied, behavior is byte-identical to today.

### Offer to implement

We'd like to contribute this rather than just file the ask — it's a small, contained seam. Would you be open to a PR? Happy to align on the exact shape (factory closure vs. a `ClockFactory` trait vs. an injected `dyn Clock` per builder call) before we start.

folknor:
## Summary

Lets a Live/Sandbox node be built with a caller-supplied clock factory instead
of the hardwired `LiveClock::default()`, used for both the kernel clock and
every per-component clock. When no factory is supplied, behavior is identical
to today.

Implements the closure-factory option (the RFC's first-listed approach) from
#4304. Closes #4304.

## Approach

A cloneable, re-invokable closure alias, mirroring the existing
`EventStoreFactory` precedent in this crate:

    pub type ClockFactory = Rc<dyn Fn() -> Rc<RefCell<dyn Clock>>>;

The key difference from `EventStoreFactory` (which is single-use, `FnOnce`):
the clock factory is invoked many times — once for the kernel clock and once
per component — so it is `Fn` and cloneable, and each component still receives
its own distinct clock instance.

## What changed

- New `ClockFactory` alias in `crates/system/src/clock_factory.rs`, re-exported
  from the crate.
- Threaded through `NautilusKernel::new_with`, `Trader::new`, and both
  `initialize_clock` / `create_component_clock`.
- A `with_clock_factory` builder method on both `NautilusKernelBuilder` and
  `LiveNodeBuilder`, mirroring `with_event_store`.
- All constructor call sites updated, including the backtest engine and test
  helpers.

## Compatibility

No factory supplied → byte-identical to today (`LiveClock::default()`). The
backtest path is unchanged (`TestClock`). The constructors `Trader::new` and
`NautilusKernel::new_with` gain a trailing parameter — a source-level signature
change, consistent with the precedent set by `EventStoreFactory`.

## Scope

In: the construction seam only. Out: any `LiveTimer`/runner change (the `Clock`
trait is already a sufficient control surface), the Sandbox
`OrderMatchingEngine`, and the Python builder surface (a clock factory is a
Rust closure not expressible from Python).

## Testing

Three new `nautilus-system` tests covering: the factory applied to the kernel
clock and every component clock; the default path unchanged when no factory is
supplied; and distinct per-component clock instances under a factory.
`cargo fmt`, `clippy -D warnings`, and the `--features live` test suite pass.

### DIFF
--- a/crates/backtest/src/engine.rs
+++ b/crates/backtest/src/engine.rs
@@ -1928,6 +1928,7 @@ mod tests {
             config,
             None,
             Some(Box::new(event_store_factory)),
+            None,
         )
         .unwrap();
         engine.instance_id = engine.kernel.instance_id;
--- a/crates/live/src/node/builder.rs
+++ b/crates/live/src/node/builder.rs
@@ -36,6 +36,7 @@ use nautilus_execution::engine::ExecutionEngine;
 use nautilus_model::identifiers::TraderId;
 use nautilus_portfolio::config::PortfolioConfig;
 use nautilus_system::{
+    clock_factory::ClockFactory,
     config::StreamingConfig,
     event_store::{EventStoreFactory, KernelEventStore},
     kernel::NautilusKernel,
@@ -85,6 +86,7 @@ pub struct LiveNodeBuilder {
     data_client_configs: HashMap<String, Box<dyn ClientConfig>>,
     exec_client_configs: HashMap<String, Box<dyn ClientConfig>>,
     event_store_factory: Option<EventStoreFactory>,
+    clock_factory: Option<ClockFactory>,
     external_msgbus_factory: Option<Box<dyn MessageBusBackingFactory>>,
     external_msgbus_egress: Option<Box<dyn MessageBusExternalEgress>>,
     external_msgbus_ingress: Option<ExternalMessageBusIngress>,
@@ -100,6 +102,7 @@ impl Debug for LiveNodeBuilder {
             .field("data_client_configs", &self.data_client_configs.keys())
             .field("exec_client_configs", &self.exec_client_configs.keys())
             .field("event_store_factory", &self.event_store_factory.is_some())
+            .field("clock_factory", &self.clock_factory.is_some())
             .field(
                 "external_msgbus_factory",
                 &self.external_msgbus_factory.is_some(),
@@ -144,6 +147,7 @@ impl LiveNodeBuilder {
             data_client_configs: HashMap::new(),
             exec_client_configs: HashMap::new(),
             event_store_factory: None,
+            clock_factory: None,
             external_msgbus_factory: None,
             external_msgbus_egress: None,
             external_msgbus_ingress: None,
@@ -171,6 +175,7 @@ impl LiveNodeBuilder {
             data_client_configs: HashMap::new(),
             exec_client_configs: HashMap::new(),
             event_store_factory: None,
+            clock_factory: None,
             external_msgbus_factory: None,
             external_msgbus_egress: None,
             external_msgbus_ingress: None,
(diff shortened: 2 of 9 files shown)


cjdsellers:
Hi @folknor

Thanks for the PR, the seam and motivation make sense. I’m happy to merge this direction, and I'll rework the constructor/API shape afterward so `new_with` / `Trader::new` don’t keep accumulating trailing optional args. The important behavior is the builder-facing clock factory with unchanged default `LiveClock` behavior and fresh clocks per component.

cjdsellers:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

folknor:
Thank you, I believe it should be fixed now. I'm not familiar with this CI workflow, it seems you need to approve the workflow before it can be tested again?

cjdsellers (approved):
Will follow up with refinements locally 👍

PR #4332, opened 2026-06-27
Fix HTTP response body size cap to prevent unbounded allocation

### LINKED ISSUE #4308: HTTP response body is read without a size limit (unbounded allocation / DoS)
### Summary

`InnerHttpClient::to_response` in `crates/network/src/http/client.rs` reads the entire HTTP response body into memory via `response.bytes().await` with no maximum-size cap. The `reqwest` client is built with pool/timeout/HTTP2 options but no body-size limit, and `reqwest` enforces no default maximum. The request timeout bounds *duration*, not *peak memory*.

```rust
let status = HttpStatus::new(response.status());
let body = response.bytes().await.map_err(HttpClientError::from)?;
```

### Impact

A compromised or malfunctioning exchange REST endpoint can stream an arbitrarily large (or chunked) response and drive the trading process to OOM. This is an availability concern: an OOM kill mid-session can leave open orders/positions unmanaged. The shared `HttpClient` is used by all exchange adapters.

Severity is low/moderate: the peer is an operator-chosen, TLS-verified exchange (semi-trusted), so it requires a hostile or buggy endpoint — but it's the one unbounded path on the network surface. The WebSocket and raw-socket paths already enforce caps (64 MiB message / 16 MiB frame; `MAX_READ_BUFFER_BYTES = 10 MiB`; 16 KiB proxy-CONNECT cap), so HTTP is the lone gap.

### Suggested fix

Bound the response body: reject up front when `Content-Length` exceeds a configurable maximum, and otherwise stream chunk-by-chunk, aborting once the cap is exceeded (so an oversized/unbounded body is never fully buffered). A default around 100 MiB would comfortably exceed legitimate exchange payloads.

I have a patch ready (with tests) and am happy to open a PR if this direction looks good. 🙏

### Environment

Found on `develop` (1.229.x). Reviewed against current `crates/network/src/http/client.rs`.

AlaeddineMessadi:
Closes #4308

## Summary

The shared `HttpClient` read the full response body via `response.bytes()` with no size limit, so a hostile or malfunctioning endpoint could exhaust memory by streaming an arbitrarily large (or chunked, unbounded) body. This mirrors the caps already enforced on the WebSocket and raw-socket paths.

As discussed in #4308, this implements the `Content-Length` precheck plus chunked body accounting with a default cap on full-body materialization.

## Changes

- Add `DEFAULT_MAX_RESPONSE_BYTES` (100 MiB) and a `max_response_bytes` field on `InnerHttpClient` (set in both the constructor and `Default`).
- Replace `response.bytes()` in `to_response` with a new `read_body_capped` that:
  - rejects up front when `Content-Length` already exceeds the cap, and
  - otherwise streams the body chunk-by-chunk, aborting as soon as the accumulated size would exceed the cap — so an oversized or unbounded body is never fully buffered into memory.

The cap is enforced internally; the public client API and the 16 adapter call sites are unchanged.

## Test plan

- [x] Added a `/large` test route serving a 1 MiB body.
- [x] `test_response_body_within_cap_is_returned` — body under the cap is returned intact.
- [x] `test_response_body_exceeding_cap_is_rejected` — body over the cap fails with an "exceeds maximum" error rather than being buffered.
- [x] `cargo fmt` clean.

> Note: I'm on a local toolchain (1.88) below the pinned 1.96, so I relied on CI to run the full `cargo check`/`nextest` for the workspace. The change is confined to `crates/network/src/http/client.rs` and depends only on existing `reqwest`/`HttpClientError` APIs.

### DIFF
--- a/crates/network/src/http/client.rs
+++ b/crates/network/src/http/client.rs
@@ -37,6 +37,13 @@ const DEFAULT_POOL_IDLE_TIMEOUT_SECS: u64 = 60;
 /// Default HTTP/2 keep-alive interval in seconds.
 const DEFAULT_HTTP2_KEEP_ALIVE_SECS: u64 = 30;
 
+/// Default maximum HTTP response body size in bytes (100 MiB).
+///
+/// Bounds peak memory per response so a hostile or malfunctioning endpoint
+/// cannot exhaust memory by streaming an arbitrarily large body. Mirrors the
+/// caps already enforced on the WebSocket and raw-socket paths.
+const DEFAULT_MAX_RESPONSE_BYTES: usize = 100 * 1024 * 1024;
+
 /// An HTTP client that supports rate limiting and timeouts.
 ///
 /// Built on `reqwest` for async I/O. Allows per-endpoint and default quotas
@@ -161,6 +168,7 @@ impl HttpClient {
             client,
             header_keys: Arc::from(valid_keys),
             header_names: Arc::from(header_names),
+            max_response_bytes: DEFAULT_MAX_RESPONSE_BYTES,
         };
 
         Ok(Self {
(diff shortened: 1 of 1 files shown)


cjdsellers:
@AlaeddineMessadi Looks like the pre-commit failed, but this is easy to fix. Just run the following and push any changes to the same PR:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

cjdsellers (approved):
Thanks again @AlaeddineMessadi

PR #4333, opened 2026-06-27
Fix VerticalHorizontalFilter and OnBalanceVolume period windows

Martingale42:
## Summary

Fix `VerticalHorizontalFilter` to respect its `period` window.

The `prices` deque is declared with a fixed `MAX_PERIOD` (1024) capacity, but the
window was never bounded to `period`. As a result the `max`/`min` used in the VHF
calculation folded over every retained price (up to 1024) instead of the last
`period` bars, so the value diverged once more than `period` inputs were seen.

This mirrors the fix applied to `DonchianChannel` in #4239: evict the oldest price
once the window reaches `period` before pushing, matching the Cython reference
which uses `deque(maxlen=period)`.

## Tests

- Added `test_value_respects_period_window` (an early spike must leave the window;
  buggy result = 33.0, fixed = 2/3).
- Existing tests pass unchanged. `test_value_with_ten_inputs` did not catch the bug
  because its min value appeared at both an evicted and a retained position.

```
cargo nextest run -p nautilus-indicators vhf   # 8 passed
```

### DIFF
--- a/crates/indicators/src/momentum/obv.rs
+++ b/crates/indicators/src/momentum/obv.rs
@@ -107,14 +107,22 @@ impl OnBalanceVolume {
             0.0
         };
 
-        let _ = self.obv.push_back(delta);
-
-        self.value = self.obv.iter().sum();
+        if self.period == 0 {
+            // Unbounded cumulative OBV (matches the Cython `deque(maxlen=None)` reference).
+            self.value += delta;
+        } else {
+            // Rolling sum over the last `period` deltas (Cython `deque(maxlen=period)`).
+            if self.obv.len() == self.period {
+                let _ = self.obv.pop_front();
+            }
+            let _ = self.obv.push_back(delta);
+            self.value = self.obv.iter().sum();
+        }
 
         if !self.initialized {
             self.has_inputs = true;
 
-            if (self.period == 0 && !self.obv.is_empty()) || self.obv.len() >= self.period {
+            if self.period == 0 || self.obv.len() >= self.period {
                 self.initialized = true;
             }
         }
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @Martingale42,

Thanks for the follow-up, and for carrying the same window bound across to `VerticalHorizontalFilter`. The fix is correct, and the new test pins it deterministically: the early `100.0` spike is exactly the kind of value the old 1024-wide deque kept folding into `max`/`min` long after it should have left the `period` window.

One related case worth folding in while you're in here, if you want to keep the sweep going. `obv.rs` has the same shape: it sums the whole deque without ever bounding it to `period`, so for `period > 0` it doesn't respect the window the way the Cython `deque(maxlen=period)` reference does. The one wrinkle is the `period == 0` mode, which the Cython side treats as an unbounded cumulative sum (`maxlen=None`). That can't be expressed with the fixed-capacity `ArrayDeque`, so it needs a bit more than a straight `pop_front` port. Happy to go into detail if you pick it up.

Otherwise this is good to land. Let me know if you hit anything else.

Martingale42:
Folded the OnBalanceVolume fix in (b2c1495). For period > 0 it now evicts the oldest delta before pushing so the sum respects the window, matching deque(maxlen=period). For period == 0 I switched to a running total instead of the deque — the fixed-capacity ArrayDeque was silently capping the cumulative sum at 1024, so a straight pop_front port wasn't enough, as you flagged. Added a windowing test plus a period == 0 test that feeds > MAX_PERIOD inputs to pin the unbounded behaviour.

cjdsellers (approved):

Issue #4297, opened 2026-06-21
[Binance Futures] Custom data: add Arrow/Parquet catalog persistence (Liquidation, Ticker, OpenInterest)


graceyangfan:
## Summary

Three Binance Futures adapter custom data types already work end-to-end for **runtime subscribe/request and JSON registration**, but they cannot be written to or queried from `ParquetDataCatalog` because they are not registered for **Arrow encoding**.

This is inconsistent with other adapter-native custom families that already support catalog persistence, and leaves a gap in the platform's documented custom-data persistence model.

## Affected types

| Type | Ingress | Current registration | Parquet write |
|------|---------|----------------------|---------------|
| `BinanceFuturesLiquidation` | WS `forceOrder` (`<symbol>@forceOrder` / `!forceOrder@arr`) | `ensure_custom_data_json_registered` only | ❌ |
| `BinanceFuturesTicker` | WS `ticker` stream | `ensure_custom_data_json_registered` only | ❌ |
| `BinanceFuturesOpenInterest` | HTTP `RequestCustomData` snapshot | `ensure_custom_data_json_registered` only | ❌ |

All three are defined with manual `CustomDataTrait` impls in `crates/adapters/binance/src/data_types.rs` and registered only via JSON in `register_binance_custom_data()`.

## Expected behavior

Parity with other adapter-native custom families that already support catalog persistence, for example:

- `DeribitVolatilityIndex` (`#[custom_data(pyo3)]` + `ensure_custom_data_registered::<T>()`)
- `HyperliquidOpenInterest` (`#[custom_data(pyo3)]` + arrow feature path)

After registration, `ParquetDataCatalog::write_custom_data_batch` should succeed and PyO3 `ParquetDataCatalog.query(...)` should return typed rows.

## Actual behavior

`prepare_custom_data_batch` fails with:

```text
Custom data type "BinanceFuturesLiquidation" is not registered for Arrow encoding;
call register_custom_data_class or ensure_custom_data_registered before writing
```

The same applies to `BinanceFuturesTicker` and `BinanceFuturesOpenInterest`.

Runtime subscribe/request paths already work; only persistence/readback is missing.

## Proposed approach

1. Add an `arrow` feature to `nautilus-binance` (same pattern as `nautilus-deribit` / `nautilus-hyperliquid`).
2. Migrate these types from manual `CustomDataTrait` impls to `#[custom_data(pyo3, stub_module = "nautilus_trader.adapters.binance")]`:
[...]

PR #4336, opened 2026-06-27
Fix Binance futures custom data typed catalog readback

graceyangfan:
## Summary

Follow-up for #4297.

The upstream Binance Arrow/Parquet work already enabled Rust-side write support for Binance futures custom data, but Python typed catalog readback was still incomplete.

In particular, these flows were still broken:

```python
catalog.query(BinanceFuturesTicker, identifiers=[...])
catalog.query(BinanceFuturesLiquidation, identifiers=[...])
catalog.query(BinanceFuturesOpenInterest, identifiers=[...])
```

Rust could write the parquet files, but Python could not reconstruct typed objects through `ParquetDataCatalog.query(...)` because the Python-side Rust custom type registration/readback path was still missing.

This PR closes that remaining gap for:

- `BinanceFuturesOpenInterest`
- `BinanceFuturesLiquidation`
- `BinanceFuturesTicker`

## What This PR Does

This PR makes Python typed catalog readback work end-to-end for the three Binance futures custom data types above.

Specifically, it:

- registers the Binance futures Rust custom data types needed for Python typed query readback
- wires the Binance adapter import path into the Rust custom type registry used by catalog reconstruction
- updates `ParquetDataCatalog` to recognize Rust custom-data layout for:
  - typed query readback
  - metadata APIs
  - `list_data_types()`
- adds regression coverage for catalog round-trip and typed Python reconstruction

## Why This Is Needed

Before this change, the feature was only half complete:

- Rust could emit Binance custom data
- Rust could write parquet rows into the catalog
- but Python `ParquetDataCatalog.query(...)` still could not read those rows back as typed objects

So from a user perspective, the Arrow/Parquet support was not fully usable yet for research/readback workflows.

## Scope

This PR is intentionally narrow and only covers the three Binance futures custom data types that already participate in the Rust Arrow/Parquet typed-query path:

- `BinanceFuturesOpenInterest`
- `BinanceFuturesLiquidation`
- `BinanceFuturesTicker`

`BinanceFuturesOpenInterestHist` is intentionally out of scope here. It is currently on a separate JSON-style path rather than the same typed Arrow query path, so including it in this PR would broaden scope beyond #4297.

## Tests / Validation

Passed:

- targeted catalog round-trip tests for Binance futures custom data
- catalog metadata/listing regression coverage
- Python strategy-facing typed query smoke test
- existing backtest node custom-data streaming tests
- targeted Binance adapter tests for ticker, liquidation, and open-interest custom data flows
- `pre-commit run --files ...` on touched files


cjdsellers:
Hi @graceyangfan,

Thanks for the PR, and for the clear root-cause write-up. I agree with the direction: closing typed Python readback for Rust-written Binance futures custom data is a natural gap closure.

One compatibility issue before this lands. `BinanceBar` is already registered into `_RUST_CUSTOM_TYPE_REGISTRY`, so the new `_make_path()` / `get_file_list_from_data_cls()` branch routes it through `data/custom/BinanceBar`, while existing Python-written catalogs use the old `data/custom_binance_bar` layout. That makes older `BinanceBar` data invisible unless this does dual-path discovery, or unless the registry separates Rust-layout query types from Python-layout custom serializers.

The same layout split also needs a pass through the broad maintenance helpers. `_extract_data_cls_and_identifier_from_path()` still parses `data/custom/BinanceFuturesTicker/<identifier>` as type `custom` and identifier `BinanceFuturesTicker`, and `delete_data_range(..., identifier=None)` still searches for `data/{class_to_filename(data_cls)}`. Could you add coverage for legacy `BinanceBar` lookup and one all-identifier maintenance path on `data/custom/{TypeName}`?

graceyangfan:
Addressed the review feedback in a24550d4a0.

Changes:
- taught ParquetDataCatalog helper paths to handle Rust custom layout consistently
- fixed reverse path extraction for data/custom/{TypeName}/...
- updated delete/file-discovery helpers to preserve legacy BinanceBar catalog compatibility
- removed the invalid register_custom_data_class(...) call for Binance PyO3 classes

Validation:
- targeted pytest coverage passed
- pre-commit --all-files passed

cjdsellers:
Hi @graceyangfan,

Thanks again for the work here, and for turning around the review feedback.

After further consideration following yesterday's roadmap and v2 planning, I don't think it makes sense to spend more energy on this catalog compatibility path this late in the piece. The new catalog for v2 is ready, but we are holding that PR until after the v2 cutover, which is now imminent, so I would rather revisit this once that lands than add another bridge in the current catalog.

We're going to hold off on merging this PR for now. Please let me know if you think I am missing a near-term blocker.

PR #4338, opened 2026-06-27
Add BorrowRate data type with live subscription support

dxwil:
# Pull Request

## Summary

Adds a new `BorrowRate` data type representing the cost of borrowing a margin-eligible currency at a venue. It's deliberately generalized across asset classes - `accrual_interval` captures the period over which interest accrues (e.g. 1h for most crypto exchanges, daily for IBKR) and `rate` is annualized. This first PR wires up the full live data path: the model type with Python bindings, msgbus topics, and live subscribe/unsubscribe, including support in the Bybit data client. These changes go in conjunction with my previous PR which adds `BorrowBalance` for accounting.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

The `BorrowRate` type follows the same patterns as the existing `Data` types (construction, equality/hash, Display, serde/msgpack/JSON round-trips, Python bindings, actor/strategy dispatch), so it slots into the existing machinery with one key difference: it is keyed by currency + venue rather than instrument_id, since a borrow rate applies to a currency at a venue rather than a single instrument.

A few deliberate design notes:
- Subscribe, not request. Although most crypto exchanges expose borrow rates via their REST API, borrow rates change on a fixed cadence (typically hourly), so modelling this as a subscription rather than a one-shot request_* is the better fit. The data client can spawn a polling task aligned to the accrual boundary and stream updates. This is implemented in the Bybit data client (hourly-aligned polling, plus an initial fetch on subscribe).
- Scope of this PR. Intentionally limited to the live data path to keep the initial addition reviewable. It does not yet add cache support or catalog/Arrow serialization, those will follow in separate PRs to avoid cluttering this one.

Manually verified live subscribe against Bybit.


cjdsellers:
Hi @dxwil,

Thanks for the PR, and clear write-up. I like this direction. Borrow and financing data is common enough across crypto spot margin, DeFi lending markets, and broker margin that it makes sense to start treating the rate as first-class data rather than adapter-specific metadata (also relates to the larger accounting changes PR).

A few things I think need resolving before this lands:

- `UnsubscribeCommand::BorrowRates` is missing from `DataEngine::topic_has_remaining_subscribers`. With two actors subscribed to the same `(currency, venue)`, the first unsubscribe removes its msgbus handler and then falls through to the client unsubscribe path, so the Bybit adapter drops the currency from `borrow_rate_subs` while the second actor is still subscribed. This should mirror funding rates by checking `get_borrow_rate_topic(currency, venue)` against `exact_subscriber_count_borrow_rates`.

- The Python public model layer is not updated. `BorrowRate` is registered on `nautilus_pyo3` and used in the new `on_borrow_rate` callbacks, but it is not importable from `nautilus_trader.model` / `nautilus_trader.model.data` like `FundingRateUpdate` and the other user-facing data types. Could you add the public export path, or make the intended low-level import explicit and test it?

One broader API point to settle while the type is still new: `BorrowRate` is keyed by `currency + venue`, which fits crypto spot margin, broker cash margin loans, and DeFi lending markets. It does not fit instrument borrow rates for securities lending and short stock borrow, where the borrowed asset is the instrument itself. That may be fine if this type is deliberately scoped to cash or coin borrow and a later `InstrumentBorrowRate` handles stock borrow, but I’d like that boundary made explicit before the API shape hardens.

I don’t think this PR needs to solve account-specific tiers, utilization curves, collateral schedules, or DeFi liquidation terms etc. But just wanted to flag this for general consideration as we're building this out.

dxwil:
Hi @cjdsellers,

Thanks for taking a look. Fixed the msgbus subscriber count. For the python export, as far as I understood Cython is no longer being developed in favor of the Rust and pyo3, therefore I didn't add a Cython type. I simply made it re-export the pyo3 type so it is accesible in the normal python namespace. If you instead prefer to still have the Cython type, I can add it, no problem.

On scope: `BorrowRate` is keyed by `currency + venue` because a borrow rate is a property of a borrowable asset, not of a tradable instrument - keying it to an `instrument_id` would wrongly narrow it. Stock borrow is the awkward case, and that's really a limitation of the current instrument model: a stock has no asset-level identity. Its `instrument_id` is venue-bound (`AAPL.NASDAQ`, verified in the IBKR adapter) and the `Equity` behind it carries a mandatory settlement currency, so there's no way to name "AAPL shares" as a standalone borrowable. A future `InstrumentBorrowRate` keyed by `instrument_id` could cover short stock borrow, but it inherits that venue/currency coupling and still assumes a single side. So perhaps some asset-level identifier is in order, but I would say that is outside the scope of this PR (if anything like this comes around in the future `BorrowRate` can be adjusted).

dxwil:
Hi @cjdsellers,

Is there anything else blocking this from being merged?

cjdsellers:
> Hi @cjdsellers,
> 
> Is there anything else blocking this from being merged?

Hi @dxwil 

Likely nothing further blocking other than bandwidth and ensuring this is the right API shape to commit to for v2. I'll have another look over the next couple of days.

cjdsellers:
Hi @dxwil,

Just to restate the current status: this remains paused while I focus on the v2 cutover over the next couple of days. We can pick it up again after that once I've had a chance to go through the API design considerations for it. Thank you for your patience!

dxwil:
Hi @cjdsellers, 

Let me know when you start looking at this so I can take care of the merge conflicts.

cjdsellers:
> Hi @cjdsellers,
> 
> Let me know when you start looking at this so I can take care of the merge conflicts.

Hi @dxwil,

Thanks for checking in. The plan is to start looking at accounting after the next RC release (likely tomorrow but we'll see). Thanks for your patience on this one!

PR #4350, opened 2026-06-29
Add Interactive Brokers PyO3 type stub annotations

dfjmax:
## Summary

The Interactive Brokers adapter is the last Rust-backed adapter with no `pyo3-stub-gen` coverage: all `60` `#[pyclass]` and `35` `#[pymethods]` bindings were exposed to Python without matching stub annotations, so every config, factory, enum, error, the historical client, and the data/execution clients resolved as untyped (`Unknown`) and carried no constructor signatures.

This brings IB in line with the other Rust-backed adapters (e.g. betfair) by enabling `pyo3-stub-gen` and adding the derive attributes per the [Type stub annotations](https://nautilustrader.io/docs/latest/developer_guide/rust/#type-stub-annotations) convention.

## Changes

- `crates/adapters/interactive_brokers/Cargo.toml`: add `pyo3-stub-gen` to the `python` feature and as an optional workspace dependency.
- `gen_stub_pyclass` on the `11` structs (configs, factories, historical client, gateway, instrument provider, execution/data cores).
- `gen_stub_pyclass_enum` on the `49` enums (error kinds, config enums, and the `common/enums/*` IB domain enums — including the set produced by the `macro_rules!` in `common/enums/misc.rs`).
- `gen_stub_pymethods` on every `#[pymethods]` impl in `python/`.

Stubs emit to `nautilus_trader.adapters.interactive_brokers`, matching the existing `nautilus_trader.core.nautilus_pyo3.interactive_brokers` runtime module already declared on each `#[pyclass]`.

## Verification

- `cargo check -p nautilus-interactive-brokers --features python` — all annotations expand cleanly (incl. the `macro_rules!`-generated enums).
- Full pre-commit suite green: `cargo fmt`, `cargo clippy`, `cargo doc`, `cargo machete` (confirms `pyo3-stub-gen` is used), and all nautilus/pyo3 convention checks.


cjdsellers:
Hi @dfjmax,

Thanks for adding the IB PyO3 stub coverage. The annotations are broadly in the right shape, but I found two generated-stub gaps to fix before this lands:

  - `interactive_brokers/__init__.pyi` only exposes Rust enum variant names like `Delayed`, while the runtime also exposes aliases like `MarketDataType.DELAYED`. Existing Python
  examples use the uppercase aliases, so the stub should include those for `MarketDataType` and the IB enums with `#[classattr]` aliases.
  - The built wheel enables `extension-module`, which enables the adapter `gateway` feature, so `DockerizedIBGateway` and `ContainerStatus` are runtime exports too. The committed stub
  should include them, or the stub generation feature shape needs to match the build shape.

If the branch has merge conflicts from a recent PR landing, please rebase on current `develop` before regenerating and pushing the stubs.

dfjmax:
Hi @cjdsellers, before I push an update, two quick questions.

The uppercase names on the IB enums (like MarketDataType.DELAYED) were missing from the stub. The other adapters get theirs with rename_all = "SCREAMING_SNAKE_CASE", so I did the same on the IB enums and the names show up now. One catch: 9 IbTickType names were hand-spelled differently from what rename_all produces (e.g. LOW_13_WEEK becomes LOW13_WEEK), and I couldn't find any code that uses them. Is rename_all the way to go here, and are those 9 name changes OK?

DockerizedIBGateway and ContainerStatus only build under the gateway feature, and that feature only gets switched on by extension-module, which the stub step drops. So they never got compiled when the stub was generated. I'd switch on nautilus-interactive-brokers/gateway in the stub step so they're included. Sound right?

faysou:
Sounds right to me

dfjmax:
@faysou @cjdsellers 

For the uppercase aliases, I added the rename_all attribute (SCREAMING_SNAKE_CASE) to the IB enums so the stub exposes them like the other adapters. The macro-generated enums in misc.rs were still coming through in PascalCase because the stub generator couldn't see rename_all through the macro, so I inlined define_ib_i32_enum there to apply it directly. That also made the hand-written classattr alias blocks redundant, so I removed those.

For the gateway exports, I enabled nautilus-interactive-brokers/gateway in the stub step so DockerizedIBGateway and ContainerStatus get included.

The rename also broke a couple of PascalCase references in the interactive_brokers_pyo3 wrapper and one test, so I updated those. The 9 IbTickType names follow rename_all, as @faysou confirmed. 

Let me know if this looks good or if you'd like anything changed.

cjdsellers (approved):
Thank you for the follow-ups @dfjmax 👌

PR #4351, opened 2026-06-29
Fix remaining indicator rolling-window bugs

Martingale42:
## Summary

Extends the indicator rolling-window bug-fix lineage (#4239 `DonchianChannel` →
#4326 `RateOfChange` → #4333 `VerticalHorizontalFilter` / `OnBalanceVolume`) with
the remaining members of the family. Each indicator stored its rolling window in an
unbounded `Vec` or a fixed-1024-capacity `ArrayDeque` instead of bounding it to
`period` / `capacity` like the Cython reference's `deque(maxlen=...)`, so each
diverged from Cython once more than the window size of observations was processed.

One commit per indicator:

- **SpreadAnalyzer** — the unbounded `Vec` of spreads made `fast_mean_iterated`
  receive a slice longer than its expected length, return `Err`, and panic on
  `.unwrap()` at the `(capacity + 1)`th quote (a live-trading crash). Bound the
  window to `capacity`.
- **EfficiencyRatio** — unbounded `inputs` and `deltas` made `net_diff` use the
  all-time-first price and `sum_deltas` sum every delta ever seen; `reset()` also
  left `deltas` populated. Bound both to `period` and clear `deltas` on reset.
  Since `AdaptiveMovingAverage` embeds this indicator, its Kaufman efficiency input
  is corrected too.
- **RelativeVolatilityIndex** — the standard deviation was computed over up to 1024
  prices while the mean came from the `period`-window SMA. Bound the price window to
  `period`.
- **FuzzyCandlesticks** — four buffers (candle lengths, body/upper-wick/lower-wick
  percents) grew to 1024, inflating the `sum / period` means and the std
  denominators. Bound all four to `period`.

All now match the Cython `deque(maxlen=period)` (or `maxlen=capacity`) semantics.

## Tests

Each fix adds a regression test feeding more than `period` / `capacity`
observations — the gap the existing per-indicator tests missed by stopping at
exactly the window size.

- `cargo nextest run -p nautilus-indicators` — all pass (including `AdaptiveMovingAverage` / AMAT unchanged).
- `cargo clippy -p nautilus-indicators --all-targets --all-features -- -D warnings`
  clean; nightly `cargo fmt`.

### DIFF
--- a/crates/indicators/src/ratio/efficiency_ratio.rs
+++ b/crates/indicators/src/ratio/efficiency_ratio.rs
@@ -79,6 +79,7 @@ impl Indicator for EfficiencyRatio {
     fn reset(&mut self) {
         self.value = 0.0;
         self.inputs.clear();
+        self.deltas.clear();
         self.initialized = false;
     }
 }
@@ -99,6 +100,13 @@ impl EfficiencyRatio {
 
     pub fn update_raw(&mut self, value: f64) {
         self.inputs.push(value);
+        // Bound the inputs window to `period`, matching the Cython
+        // `deque(maxlen=period)`; otherwise the net change below is measured from
+        // the all-time-first price instead of `period` bars back.
+        if self.inputs.len() > self.period {
+            self.inputs.remove(0);
+        }
+
         if self.inputs.len() < 2 {
             self.value = 0.0;
             return;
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @Martingale42,

Thanks for the PR, and for carrying the rolling-window sweep across the remaining indicators. Most of this reads right to me, but one `SpreadAnalyzer` detail still needs fixing before this lands.

The new eviction keeps `spreads` at `capacity`, but `average` still updates through `fast_mean_iterated(..., drop_left=false)` after removing index 0. At capacity, that subtracts `values[length - 1]`, which is the current spread, so the average stops changing for non-constant spreads. A regression with varying spreads past capacity should catch this. Please either carry the dropped value into the update path or recompute the average from the bounded window after eviction.

Martingale42:
Fixed. After eviction the average now recomputes from the bounded window (`fast_mean(&self.spreads)`) instead of the incremental `fast_mean_iterated` update. The incremental path subtracted `values[length - 1]` — the spread just pushed — once the window was at capacity, so the average froze for non-constant spreads. The Cython `SpreadAnalyzer` shares this quirk: it appends to the `deque(maxlen)` and then calls `fast_mean_iterated(..., drop_left=false)`, so the recompute intentionally diverges from that reference to stay correct. Recompute is O(capacity), the same cost as the existing `Vec::remove(0)`.

Added `test_average_tracks_varying_spreads_past_capacity`, which pushes 15 increasing spreads through a capacity-10 window and asserts the average equals the mean of the last 10 (`0.105`) rather than the frozen `0.055`. Also updated the constant-spread test's baked value to the recomputed float.

cjdsellers (approved):
Thanks for the follow-ups @Martingale42 👌

PR #4352, opened 2026-06-29
Add Ulcer Index, Omega Ratio, VaR and Expected Shortfall statistics

Martingale42:
## Summary

Adds four portfolio statistics to the Rust `analysis` crate, each authored directly
in Rust, and wires them into the PyO3 `PortfolioAnalyzer` dispatch.

- **Ulcer Index** — downside risk as the root-mean-square of the percentage
  drawdowns of the cumulative-return equity curve (`sqrt(mean(D_i^2))`). Reuses
  `MaxDrawdown`'s cumulative-equity convention (fractional drawdowns, same scale).
- **Omega Ratio** — probability-weighted ratio of gains to losses about a threshold
  `θ` (`sum(max(r - θ, 0)) / sum(max(θ - r, 0))`), default `θ = 0`. Parameterized
  like `SharpeRatio`.
- **Value at Risk** — historical (empirical) VaR: the return-distribution quantile
  at `1 - confidence`, linear interpolation matching `numpy.percentile`. Default
  confidence `0.95`.
- **Expected Shortfall** (Conditional VaR) — mean of returns at or below the VaR
  threshold; composes `ValueAtRisk`, mirroring how `CalmarRatio` composes `CAGR`
  and `MaxDrawdown`.

All four are added to the `register_statistic` / `deregister_statistic` PyO3
dispatch in `crates/analysis/src/python/analyzer.rs` so Python users can register
them with a `PortfolioAnalyzer` (otherwise they fail with `Unknown statistic type`).

## Details

- Convention: the returns-performance ratio (`OmegaRatio`) is registered in the
  default analyzer alongside Sharpe/Sortino; the downside-risk metrics
  (`UlcerIndex`, `ValueAtRisk`, `ExpectedShortfall`) are exposed but not registered
  in the default analyzer, consistent with `MaxDrawdown`.
- Results are expressed as returns / on the same fractional scale as `MaxDrawdown`.

## Tests

- `cargo nextest run -p nautilus-analysis` — all pass (per-statistic cases: name,
  empty, known value, plus an `ES <= VaR` invariant).
- Known values cross-checked against numpy (Ulcer Index; VaR `-0.091` / ES `-0.10`).
- Python registration tests for all four in
  `python/tests/unit/analysis/test_analysis.py`.
- `cargo clippy -p nautilus-analysis --all-targets --all-features -- -D warnings`
  clean; nightly `cargo fmt`; v2 stubs regenerated via `make build-debug-v2`.


cjdsellers:
Hi @Martingale42,

Thanks for the PR. The new statistics are useful, and I think they are reasonable to expose as opt-in analyzer stats. I would keep them out of the default analyzer for now though: `ValueAtRisk`, `ExpectedShortfall`, and `UlcerIndex` are tail/downside risk views, and `OmegaRatio` is threshold-sensitive enough that I do not think it should appear in default returns stats.

A few things to fix before this lands:

- The branch is currently conflicting with `develop`. When rebasing, please keep the existing `ReturnsSkewness` and `ReturnsKurtosis` wiring in `PortfolioAnalyzer::default`, `analysis` PyO3 exports, Python stubs, tests, and `register_statistic` / `deregister_statistic`. The current PR head appears to drop some of that wiring.
- `ValueAtRisk::new` and `ExpectedShortfall::new` document confidence as `(0, 1)` but accept any `f64`. Invalid values can flow into `percentile_linear` and panic via out-of-range indexing. Please validate finite confidence in `(0.0, 1.0)` and add Rust/Python coverage for invalid confidence values.
- Please remove `OmegaRatio` from `PortfolioAnalyzer::default` and keep it registerable like the other non-default risk stats.

The formulas and registration direction otherwise look sensible to me. Let me know if anything is unclear.

Martingale42:
Addressed all three points, and did a self-review pass that tightened a few more things.

**Requested changes**

- Rebased onto `develop`. The `ReturnsSkewness`/`ReturnsKurtosis` wiring is preserved in `PortfolioAnalyzer::default`, the analysis PyO3 exports, the stubs, `test_analysis.py`, and `register_statistic`/`deregister_statistic`; the four new statistics are added alongside it.
- `ValueAtRisk` and `ExpectedShortfall` now validate `confidence` via `new_checked`, which errors unless the value is finite and in `(0, 1)`. `new` delegates to it and panics on invalid input; the PyO3 `__new__` maps the error to `ValueError`. Covered with Rust cases (`0.0`, `1.0`, `1.5`, `-0.5`, NaN, inf) and a parametrized Python test.
- Removed `OmegaRatio` from `PortfolioAnalyzer::default`; it stays registerable through the dispatch like the other non-default risk stats.

**Self-review follow-ups**

- `ExpectedShortfall` now downsamples and sorts the returns once, taking both the VaR threshold (via a shared `percentile_linear`) and its tail from the same value-sorted sample, instead of composing `ValueAtRisk` (which downsampled a second time) and re-filtering an unsorted map. Threshold and tail are now consistent by construction, and a redundant downsample + sort and a tail `Vec` allocation are gone. Added a multi-element-tail regression (the prior tests only exercised a single-element tail).
- Applied the same finite-value validation to `OmegaRatio`'s `threshold` (rejects NaN/inf), for symmetry with the confidence stats.
- Dropped two unreachable branches: `percentile_linear`'s `n == 1` case (already covered by the `lower == upper` check) and `ExpectedShortfall`'s empty-tail branch.

One pre-existing thing I noticed but left as-is: deregistering a parameterized statistic with a different parameter than it was registered with silently no-ops, since the analyzer keys off `name()` (e.g. register `ValueAtRisk(0.99)`, deregister `ValueAtRisk()` resolves to a different key). This already applies to `SharpeRatio`'s period and now extends to these; happy to open a separate issue/PR if you'd like it addressed.

cjdsellers:
Hey @Martingale42 

Thanks for the follow-ups and for going above and beyond with the self-review.

Looks like there are a few conflicts after merging some previous PRs. If you could rebase and resolve those when you get a chance, then this is good to merge!

PR #4353, opened 2026-06-29
Fix Polymarket RTDS retained-subscription recovery and simplify live multi-symbol subscriptions

### LINKED ISSUE #4343: Polymarket RTDS reconnect replay can be skipped and retained subscriptions can stay dark
## Summary

Polymarket RTDS reconnect handling currently has a lifecycle gap that can leave retained custom-data subscriptions dark after a transport flap.

## Confirmed issues

1. `crates/adapters/polymarket/src/rtds.rs` checks for the text literal `"reconnected"`, but the shared `nautilus-network` WebSocket client emits the Rust-side reconnect sentinel `__RECONNECTED__` to channel consumers. That means retained RTDS replay is skipped on normal transparent reconnects.
2. `run_message_loop()` exits permanently when its raw message channel closes. If retained RTDS subscriptions still exist, recovery is then dependent on a future custom-data subscribe/unsubscribe call rather than being proactively restored.
3. The main Polymarket data WebSocket `Reconnected` path currently logs only. It does not ensure the RTDS side is re-established when retained RTDS subscriptions exist.

## Additional finding

Live RTDS multi-symbol filtered subscriptions are not reliable in practice. When multiple per-symbol filtered RTDS subscriptions are active at the same time, later symbols can stop receiving live updates. Because of that, the correct fix is not only to repair reconnect replay, but also to intentionally consolidate live RTDS transport state at the topic boundary and keep symbol selection in retained local state.

In other words, this is not just a convenience refactor. It is part of making RTDS custom-data subscriptions correct and stable under the real venue behavior.

## Expected behavior

- RTDS should continue to rely on the existing `WebSocketClient` auto-reconnect behavior.
- RTDS should consume the standard Rust reconnect sentinel and replay retained subscriptions through the existing retained-state reconciliation path.
- If the RTDS message loop terminates unexpectedly while retained RTDS subscriptions still exist, the feed should invalidate the dead generation and schedule recovery through the same retained-state sync path.
- The main Polymarket data WebSocket reconnect event should nudge RTDS recovery when retained RTDS subscriptions exist, without introducing a parallel reconnect mechanism.
- `crypto_prices` and `equity_prices` live transport should be maintained at one topic-level subscription per topic, with symbol filtering handled locally from retained desired state.

## Why this matters

This is a correctness issue for Polymarket RTDS custom data. After a disconnect/reconnect cycle, existing subscriptions can stop emitting until another subscription mutation happens to wake the feed back up. Separately, per-symbol filtered live RTDS subscriptions can also leave multi-symbol consumers partially dark even without a fresh subscription mutation.

## Suggested validation

- Regression test proving the RTDS reconnect sentinel triggers replay.
- Regression test proving message-channel closure with retained subscriptions re-establishes the RTDS session without requiring a new subscribe call.
- Regression test proving server-side disconnect/reconnect restores multi-symbol RTDS delivery.
- Regression test proving reconnect restores only still-retained symbols, not previously unsubscribed ones.

graceyangfan:
## Summary

Fix Polymarket RTDS retained-subscription recovery after reconnects and simplify live multi-symbol RTDS transport.

Closes #4343.

## What changed

- consume the shared Rust `__RECONNECTED__` sentinel instead of the stale string literal check
- keep RTDS reconnect ownership in the shared `WebSocketClient`; the feed only reconciles retained desired state
- self-heal when the RTDS message channel closes while retained subscriptions still exist
- nudge RTDS recovery from the main Polymarket WS reconnect path instead of depending on a future subscribe mutation
- collapse live `crypto_prices` and `equity_prices` transport state to one topic-level subscription per topic, then filter per symbol locally

## Why this design

Polymarket RTDS reconnects are normal, not exceptional, so the safest design is the one with the fewest transport states.

This PR does not add a second reconnect mechanism. It removes adapter-side fragility by reducing RTDS live transport to topic-level state and replaying retained symbol intent through a single reconciliation path.

That shape is also driven by real venue behavior: live RTDS multi-symbol filtered subscriptions are not reliable in practice. Consolidating live transport at the topic boundary avoids per-symbol wire churn, prevents retained subscriptions from staying dark after reconnect, and keeps replay identical for fresh connect and reconnect.

## Behavioral changes

- existing RTDS retained subscriptions now recover without requiring a new subscribe call
- `crypto_prices` and `equity_prices` multi-symbol subscriptions now share one live wire subscription per topic
- unsubscribe tears down the topic only after the last retained symbol for that topic is removed
- reconnect replay and fresh-connect sync both use the same retained-state reconciliation path

## Validation

- Rust regression tests for reconnect sentinel replay, channel-close self-heal, closing-vs-reconcile races, and topic-level incremental subscribe/unsubscribe behavior
- added tests proving a newly added symbol still emits its first live update under shared topic-level transport
- added fault-injection regressions proving:
  - server-side disconnect/reconnect restores multi-symbol RTDS delivery
  - reconnect restores only still-retained symbols, not previously unsubscribed ones
- rebuilt the root PyO3 module from this branch and exercised a TradingNode probe with `btc,eth`; both streams remained live and balanced, with real upstream reconnect events observed during the run
- long-run live TradingNode validation completed successfully:
  - runtime: `00:05 -> 10:25` (`10h20m`)
  - RTDS steady state: `5` assets, `179` points, latest age `~1s`
  - reconnects observed: `6`, failures: `0`
  - crashes: `0`
  - entries: `0` during this run because the current gap threshold remained too strict for signal generation


cjdsellers:
Hi @graceyangfan,

Thanks for the PR, and for the clear RTDS reconnect write-up. The retained-state reconcile path and topic-level transport direction make sense to me.

I found one shutdown race before this lands. In `crates/adapters/polymarket/src/rtds.rs:484`, the reconcile worker waits on `reconcile_notify.notified().await` before checking `closing`. If `disconnect()` sets `closing = true` and calls `notify_waiters()` while the worker is inside `reconcile_once()`, the worker misses that wakeup and can sleep forever when it returns to the loop, so normal shutdown waits 2s and then aborts it. Could you check `closing` before awaiting, or use a notification path that leaves a permit for late waiters?

graceyangfan:
Thanks, good catch. You were right about the missed-notify shutdown race in `run_reconcile_loop()`.

I updated the worker loop so it checks `closing` before waiting on `reconcile_notify.notified().await`, and still checks again immediately after wakeup. That closes the specific window where `disconnect()` can set `closing = true` and call `notify_waiters()` while the worker is still inside `reconcile_once()`, then have the worker return to the top of the loop and sleep forever waiting for another notify.

I also added a regression test that exercises that exact sequence: the reconcile worker is already inside `reconcile_once()`, shutdown arrives, no second notify is sent, and the worker now exits cleanly instead of depending on the 2s abort timeout.

cjdsellers (approved):
Thank you for the follow-ups @graceyangfan

PR #4251, opened 2026-06-11
Add benchmark-relative portfolio statistics

mahimn01:
# Pull Request

- [x] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

Implements the benchmark-relative performance statistics agreed in #4229: `BetaRatio`, `Alpha` (Jensen's), `InformationRatio`, `TrackingError`, and `TreynorRatio`, as Rust-core `PortfolioStatistic`s with PyO3 bindings.

- Adds an optional trait method `calculate_from_returns_with_benchmark` (default `None`, so analyzer loops filter by `Option` like they already do for `calculate_from_positions` results) plus a shared `align_returns` helper that daily-bins both series and inner-joins on shared timestamps (non-overlapping points dropped, not zero-filled).
- Each statistic overrides the single-source methods to return `None` (panic-safe per the #3941/#4174 precedent) and is **not** default-registered — existing analyzer and tearsheet output is unchanged when no benchmark is provided.
- Adds a stateless `get_performance_stats_returns_vs_benchmark(benchmark)` to the Rust and PyO3 analyzers; the pure-Python `PortfolioAnalyzer` gains a thin getter that delegates to the registered PyO3 statistics (no parallel Python implementations).
- Thin tearsheet wiring: when `benchmark_returns` is provided, benchmark statistics are computed from the same resolved returns series used for the equity chart and merged into the returns stats.
- Includes PyO3 class registration, `nautilus_trader.analysis` exports, and stub updates.

### Conventions

| Statistic | Per-period | Annualization |
|---|---|---|
| `BetaRatio` | Cov(r, b) / Var(b), ddof=1 | none |
| `Alpha` | (mean_r − rf) − β·(mean_b − rf) | geometric: (1 + α)^period − 1 |
| `InformationRatio` | mean(r − b) / std(r − b, ddof=1) | × √period |
| `TrackingError` | std(r − b, ddof=1) | × √period |
| `TreynorRatio` | (CAGR − rf_annual) / β | CAGR numerator, rf annualized internally |

Per-period formulas match `empyrical`; IR and TE are annualized for consistency with `SharpeRatio` / `ReturnsVolatility`, with the period in the display name (e.g. `Information Ratio (252 days)`). Degenerate inputs (<2 aligned points, flat benchmark, zero active std for IR) return `NaN`, matching existing conventions.

Up/Down Capture is left as the follow-up discussed in #4229.

### Verification

[...]

Issue #4229, opened 2026-06-08
Add benchmark-relative performance statistics (Beta, Alpha, Information Ratio, Tracking Error, Treynor)


mahimn01:
## Motivation

The Rust analysis crate (`crates/analysis`) has solid absolute return/risk statistics (Sharpe, Sortino, Calmar…) but no **benchmark-relative** ones. Meanwhile the Python tearsheet layer (`nautilus_trader/analysis/`) already carries a benchmark series — `TearsheetConfig.include_benchmark`/`benchmark_name` and `create_tearsheet(benchmark_returns=…)` — but uses it only for a plot overlay (`(1 + benchmark_returns).cumprod()`); no statistic consumes it. I'd like to close that gap with a set of CAPM / active-management metrics, building on the `PortfolioStatistic` framework I worked in for #4141 / #4174.

## Proposed metrics

(Up/Down Capture deferred — see below.)

| Metric | Per-period | rf | Annualization |
|---|---|---|---|
| Beta | Cov(r,b) / Var(b) | – | none |
| Alpha (Jensen) | mean(r−rf) − β·mean(b−rf) | 0 default | geometric, (1 + α)^period − 1 |
| Information Ratio | mean(r−b) / std(r−b, ddof=1) | – | ×√period |
| Tracking Error | std(r−b, ddof=1) | – | ×√period |
| Treynor | (annualized return − rf) / β | 0 default | CAGR numerator |

## Conventions

I'd match [`empyrical`](https://github.com/quantopian/empyrical) (the most-cited reference; pyfolio/ffn re-export it) where it has an opinion, and be explicit where I'd deliberately diverge:

- **Beta** — `Cov(r,b)/Var(b)`; the ratio is ddof-invariant, so it matches empyrical regardless of population vs sample moments (rf=0 specialization of its excess-return beta).
- **Alpha (Jensen)** — annualized **geometrically**, `(1 + per-period α)^period − 1`, matching `empyrical.alpha_aligned` exactly.
- **Information Ratio & Tracking Error** — per-period formulas match empyrical's `excess_sharpe` (mean/std of active return, ddof=1), but I'd annualize both **×√period** for consistency with this crate's own `SharpeRatio` / `ReturnsVolatility`. Note **empyrical leaves IR/TE periodic**, so anyone validating against empyrical would divide our value by √period. (Happy to ship the periodic form instead if you'd prefer empyrical parity.)
[...]

PR #4354, opened 2026-06-29
Add Up/Down Capture ratio portfolio statistics

mahimn01:
# Pull Request

## Summary

Adds the Up/Down Capture ratios as the follow-up to #4251 (left for later in #4229): `UpCaptureRatio` and `DownCaptureRatio`, as Rust-core `PortfolioStatistic`s with PyO3 bindings.

- Both reuse the benchmark-aware `calculate_from_returns_with_benchmark` path and the `align_returns` helper from #4251 (daily-bin + inner-join on shared timestamps), and override the single-source methods to return `None` (panic-safe, per the #4174 precedent). Like the #4251 statistics they are **not** default-registered, so existing analyzer/tearsheet output is unchanged unless a benchmark is provided.
- A shared `capture_ratio(returns, benchmark, period, MarketSide)` plus `geometric_annualized_return` live in `up_capture_ratio.rs` and are reused by `down_capture_ratio.rs`, mirroring how `beta()` and `active_return_stats()` are shared between sibling statistics.
- Includes PyO3 class registration, `nautilus_trader.analysis` exports, and stub updates.

### Convention

Both follow `empyrical`'s `up_capture` / `down_capture`: filter the aligned periods to those where the benchmark return is positive (up) or negative (down), then take the ratio of the strategy's geometric annualized return to the benchmark's over that subset (annualization exponent `period / m`, with `m` the size of the filtered subset — the same geometric CAGR `TreynorRatio` already uses). Display names carry the period, e.g. `Up Capture Ratio (252 days)`.

This is the empyrical convention; it differs from Morningstar's (a ratio of cumulative, non-annualized returns) — the two coincide only when the up and down subsets are equal length. The doc comments note this. Degenerate inputs (<2 aligned points, an empty up/down subset, or a benchmark subset whose annualized return is ~0) return `NaN`, matching the existing statistics.

### Verification

- `cargo test -p nautilus-analysis --lib` — 198 passed; the 16 new capture tests use known values derived from scratch in Python and cross-checked against `empyrical` 0.5.5 (they agree to full f64 precision), plus partial-overlap inner-join and NaN-edge cases
- `cargo clippy -p nautilus-analysis --lib --tests --all-features --no-deps -- -D warnings` — clean
- `cargo fmt -p nautilus-analysis --check`
- `ruff check` on the touched Python exports/stubs

Follow-up to #4251 (#4229).


cjdsellers:
Hi @mahimn01,

Thanks for the PR. The capture-ratio implementation reads well to me, and keeping these as opt-in benchmark-relative stats makes sense.

One small integration gap before this lands: the v2 generated analysis stub still needs to be refreshed so `python/nautilus_trader/analysis/__init__.pyi` includes `UpCaptureRatio` and `DownCaptureRatio`. Since the branch is currently conflicting with `develop`, could you please rebase, run `make py-stubs-v2`, and move the `RELEASES.md` entry into the current unreleased section?

mahimn01:
Thanks @cjdsellers, all three are done and pushed.

Rebased onto develop (resolved the register/deregister arms alongside the new returns skewness/kurtosis stats), moved the `RELEASES.md` entry into the 1.231.0 unreleased section, and refreshed the v2 analysis stub so `python/nautilus_trader/analysis/__init__.pyi` now exports `UpCaptureRatio` and `DownCaptureRatio`.

One honest note on the stub. I couldn't run `make py-stubs-v2` locally because the repo pins uv `==0.11.25` and my machine is on 0.11.14, so I added the two entries by hand to match the generated format (same shape as the sibling benchmark stats like `InformationRatio`). If a regen on your side shows any diff I'd expect it to be environment-only, and I'm glad to sort out the uv pin and rerun if you'd prefer.

cargo test (208 passed), clippy with `-D warnings`, fmt, and the check-formatting hooks are all green.

PR #4366, opened 2026-07-01
Measure position-reconciliation grace on the monotonic clock

folknor:
### The bug

`check_position_discrepancy` has a grace window. Idea's simple and correct: if we *just* touched a position locally, don't panic about a cache-vs-venue mismatch yet — the local pipeline is probably mid-flight catching up. It suppresses the discrepancy for `position_check_threshold_ns` and gives things a moment to settle before it synthesizes an `EXTERNAL` reconciliation position.

The catch is *which clock measures that moment*. The window is a real-time settling period — you want N real seconds of cover, full stop. But it was measured off `self.clock`: `last_activity` stamped from `self.clock`, threshold compared against a `self.clock` delta.

`self.clock` is not guaranteed to tick at wall rate. A live node can be handed a non-wall clock through a clock factory — a simulated or replay venue driving the node's clock. Let that clock run fast, or sit it on a different epoch, and `ts_now - last_activity` stops being real elapsed time. The threshold stops meaning what it says. Run the clock at speed and the window shrinks by that factor; it can be gone before the first reconciliation poll even fires.

And the runtime already has a clock for exactly this. The live runner loop (`LiveNode::run`) schedules its wake-ups on the monotonic `dst::time` clock — that's what the codebase reaches for whenever it needs real elapsed time. The grace is a real-time window too; it was just the one spot still measuring that window on `self.clock`.

### The fix

Measure the grace on the monotonic clock. Stamp `last_activity` from `dst::time::Instant`, compare with `elapsed()` against the threshold as a `Duration`. Now N seconds means N real seconds of cover at any clock speed or epoch. `record_position_activity` stops taking a `ts_event` and stops reading `self.clock`; the activity map holds monotonic instants.

This is a live-only path — the continuous position check never runs under backtest — so a monotonic clock here costs nothing in backtest determinism.

### The latent panic it also removes

The old code had a second, worse bug that this removes on the way past — @cjdsellers spotted it in review. `last_activity` was stamped from venue timestamps (`ts_event` / `ts_last`) while `ts_now` came from `self.clock`: two different axes, subtracted with plain `UnixNanos` arithmetic, which panics on underflow. A venue clock running even slightly ahead of local time is enough — `ts_now - last_activity` underflows and takes the reconciliation task down with it.

Stamping on the monotonic clock and comparing with `elapsed()` removes the mixed-axis subtraction entirely, so there's nothing left to underflow. The order-side `record_local_activity` already sidestepped both problems — receipt-time stamp, saturating `duration_since` — so this brings the position side in line with it.

### Test

Added a regression test. It records local activity, then jumps `self.clock` ~954 days forward in a single step — a trading clock that's raced far ahead while no real time has passed — and asserts the grace still holds. I pointed the grace back at `self.clock`, watched the test fail (two phantom reconciliation events), put the monotonic clock back, watched it pass.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -18,7 +18,7 @@
 //! This module provides the execution manager for reconciling execution state between
 //! the local cache and connected venues, as well as purging old state during live trading.
 
-use std::{cell::RefCell, fmt::Debug, rc::Rc, str::FromStr, sync::LazyLock};
+use std::{cell::RefCell, fmt::Debug, rc::Rc, str::FromStr, sync::LazyLock, time::Duration};
 
 use indexmap::{IndexMap, IndexSet};
 use nautilus_common::{
@@ -26,6 +26,7 @@ use nautilus_common::{
     clients::ExecutionClient,
     clock::Clock,
     enums::{LogColor, LogLevel},
+    live::dst,
     log_info,
     messages::{
         ExecutionReport,
@@ -271,7 +272,8 @@ pub struct ExecutionManager {
     recon_check_retries: IndexMap<ClientOrderId, u32>,
     ts_last_query: IndexMap<ClientOrderId, UnixNanos>,
     order_local_activity_ns: IndexMap<ClientOrderId, UnixNanos>,
-    position_local_activity_ns: IndexMap<InstrumentAccountKey, UnixNanos>,
+    // Monotonic (`dst::time`) instants, not `self.clock`; see `record_position_activity`.
+    position_local_activity: IndexMap<InstrumentAccountKey, dst::time::Instant>,
     position_recon_retries: IndexMap<InstrumentAccountKey, u32>,
     recent_fills_cache: IndexMap<TradeId, UnixNanos>,
 }
@@ -305,7 +307,7 @@ impl ExecutionManager {
             recon_check_retries: IndexMap::new(),
             ts_last_query: IndexMap::new(),
             order_local_activity_ns: IndexMap::new(),
-            position_local_activity_ns: IndexMap::new(),
+            position_local_activity: IndexMap::new(),
             position_recon_retries: IndexMap::new(),
             recent_fills_cache: IndexMap::new(),
         }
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR, and for this careful clock-axis analysis.

The direction reads right to me: as you pointed out, the reconciliation loop already schedules on `dst::time`, so measuring the grace there lines the two up. I applied the branch locally and the position-check suite passes, including the new regression test.

Worth adding to the description: this also fixes a more severe latent bug. `last_activity` was stamped from venue timestamps (`ts_event` / `ts_last`) while `ts_now` came from `self.clock`, and `UnixNanos` subtraction panics on underflow, so a venue clock slightly ahead of local time could panic the reconciliation task. The order-side `record_local_activity` already avoids both problems by stamping receipt time and comparing with saturating `duration_since`.

A couple of things to fix before this lands:

- The updated comment in `test_position_check_activity_throttle_independent_per_account` says the activity is "stamped from the manager's clock", but it is stamped from the monotonic `dst::time` clock.
- The same rationale is currently stated four times (field comment, `record_position_activity` doc, check-site comment, test preamble). Please keep one trimmed version in the `record_position_activity` doc, reduce the others to a line each, and drop the em dashes per repo style.

One follow-up to note rather than block on: the order-side graces (`open_check_threshold_ns`, inflight, recent-fills TTL) still measure on `self.clock`, so the same argument applies there. I'd keep this PR scoped to positions and move the siblings separately.

folknor:
Yes, I discovered the latent bug while working on this, but I didn't mention it because I wasn't 100% sure - only 97% :joy: 

Thank you, I will fix up the code later today.

folknor:
> A couple of things to fix before this lands:

Thank you, I've addressed your concerns and fast-forwarded the branch.

> One follow-up to note rather than block on: the order-side graces (`open_check_threshold_ns`, inflight, recent-fills TTL) still measure on `self.clock`, so the same argument applies there. I'd keep this PR scoped to positions and move the siblings separately.

Indeed! I'm working on that now, separate PR incoming. It's actually a bit more than you scoped here, I hope you will be pleasantly surprised.

PR #4372, opened 2026-07-02
V2 wranglers report opaque errors for fixed-point raw overflow

MandalorianBatman:
## Summary

The v2 wranglers (`BarDataWranglerV2`, `QuoteTickDataWranglerV2`, `TradeTickDataWranglerV2`, `OrderBookDepth10DataWranglerV2`) compute
`raw = round(source * 10**p) * 10**(FIXED_PRECISION - p)` which is INVARIANT in `p` and always produces `source * 10**FIXED_PRECISION`. With `FIXED_PRECISION=16`, a typical NSE equity at 3812 INR (`price_precision=2`) produces a raw of 3.81e17 (fits int64), but `price_precision=0` produces 3.81e19 (overflows). Same problem for active 1-minute volumes: 140k shares × 10**16 = 1.4e21.

Previously the int64 overflow surfaced downstream as an opaque polars/pyarrow "Int128" error in the `.to_bytes()` conversion or in the catalog write, with no indication of which column or row was the source. This patch adds a pre-bytes overflow check that raises a clear `ValueError` identifying the column, the offending value, and the precision knob the caller can adjust.

**Why not change the formula**: the bytes produced here are read by `PriceRaw::from_le_bytes` / `QuantityRaw::from_le_bytes` on the Rust side, which assumes the value is at the 10**FIXED_PRECISION scale. Changing the formula would silently break every existing catalog and live consumer. The explicit overflow check is the right seam — it surfaces a clear error at the right place without changing wire format.

## Related Issues/PRs

Discovered via the alpha-miner project (`MandalorianBatman/alpha-miner`) when downloading NSE equity data. The downstream caller works around the bug locally with a pre-scale divisor; once this PR lands upstream, the workaround becomes belt-and-suspenders and can be removed.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Breaking change details

None. The change is additive — code that previously succeeded continues to succeed; code that previously failed with a cryptic downstream error now fails with a clear `ValueError` at the wrangler boundary.

## Documentation

- [x] Release notes entry added to `RELEASES.md` under 1.231.0 Fixes

## Release notes

Entry added:

> Fixed v2 wranglers silently overflowing int64 for realistic source values; now raises a clear `ValueError` with the offending column/value and a hint to adjust the precision knob (e.g. NSE equity at 3812 INR with `price_precision=0`, or 140k-share 1-minute volume)

## Testing

- [x] Affected code paths are already covered by the test suite (existing `test_wranglers_v2.py` exercises all four wranglers on small well-formed inputs)

Two new regression tests in `tests/unit_tests/persistence/test_wranglers_v2.py`:

- `test_bar_wrangler_overflow_raises_clear_error` — `BarDataWranglerV2` with `price_precision=0` against a realistic source must raise `ValueError` matching `"Wrangler overflow"`
- `test_bar_wrangler_normal_values_still_work` — `BarDataWranglerV2` with `price_precision=2` (paise) against the same source must succeed (sanity check that the check doesn't break the normal path)

Local test run:

```
$ pytest tests/unit_tests/persistence/test_wranglers_v2.py -v
test_bar_wrangler_overflow_raises_clear_error PASSED
test_bar_wrangler_normal_values_still_work PASSED
... (all pre-existing tests still pass)
```

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -34,6 +34,7 @@ releases as feedback arrives, before the final `2.0.0` release.
 - Renamed Python v2 `RedisMessageBusDatabase` to `RedisMessageBusBacking` (documenting a previous break)
 
 ### Fixes
+- Fixed v2 wranglers silently overflowing int64 for realistic source values; now raises a clear `ValueError` with the offending column/value and a hint to adjust the precision knob (e.g. NSE equity at 3812 INR with `price_precision=0`, or 140k-share 1-minute volume)
 - Fixed `LiveNode` external order claims bypassing the execution engine (#4347), thanks for reporting @linimin
 - Fixed `PerContractFeeModel` generic spread fees to charge per leg ratio (#4360), thanks for reporting @pjlegato
 - Fixed `HEDGING` reduce-only orders without cached position IDs (#4312), thanks for reporting @luckykefu
--- a/nautilus_trader/persistence/wranglers_v2.py
+++ b/nautilus_trader/persistence/wranglers_v2.py
@@ -33,6 +33,51 @@ from nautilus_trader.model.objects import FIXED_SCALAR
 RAW_BYTE_ORDER: Final = "little"
 
 
+# Maximum signed 64-bit integer. Fixed-point raw values are stored as
+# little-endian 16-byte (128-bit) signed integers (FIXED_PRECISION_BYTES),
+# but the meaningful range is int64 because the Rust PriceRaw / QuantityRaw
+# types are 64-bit. Wrangler operations that would produce a raw value
+# outside this range silently overflowed prior to 1.230.1, surfacing as
+# an obscure polars/pyarrow Int128 error downstream.
+_INT64_MAX: Final = 2**63 - 1
+_INT64_MIN: Final = -(2**63)
+
+
+def _check_raw_overflow(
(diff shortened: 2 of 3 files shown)


cjdsellers:
Hi @MandalorianBatman,

Thanks for the PR and for the clear write-up. I agree this path should fail before the Arrow/catalog boundary, but I found a few blockers before this can land:

- `_check_raw_overflow()` currently returns `None`, and every caller then calls `.to_bytes()` on that result. That means the normal, non-overflow path now fails with `AttributeError`; the check needs to return the raw value after validating it.
- The guard uses signed int64 limits for all raw values, but high-precision builds use `PriceRaw = i128` and `QuantityRaw = u128`. This would reject valid high-precision values, and the `price_precision=2` test case still scales `3812.25` to about `3.81225e19`, not `3.81e17`.
- Most new call sites pass `col_name="<inline: see caller>"`, so the new error does not identify the offending column as described.

cjdsellers:
Now implemented from https://github.com/nautechsystems/nautilus_trader/commit/a1c7c078fb1877a5d194cb264dfdd1f4939ad0da, thanks again.

PR #4384, opened 2026-07-04
Remove stale @typing.final from subclassable Strategy and DataActor stubs

bebop23:
# Pull Request

## Summary

The v2 pyo3 `Strategy` (`nautilus_trader.trading`) and `DataActor` (`nautilus_trader.common`) wrappers are declared `#[pyclass(subclass)]`, and subclassing them is the documented public API. Their generated `.pyi` stubs, however, still carry `@typing.final`, so the documented usage fails static type checking:

```
error: Base class "Strategy" is marked final and cannot be subclassed (reportGeneralTypeIssues)
```

This degrades static checking for every typed v2 user (any `class MyStrategy(Strategy): ...`).

## Root cause

`pyo3_stub_gen` already suppresses `@typing.final` for `#[pyclass(subclass)]` types — the `ClassDef.subclass` guard in `patches/pyo3-stub-gen/src/generate/class.rs` (`if !self.subclass { writeln!(f, "@typing.final")? }`), fed by `PyClassInfo.subclass` which the derive macro parses from the `#[pyclass(..., subclass)]` attribute. That support landed after these two stubs were last generated, and the stubs were never regenerated, so the stale `@typing.final` shipped in `2.0.0rc1`. Removing the marker matches exactly what the generator now emits for these classes (nothing else about the class definitions changes).

Scope note: `Indicator` has no subclassable base (concrete indicators are not `subclass` and are correctly final). This change is deliberately limited to the two public, user-subclassable base classes and does not touch other generated output.

## Related Issues/PRs

Reported for the v2 release wheel; the fix is the stub side of the `#[pyclass(subclass)]` types `Strategy` and `DataActor`.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

Verified with **pyright**: a module that does `class MyStrategy(Strategy): ...` and `class MyActor(DataActor): ...` reports **0 errors** after this change, and reproduces the `reportGeneralTypeIssues` "marked final and cannot be subclassed" error before it (for both classes). The change is exactly the two `@typing.final` deletions the stub generator produces for these `#[pyclass(subclass)]` types, so it stays stable across future `make build-debug-v2` / `py-stubs-v2` regenerations.


cjdsellers:
Hi @bebop23,

Thanks for the clear write-up and for checking this against `#[pyclass(subclass)]`. I've landed this by teaching the v2 stub generator to carry PyO3 `subclass` metadata through post-processing, so generated stubs now drop stale `@typing.final` markers for `Strategy`, `DataActor`, their configs, and the other subclassable PyO3 classes it finds. I also added regression tests, regenerated the stubs, and credited you in `RELEASES.md` https://github.com/nautechsystems/nautilus_trader/commit/678b8eef61219d947160670b53333dc2c1af3403.

Let me know if you spot any other stale typing markers.

PR #4386, opened 2026-07-04
Consolidate reconciliation recency maps into RecencyMap<K>

folknor:
Right now the monotonic-clock choice is a discipline, not a rule. Every one of
those four sites - `ts_last_query`, `order_local_activity`,
`position_local_activity`, `recent_fills_cache` - has to remember to stamp from
`dst::time::Instant::now()` and not `self.clock`, and to handle a future-marked
instant in the safe direction by hand.

So make it structural. `RecencyMap<K>` owns `Instant::now()` internally - you
`mark(key)`, ask `within(key, window)`, `prune_older_than(ttl)`, and there is no
seam left to thread the wrong clock through.

### The shape

```
struct RecencyMap<K> { inner: IndexMap<K, dst::time::Instant> }
mark / contains_key / remove / last_marked
within / within_at / elapsed / elapsed_at    // _at variants take the pass's `now`
prune_older_than
```

The `_at` variants exist so the reconciliation loop passes its loop-top `now`
once instead of resampling `Instant::now()` per order - the second half of that
same review note. Every `checked_duration_since` resolves a `None`
(marked-in-the-future) to the safe direction, matching what each original site
did by hand, so a garbage future instant reads as "0 elapsed, within window"
rather than panicking.

### Renames

The `ts_*`-named fields you flagged in #4376 - `ts_*` repo-wide means a domain
`UnixNanos`, but these hold monotonic `Instant`s, so the names lie:

- `InflightCheck::ts_submitted` -> `submitted_at`
- `InflightCheck::last_query_ts` -> `last_query_at`
- the node scheduler's `ts_last_*` locals and `ReconciliationCheckState` fields
  -> `last_*_check`
- `ts_last_query` -> `order_query_recency` (keyed on `ClientOrderId`, parallels
  `order_local_activity`; the other three map fields self-document under
  `RecencyMap<K>`, but this one's `ts_` prefix did not survive the type change
  gracefully, so it got a real name too)

### Tests

Focused `RecencyMap` unit tests: mark/contains/remove, monotonic expiry under
paused virtual time, pruning older-than-TTL, and the future-marked-instant safe
direction.

### Still to come

The second promised follow-up - close the class: `handle_missing_order`'s
venue-`ts_last` gate onto receipt time, plus the wider timer/throttle audit.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -67,6 +67,8 @@ use nautilus_model::{
 use rust_decimal::{Decimal, prelude::ToPrimitive};
 use ustr::Ustr;
 
+use super::recency::RecencyMap;
+
 /// Tag for orders originating from venue (external orders).
 static TAG_VENUE: LazyLock<Ustr> = LazyLock::new(|| Ustr::from("VENUE"));
 
@@ -231,11 +233,11 @@ impl ExecutionManagerConfig {
 struct InflightCheck {
     #[allow(dead_code)]
     pub client_order_id: ClientOrderId,
-    pub ts_submitted: dst::time::Instant,
+    pub submitted_at: dst::time::Instant,
     pub retry_count: u32,
     // `Instant` debug output is runtime-specific and intentionally only useful
     // as an opaque monotonic offset.
-    pub last_query_ts: Option<dst::time::Instant>,
+    pub last_query_at: Option<dst::time::Instant>,
 }
 
 /// Manager for execution state.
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR. I like the naming convention here: dropping the Hungarian-notation-esque domain `ts_` component when the value is not a `UnixNanos` and does not represent a domain timestamp makes the clock axis much clearer. Names like `submitted_at`, `last_query_at`, and `last_*_check` read as real-time instants without implying domain time, while `RecencyMap` keeps the monotonic clock choice structural.

I traced the converted call sites and ran the focused recency, open-order query, inflight, fill-cache, and position-grace test slices. This looks good to me.

Will get to the next PR shortly!

PR #4397, opened 2026-07-06
Fix IB v2 fills dropped on non-UTC execution timestamps

### LINKED ISSUE #4396: [Interactive Brokers V2] Fills not registered: execution timestamp rejected unless zone is UTC
Currently I'm migrating a live IB strategy to v2 (rc1). Market orders fill at IB (visible in TWS and in position reports) but Nautilus never produces an **OrderFilled**. The strategy keeps seeing the position as flat and re-enters every bar.

The fill report dies at parse time:

```
[ERROR] nautilus_interactive_brokers::execution::core::core_updates:
Unsupported non-UTC execution timezone 'MET' in '20260706 14:05:00 MET'.
Configure TWS / IB Gateway to emit UTC timestamps
```

The account is a European paper account, so IB stamps executions with `MET`.

What I've tried to get UTC out of the gateway:

- **Send status updates with UTC timestamps** GUI-only setting; my docker image (`gnzsnz/ib-gateway`) exposes no env for it, and IBC has no config key for it.
- **TIME_ZONE=UTC / TZ=UTC on the container** only sets `jts.ini [Logon] TimeZone` (confirmed: renders `TimeZone=Africa/Abidjan`). Execution stamps still come back as `MET`; they carry the IB server's zone, not the gateway session tz.

So the gateway can't be coerced into emitting UTC for executions. `parse_execution_time` in `crates/adapters/interactive_brokers/src/execution/parse.rs` accepts only `{UTC, Universal, Etc/UTC, GMT, Z}` and bails on anything else.

For reference, v1 handled this `timestring_to_timestamp` parsed the zone with pandas (`pd.Timestamp(dt, tz=tz)`), so `MET`, `America/New_York`, etc. all worked. The Rust port kept a UTC-only whitelist.

Repro: v2 `LiveNode` against an IB paper account whose server reports `MET`, submit a market order, watch the fill. Reproduced on `2.0.0rc1.dev20260706`.

dfjmax:
## Problem

Market-order fills registered at IB, but Nautilus never produced an `OrderFilled`. The strategy kept seeing the position as flat and re-entered every bar.

Root cause: `parse_execution_time` in the IB v2 adapter accepted only a UTC whitelist (`Universal`, `UTC`, `Etc/UTC`, `GMT`, `Z`). Any execution stamped with a different timezone was rejected at parse time and the fill report was dropped. Accounts whose server reports a regional zone hit this on every fill, and the gateway can't be coerced into emitting UTC for executions.

The v1 adapter never had this problem — `timestring_to_timestamp` handed the timezone to pandas, which resolved it via the IANA database.

## Fix

Resolve the timezone token through `chrono-tz` against the IANA database instead of the whitelist, so `MET`, `America/New_York`, `Asia/Tokyo`, and friends all work. DST is handled correctly (same offset for a given date as pandas/zoneinfo), and genuinely ambiguous or non-existent local times still error out rather than silently picking an instant.

Parity with v1 is structural: both lean on the same IANA tz database, so any zone v1 accepted now resolves to the same instant in v2.

## Testing

New unit tests in `parse.rs`:
- `MET` timestamp (the reported case), asserting the correct UTC+1 offset
- `America/New_York` across winter (EST) and summer (EDT) to prove DST is applied
- `Asia/Tokyo` for the fixed-offset, no-DST path
- an unrecognised timezone still errors

Closes #4396

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6233,6 +6233,7 @@ dependencies = [
  "bollard",
  "bon",
  "chrono",
+ "chrono-tz",
  "dashmap",
  "futures-util",
  "ibapi",
--- a/crates/adapters/interactive_brokers/Cargo.toml
+++ b/crates/adapters/interactive_brokers/Cargo.toml
@@ -78,6 +78,7 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 bon = { workspace = true }
 chrono = { workspace = true }
+chrono-tz = { workspace = true }
 dashmap = { workspace = true }
 futures-util = { workspace = true }
 ibapi = { workspace = true }
--- a/crates/adapters/interactive_brokers/src/execution/parse.rs
+++ b/crates/adapters/interactive_brokers/src/execution/parse.rs
@@ -18,7 +18,7 @@
 use std::str::FromStr;
 
 use anyhow::Context;
-use chrono::{DateTime, NaiveDateTime, Utc};
+use chrono::{DateTime, LocalResult, NaiveDateTime, Utc};
 use chrono_tz::Tz;
 use ibapi::orders::{Execution, OrderStatus};
 use nautilus_core::UnixNanos;
@@ -381,8 +381,9 @@ fn decimal_from_f64(value: f64) -> anyhow::Result<Decimal> {
 ///
 /// # Errors
 ///
-/// Returns an error if the execution timestamp is malformed, uses an
-/// unrecognized timezone, or is ambiguous / non-existent in its timezone.
+/// Returns an error if the timestamp is malformed, the timezone is
+/// unrecognized, or the local time is non-existent (a DST spring-forward gap).
+/// DST fall-back folds resolve to the earliest matching instant.
 pub fn parse_execution_time(time_str: &str) -> anyhow::Result<UnixNanos> {
     const NAIVE_FORMAT: &str = "%Y%m%d %H:%M:%S";
 
@@ -420,12 +421,7 @@ pub fn parse_execution_time(time_str: &str) -> anyhow::Result<UnixNanos> {
 /// Localize a naive timestamp against an IB timezone token and convert to UTC.
 ///
 /// `Z` is normalized to `UTC`; everything else is resolved through the IANA tz
-/// database via `chrono-tz`.
-///
-/// # Errors
-///
-/// Returns an error if the timezone is unrecognized, or the wall-clock time is
-/// ambiguous or non-existent in that timezone (e.g. inside a DST fold/gap).
+/// database via `chrono-tz`. Error and fold behavior is documented on [`parse_execution_time`].
 fn localize_with_zone(
     dt: NaiveDateTime,
     tz_str: &str,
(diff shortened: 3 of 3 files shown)


faysou:
Is there some caching for querying this IANA database or it does it for every market data received ?

dfjmax:
Hi @faysou as far as I've checked it only runs on fills, not market data. The call sites are all in the execution path reading **Execution.time**, nothing under src/market/.

I think that we don't need cache here. chrono-tz ships the tzdata compiled into the binary, so there's no runtime lookup to memoize, just lookups against static tables.

faysou:
Ok and are market data timestamps UTC when configuring tws or gateway to use UTC ?

dfjmax:
Didn't touch market data in this PR, but from a quick look: for trades, quotes, and bars Nautilus just uses the timestamp IB sends (through ibapi) without re-reading the timezone, so it's whatever IB returns under that setting. 

Order-book depth is stamped from the local clock, so that one's UTC regardless.

faysou:
Yes because I was assuming that tws or gateway would be configured as UTC and that all data from it would be UTC. Can you please check if you can do something to get market data converted to UTC when timestamps are not UTC ?

dfjmax:
Had a look, and I don't think market data needs anything here. Those timestamps come over the wire as unix epoch seconds, and ibapi decodes them straight into a UTC instant, which Nautilus reads as-is. So trades, quotes, and bars are already correct UTC regardless of the TWS/gateway display setting. Market data is numeric epoch on the wire, so there's no timezone token to misread.

The execution time was the one that bit because IB sends it as formatted text with a timezone tag, and there's no numeric timestamp field for fills, so it has to be resolved rather than just read. That's why this PR is scoped to it.

faysou:
Ok thank you

bebop23:
Nice fix — resolving through the IANA db via `chrono-tz` is the right call, and matches what I independently landed on chasing the same dropped-fill → dangling-position symptom (mine was a CME `US/Central` account rather than an EU `MET` one).

One edge case in `localize_with_zone` worth a look. `dt.and_local_timezone(zone).single()` returns `None` for **two** distinct situations — the DST spring-forward gap *and* the fall-back fold — and today both bail with an error. That error propagates back out through the order-update handler and drops the fill, which is the exact failure this PR is closing, just narrowed to the DST-transition window.

The gap case erroring is fine (an exchange clock never emits a nonexistent local time). The **fold** case is the real one: on fall-back night the wall clock reads the same hour twice, so a bare `"... 01:30:00 US/Central"` is genuinely two valid UTC instants —

| Real instant (UTC) | US/Central wall clock |
|---|---|
| 06:30 UTC | **01:30** (CDT, −5) |
| 07:30 UTC | **01:30** (CST, −6) |

— which is why `chrono` hands back `Ambiguous(earliest, latest)` instead of one answer: the input string simply doesn't carry the offset needed to disambiguate. IB won't stamp a gap time, but it *will* stamp a fold time, so erroring there costs a real fill (at most 1h of skew if you just pick one).

Since the whole premise here is "a shifted timestamp is recoverable, a dropped fill isn't," resolving the ambiguous case to the earliest instant rather than erroring seems more in keeping:

```rust
use chrono::offset::LocalResult;

match dt.and_local_timezone(zone) {
    LocalResult::Single(local) => Ok(local.with_timezone(&Utc)),
    // Fall-back fold: wall time occurs twice. Take the earliest instant
    // rather than dropping the fill — worst case ~1h skew for that hour.
    LocalResult::Ambiguous(earliest, _latest) => Ok(earliest.with_timezone(&Utc)),
    // Spring-forward gap: a venue clock never emits this, erroring is fine.
    LocalResult::None => anyhow::bail!(
        "Execution timestamp '{time_str}' is non-existent in timezone '{tz_str}'"
    ),
}
```

It's a rare, market-hours-dependent, once-a-year window, so entirely your call whether it's worth the extra arm — flagging it since it's the same failure mode the PR is fixing.

dfjmax:
Good catch .single() folded both the gap and the fold into None, so the fall-back night dropped the fill exactly as you describe.

Switched to an explicit LocalResult match:
- Ambiguous(earliest, _) → earliest instant (the offset in force just before the clocks roll back)
- None → still errors, since a venue clock won't emit a gap time

Added tests on America/Chicago for both, the fold resolves to 06:30 UTC (CDT) and not 07:30 UTC (CST), and the spring-forward 02:30 errors.

cjdsellers (approved):
Thank you for the fix @dfjmax and for the input @faysou + @bebop23 🙏

PR #4400, opened 2026-07-07
Make timer_exists consistent with timer_names and timer_count on expired timers

folknor:
The `Clock` trait has three introspection surfaces for timers, and they disagreed with each other. `timer_names` and `timer_count` both filter on `!timer.is_expired()` - they report only *active* timers. But `timer_exists` was a bare `contains_key`, so it counted expired timers too.

Expired entries are not removed the instant a timer fires; they are purged lazily. `LiveClock` clears them only on the next `set_time_alert_ns` / `set_timer_ns` / cancel call. So there is an unbounded window where a fired-and-done timer still sits in the map - and during that window `timer_exists(name)` returns `true` while `timer_count()` and `timer_names()` behave as though it is gone.

`TestClock` removes one-shot timers eagerly as they fire, so the disagreement is only reachable on the live clock in practice - but both impls carried the identical bare `contains_key`, so both are fixed here to keep the trait contract uniform across clocks.

### The fix

Filter `timer_exists` on `!is_expired()` in both `LiveClock` and `TestClock`, so all three surfaces answer from the same active-timer view. The trait method's doc and the `Clock` API wrapper are updated to say they report an *active (not expired)* timer.

### Consumers are unaffected

Every caller either already tolerates the new answer or improves under it:

- The cancel-before-set callers (`throttler`, option-chain `manager`, data `engine`) do `if timer_exists { cancel_timer }`. With the fix they skip cancelling an already-expired lingering entry - which the immediately following `set_*` overwrites via `replace_existing_timer` anyway, so the outcome is identical.
- The backtest instrument-expiration guard (`if timer_exists { return }`) runs on `TestClock`, which never leaves an expired timer in the map, so its behavior does not change.

### Tests

- `TestClock`: a repeating timer with a stop time is advanced past expiry; the test asserts all three surfaces agree it is gone (`timer_exists` false, `timer_count` 0, `timer_names` empty).
- `LiveClock`: a timer is run past its stop time on the real Tokio runtime; once it has finished, the test asserts the entry is *still physically in the map* (`timers.contains_key` true - lazy purge has not run) yet `timer_exists` reports false, matching `timer_count`/`timer_names`. This is the exact expired-but-present state the fix targets.

### DIFF
--- a/crates/common/src/clock.rs
+++ b/crates/common/src/clock.rs
@@ -70,7 +70,7 @@ pub trait Clock: Debug + Any {
     /// Returns the count of active timers in the clock.
     fn timer_count(&self) -> usize;
 
-    /// If a timer with the `name` exists.
+    /// If an active (not expired) timer with the `name` exists.
     fn timer_exists(&self, name: &Ustr) -> bool;
 
     /// Register a default event handler for the clock. If a timer
@@ -608,7 +608,7 @@ impl<'a> ClockApi<'a> {
         }
     }
 
-    /// Returns whether the timer `name` exists.
+    /// Returns whether an active (not expired) timer `name` exists.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Thank you @folknor. It's a tight minimal change with good reasoning 👍

PR #4401, opened 2026-07-07
Stop LiveTimer firing outside its stop_time_ns bound

folknor:
`TestTimer` enforces the invariant `ts_event <= stop_time_ns` - its property test asserts every fired event obeys it. `LiveTimer` measured the same bound differently, and disagreed with `TestTimer` in two ways.

The live loop fired an event first, then incremented `next_time_ns` and broke on `max(next_time_ns, now_ns) >= stop_time_ns`. That fire-then-check ordering, plus folding the wall-clock read `now_ns` into the comparison, produced:

- **Fires one event past the stop.** A repeating timer with `stop_time_ns < start_time_ns + interval_ns` fired a single event whose `ts_event` was already beyond the stop time. `TestTimer` fires nothing in that case.
- **Drops the boundary event.** An event scheduled exactly at `stop_time_ns` was skipped: after firing the previous event, the loop incremented `next_time_ns` up to the boundary, and the `>= stop_time_ns` check broke *before* the tick that would have fired it. `TestTimer` fires the boundary event, then expires - so the two clocks disagreed in opposite directions at the two edges.

Mixing `now_ns` into the bound was independently wrong. `ts_event` is always the scheduled `next_time_ns`, never the wall-clock read - so under slow delivery (real `now_ns` past the stop while the scheduled time had not yet reached it) the scheduled event set could be truncated early.

### The fix

Enforce the bound on the scheduled time and check-then-fire, matching `TestTimer`: skip firing when `next_time_ns > stop_time_ns`, fire the event scheduled exactly at `stop_time_ns` (inclusive boundary), then expire. `now_ns` remains the event's `ts_init` only, out of the stop decision. One-shot alerts - where `next_time_ns == stop_time_ns` on the first tick - fire exactly once, unchanged.

### Tests

The stop-bound decision is factored into small pure helpers - `should_fire_scheduled_time` and `expires_after_scheduled_time`, alongside the start-time helpers `normalize_start_time_ns` and `timer_start_delay` - and each is unit-tested directly instead of through a real Tokio timer. This covers the cases the fix targets, deterministically and with no wall-clock waits:

- **Unbounded** - a `None` stop time never suppresses a fire and never expires.
- **Past the stop** - `next_time_ns > stop_time_ns` does not fire.
- **Before the stop** - fires without expiring.
- **Exact boundary** - `next_time_ns == stop_time_ns` fires, then expires.
- Start-time normalization (past-due clamps to now, a future time is kept) and startup-delay math (overhead subtracted, saturating to zero below it).

An earlier revision of this PR asserted the stop-bound behavior through live timers with wall-clock timing. Those tests proved flaky on loaded CI runners - one capped a timer at `now + 2ms`, but `start()` re-reads a fresh `now_ns` and clamps the scheduled time up to it, so a read landing past the bound suppressed the first fire and the test timed out. They are replaced by the pure-function tests above. The Python-callback deferral test is kept but no longer pins a stop time, which was incidental to what it checks and could suppress the first fire under a slow scheduler.


cjdsellers (approved):
Thank you @folknor, good changes and this aligns the live and test timers well.

Issue #4405, opened 2026-07-07
[Rust] LiveNode initial reconciliation failures - fail fast


TheoBabilon:
# Feature Request

Hi @cjdsellers, I understand that this maybe not be a good time for this considering the core dev focus on V2 Rust port and the core engine stability, so feel free to just close.

Fail-fast in Rust `LiveNode` and prevent kernel from starting trader/strategy on any initial reconciliation failure/timeouts.

## Problem Statement

The Rust `LiveNode` currently behaves differently vs. the Python startup procedure: in Rust we [just logs and continue](https://github.com/nautechsystems/nautilus_trader/blob/af32cf301a27cefad11850afcc9e696d56807129/crates/live/src/node/mod.rs#L668-L676) (as opposed to the indication in the function docstring) when startup execution reconciliation cannot retrieve mass status from any execution client. This can allow the node to initialize and start the trader even though startup execution state was not reconciled. 
In Python (not using it myself so this is just from looking at the code) the `NautilusKernel` will (correctly, in my opinion) [exit early](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/system/kernel.py#L1029) if any error is [raised during the initial recon](https://github.com/nautechsystems/nautilus_trader/blob/develop/nautilus_trader/live/execution_engine.py#L1805).

### Proposed Solution

I believe that startup execution reconciliation should be "fail-fast".

If the Rust live node cannot obtain mass status during startup reconciliation, it should return an error and stop startup before portfolio initialization and trader start.

PR #4406, opened 2026-07-07
Exit startup early when execution reconciliation fails in Rust

TheoBabilon:
Return an error from Rust LiveNode startup reconciliation when mass status generation times out, returns None, or fails, instead of warning and continuing.

# Pull Request

## Summary

This prevents ignoring initial reconciliation failures/timeouts and blindly starting the trading node.

## Related Issues/PRs

Relates to https://github.com/nautechsystems/nautilus_trader/issues/4405

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Will [error out](https://github.com/nautechsystems/nautilus_trader/blob/8160730c7c550480b0a439fb11086a4c4de15f0b/crates/live/src/node/mod.rs#L415) the node in case of failure instead of just logging and starting.

## Testing

Did not see specific tests for this scenario but some could be added
We could also create an error enum to identify timeouts, missing mass status or failures during generation, but wanted to keep things simple

### DIFF
--- a/crates/adapters/sandbox/src/execution.rs
+++ b/crates/adapters/sandbox/src/execution.rs
@@ -1164,7 +1164,14 @@ impl ExecutionClient for SandboxExecutionClient {
         &self,
         _lookback_mins: Option<u64>,
     ) -> anyhow::Result<Option<ExecutionMassStatus>> {
-        // Sandbox doesn't need reconciliation
-        Ok(None)
+        let core = self.core.borrow();
+        let ts_init = self.clock.borrow().timestamp_ns();
+        Ok(Some(ExecutionMassStatus::new(
+            core.client_id,
+            core.account_id,
+            core.venue,
+            ts_init,
+            None,
+        )))
     }
 }
--- a/crates/adapters/sandbox/tests/execution.rs
+++ b/crates/adapters/sandbox/tests/execution.rs
@@ -2418,6 +2418,28 @@ fn test_account_id(execution_client: SandboxExecutionClient, account_id: Account
 }
 
 #[rstest]
+#[tokio::test]
+async fn test_generate_mass_status_returns_empty_report(
+    execution_client: SandboxExecutionClient,
+    client_id: ClientId,
+    account_id: AccountId,
+    venue: Venue,
+) {
+    let mass_status = execution_client
+        .generate_mass_status(None)
+        .await
+        .unwrap()
+        .unwrap();
+
+    assert_eq!(mass_status.client_id, client_id);
+    assert_eq!(mass_status.account_id, account_id);
+    assert_eq!(mass_status.venue, venue);
+    assert!(mass_status.order_reports().is_empty());
+    assert!(mass_status.fill_reports().is_empty());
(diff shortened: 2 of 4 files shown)


cjdsellers:
Hi @TheoBabilon,

Thanks for the PR. The fail-fast direction makes sense, but I found two lifecycle issues that should be fixed before this lands.

The new `bail!` / `Err` paths in `perform_startup_reconciliation()` propagate straight through `start()` and `run()` after the kernel has started and clients have connected. That skips `abort_startup()` / `finalize_stop()`, so the node can be left in `Starting` with live clients still connected. The call sites should abort startup, drain channels in `run()`, and then return the reconciliation error.

Also, `timeout_reconciliation` still does not wrap the `generate_mass_status()` await. A single slow client can exceed the timeout and still let startup continue if it eventually returns `Ok(Some(_))`. I think this needs a remaining-budget `tokio::time::timeout` around the mass-status call.

TheoBabilon:
Thanks for the quick review @cjdsellers - I've added tests that highlight both lifecycle and timeout issues; had to create a `StartupMassStatusExecutionClient` setup for this, let me know if that is ok with you.

Now there's a question on the `Ok(None)` branch: this PR makes it fail-fast but from what I see that means e.g. existing `SandboxExecutionClient` would [now fail](https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/adapters/sandbox/src/execution.rs#L1168) - so depending on how you see things we can either return a valid (empty, so `ExecutionMassStatus::new(...)`) `ExecutionMassStatus` from the sandbox (which I think would match what was done in Python [as it was a condition for fast-exit](https://github.com/nautechsystems/nautilus_trader/blob/c22f0a61268d86d397d41f2d78ddaa32238d7ac5/nautilus_trader/live/execution_engine.py#L1721)) or make the `Ok(None)` branch non-critical and just warn/ignore (I'd lean for the former, which seems more correct to me)

cjdsellers:
Hi @TheoBabilon,

Thanks for the updates. The lifecycle cleanup, timeout wrapping, and clippy fix all look right to me, and I agree with your lean on the `Ok(None)` branch. I think `Ok(None)` should stay fail-fast, and the sandbox should return an empty `ExecutionMassStatus` instead.

```bash
make pre-commit
```

or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

TheoBabilon:
Apologies for that clippy quirk @cjdsellers, should've been more careful with this. I'll push the empty ExecutionMassStatus change for the sandbox and make sure `make pre-commit` comes clean again

TheoBabilon:
should be ready @cjdsellers , `make pre-commit` is all green besides an unrelated error on the `uv-lock` hook which I believe is related to internal/personal environment setup. Let me know if anything seems off to you or if you want me to change something else

cjdsellers:
Thanks again for the follow-ups @TheoBabilon 👌

cjdsellers (approved):

PR #4408, opened 2026-07-07
Wire live client routing config into Rust nodes

### LINKED ISSUE #4404: [Interactive Brokers v2] IB client fails to route exchange-MIC instruments for data and execution without an explicit client_id
# Bug Report

## Confirmation

- [x] I've reviewed the source code to confirm the behavior is not by design.
- [x] I've tested this on `2.0.0rc1` and can reproduce it.

### Expected Behavior

An IB (Interactive Brokers) client registered under the broker venue `IB` should implicitly route data subscriptions, data requests, and order commands for instruments whose venue is a real exchange MIC (`IBIS`, `IDEALPRO`, `ARCA`, …) the same way the v1 IB adapter did. No per-call `client_id` should be required.

### Actual Behavior

In v2, both the data and execution engines resolve a client by exact `routing_map[venue]` match, then fall back to `default_client`. The IB client is registered with `venue = "IB"` (the broker venue), so it is never promoted to `default_client` (its venue is `Some`, not `None`), and there is no `routing_map` entry for the exchange MICs.

Therefore any subscribe/request/order for an exchange-MIC instrument without an explicit `client_id` fails:

```
[ERROR] nautilus_data::engine: Cannot handle request: no client found for client_id=None, venue=Some("IBIS")
[ERROR] nautilus_data::engine: Cannot handle command: no client found for client_id=None, venue=Some("IBIS")
[ERROR] nautilus_execution::engine: No execution client found for command: client_id=None, venue=IBIS
```

The execution client already implements `handles_order_venue() -> true` ("IB uses a broker venue for the client while routing exchange-MIC instruments"), but it is only consulted as a **post-find validation** in `crates/execution/src/engine/mod.rs`, never as a routing fallback. The data client has no `handles_venue` equivalent at all.

1. Start a v2 `LiveNode` with an IB data + execution client.
2. Load an instrument whose venue is not `IB` (e.g. `VWCE.IBIS`).
3. From a strategy, call `subscribe_bars` / `request_bars` / `submit_order` without passing `client_id`.
4. Observe the "no client found for venue=IBIS" errors above.

Workaround: pass `client_id="IB"` explicitly on every call. This is fragile, any path that omits it (e.g. `close_all_positions`) silently fails to route.

### Root cause (v1 vs v2)

- **v1** (Python): the IB client was registered with `venue=None`, which the engine auto-promoted to `default_client`. Resolution for an unmapped venue fell through to the default → IB handled it. No per-call `client_id` needed.
- **v2** (Rust): `ExecutionClientCore.venue` is required (non-optional), so the IB factory sets it to the broker venue `"IB"`. The auto-default trick no longer applies. The new "I route multiple MICs" signal, handles_order_venue`, was added on the execution client but wired only as post-find validation, and the data client has no equivalent. As a result the signal that should replace the v1 default trick is unreachable during client resolution.

### Fix

A fix is available (PR to follow). It makes both engines consult the multi-venue signal during resolution, mirroring v1's implicit routing:

- Add `handles_venue()` to the `DataClient` trait (`crates/common/src/clients/data.rs`), default `self.venue() == Some(venue)`, the data-side twin of the existing `handles_order_venue` on `ExecutionClient`.
- Override `handles_venue() -> true` on the IB data client (`crates/adapters/interactive_brokers/src/data/core.rs`).
- In the data engine `get_client` (`crates/data/src/engine/mod.rs`): after the `routing_map` miss and before the default fallback, scan registered clients for one whose `handles_venue(venue)` is true. `IndexMap` insertion order gives deterministic first-match.
- In the execution engine `find_client_for_command` (`crates/execution/src/engine/mod.rs`): same scan via `handles_order_venue` between the instrument-venue map lookup and the `default_client` fallback.

Both the existing post-find validation and the explicit-`client_id` path are unaffected. Verified end-to-end on a live IB Gateway: subscribe bars, request bars, submit buy, submit sell / square-off, unsubscribe — all route for `VWCE.IBIS` without `client_id`.

### Specifications

- OS platform: macOS 26.5.1 (Darwin 25.5.0, arm64)
- Python version: 3.13.12
- `nautilus_trader` version: 2.0.0rc1

dfjmax:
Closes #4404.

## Problem

The engines resolved a client by exact `routing_map[venue]` match, then fell back to a single default. The IB client registers under the broker venue `IB` (`Some`), so it can never become the default, and there's no `routing_map` entry for the real exchange MICs (`IBIS`, `IDEALPRO`, …). Anything targeting an exchange-MIC instrument without an explicit `client_id` failed:

```
NO_EXECUTION_CLIENT: client_id=None, venue=IBIS
```

`RoutingConfig` (`default` / `venues`) existed on the live client configs and was exposed in Python, but the live builder never read it — dead config.

## Fix

Make the Rust/PyO3 live builder apply explicit routing config (mirrors the Cython live path):

- `add_data_client` / `add_exec_client` gain additive `*_with_routing` variants taking a `RoutingConfig` — live-node metadata beside the adapter config; factories still receive the plain IB config. At build the client is registered, then marked default (`routing.default`) or venue-routed (`routing.venues`).
- `DataEngine::register_venue_routing(client_id, venue)` — listed exchange MICs route to a registered client by id.
- `ExecutionEngine::set_default_client(client_id)` — marks an already-registered client default **without cloning a second instance** (default tracked by id, resolved through the `clients` map).

Routing order stays: explicit `client_id`, `routing.venues`, `routing.default`, then no route.

## Testing

Validated on a live IB gateway with `VWCE.IBIS`:
- Without `routing` → `NO_EXECUTION_CLIENT: client_id=None, venue=IBIS` (bug reproduced).
- With `routing=RoutingConfig(default=True)` → subscribe/bars, buy, fill, position — all route without `client_id`.

Engine unit tests added for both default-fallback and venue-routing paths; both suites + clippy clean.

### DIFF
--- a/crates/adapters/interactive_brokers/examples/node_data_tester.rs
+++ b/crates/adapters/interactive_brokers/examples/node_data_tester.rs
@@ -37,7 +37,7 @@ use nautilus_interactive_brokers::{
     },
     factories::InteractiveBrokersDataClientFactory,
 };
-use nautilus_live::node::LiveNode;
+use nautilus_live::{config::RoutingConfig, node::LiveNode};
 use nautilus_model::{
     data::bar::BarType,
     identifiers::{ClientId, InstrumentId, TraderId},
@@ -79,13 +79,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         ..Default::default()
     };
 
+    let routing = RoutingConfig::builder().default(true).build();
+
     let mut node = LiveNode::builder(TraderId::from(TRADER_ID), Environment::Live)?
         .with_name(NODE_NAME.to_string())
         .with_delay_post_stop_secs(2)
-        .add_data_client(
+        .add_data_client_with_routing(
             None,
             Box::new(InteractiveBrokersDataClientFactory::new()),
             Box::new(data_config),
+            routing,
         )?
         .build()?;
 
--- a/crates/adapters/interactive_brokers/examples/node_exec_tester.rs
+++ b/crates/adapters/interactive_brokers/examples/node_exec_tester.rs
@@ -38,7 +38,10 @@ use nautilus_interactive_brokers::{
     },
     factories::{InteractiveBrokersDataClientFactory, InteractiveBrokersExecutionClientFactory},
 };
-use nautilus_live::{config::LiveExecEngineConfig, node::LiveNode};
+use nautilus_live::{
+    config::{LiveExecEngineConfig, RoutingConfig},
+    node::LiveNode,
+};
 use nautilus_model::{
     enums::{OrderType, TimeInForce},
     identifiers::{AccountId, ClientId, InstrumentId, StrategyId, TraderId},
@@ -80,6 +83,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let market_data_type = parse_market_data_type(MARKET_DATA_TYPE);
     let order_qty = Quantity::from(ORDER_QTY);
 
+    let routing = RoutingConfig::builder().default(true).build();
+
     let data_config = InteractiveBrokersDataClientConfig {
         host: HOST.to_string(),
         port: PORT,
@@ -108,17 +113,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .with_exec_engine_config(exec_engine_config)
         .with_delay_post_stop_secs(5)
         .with_reconciliation(true)
-        .add_data_client(
+        .add_data_client_with_routing(
             None,
             Box::new(InteractiveBrokersDataClientFactory::new()),
             Box::new(data_config),
+            routing.clone(),
         )?
-        .add_exec_client(
+        .add_exec_client_with_routing(
             None,
             Box::new(InteractiveBrokersExecutionClientFactory::new(
                 trader_id, account_id,
             )),
             Box::new(exec_config),
+            routing,
         )?
         .build()?;
 
--- a/crates/data/src/engine/mod.rs
+++ b/crates/data/src/engine/mod.rs
@@ -90,9 +90,7 @@ use nautilus_common::{
 };
 use nautilus_core::{
     Params, UUID4, UnixNanos, WeakCell,
-    correctness::{
-        FAILED, check_key_in_map, check_key_not_in_map, check_predicate_false, check_predicate_true,
-    },
+    correctness::{FAILED, check_key_in_map, check_key_not_in_map, check_predicate_true},
     datetime::{NANOSECONDS_IN_DAY, millis_to_nanos_unchecked},
 };
 #[cfg(feature = "defi")]
@@ -152,7 +150,7 @@ pub struct DataEngine {
     pub(crate) cache: Rc<RefCell<Cache>>,
     pub(crate) external_clients: AHashSet<ClientId>,
     clients: IndexMap<ClientId, DataClientAdapter>,
-    default_client: Option<DataClientAdapter>,
+    default_client_id: Option<ClientId>,
     routing_map: IndexMap<Venue, ClientId>,
     book_intervals: AHashMap<NonZeroUsize, BookSnapshotInfos>,
     book_snapshot_counts: IndexMap<BookSnapshotKey, usize>,
@@ -235,7 +233,7 @@ impl DataEngine {
             cache,
             external_clients,
             clients: IndexMap::new(),
-            default_client: None,
+            default_client_id: None,
             routing_map: IndexMap::new(),
             book_intervals: AHashMap::new(),
             book_snapshot_counts: IndexMap::new(),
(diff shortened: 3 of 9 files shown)


cjdsellers:
Hi @dfjmax,

Thanks for digging into this. I agree that IB needs a way to route exchange-MIC instruments without requiring `client_id` everywhere, but I want to check the routing shape before we merge this.

We already have routing config for this class of behavior: explicit `routing.venues` and `routing.default`. With this patch, `handles_venue` / `handles_order_venue` becomes a second implicit routing path, and because IB returns `true` for every venue it can win before a configured default client. That means a user with IB registered plus another default client could see unmapped venues route to IB instead of the default they configured.

Would you be open to preserving the existing priority order here? The options I see are either moving the `handles_*` scan after the default fallback, or wiring IB through the existing default/venue routing path for the Rust/PyO3 setup. 

Curious to understand how you see `handles_*` fitting with existing configured routing?

dfjmax:
Yeah, good call on the priority. handles_* really shouldn't outrank a configured default.

While digging into how to reorder it, I noticed the configured path you mentioned isn't actually wired up in v2, and I think that's the real reason IB has no clean route right now.

RoutingConfig with default and venues fields is on both the data and execution client configs and even exposed in Python, but the builder never reads it. It just passes the client's own venue into register_client. And there's a register_default_client on both engines that nothing in production ever calls. So routing.default and routing.venues are basically dead config today.

That also explains why the v1 trick stopped working. v1 registered IB with venue=None and the data engine promoted the first client with no venue to default, so unmapped venues fell through to IB. v2 made venue required and the factory sets it to IB, so it can never become that default, and the routing.default flag that should have replaced the trick was never connected.

So I'd lean toward wiring routing.default and routing.venues into the builder and dropping the handles_* scan along with the new handles_venue on the data client. Then IB routes exchange MIC instruments just by setting routing.default to true, or by listing the MICs under routing.venues. Configured routing stays authoritative and there's no second implicit path. handles_order_venue keeps doing what it already does, validating a client once it's been found.

Does that direction work for you? If so I'll rework the PR around it.

dfjmax:
@cjdsellers, quick steer wanted. The routing.default and routing.venues fields sit on LiveDataClientConfig and LiveExecClientConfig, but those configs aren't actually in the registration path. What gets passed to add_data_client and add_exec_client are the adapter configs like InteractiveBrokersDataClientConfig, which don't carry routing, and I'd rather not push a live-level concept down into adapter crates.

I've got a working version that passes routing as an optional param on add_data_client and add_exec_client, which keeps it a live-node concern. The alternative is reviving those two live configs as wrappers that carry routing next to the adapter config, which feels more correct but is a bigger change. Which would you prefer here?

cjdsellers:
Hi @dfjmax,

Yes, that direction works. I think the right fix is to make the Rust/PyO3 live builder apply explicit routing config, matching the Cython live path, and drop the new `handles_*` resolution scan.

I would keep `routing` as live-node metadata beside the adapter config, not push it into adapter configs. So I prefer the `add_data_client` / `add_exec_client` routing-param direction over reviving `LiveDataClientConfig` / `LiveExecClientConfig` wrappers as the factory input. For Rust API shape, this may need additive methods rather than optional params, but the important bit is that factories still receive `InteractiveBrokersDataClientConfig` / `InteractiveBrokersExecClientConfig`.

One implementation caveat: data and execution differ here. Cython can register one client object and then also mark that same object as default or venue-routed. Rust should avoid registering or cloning a second live client instance just to make it default. If the current engine APIs make that awkward, please add the small engine-side wiring needed to mark an already-registered `client_id` as default or venue-routed.

The routing order should stay: explicit `client_id`, configured `routing.venues`, configured `routing.default`, then no route. No implicit IB catch-all before the configured default.

dfjmax:
Hi @cjdsellers, I have applied the changes as we discussed.

Routing stays as live-node metadata beside the adapter config, and for exec the default is tracked by id via `set_default_client`. Order is client_id → `routing.venues` → `routing.default`.

I've tested using a live gateway with VWCE.IBIS and passing `routing=RoutingConfig(default=True)` and it works as intended.

Please let me know if anything else needs to be worked on.

cjdsellers:
Hi @dfjmax,

Thanks for reworking this. The execution side looks right, but the data `routing.venues` path still fails for IB. `InteractiveBrokersDataClient::venue()` returns `None`, so `register_client(adapter, venue)` moves it into `default_client`. The following `register_venue_routing` only accepts IDs in `clients`, so `RoutingConfig(venues=["IBIS"])` makes `LiveNodeBuilder::build()` return `No client registered with ID IB`. When `default=true`, the `if/else` also skips `routing.venues` entirely.

Please register the data client once and track default routing by `client_id`, as the execution side now does, then apply both routing fields independently. A builder-level regression test with a `None`-venue client should cover venue-only routing and `default + venues`.

dfjmax:
Hi @cjdsellers, the data side is now reworked too, register_client always keeps the client in the clients map (so None-venue clients like IB are addressable by **register_venue_routing**), and default routing is tracked by **client_id** via **set_default_client**, matching the execution side. 

**routing.default** and **routing.venues** now apply independently. 

Added a builder-level regression test with a None-venue client covering venue-only routing and default + venues.

cjdsellers:
Hi @dfjmax,

Thanks for all the follow-ups. The original IB routing issue and the data registration failure from my last review are now fixed. The focused tests and CI are green, so I’m going to merge this.

There is still a multi-client default-order edge case, plus some checked venue parsing work. I don’t think those need to hold this PR up after the rework already done here. I’ll follow up after merge.

Thanks again for sticking with the review!

cjdsellers (approved):

Issue #3733, opened 2026-03-19
IBKR: Historical bar subscriptions not restored after daily gateway restart


bomber555:
# Bug Report

## Confirmation

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [ ] I've tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

> Note: The relevant code path (`_resubscribe_all` in `client.py` and `_subscribe` in `market_data.py`) is unchanged in the latest `develop` branch as of 2026-03-19.

### Expected Behavior

After IBKR's daily gateway restart (~21:00-21:15 UTC / 06:00-06:15 JST), `InteractiveBrokersClient` should reconnect **and** re-send `reqHistoricalData()` to the gateway for all active `keepUpToDate=True` bar subscriptions, so that bars resume flowing without manual intervention.

### Actual Behavior

After IBKR daily restart, the client reconnects successfully (all data farms report OK), but **no historical bars are delivered**. The log shows `"Reusing existing Subscription instance"` for each subscription, and `reqHistoricalData()` is never re-sent to the gateway. Bars stop permanently until the entire `TradingNode` is restarted.

**Root cause:** `_resubscribe_all()` (client.py:344) calls `subscription.handle()` which re-enters `subscribe_historical_bars()`, which calls `_subscribe()` (market_data.py:519). Since the `Subscription` object still exists in the cache, `_subscribe()` returns early at line 184 (`"Reusing existing Subscription instance"`) without ever reaching `reqHistoricalData()` at line 541. Meanwhile, IBKR has cancelled all pending requests server-side during its restart, so the old `req_id` is dead.

### Steps to Reproduce the Problem

1. Subscribe to historical bars with `keepUpToDate=True` (e.g. 15-minute bars on any instrument)
2. Wait for IBKR daily gateway restart (occurs ~21:00-21:15 UTC / 06:00-06:15 JST every day)
3. Observe logs: connection restores, data farms reconnect OK, but `"Reusing existing Subscription instance"` appears and no new bars arrive after the restart
4. Confirm via IB Gateway GUI: API connection is green, data farms are ON, API clients are connected — but no data flows

### Code Snippets or Logs

**Log output showing the problem:**

```
[...]

PR #4383, opened 2026-07-04
Fix IB market order update price normalization

faysou:
# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [ ] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

- Normalizes Interactive Brokers `OpenOrder` price fields before emitting `OrderUpdated` events.
- Prevents IB echoed `limit_price` and `aux_price` values from reaching market-style Nautilus orders, avoiding the `MarketOrder::update()` invalid-event assertion reported in issue #4375.
- Preserves valid price fields for limit and stop-style order updates by mapping the IB order type back to the Nautilus order type before constructing the event.

### Design decisions
- Kept the fix at the adapter boundary in `crates/adapters/interactive_brokers/src/execution/core_updates.rs` so malformed venue echoes are cleaned before model order invariants are applied.
- Reused the existing `IbOrderType` to Nautilus `OrderType` mapping rather than adding string checks for only `MKT`.
- Left the order model assertions unchanged because they correctly protect order-type invariants.

### Verification
- `cargo test -p nautilus-interactive-brokers test_process_order_update_stream_clears_market_order_update_prices --lib`
- `cargo test -p nautilus-interactive-brokers --lib`
- `make build-debug`
- `make format`
- `make pre-commit`

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/4375

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Improvement (non-breaking)
- [ ] Breaking change (impacts existing behavior)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

- [ ] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)

## Release notes

- [ ] I added a concise entry to `RELEASES.md` that follows the existing conventions (when applicable)

## Testing

**Ensure new or changed logic is covered by tests.**

- [ ] Affected code paths are already covered by the test suite
- [x] I added/updated tests to cover new or changed logic

PR #4412, opened 2026-07-07
Fix Interactive Brokers over-reaction to data-farm flaps

dinana:
## Problem

During IB Gateway's nightly maintenance restart, IB's data farms flap broken↔OK for several
minutes while the gateway re-establishes its upstream link. The adapter treats a transient
market-data farm "broken" (2103) and a single subscription disconnect (10182) as full
connectivity loss: it clears `_is_ib_connected`, so the connection watchdog tears down the
**entire** socket — including the order/execution channel — for what is only a market-data
notification. Each teardown then reconnects before IB releases the previous session, hitting
error 326 ("client id already in use"), and the adapter falls back to a **random** client id
whose flag resets on each connect — cycling configured→random→configured→random for the whole
farm-stabilization window.

Per IB's message-code documentation, 2103/2105 are transient/expected during daily maintenance
and should not trigger a socket teardown.

## Root cause

- `client/error.py`: `2103` was in `CONNECTIVITY_LOST_CODES` and `10182` cleared
  `_is_ib_connected`; both trip the `client.py` watchdog into a full `_stop_async` +
  `_start_async` reconnect of the whole socket. The market-data farm (2103/2104) and HMDS farm
  (2105/2106) were handled inconsistently — `2103` was "lost" but `2105` unhandled; `2104` was
  "restored" but `2106` missing.
- `client/connection.py`: error 326 armed a random client-id fallback whose flag reset each
  connect, so the fallback never "stuck" — producing the oscillation.

## Fix

1. **Decouple data-farm state from socket state.** New `DATA_FARM_BROKEN_CODES` {2103, 2105} and
   `DATA_FARM_OK_CODES` {2104, 2106} (resolving the asymmetry). A farm "broken" and a single
   10182 mark the data feeds degraded and record the degrade time **without** clearing
   `_is_ib_connected`, so the socket and the order/execution channel stay up. A farm "OK"
   resubscribes the degraded feeds without a socket teardown.
2. **Preserve bar-gap recovery.** `subscribe_historical_bars` moves the historical-bar backfill
   cutoff back to the earliest data-loss point (socket disconnect **or** farm degrade), so bars
   completed during a farm outage are still recovered via the backfill.
3. **Deterministic, tiered client-id handling** for any residual error 326 (removing the needless
   teardowns removes most collisions outright):
   - A single/occasional 326 backs off and retries the **same** configured id; IB scopes
     open-order visibility by client id, so reusing it preserves order isolation.
   - A consistent 326 (beyond a small reuse limit) increments the id **deterministically** within
     a bounded band — never randomly — and re-fetches all open orders, since order isolation does
     not hold under a fallback id.
   - The collision counter resets on a clean connect.
4. **Docs.** The config docstrings and the IB integration guide now recommend reserving a
   contiguous band of client ids per client (e.g. 1-5, 6-10) so a fallback stays within range,
   and note that client-id order isolation is not guaranteed while running under a fallback id.

## Tests

New/updated unit tests under `tests/integration_tests/adapters/interactive_brokers/client/`:

- 2103/2105 and 10182 keep the socket up and only mark feeds degraded.
- 2104/2106 resubscribe feeds without a teardown; a farm-OK with no outstanding degradation is a
  no-op.
- A full farm-flap sequence (2105/2103 broken → 2104/2106 OK) leaves the socket and the client id
  intact and resubscribes once.
- A single 326 keeps the configured id and does not fetch orders; a consistent 326 increments
  within the bounded band and re-fetches orders; the offset is capped.

All existing IB adapter tests pass (481 passed).

## Testing against a live gateway

Validated against a live IB Gateway over two consecutive nightly maintenance windows. A dedicated
client id ran the patched adapter, subscribed to 1-minute bars (the `reqHistoricalData`
keepUpToDate path) plus quotes on a liquid CME future, bracketing the ~03:59 UTC restart and the
farm-stabilization window each night:

| | Night 1 | Night 2 |
|---|---|---|
| Data-farm flaps (2103/2105 → 2104/2106) | 1 | 2 |
| …caused a socket teardown? | No — resubscribed | No — resubscribed |
| Farm-induced teardowns (the bug) | 0 | 0 |
| Genuine disconnects (gateway restart / IB code 1100) | 1 (restart) | 2 (restart + a real 1100) |
| Error 326 | 0 | yes — backed off, retried the **same** id |
| Client-id changes | 0 (stayed on the configured id) | 0 (stayed on the configured id) |
| Bar gaps / data loss | 0 | 0 |

Night 2 hit a rougher window: a genuine IB "connectivity lost" (code 1100) on top of the restart
forced a reconnect that raced IB's client-id release and produced error 326 — exactly the
scenario the old code turned into a `configured→random→configured` client-id churn. The patched
adapter instead backed off and retried on the **same** configured id, rode through every farm
flap without tearing down the socket, and lost no bars.

## References

- IB message codes (2103/2105 transient; 2104/2106 OK): https://interactivebrokers.github.io/tws-api/message_codes.html
- Error 326 / socket-release race: https://github.com/erdewit/ib_insync/issues/376
- Related: #3968 (farm reconnect events), #4210 (reconnect startup), #3733 (frozen subscriptions)

### DIFF
--- a/docs/integrations/ib.md
+++ b/docs/integrations/ib.md
@@ -122,6 +122,14 @@ The adapter includes connection management features:
 - **Connection timeout**: Adjust the timeout with the `connection_timeout` parameter (default: 300 seconds).
 - **Connection watchdog**: Monitor connection health and trigger reconnection automatically when required.
 - **Graceful error handling**: Handle diverse connection scenarios with error classification.
+- **Transient data-farm tolerance**: Transient IB data-farm status notifications (codes 2103/2105 "broken" followed by 2104/2106 "OK", common during the nightly gateway restart) degrade only the affected data feeds and are recovered by resubscribing when the farm returns. They do **not** tear down the socket or the order/execution channel.
+
+#### Client ID allocation
+
+IB scopes open-order visibility by client id: a given client id only sees, and can manage, the orders it placed. Keep this in mind when allocating ids:
+
+- **Reserve a band per client.** On reconnect during a gateway restart the configured client id can still be held by the not-yet-released previous session (IB error 326). The adapter backs off and retries the *same* configured id first; only if the collision is consistent does it fall back to a nearby id within a bounded band (incremented deterministically, never randomly). Assign each client a contiguous band of ids (e.g. `1-5`, `6-10`, `11-15`) so a fallback stays within that client's range and never collides with a sibling process.
+- **Order isolation is not guaranteed at all times.** While running under a fallback id, order isolation by client id does not hold, and the adapter re-fetches all open orders in that case. Do not rely on client-id order isolation as a hard guarantee.
 
 ## Overview
 
--- a/nautilus_trader/adapters/interactive_brokers/client/client.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/client.py
@@ -172,7 +172,17 @@ class InteractiveBrokersClient(
         self._reconnect_jitter_secs: int = secrets.randbelow(4)
         self._had_ib_connection: bool = False
         self._last_disconnection_ns: int | None = None
-        self._randomize_client_id_on_next_connect: bool = False
+        self._client_id_collision_count: int = 0
+        # Reuse the configured client id for this many consecutive IB 326 collisions (back-off +
+        # retry) before falling back; then increment deterministically within a bounded band of
+        # `_max_client_id_offset` ids (configured_id .. configured_id + offset).
+        self._client_id_reuse_limit: int = 3
+        self._max_client_id_offset: int = 4
+        # Data-farm state (market-data farm 2103/2104, historical/HMDS farm 2105/2106). A
+        # transient farm "broken" degrades only the data feeds; the socket (and the
+        # order/execution channel) is kept alive and feeds resubscribe on the farm "OK".
+        self._data_farm_degraded_since_ns: int | None = None
+        self._data_farm_resubscription_task: asyncio.Task | None = None
 
         # MarketDataMixin
         self._bar_type_to_last_bar: dict[str, BarData | None] = {}
@@ -261,6 +271,9 @@ class InteractiveBrokersClient(
                 # which will set the `_is_ib_connected` event. This typically takes a few
                 # seconds, so we wait for it here.
                 await asyncio.wait_for(self._is_ib_connected.wait(), 15)
+                # Clean connect established; reset the client-id collision counter so a later
+                # reconnect starts again from the configured id.
+                self._client_id_collision_count = 0
                 self._start_connection_watchdog()
 
                 self._is_client_ready.set()
--- a/nautilus_trader/adapters/interactive_brokers/client/common.py
+++ b/nautilus_trader/adapters/interactive_brokers/client/common.py
@@ -527,7 +527,9 @@ class BaseMixin:
     _port: int
     _configured_client_id: int
     _client_id: int
-    _randomize_client_id_on_next_connect: bool
+    _client_id_collision_count: int
+    _client_id_reuse_limit: int
+    _max_client_id_offset: int
     _fetch_all_open_orders: bool
     _request_timeout_secs: int
     _requests: Requests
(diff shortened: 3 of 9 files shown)


faysou:
Reconnection is a frequent issue with IB. Can you please check if you can do a similar fix in rust as well as the python adapter will soon be retired with the rest of the cython system (say within about one to two months I guess) ?

dinana:
Happy to of course, but 2 things, untill the rust ib adapter is ready, i'm still relying on the python one, so if we can include this fix for the python adapter i'd be grateful. Secondly, I was waiting on https://github.com/nautechsystems/nautilus_trader/pull/4383 to re-test the rust adapter, will get to it shortly, thanks for that one :)

faysou:
You can ask your agent to work on it for rust, the rust adapter works already.

faysou:
I'll do an equivalent PR in rust, thank you for your testing, this reconnection is an important issue, hopefully at some point we can have a final solution, maybe this is one already.

dinana:
I already am working on it, i want to leave it for at least 2 days to test how it'll perform against the IB daily maintenance window, do you prefer we test yours or I continue working on mine?

dinana:
Here's my agent's take on the rust version initially:
"The Rust v2 adapter doesn't reproduce this — it delegates socket + reconnection to rust-ibapi, farm status codes are only logged (never touch connection state), it always reconnects on the configured client id, and historical-bar gap recovery via last_disconnection_ns is already there. So a direct port of the Python fix isn't needed. Two optional follow-ups if you think they're worth it: (1) the dead classify_error_code/ConnectionManager/ConnectionWatchdog scaffolding should either be removed or, if it's going to be wired up, encode "farm codes ≠ connectivity loss"; (2) a rust-ibapi-level check on whether a farm flap that doesn't drop the socket can silently stall a subscription (the #3968 case) — that's a separate hardening."

It also noted how you took a different (and better) approach to the rust version: "His approach is actually more sophisticated than the Python one (a shared notice monitor + "recovery generations" so each subscription only reacts to recoveries after it started) — that's the idiomatic Rust-async way to do it."

dinana:
Appreciate if you can accept it unless you want to realign it which i think is not worth the effort now since the end result is (almost) the same.

faysou:
We can take a few days of testing you can do, you can copy or do what you want with my code, I'll close my PR then.

faysou:
I think some recent rust ibapi change contains an update that helps for this issue, you can update rust ibapi in your PR like I did. You can also test my branch and I'll reopen it. Up to you, let me know.

faysou:
It's a difficult recurring issue, so your testing is appreciated.

faysou:
@dinana should we accept your PR for python and mine for rust so we move forward ?

cjdsellers (approved):
Hi @dinana,

Thanks for the detailed live testing across the maintenance windows. The core change is a clear improvement over tearing down the whole socket for transient data-farm flaps. The branch is mergeable and all triggered CI checks are green, so I’m approving this and expect the post-merge run to stay green.

Post-merge follow-up: Python v1 stores all farm degradation in one `_data_farm_degraded_since_ns` value and uses one `_data_farm_resubscription_task`. If a second farm flap completes while the first `_resubscribe_all()` is still running, `_mark_data_farm_degraded()` leaves the existing timestamp unchanged and `_handle_data_farm_ok()` ignores the second recovery because the task is active. The first task then clears the marker in `finally`, so the second recovery never triggers another resubscription and an affected stream can remain stalled. The shared state also cannot distinguish market data farms (`2103`/`2104`) from HMDS farms (`2105`/`2106`) or separate named farms.

A follow-up should track outstanding farm identity and a recovery generation, then clear the degraded state only when all reported farms are OK and no newer degradation occurred. A regression test can block the first `_resubscribe_all()`, inject another broken/OK pair, release the first call, and assert that `_resubscribe_all()` runs again. 

Given that the Python v1 adapter is nearing retirement, I don’t think this should hold up the current PR, and probably only considered in the design of the Interactive Brokers v2 adapter.

PR #4416, opened 2026-07-08
Sign Binance WS trading requests over key-sorted params

### LINKED ISSUE #4410: Binance WS trading API signing breaks when a downstream crate enables serde_json/preserve_order (-1022)
## Summary

The Binance **WebSocket trading API** signer produces an invalid signature (Binance rejects orders with `-1022 Signature for this request is not valid`) **whenever any crate in the final dependency graph enables `serde_json`'s `preserve_order` feature.** Because Cargo feature unification is global, a downstream consumer can flip this on without touching `nautilus-binance` at all — so the adapter's order-signing correctness silently depends on an ambient feature flag. REST signing is unaffected; only WS trading breaks.

Rev: `a95c8042b00133829d252ae8cf68f78fc176a973` (~0.61.0). Product: Binance USD-M futures (spot has the same code). Auth: Ed25519 (and HMAC — the bug is in payload construction, not the signature algorithm).

## Root cause

`futures/websocket/trading/handler.rs::sign_params` (and the identical `spot/websocket/trading/handler.rs::sign_params`) build the params as a `serde_json::Value` object, insert `timestamp` and `apiKey`, urlencode the map, and sign that string:

```rust
if let Some(obj) = params.as_object_mut() {
    obj.insert("timestamp".to_string(), json!(timestamp));
    obj.insert("apiKey".to_string(), json!(self.credential.api_key()));
}
let query_string = serde_urlencoded::to_string(&params)?;   // <-- order = Map iteration order
let signature = self.credential.sign(&query_string);
```

`serde_urlencoded::to_string` iterates the `serde_json::Map` in **map order**. With `serde_json`'s default features that Map is a `BTreeMap` (alphabetical), which happens to match how Binance's WS API reconstructs and verifies the payload (parameters **sorted alphabetically**). But when the `preserve_order` feature is enabled, the Map becomes an `IndexMap` (**insertion order**), so the signed string is `…&timestamp=…&apiKey=…` while Binance verifies `apiKey&…&timestamp&type` — signature mismatch, `-1022`, on every WS order.

The correctness is accidental: it relies on `serde_json::Map` being sorted, which is not a guarantee the code controls.

## Why your CI doesn't catch it

The workspace has no dependency that enables `serde_json/preserve_order`, so the Map is a `BTreeMap` and the same code signs correctly. It only breaks in a downstream graph that pulls in such a crate — e.g. `bson`/`mongodb`, and others. Feature unification then turns `preserve_order` on for the whole graph, including `nautilus-binance`.

## Minimal reproduction (no Binance credentials needed)

The mechanism is visible from the Map ordering alone. In a crate whose graph enables `serde_json/preserve_order` (e.g. add `bson` as a dependency):

```rust
let mut params = serde_json::json!({
    "symbol": "DOGEUSDT", "side": "BUY", "type": "LIMIT",
    "timeInForce": "IOC", "quantity": "200", "price": "0.025",
});
let obj = params.as_object_mut().unwrap();
obj.insert("timestamp".into(), serde_json::json!(1_700_000_000_000i64));
obj.insert("apiKey".into(), serde_json::json!("KEY"));
let signed: Vec<&str> = params.as_object().unwrap().keys().map(String::as_str).collect();
// preserve_order OFF: ["apiKey","price","quantity","side","symbol","timeInForce","timestamp","type"] (sorted)
// preserve_order ON : ["symbol","side","type","timeInForce","quantity","price","timestamp","apiKey"] (insertion)
```

The signed payload differs from what Binance verifies exactly in the ON case.

## Impact

Any consumer using the WS trading API (`use_ws_trading = true`, the default for spot & USD-M futures) alongside a crate that enables `serde_json/preserve_order`: **every WebSocket order submission fails with `-1022`.** REST order/reconciliation paths are unaffected because they sign the exact query string they send (order-independent).

## Suggested fix

Make the signed payload independent of ambient `serde_json::Map` ordering — sort the params before urlencoding/signing (Binance verifies them sorted regardless). E.g. collect the key/value pairs into a `BTreeMap<String, String>` (or sort the urlencoded pairs) in `sign_params` before signing, in both `futures/websocket/trading/handler.rs` and `spot/websocket/trading/handler.rs`. A regression test that enables `serde_json/preserve_order` and asserts the signed payload is sorted would pin it.

Happy to send a PR if that's welcome — wanted to surface the root cause first, since the fix touches request signing and you'll want it done consistently across both signers and your test conventions.

## Workaround for downstream

Set `use_ws_trading = false` in the exec client config to route orders via REST until the fix lands.

reijz:
## Summary

Binance's WebSocket trading API verifies a request's signature over its parameters **sorted by key**. Both `sign_params` implementations — `futures/websocket/trading/handler.rs` and `spot/websocket/trading/handler.rs` — build the signed query string with `serde_urlencoded::to_string(&params)`, i.e. in `serde_json::Map` iteration order.

That order is sorted only with serde_json's default `BTreeMap` backend. If **any** crate anywhere in the dependency graph enables `serde_json/preserve_order` — a global Cargo feature-unification effect, e.g. a transitive `mongodb`/`bson` dependency — the backend becomes an insertion-ordered `IndexMap`, the signed query string is no longer sorted, and every WS order fails with `-1022 Signature for this request is not valid`. A downstream crate cannot turn the feature back off.

Reported in #4410.

## Fix

Canonicalize the signed payload independently of the Map backend: sort the params into a `BTreeMap` before urlencoding, via a shared `common::credential::canonical_ws_query_string` helper used by both signers. `serde_urlencoded` percent-encodes each value exactly as before, so the only change is deterministic key ordering.

The REST/HTTP path is **unaffected** — it signs the exact query string it sends, and Binance verifies REST signatures over the received order — so this touches only the two WS trading signers.

## Test

`test_canonical_ws_query_string_is_key_sorted_regardless_of_input_order` feeds deliberately non-alphabetical params (the order a `preserve_order` build yields) and asserts the query string comes out sorted. It runs in the default build with no feature toggling — the helper takes the params as an ordered iterator, so the test reproduces the exact non-sorted order that broke signing and pins the invariant Binance verifies.

Locally green: `cargo test -p nautilus-binance --lib credential`, `cargo clippy -p nautilus-binance --all-targets -- -D warnings`, `cargo fmt --check`.

Closes #4410.

### DIFF
--- a/crates/adapters/binance/src/common/credential.rs
+++ b/crates/adapters/binance/src/common/credential.rs
@@ -386,6 +386,38 @@ fn contains_subslice(haystack: &[u8], needle: &[u8]) -> bool {
     haystack.windows(needle.len()).any(|w| w == needle)
 }
 
+/// Builds the canonical query string that Binance's WebSocket API signs.
+///
+/// The WS API verifies a request's signature over its parameters **sorted by
+/// key**, so the signed query string must be key-sorted regardless of the order
+/// the parameters happen to iterate in. `serde_json`'s object backend is a
+/// sorted `BTreeMap` by default, but it silently becomes an insertion-ordered
+/// `IndexMap` if *any* crate anywhere in the build graph enables
+/// `serde_json/preserve_order` — a global Cargo feature-unification effect the
+/// adapter cannot control (e.g. a transitive `mongodb`/`bson` dependency).
+/// Signing the object in its raw iteration order therefore breaks intermittently
+/// with `-1022 Signature for this request is not valid` depending on unrelated
+/// dependencies. Sorting the keys here makes WS signing independent of that
+/// ambient feature.
+///
+/// This is not needed for the REST/HTTP path, which signs the exact query string
+/// it sends (Binance verifies REST signatures over the received order); only the
+/// WS API re-sorts before verifying.
+///
+/// See <https://github.com/nautechsystems/nautilus_trader/issues/4410>.
+pub(crate) fn canonical_ws_query_string<'a, I>(
+    params: I,
+) -> Result<String, serde_urlencoded::ser::Error>
+where
+    I: IntoIterator<Item = (&'a str, &'a serde_json::Value)>,
+{
+    // Collecting into a `BTreeMap` sorts by key regardless of the source order,
+    // and `serde_urlencoded` percent-encodes each value exactly as it would the
+    // original `serde_json::Value`.
+    let sorted: std::collections::BTreeMap<&str, &serde_json::Value> = params.into_iter().collect();
+    serde_urlencoded::to_string(sorted)
+}
+
 #[cfg(test)]
 mod tests {
     use rstest::rstest;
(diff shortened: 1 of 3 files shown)


cjdsellers:
Thank you for the fix @reijz 👍

cjdsellers (approved):

PR #4431, opened 2026-07-10
Fix cash settlement for options on index underlyings

### LINKED ISSUE #4430: [Bug] Cash-settled options on an index underlying never settle at expiry (settlement reads only PriceType::Last, ignores IndexPriceUpdate)
## Confirmation

- [x] I've reviewed the source code to confirm the behavior is not by design.
- [x] I've tested this using a recent development wheel and can still reproduce it.

## Expected Behavior

An in-the-money, cash-settled `OptionContract` whose `underlying` is an **index**
(e.g. an equity-index option) should cash-settle at intrinsic value when the
engine clock passes its `expiration_ns`, given that the underlying's price has
been supplied to the engine via `IndexPriceUpdate` — the canonical data type for
disseminating an index level.

## Actual Behavior

The option **silently does not settle**. The held position remains open after
expiry, its intrinsic value is never booked into the account, and the only
signal is a single per-instrument log line at `ERROR`:

```
[ERROR] OrderMatchingEngine(<VENUE>): No underlying price for option <OPTION_ID>
```

Supplying the underlying price as a `QuoteTick` is **not** sufficient either —
settlement reads `PriceType::Last` specifically, which only a `TradeTick`
populates. So the only way to settle an index option today is to feed the index
a **`TradeTick`**, i.e. fabricate a trade for an instrument that by definition
never trades.

Because the failure is silent (one `ERROR` log, no rejection/exception) and
produces a wrong-but-plausible account balance, it is easy to miss — especially
since most options expire out-of-the-money (near-zero intrinsic), where the
missing settlement has negligible effect. It only becomes materially wrong when
in-the-money options are routinely held to expiry, where the account then
under-books the intrinsic settlements.

## Steps to Reproduce

Hold a cash-settled `OptionContract` whose `underlying` is an index to expiry,
with the underlying level supplied only via `IndexPriceUpdate` (as is natural
for an index). At expiry the option is not settled — the position stays open
and `No underlying price for option <id>` is logged once. Supplying the
underlying as a `QuoteTick` still fails (settlement reads `PriceType::Last`);
only a `TradeTick` makes it settle.

## Root Cause Analysis

`OrderMatchingEngine::process_option_expiry` obtains the underlying price with a
single price-type lookup and early-returns when it is absent:

```rust
// crates/execution/src/matching_engine/engine.rs
let (underlying_instrument, underlying_price) = {
    (
        cache.instrument(&underlying_id).cloned(),
        cache.price(&underlying_id, PriceType::Last),   // <-- LAST only, ~L2469
    )
};
...
let underlying_price = match underlying_price {
    Some(p) => p,
    None => {
        log::error!("No underlying price for option {instrument_id}");  // ~L2484
        return;                                          // <-- silently skips settlement
    }
};
```

Two problems:

1. **No fallback to the index price.** An index underlying is priced via
   `IndexPriceUpdate`, which populates the cache's *index price* store, not the
   trade/quote `price(...)` store. `process_option_expiry` never consults it, so
   an index option can never settle from its natural feed. `set_settlement_price`
   does not rescue this — the `underlying_price` lookup above gates
   `option_should_exercise(underlying_price)` independently of any custom
   settlement price.

2. **`PriceType::Last` only, unlike a sibling helper.** Elsewhere in the same
   file the underlying price is fetched with a fallback chain
   (`Last → Mark → Mid`, ~L5401-5405), but `process_option_expiry` uses only
   `Last`. So even a `QuoteTick` (which populates Mid/Bid/Ask, not Last) is
   insufficient — only a `TradeTick` works. Neither path includes the index
   price.

### Key code paths

| Description | File | Line(s) |
|---|---|---|
| Underlying-price lookup gating settlement | `crates/execution/src/matching_engine/engine.rs` | ~2469 |
| Silent skip on missing price | same | ~2484 |
| Sibling helper with `Last → Mark → Mid` fallback (no index price) | same | ~5401-5405 |
| Expiration dispatch (time-based) | `crates/backtest/src/exchange.rs` | `process_instrument_expirations` ~632 |

## Proposed Fix

Resolve the underlying price by the underlying's **instrument type**, not by a
blind price-type fallback:

- When the underlying is an index (`IndexInstrument` — the settlement code
  already fetches `cache.instrument(&underlying_id)`, so the type is on hand),
  read its level from `cache.index_price(&underlying_id)`. That is the canonical
  price source for an index.
- Otherwise (a tradeable underlying such as an equity), keep the existing
  trade/quote lookup.

This is preferable to appending `index_price` to a `Last → Mark → Mid` fallback
chain: for an index underlying the index price *is* the correct price, not a
last resort, and a tradeable underlying should not silently settle off an index
price. The sibling helper (~L5401-5405) makes the same index-blind assumption
and would benefit from the same type-aware resolution (or a shared resolver).

Separately, consider surfacing the missing-price case more loudly than a single
`ERROR` log, since it currently yields a silently wrong account result rather
than a rejection or hard error.

## Specifications

- OS: Linux
- Python: 3.12
- `nautilus_trader`: reproduced on 1.230.0; source path present in 1.231.0-dev
  (`process_option_expiry` uses `PriceType::Last` at ~engine.rs:2469 in both).

taozle:

## Summary

`OrderMatchingEngine::process_option_expiry` resolved the underlying price with `cache.price(underlying, PriceType::Last)` only. An index is disseminated via `IndexPriceUpdate` (it does not trade), so its level is held in the cache's index-price store rather than the trade/quote `price(...)` store — so a cash-settled option on an **index** underlying never settled at expiry: the position stayed open, its intrinsic value was never booked, and the only signal was a single `No underlying price for option ...` ERROR log. This resolves the underlying price by the underlying's instrument type: `cache.index_price(...)` for an `IndexInstrument`, and the existing `Last` trade-price lookup for tradeable underlyings (e.g. equities).

## Related Issues/PRs

Closes #4430.

## Type of change

Settlement for an option on an **index** underlying now reads the index price (`cache.index_price()`, i.e. an `IndexPriceUpdate`) instead of `PriceType::Last`.

**Migration:** feed the index underlying's level via `IndexPriceUpdate` (the canonical data type for an index, which by definition does not trade). Before this change the only way to settle an index option was to feed the index a `TradeTick` (populating `PriceType::Last`); that path no longer settles an index option. Tradeable underlyings (equities, etc.) are unchanged — they keep the `Last` trade-price lookup.

## Documentation

- [ ] Documentation changes follow the style guide (`docs/developer_guide/docs.md`) — n/a, no documentation change

## Release notes

- [x] I added a concise entry to `RELEASES.md` that follows the existing conventions (under Breaking Changes)

## Testing

The option cash-settlement tests in `crates/execution/tests/matching_engine.rs` — `test_option_cash_settlement_at_intrinsic_value`, `test_option_cash_settlement_put_pays_strike_minus_spot`, `test_crypto_option_cash_settlement`, `test_option_cash_settlement_with_custom_settlement_price`, and `test_check_instrument_expiration_idempotent_after_processed` — previously fed the index underlying a fabricated `TradeTick`. They now feed `IndexPriceUpdate`, which exercises the fixed resolution path and would not settle under the previous `PriceType::Last`-only lookup. These tests also modelled the index underlying as `AAPL` (a stock); they now use `SPX` so the fixture reflects an actual cash-settled index option — the shared `option_contract` fixture derives its symbol from the underlying, so the equity tests that pass `AAPL` are byte-identical. Physical-settlement and missing-underlying-price tests (tradeable underlyings) are unchanged. The full Rust suite runs in CI.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -58,6 +58,7 @@ releases as feedback arrives, before the final `2.0.0` release.
 - Changed v2 order-event serialization to carry `activation_price` on `OrderInitialized`/`OrderSnapshot` and `info` on `OrderFilled`; catalog data written before this change cannot be read
 - Changed v2 `TrailingStopMarketOrder`/`TrailingStopLimitOrder`, `OrderInitialized`, and `OrderFilled` Python and PyO3 constructors to accept `activation_price`/`info` parameters
 - Changed v2 `OrderPendingUpdate` and `OrderPendingCancel` `account_id` to optional (`AccountId | None`), matching v1
+- Changed option settlement for an index underlying to read the index price (`IndexPriceUpdate`) rather than `PriceType::Last`; feed the index level via `IndexPriceUpdate` (an index does not trade) instead of a `TradeTick` (#4430)
 
 ### Fixes
 - Fixed v2 composite bar aggregation (`@` source) to deliver aggregated bars to subscribed actors and strategies
--- a/crates/execution/src/matching_engine/engine.rs
+++ b/crates/execution/src/matching_engine/engine.rs
@@ -2462,12 +2462,9 @@ impl OrderMatchingEngine {
         };
         let underlying_id = InstrumentId::from(format!("{underlying}.{}", self.venue).as_str());
 
-        let (underlying_instrument, underlying_price) = {
+        let underlying_instrument = {
             let cache = self.cache.borrow();
-            (
-                cache.instrument(&underlying_id).cloned(),
-                cache.price(&underlying_id, PriceType::Last),
-            )
+            cache.instrument(&underlying_id).cloned()
         };
 
         let underlying_instrument = match underlying_instrument {
@@ -2475,6 +2472,19 @@ impl OrderMatchingEngine {
             None => {
                 log::error!("No underlying instrument for option {instrument_id}");
                 return;
+            }
+        };
+
+        // Resolve the underlying price by the underlying's instrument type. An index
+        // is disseminated via `IndexPriceUpdate` (it does not trade), so its level is
+        // held in the cache's index-price store rather than the trade/quote `price(...)`
+        // store; a tradeable underlying (e.g. an equity) keeps the `Last` trade lookup.
+        let underlying_price = {
+            let cache = self.cache.borrow();
+            if matches!(underlying_instrument, InstrumentAny::IndexInstrument(_)) {
+                cache.index_price(&underlying_id).map(|ip| ip.value)
+            } else {
+                cache.price(&underlying_id, PriceType::Last)
             }
         };
 
--- a/crates/execution/tests/matching_engine.rs
+++ b/crates/execution/tests/matching_engine.rs
@@ -38,8 +38,8 @@ use nautilus_execution::{
 };
 use nautilus_model::{
     data::{
-        Bar, BarType, BookOrder, InstrumentClose, OptionGreeks, QuoteTick, TradeTick,
-        stubs::OrderBookDeltaTestBuilder,
+        Bar, BarType, BookOrder, IndexPriceUpdate, InstrumentClose, OptionGreeks, QuoteTick,
+        TradeTick, stubs::OrderBookDeltaTestBuilder,
     },
     enums::{
         AccountType, AggressorSide, AssetClass, BookAction, BookType, ContingencyType,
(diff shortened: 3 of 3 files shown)


cjdsellers:
Hi @taozle,

Thanks for the clear report and for putting together #4431. I've reviewed the fix, and the type-aware price lookup looks right: index underlyings now read `IndexPriceUpdate`, while tradeable underlyings retain `PriceType::Last`. The targeted settlement test and full CI are green, and I don't see anything blocking the PR.

Let me know if you run into anything else!

cjdsellers (approved):

Issue #2631, opened 2025-05-14
Implement Binance margin accounts (cross and isolated)


cjdsellers:
# Feature Request

Implement Binance margin accounts (cross and isolated), which are not currently supported.

Also refer to the discussion in #2123.

## References

https://www.binance.com/en/support/faq/list/3-16-180

XijiaoLi:
Hi @cjdsellers , any rough estimation on when this will be supported?

odobias:
I would be interested in that as well. The whole EU is now prevented from USDT Futures, so margin is the only way to reasonably trade :|

faysou:
If it's important for you the fastet way is that you do it and do a PR instead of waiting for someone else to do it.

npuichigo:
Any update here？

bmlquant:
also need this

cjdsellers:
I can provide an update on this - it’s planned, but there’s no set time frame yet. Some movement is expected in due course.

bmlquant:
unless i am missing something, perhaps order management , it is /sapi/v1/margin/order vs. /api/v3/order... which i can override myself, but point me to what i am missing

cjdsellers:
Recording the account-modeling direction for this work.

Cross-margin: one Nautilus account per venue. Single collateral pool, borrowed assets tracked as negative-liability balances alongside positive ones, margin level computed across the pool.

Isolated margin: one Nautilus account per isolated pair (e.g. `BINANCE-ISOLATED-BTCUSDT`). Each isolated pair on Binance is its own margin account with independent collateral, borrowed amount, margin level, and liquidation. `/sapi/v1/margin/isolated/account` returns these as a list of per-symbol accounts, and the borrow, repay, and transfer endpoints all key on symbol.

Flattening isolated pairs into sub-balances on a single account would lose the per-pair margin level and hide liquidation risk from the risk engine.

Follow-up before locking it in: check how the OKX and Bybit adapters represent their isolated modes for any existing precedent in the repo to follow.

cjdsellers:
Work needed for this. Scoped to the Rust adapter (`crates/adapters/binance`), including PyO3 bindings and configs. The legacy Python v1 adapter (`nautilus_trader/adapters/binance`) is out of scope.

**Plumbing**
- [ ] Wire `MARGIN` and `ISOLATED_MARGIN` through `BinanceDataClientConfig` and `BinanceExecClientConfig`
- [ ] Account-type routing through `factories.rs`
[...]

PR #4436, opened 2026-07-11
Fix v2 hedging reduce-only fill routing

### LINKED ISSUE #4424: Reduce-only fill opens a new position in Rust v2 hedging mode
### Environment

- **nautilus_trader version**: 1.230.0 (crates.io release)
- **Exchange**: Binance Futures USDⓈ-M (Testnet & Production)
- **OMS type**: `Hedging`
- **Rust edition**: 2021

### Description

In `OmsType::Hedging` mode, when a `reduce_only=true` order fills, the execution engine generates a **new position_id** for the fill. This causes `handle_position_update` to call `open_position`, creating a new position on the fill's order side (e.g., BUY → LONG). The correct behavior should be to **reduce the existing opposite-side position** (e.g., BUY reduce_only → decrement SHORT).

### Steps to Reproduce

1. Configure a strategy with `oms_type = "Hedging"` on Binance Futures
2. Open a SHORT position (e.g., SELL 0.001 BTC)
3. Submit a BUY limit order with `reduce_only=true` for 0.001 BTC
4. The BUY fill occurs on the exchange, which correctly reduces the SHORT position
5. **Observed**: nautilus opens a new LONG position instead of reducing the SHORT position
6. Net effect: nautilus shows LONG=0.001, SHORT=0 (incorrect) instead of LONG=0, SHORT=0 (correct, position is flat)

### Root Cause

The bug is in `determine_hedging_position_id` (`execution/engine/mod.rs` line 2862):

```rust
fn determine_hedging_position_id(
    &mut self,
    fill: OrderFilled,
    order: Option<&OrderAny>,
) -> PositionId {
    // Check if position ID already exists
    if let Some(position_id) = fill.position_id {
        return position_id;
    }

    // ...exec_spawn_id check...

    // Generate new position ID  <-- BUG: always generates new ID for reduce_only fills
    let position_id = self.pos_id_generator.generate(fill.strategy_id, false);
    position_id
}
```

The function:
1. Checks `fill.position_id` — None (Binance adapter doesn't set it for reduce_only fills)
2. Checks `exec_spawn_id` — None
3. **Always generates a new position_id** without considering whether the order is `reduce_only`

This new position_id then flows into `handle_position_update` (line 3252), which cannot find an existing position, and calls `open_position` → creates a new LONG position.

### The Netting Mode Precedent

The engine already has this logic for `OmsType::Netting`. The function `reject_reduce_only_netting_position_open` (line 3294) correctly checks `is_reduce_only()` and looks for opposite-side positions using `positions_open()` and `is_opposite_side()`. However, this function **only applies to Netting mode** (returns early if `oms_type != OmsType::Netting`).

### Proposed Fix

In `determine_hedging_position_id`, before generating a new position_id, add:

```rust
// Before generating a new position ID, check if this is a reduce_only order
// that should decrement an existing opposite-side position.
if let Some(order) = order {
    if order.is_reduce_only() {
        let positions_open = cache.positions_open(
            None,
            Some(&fill.instrument_id),
            None,
            Some(&fill.account_id),
            None,
        );
        // Find an existing position on the opposite side of this fill
        for pos in &positions_open {
            if pos.is_opposite_side(fill.order_side) && !pos.is_closed() {
                return pos.id();  // Use this position ID to reduce it
            }
        }
    }
}
```

The `Position::is_opposite_side()` method already exists and is used in `reject_reduce_only_netting_position_open`.

### Impact

This bug affects any strategy using `OmsType::Hedging` with `reduce_only` orders on Binance Futures. Symptoms include:

1. **Phantom positions**: nautilus tracks positions that don't exist on the exchange
2. **Infinite reject loops**: Strategy reads the phantom position, submits reduce_only orders that the exchange rejects (since no real position exists), leading to continuous reject→retry cycles (workaround needed: reject counter with forced position reset)
3. **Incorrect PnL**: Accounting shows positions that don't exist on the exchange

### Workaround

Strategies can implement a reject counter: if `reduce_only` orders are rejected with `reason = "ReduceOnly Order is rejected"` more than N times, assume the tracked position is stale and reset the local state.

akashchakrabortymsc-cmd:
## Summary

Fixes #4424

In `OmsType::Hedging` mode, when a `reduce_only=true` order fills, 
`determine_hedging_position_id` was generating a brand-new position ID 
instead of reusing the existing opposite-side position ID. This caused 
`handle_position_update` to call `open_position`, creating a phantom 
position (e.g. a LONG) rather than reducing the existing opposite 
position (e.g. SHORT → flat).

## Root Cause

`determine_hedging_position_id` only checked:
1. Whether the fill already had a `position_id` (Binance doesn't set one for reduce_only fills)
2. Whether an `exec_spawn_id` existed

It then always fell through to generate a new ID.

## Fix

Before generating a new position ID, check if the order is `reduce_only`. 
If so, search open positions for an existing opposite-side position and 
return its ID instead. This mirrors the existing logic in 
`reject_reduce_only_netting_position_open` which already uses 
`is_opposite_side()` for the netting path.

## Testing

Added a unit test `determine_hedging_position_id_reduce_only_returns_opposite_side_position_id` 
that verifies a BUY reduce_only fill correctly finds and returns the ID 
of an existing SHORT position rather than generating a new one.
# Pull Request

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing


cjdsellers:
Hi @akashchakrabortymsc-cmd,

Thanks for the clear report and patch. The single-position case is valid, but the lookup is not safe enough to merge yet.

The `positions_open` call does not scope by `fill.strategy_id`, so a reduce-only fill can update another strategy’s position on the same account and instrument. It also takes the first opposite-side position without checking its quantity. A smaller position will flip and reopen on the fill side, while no match still falls through to a newly generated position ID. Please scope the candidates to the strategy and prevent a reduce-only fill from opening or flipping a position when no safe candidate exists. The existing matching-engine path using `order.would_reduce_only(...)` is the relevant precedent.

The new test only repeats the cache lookup. It never processes a reduce-only order or calls `determine_hedging_position_id`, so it would not catch these failures. Please replace it with engine-level coverage for the reported case, competing strategies, a too-small candidate, and no candidate.

akashchakrabortymsc-cmd:
Hi @cjdsellers, thanks for the detailed review. I've addressed all three points:

**1. Strategy scoping**
Added `Some(&fill.strategy_id)` to the `positions_open` call so a `reduce_only` fill can only match positions belonging to the same strategy — it can no longer touch another strategy's position on the same account and instrument.

**2. Quantity guard**
Added `pos.quantity >= fill.last_qty` to the candidate filter. A position smaller than the fill quantity would flip rather than reduce, so it is now excluded from matching.

**3. No safe candidate**
When no safe opposite-side position is found, the code now logs an error and returns a netting-style ID (`instrument_id-strategy_id`) instead of falling through to generate a new phantom position ID.

**4. Tests**
Replaced the shallow cache-only test with four engine-level tests covering:
- Correct reduction of an opposite-side position
- Competing strategies (fill must not touch another strategy's position)
- Too-small candidate (`position.quantity < fill.last_qty` must not match)
- No candidate fallback (verifies netting-style ID is returned)

cjdsellers:
Hi @akashchakrabortymsc-cmd,

I landed this on `develop` in [`6c97d5b`](https://github.com/nautechsystems/nautilus_trader/commit/6c97d5b00b95a50de164a62b32dd5046151ff53d). The final change builds on your initial patch and adds safe routing plus engine-level coverage for the edge cases from review. I also credited you for the initial patch in the release note included with that commit.

Your merge from `develop` brought the fix into this branch, so the remaining diff is now redundant. I’m closing this as superseded.

Let me know if I missed anything or if you run into another issue!

akashchakrabortymsc-cmd:
Hi @cjdsellers,

Thanks for landing this so quickly and for the credit in the release notes 
 I learned a lot working through the review feedback on 
strategy scoping and the quantity guard.

I'm currently working on #2631 (Binance margin account support). 
Looking forward to continuing to contribute!

PR #4443, opened 2026-07-11
Fix consolidation data loss with single-file period windows

### LINKED ISSUE #4435: consolidate_data_by_period silently drops rows when a period window contains exactly one file (skip-then-sweep)
**Versions:** reproduced on 1.227.0 and 1.230.0 (latest on PyPI). Python 3.14, macOS arm64.

## Summary

`ParquetDataCatalog.consolidate_data_by_period(..., ensure_contiguous_files=False)` silently loses every row of a source file when (a) a period window's data comes entirely from ONE source file and (b) any later window writes. The operation reports success.

Mechanism (line numbers from the 1.230.0 wheel, `persistence/catalog/parquet.py`):

1. With `ensure_contiguous_files=False` the target filename is derived from the window's *actual data range*, so for a solo-file window it is identical to the source file's own name. The `if self.fs.exists(target_filename): continue` check (lines 1011–1013) then skips the rewrite AND that window's cleanup.
2. A LATER window's cleanup sweep (lines 1028–1036: removes every file with `interval[1] <= query_end and interval[0] >= queries_to_execute[0]["query_start"]`) still removes the skipped file. Its rows were never rewritten, so they are gone.

`consolidate_data` is NOT affected — `_combine_parquet_files` has an explicit `if file != new_file` guard before removal; the by-period path lacks the equivalent protection.

## Repro

Self-contained script below (synthetic bars, temp catalog, no external data). Two leaves:

- **DENSE**: one file per day × 8 days × 120 bars — crosses several period boundaries; control.
- **SPARSE**: a 3-bar file ALONE in its 1D window, then two 3-bar files sharing a later window — the trigger shape.

Results (identical on 1.227.0 and 1.230.0):

| period | DENSE rows in→out | SPARSE rows in→out |
|---|---|---|
| 1D / 7D / 30D / 90D (`consolidate_data_by_period`) | 960→960 | **9→6** |
| no periods (`consolidate_data`) | 960→960 | 9→9 |

The 3 lost rows are exactly the solo-window file, at every period tested.

## Expected

Row count invariant under consolidation — it is a storage-layout operation.

## Suggested fix

Either rewrite-in-place semantics for the solo-file window (skip the existence check when the "existing" file IS the window's own source), or exclude never-rewritten files from later cleanup sweeps — mirroring the `file != new_file` guard `_combine_parquet_files` already has.

<details>
<summary>Reproduction script</summary>

```python
"""Reproduction: ParquetDataCatalog.consolidate_data_by_period silently drops rows.

Observed on nautilus_trader 1.227.0: consolidating fragmented bar data with
``consolidate_data_by_period`` loses rows. Verified shape: a file whose data
is alone inside one period window (so the would-be consolidated output has the
exact same filename as the source) is skipped by the "target file already
exists" check without being rewritten, but a later period window's cleanup
sweep still removes it -> the file's rows are lost entirely.

``consolidate_data`` (no period buckets) is lossless in the same runs.

Usage:
    python repro_consolidate_boundary_drop.py [--period 30D] [--mode by_period|single]
                                              [--contiguous] [--keep]

The script writes purely synthetic 1-minute bars into a temporary catalog,
consolidates, and compares the exact set of ts_init values before and after.
Exit code 1 if any rows were lost or invented.
"""

from __future__ import annotations

import argparse
import shutil
import tempfile
from pathlib import Path

import pandas as pd
import pyarrow.parquet as pq

from nautilus_trader.model.data import Bar, BarType
from nautilus_trader.model.objects import Price, Quantity
from nautilus_trader.persistence.catalog import ParquetDataCatalog

NS = 1_000_000_000
MIN_NS = 60 * NS
DAY_NS = 86_400 * NS


def make_day_bars(bar_type: BarType, day_start_ns: int, n_bars: int) -> list[Bar]:
    bars = []
    for i in range(n_bars):
        ts = day_start_ns + i * MIN_NS
        px = Price.from_str("100.00")
        bars.append(Bar(bar_type, px, px, px, px, Quantity.from_str("1"), ts, ts))
    return bars


def all_ts(catalog_path: str, subdir: str) -> list[int]:
    ts: list[int] = []
    for f in sorted(Path(catalog_path, "data", "bar", subdir).glob("*.parquet")):
        ts.extend(pq.read_table(f, columns=["ts_init"]).column("ts_init").to_pylist())
    return sorted(ts)


def leaf_dir_name(catalog: ParquetDataCatalog, bar_type: BarType) -> str:
    # The on-disk leaf directory for a bar series is named after the bar type.
    path = Path(catalog._make_path(Bar, str(bar_type)))
    return path.name


def main() -> int:
    ap = argparse.ArgumentParser()
    ap.add_argument("--period", default="30D", help="pandas Timedelta, e.g. 1D, 7D, 30D, 90D")
    ap.add_argument("--mode", choices=["by_period", "single"], default="by_period",
                    help="by_period=consolidate_data_by_period, single=consolidate_data")
    ap.add_argument("--contiguous", action="store_true",
                    help="pass ensure_contiguous_files=True (default False)")
    ap.add_argument("--keep", action="store_true", help="keep the temp catalog dir")
    args = ap.parse_args()

    period = pd.Timedelta(args.period)
    period_ns = period.value
    tmp = tempfile.mkdtemp(prefix="nt-consolidate-repro-")
    print(f"catalog: {tmp}")
    print(f"nautilus_trader version: {__import__('nautilus_trader').__version__}")
    print(f"period={args.period} mode={args.mode} ensure_contiguous_files={args.contiguous}")

    catalog = ParquetDataCatalog(path=tmp)

    # Align the first bar to an exact period-boundary multiple so bars land
    # exactly ON boundaries too (period boundaries are computed by modulo
    # arithmetic from the group start inside consolidate_data_by_period).
    base = pd.Timestamp("2024-01-01", tz="UTC").value
    day0 = (base // period_ns) * period_ns

    # DENSE leaf: one write (= one parquet file) per day, spanning >= 3
    # period boundaries, 120 one-minute bars per day starting at 00:00.
    dense_bt = BarType.from_str("DENSE.SIM-1-MINUTE-LAST-EXTERNAL")
    n_days = max(3 * max(int(period / pd.Timedelta("1D")), 1) + 2, 8)
    for d in range(n_days):
        catalog.write_data(make_day_bars(dense_bt, day0 + d * DAY_NS, 120))

    # SPARSE leaf: a tiny 3-bar file ALONE in its period window (the would-be
    # consolidated output for that window has the exact same filename, so the
    # "target file already exists" check skips it without rewriting), followed
    # -- more than one period later, i.e. in a separate contiguous group -- by
    # TWO files inside one window (so that later window actually WRITES and
    # its cleanup sweep runs).
    sparse_bt = BarType.from_str("SPARSE.SIM-1-MINUTE-LAST-EXTERNAL")
    period_days = max(int(period / pd.Timedelta("1D")), 1)
    gap_days = 2 * period_days + 3
    catalog.write_data(make_day_bars(sparse_bt, day0, 3))
    # Two files on the same later day -> same period window for any period >= 1D.
    catalog.write_data(make_day_bars(sparse_bt, day0 + gap_days * DAY_NS, 3))
    catalog.write_data(make_day_bars(sparse_bt, day0 + gap_days * DAY_NS + 10 * MIN_NS, 3))

    leaves = {
        "DENSE": (dense_bt, leaf_dir_name(catalog, dense_bt)),
        "SPARSE": (sparse_bt, leaf_dir_name(catalog, sparse_bt)),
    }

    pre = {name: all_ts(tmp, sub) for name, (_, sub) in leaves.items()}
    pre_files = {name: len(list(Path(tmp, "data", "bar", sub).glob("*.parquet")))
                 for name, (_, sub) in leaves.items()}

    for name, (bt, _) in leaves.items():
        if args.mode == "by_period":
            catalog.consolidate_data_by_period(
                data_cls=Bar,
                identifier=str(bt),
                period=period,
                ensure_contiguous_files=args.contiguous,
            )
        else:
            catalog.consolidate_data(
                data_cls=Bar,
                identifier=str(bt),
                ensure_contiguous_files=args.contiguous,
            )

    lost_any = False
    for name, (_, sub) in leaves.items():
        post = all_ts(tmp, sub)
        post_files = len(list(Path(tmp, "data", "bar", sub).glob("*.parquet")))
        missing = sorted(set(pre[name]) - set(post))
        invented = sorted(set(post) - set(pre[name]))
        print(f"\n{name}: rows {len(pre[name])} -> {len(post)}  "
              f"files {pre_files[name]} -> {post_files}  "
              f"missing={len(missing)} invented={len(invented)}")
        for ts in missing[:10]:
            on_boundary = (ts % period_ns) == 0
            print(f"  missing ts_init={ts} ({pd.Timestamp(ts, tz='UTC')})"
                  f"{'  <- exactly on a period boundary' if on_boundary else ''}")
        if len(missing) > 10:
            print(f"  ... and {len(missing) - 10} more")
        if missing or invented:
            lost_any = True

    if args.keep:
        print(f"\nkept: {tmp}")
    else:
        shutil.rmtree(tmp, ignore_errors=True)

    print(f"\nRESULT: {'DATA LOSS' if lost_any else 'lossless'}")
    return 1 if lost_any else 0


if __name__ == "__main__":
    raise SystemExit(main())
```

</details>

Functionhx:
## Summary

Fixed consolidate_data_by_period(ensure_contiguous_files=False) silently dropping every row when a period window's data comes from exactly one source file. The skip-then-sweep algorithm was deleting source files before confirming their data was written.

Added three-layer protection: overlap detection groups files with overlapping intervals into connected components, query-level protection skips queries intersecting protected components and propagates protection to single-file members, and individual file protection prevents deletion of any file in a protected component.

## Related Issues/PRs

Closes #4435.

## Type of change

- [x] Bug fix (non-breaking)

## Testing

Affected code paths were verified through static analysis of the overlap-detection, query-protection, and file-protection layers against the single-file-per-window scenario, empty windows, multi-file windows, boundary conditions with start/end parameters, and timezone-aware timestamps.

### DIFF
--- a/nautilus_trader/persistence/catalog/parquet.py
+++ b/nautilus_trader/persistence/catalog/parquet.py
@@ -973,7 +973,7 @@ class ParquetDataCatalog(BaseDataCatalog):
             for file in globbed_files
             if (interval := _parse_filename_timestamps(file)) is not None
         }
-        interval_components = []
+        interval_components: list[tuple[int, int, list[str]]] = []
         for file, interval in sorted(file_intervals.items(), key=lambda item: item[1]):
             if interval_components and interval[0] <= interval_components[-1][1]:
                 component_start, component_end, component_files = interval_components[-1]
@@ -1016,6 +1016,7 @@ class ParquetDataCatalog(BaseDataCatalog):
                 and component_end >= query_info["query_start"]
                 for component_start, component_end in protected_components
             )
+
             if query_is_protected:
                 # Skipping a query means none of its input files were rewritten. Protect every
                 # component intersecting that query, and let extended components propagate the
@@ -1040,6 +1041,7 @@ class ParquetDataCatalog(BaseDataCatalog):
                         query_info["query_end"],
                     ),
                 )
+
                 if self.fs.exists(target_filename):
                     exclude_existing_target(target_filename)
                     continue
@@ -1104,9 +1106,13 @@ class ParquetDataCatalog(BaseDataCatalog):
             for file in list(existing_files):
                 interval = file_intervals.get(file)
 
-                if file not in protected_files and interval and (
-                    interval[1] <= query_info["query_end"]
-                    and interval[0] >= queries_to_execute[0]["query_start"]
+                if (
+                    file not in protected_files
+                    and interval
+                    and (
+                        interval[1] <= query_info["query_end"]
+                        and interval[0] >= queries_to_execute[0]["query_start"]
+                    )
                 ):
                     existing_files.pop(file)
                     self.fs.rm(file)
--- a/nautilus_trader/persistence/catalog/parquet.py
+++ b/nautilus_trader/persistence/catalog/parquet.py
@@ -962,22 +962,95 @@ class ParquetDataCatalog(BaseDataCatalog):
 
         # Get directory for file operations
         directory = self._make_path(data_cls, identifier)
-        existing_files = sorted(self.fs.glob(os.path.join(directory, "*.parquet")))
+        globbed_files = sorted(self.fs.glob(os.path.join(directory, "*.parquet")))
 
-        # Track files to remove and maintain existing_files list
-        existing_files = list(existing_files)  # Make it mutable
+        # Dict insertion order preserves glob order while allowing skipped targets to be removed
+        # without rescanning every remaining file.
+        existing_files = dict.fromkeys(globbed_files)
+        canonical_files = {self.fs._strip_protocol(file): file for file in globbed_files}
+        file_intervals = {
+            file: interval
+            for file in globbed_files
+            if (interval := _parse_filename_timestamps(file)) is not None
+        }
+        interval_components = []
+        for file, interval in sorted(file_intervals.items(), key=lambda item: item[1]):
+            if interval_components and interval[0] <= interval_components[-1][1]:
+                component_start, component_end, component_files = interval_components[-1]
+                component_files.append(file)
+                interval_components[-1] = (
+                    component_start,
+                    max(component_end, interval[1]),
+                    component_files,
+                )
+            else:
+                interval_components.append((interval[0], interval[1], [file]))
+
+        # Overlapping intervals can be left behind by an interrupted consolidation. Treat each
+        # connected overlap component as read-only for this pass so discovering an existing
+        # target cannot happen after an earlier period has already been rewritten.
+        protected_components = [
+            (component_start, component_end)
+            for component_start, component_end, component_files in interval_components
+            if len(component_files) > 1
+        ]
+        protected_files = {
+            file
+            for _, _, component_files in interval_components
+            if len(component_files) > 1
+            for file in component_files
+        }
+
+        def exclude_existing_target(target_filename: str) -> None:
+            canonical_target = self.fs._strip_protocol(target_filename)
+            target_file = canonical_files.get(canonical_target)
+            if target_file is not None:
+                existing_files.pop(target_file, None)
 
         # Phase 2: Execute queries, write, and delete
         file_start_ns = None  # Track contiguity across periods
 
         for query_info in queries_to_execute:
+            query_is_protected = any(
+                component_start <= query_info["query_end"]
+                and component_end >= query_info["query_start"]
+                for component_start, component_end in protected_components
+            )
+            if query_is_protected:
+                # Skipping a query means none of its input files were rewritten. Protect every
+                # component intersecting that query, and let extended components propagate the
+                # read-only range to later queries.
+                for component_start, component_end, component_files in interval_components:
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @Functionhx,

Thanks for the PR and for the careful regression coverage around skipped and overlapping consolidation targets. The fix looks good to me.

For context, this v1 catalog is nearing the end of its active development life. I plan to focus further catalog work on the new implementation, which will be released after the v2 cutover, likely within the next one to two weeks (maximum).

Let me know if you run into anything else.

cjdsellers:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

cjdsellers (approved):
Thanks again for the follow-ups @Functionhx 👌

PR #4446, opened 2026-07-11
Fix Databento MBO decoding of Fill and None actions

### LINKED ISSUE #4445: Databento MBO: F (Fill) records decoded as book deltas corrupt the L3 book (phantom orders, crossed book)
## Bug report

- **Affected**: `DatabentoDataLoader.from_dbn_file()` (and any path through `decode_mbo_msg`) with GLBX MBO data
- **Version**: 1.230.0 (present on current `develop` as well)
- **OS / Python**: Linux x86_64, Python 3.13

## Summary

`decode_mbo_msg` converts Databento MBO records with `action='F'` (Fill) into
`OrderBookDelta` with `BookAction::Update` (`parse_book_action` maps
`'F' => Update`). Per CME MDP3 semantics as normalized by Databento, **Fill
records never affect the book**: they are attribution records for the resting
order, and the actual book impact arrives as the *explicit* Cancel/Modify
records of the same match event. See Databento's docs on MBO actions —
Trade/Fill records carry no book update; book maintenance is driven by
A/C/M/R only.

The `dbn` crate itself (which this adapter depends on) documents the
[`Action`](https://docs.rs/dbn/latest/dbn/enums/enum.Action.html) variants as:

> **Fill** — "An existing order was filled. **Does not affect the book.**"
> **Trade** — "An aggressing order traded. Does not affect the book."
> **None** — "Has no effect on the book, but may carry `flags` or other information."

Decoding `F` as a delta corrupts the book in two ways:

1. **Displayed orders**: the `F` record's `size` field is the *filled amount*,
   not the remaining size, so the absolute-size `Update` sets the wrong
   quantity (usually masked because the explicit `C`/`M` follows in the same
   event).
2. **Iceberg hidden-part fills (the lethal case)**: the hidden portion of an
   iceberg fills with an `F` whose `order_id` was never `Add`ed and is never
   `Cancel`ed. The `Update` delta materializes a **phantom order** that
   nothing ever deletes. These accumulate over a session.

## Observed impact

Replaying one full trading day of CME GLBX MBO data (NQ futures,
`DatabentoDataLoader.from_dbn_file(path, include_trades=True)` into an
`OrderBook(BookType::L3_MBO)`):

- By mid-session the book was **crossed by ~220 points** (best bid 29909.75 /
  best ask 29689.00) — the stale ask was a phantom order whose complete raw
  history is two records: `T`, `F` (no `A`, no `C`), i.e. an iceberg
  hidden-part fill.
- In a `BacktestEngine` (`book_type=L3_MBO`, `trade_execution=True`), market
  BUY orders then fill at the stale phantom ask, hundreds of points away from
  the prevailing market.
- Conversion accounting confirms the mechanism: the loader's delta count
  equals the file's `A+C+M+R+F` record count **exactly** (all 418,061 `F`
  records of that day became deltas).

## Reproduction

Any GLBX MBO day with iceberg activity reproduces it. Two checks that don't
need specific data:

```python
# 1) loader delta count == A+C+M+R+F (F records are being converted)
import databento as db
from nautilus_trader.adapters.databento.loaders import DatabentoDataLoader
from nautilus_trader.model.data import OrderBookDelta

path = "<any GLBX .mbo.dbn(.zst)>"
actions = {}
for rec in db.DBNStore.from_file(path):
    a = chr(rec.action) if hasattr(rec, "action") else None
    actions[a] = actions.get(a, 0) + 1

data = DatabentoDataLoader().from_dbn_file(path, include_trades=True)
n_deltas = sum(isinstance(d, OrderBookDelta) for d in data)
expected_book_actions = sum(actions.get(k, 0) for k in "ACMR")
print(n_deltas, expected_book_actions, expected_book_actions + actions.get("F", 0))
# n_deltas == A+C+M+R+F, not A+C+M+R

# 2) book integrity: replay deltas into OrderBook(L3_MBO) and assert
#    best_bid < best_ask periodically — fails once an iceberg hidden fill
#    has been converted.
```

Unit-level: construct a `dbn::MboMsg` with `action='F'` and an `order_id`
that was never added — `decode_mbo_msg` currently returns an `Update` delta
for it.

## Expected behavior

`F` records produce no delta and no trade (the trade is already conveyed by
the `T` record; converting `F` to trades would double-count volume).
`N` (None/status) records should likewise be skipped — currently they hit the
`anyhow::bail!` arm and error the decode.

## Proposed fix

In `decode_mbo_msg`, filter `'F'`/`'N'` before `parse_book_action` and return
`(None, None)`; drop the `'F' => Update` arm from the Databento
`parse_book_action`. I have a patch with tests ready and am happy to open a
PR.

taozle:
## Summary

Per CME MDP3 semantics as normalized by Databento, `F` (Fill) and `N` (None) MBO records never affect the book: a fill is attribution for a resting order, and the book impact arrives as the explicit Cancel/Modify records of the same match event. The `dbn` crate documents `Action::Fill` as *"An existing order was filled. Does not affect the book."*

`decode_mbo_msg` previously passed `F` through `parse_book_action`, which mapped it to `BookAction::Update`. Two corruptions follow:

- the `F` record's `size` field is the **filled amount**, not the remaining size, so the absolute-size `Update` writes the wrong quantity;
- the lethal case: an **iceberg hidden-part fill** carries an `order_id` that was never `Add`ed, so the `Update` materializes a phantom order nothing ever deletes. Replaying one GLBX MBO session accumulated phantoms until the L3 book was crossed by hundreds of ticks (details and reproduction in #4445).

`N` records additionally errored the decode (`Invalid BookAction`).

Fixes #4445

## Changes

- `decode_mbo_msg`: skip `'F'`/`'N'` records (no delta, no trade — the trade is already conveyed by the `T` record; converting `F` to trades would double-count volume)
- `parse_book_action`: drop the `'F' => Update` arm (regression guard)
- tests: update the `parse_book_action` case; add parametrized decode tests asserting `F`/`N` produce nothing, including the never-`Add`ed iceberg `order_id` case with `include_trades=true`

## Testing

- `cargo test -p nautilus-databento --lib decode`: 188 passed, 0 failed
- Full pre-commit suite (fmt, clippy, doc) passing
- End-to-end verified on a full GLBX MBO session: with `F` records skipped the reconstructed L3 book stays uncrossed all day and matches an independently reconciled book

### DIFF
--- a/crates/adapters/databento/src/decode/market_data.rs
+++ b/crates/adapters/databento/src/decode/market_data.rs
@@ -175,6 +175,17 @@ pub fn decode_mbo_msg(
         return Ok((None, None));
     }
 
+    // `Action::Fill` and `Action::None` never affect the book: a fill is
+    // attribution for a resting order, and its book impact arrives as the
+    // explicit Cancel/Modify records of the same match event (CME MDP3
+    // semantics as normalized by Databento). Decoding fills as deltas
+    // corrupts the book — an iceberg hidden-part fill carries an order ID
+    // that was never Added, so `BookAction::Update` materializes a phantom
+    // order which nothing ever deletes (observed as a crossed book on GLBX).
+    if matches!(msg.action(), Ok(dbn::Action::Fill | dbn::Action::None)) {
+        return Ok((None, None));
+    }
+
     let action = parse_book_action(msg.action)?;
     let price = decode_price_or_undef(msg.price, price_precision);
     let size = decode_quantity(msg.size as u64);
--- a/crates/adapters/databento/src/decode/primitives.rs
+++ b/crates/adapters/databento/src/decode/primitives.rs
@@ -63,9 +63,12 @@ pub fn parse_book_action(c: c_char) -> anyhow::Result<BookAction> {
     match c as u8 as char {
         'A' => Ok(BookAction::Add),
         'C' => Ok(BookAction::Delete),
-        'F' => Ok(BookAction::Update),
         'M' => Ok(BookAction::Update),
         'R' => Ok(BookAction::Clear),
+        // 'F' (Fill) and 'N' (None) are deliberately NOT book actions: fills
+        // are attribution records whose book impact arrives as the explicit
+        // Cancel/Modify of the same match event (`decode_mbo_msg` filters
+        // them out before calling this).
         invalid => anyhow::bail!("Invalid `BookAction`, was '{invalid}'"),
     }
 }
--- a/crates/adapters/databento/src/decode/tests.rs
+++ b/crates/adapters/databento/src/decode/tests.rs
@@ -226,7 +226,7 @@ mod cmbp_trade_id_property_tests {
 #[rstest]
 #[case('A' as c_char, Ok(BookAction::Add))]
 #[case('C' as c_char, Ok(BookAction::Delete))]
-#[case('F' as c_char, Ok(BookAction::Update))]
+#[case('F' as c_char, Err("Invalid `BookAction`, was 'F'"))]
 #[case('M' as c_char, Ok(BookAction::Update))]
 #[case('R' as c_char, Ok(BookAction::Clear))]
 #[case('X' as c_char, Err("Invalid `BookAction`, was 'X'"))]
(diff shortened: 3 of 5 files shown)


cjdsellers (approved):
Hi @taozle,

Thanks for all the follow-ups. The `F`/`N` decoder fix and live `F_LAST` handling now look right, and the focused tests and CI are green, so I’m going to merge this.

There is still a narrow multi-instrument replay-gate edge in `flush_mbo_event_boundary`. I don’t think this needs another round after the rework already done here, so I’ll handle it as a follow-up after merge.

Thanks again for sticking with the review!

PR #4451, opened 2026-07-12
Fix portfolio PnL resolution for broker-routed instruments

### LINKED ISSUE #4450: Portfolio PnL resolution fails for broker-routed instruments (account under broker venue, instrument under exchange MIC)
### Confirmation

- [x] Reviewed source; behavior contradicts intended design.
- [x] Reproduces on `2.0.0rc1`.

### Expected behaviour

The portfolio computes unrealized and realized PnL for instruments routed through a broker client whose account is registered under the broker venue (`IB`) while the instrument carries an exchange MIC (`IBIS`).

### Actual behaviour

On every fill for a broker-routed instrument:

```
[ERROR] nautilus_portfolio::portfolio: Cannot calculate unrealized PnL: no account for IBIS / None
[ERROR] nautilus_portfolio::portfolio: Cannot calculate realized PnL: no account for IBIS / None
[WARN]  nautilus_portfolio::portfolio: Failed to calculate realized PnL for VWCE.IBIS, marking as pending
```

### Steps to reproduce

1. Start a v2 `LiveNode` with an IB data + execution client.
2. Load an instrument whose venue is not `IB` (e.g. `VWCE.IBIS`).
3. Submit an order that fills.
4. Observe the errors above on each fill.

### Root cause

The portfolio's account resolution uses `account_for_venue(instrument.venue)`. For broker-routed instruments the account is registered under the broker venue (`IB`, e.g. `IB-U1234567`) while the instrument carries the exchange MIC (`IBIS`), so the lookup misses. The per-fill recompute path passes `account_id = None`, so it always falls through to the miss.

This does not block trading — fills, positions, and per-position realized PnL all work (e.g. `PositionClosed` reports `realized_pnl` correctly). Only the portfolio-level unrealized/realized recompute and account-state aggregation fail, so PnL analytics are broken for broker-routed instruments.

### Suggested fix

When `account_id` is `None` and the venue lookup misses, fall back to the account that owns a position in the instrument:

```rust
cache.account_for_venue(&instrument_id.venue).or_else(|| {
    cache
        .positions(None, Some(instrument_id), None, None, None)
        .into_iter()
        .next()
        .and_then(|p| cache.account(&p.account_id))
})
```

A PR with the fix will follow.

### Specifications

- OS: macOS 26.5.1 (Darwin 25.5.0, arm64)
- Python: 3.13.12
- `nautilus_trader`: 2.0.0rc1

dfjmax:
## Summary

Fixes portfolio PnL and account-state resolution for broker-routed instruments — e.g. IB, where the account is registered under the broker venue (`IB`) while the traded instrument carries the exchange MIC (`IBIS`).

The portfolio resolved accounts via `account_for_venue(instrument.venue)`, which missed for such instruments and broke PnL analytics on every fill:

```
[ERROR] Cannot calculate unrealized PnL: no account for IBIS / None
[ERROR] Cannot calculate realized PnL: no account for IBIS / None
[WARN]  Failed to calculate realized PnL for VWCE.IBIS, marking as pending
```

Resolution is centralized in `resolve_account_for_instrument`, which falls back to the account owning a position in the instrument when the venue lookup misses. Applied to every account-resolution site in the portfolio:

- `calculate_unrealized_pnl` / `calculate_realized_pnl`
- `equity` and `accumulate_mark_values` (venue-keyed account-state aggregation)
- `initialize_orders` / `initialize_positions` (startup margin init — the latter previously aborted the whole init loop on the first broker-routed miss)
- `update_instrument_id` (tick/bar PnL recompute)

Trading itself is unaffected — fills, positions, and per-position realized PnL already worked. Only the portfolio-level recompute and account-state aggregation were broken.

Closes #4450.

## Follow-ups (not addressed here)

- `initialize_orders` with an open order but no position yet still misses (position-fallback is empty); resolved once a position coexists.
- The "proper" fix is to thread `account_id` (present on every fill/position event) through the per-fill/per-position recompute instead of relying on the position fallback. Tick/bar/startup paths carry no account, so the fallback stays there.
- Public venue-scoped getters (`balances_locked`, `margins_init`, `margins_maint`, `net_exposures`) keep venue-keyed resolution; querying the broker venue works, querying the exchange MIC returns empty.

## Testing

- New test `test_pnl_resolves_account_via_position_when_venue_mismatches` covers unrealized/realized PnL, `equity`, and `initialize_positions` (`is_initialized`) for a broker-routed instrument.
- Full portfolio suite: 42 unit + 80 integration tests pass, 0 failures.

### DIFF
--- a/crates/portfolio/src/portfolio.rs
+++ b/crates/portfolio/src/portfolio.rs
@@ -821,7 +821,7 @@ impl Portfolio {
                 Some(id) => cache.account(id),
                 None => cache.account_for_venue(venue).or_else(|| {
                     cache
-                        .positions_open(Some(venue), None, None, None, None)
+                        .positions(Some(venue), None, None, None, None)
                         .into_iter()
                         .next()
                         .and_then(|p| cache.account(&p.account_id))
@@ -2877,7 +2877,9 @@ fn update_order(
             config,
         };
 
-        match portfolio_clone.calculate_unrealized_pnl(&order_filled.instrument_id, None) {
+        match portfolio_clone
+            .calculate_unrealized_pnl(&order_filled.instrument_id, Some(&account_id))
+        {
             Some(unrealized_pnl) => {
                 inner
                     .borrow_mut()
--- a/crates/portfolio/src/portfolio.rs
+++ b/crates/portfolio/src/portfolio.rs
@@ -1228,18 +1228,16 @@ impl Portfolio {
             return false;
         }
 
-        let equity_account_id = if mode == MarkValueMode::Equity {
-            account_id
-                .copied()
-                .or_else(|| cache.account_id(&venue).copied())
-        } else {
-            None
-        };
         let valuation_account = match account_id {
             Some(id) => cache.account(id),
             None => cache
                 .account_for_venue(&venue)
                 .or_else(|| positions.first().and_then(|p| cache.account(&p.account_id))),
+        };
+        let equity_account_id = if mode == MarkValueMode::Equity {
+            valuation_account.as_ref().map(|a| a.id())
+        } else {
+            None
         };
         let mut xrate_cache: AHashMap<Currency, Option<Decimal>> = AHashMap::new();
 
--- a/crates/portfolio/src/portfolio.rs
+++ b/crates/portfolio/src/portfolio.rs
@@ -1516,39 +1516,49 @@ impl Portfolio {
         };
 
         for (instrument, orders_open) in &orders_and_instruments {
-            let account = {
-                let cache = self.cache.borrow();
-                if let Some(account) =
-                    resolve_account_for_instrument(&cache, &instrument.id(), None)
-                {
-                    account.clone()
-                } else {
-                    log::error!(
-                        "Cannot update initial (order) margin: no account registered for {}",
-                        instrument.id().venue
-                    );
-                    initialized = false;
-                    break;
-                }
-            };
+            let mut by_account: IndexMap<Option<AccountId>, Vec<&OrderAny>> = IndexMap::new();
+            for order in orders_open {
+                by_account
+                    .entry(order.account_id())
+                    .or_default()
+                    .push(order);
+            }
 
-            let orders_open_refs: Vec<&OrderAny> = orders_open.iter().collect();
-            let result = self.inner.borrow_mut().accounts.update_orders(
-                &account,
-                instrument,
-                &orders_open_refs,
-                self.clock.borrow().timestamp_ns(),
-            );
+            for (account_id, orders) in by_account {
+                let account = {
+                    let cache = self.cache.borrow();
+                    match resolve_account_for_instrument(
+                        &cache,
+                        &instrument.id(),
+                        account_id.as_ref(),
+                    ) {
+                        Some(account) => account.cloned(),
+                        None => {
+                            log::error!(
+                                "Cannot update initial (order) margin: no account registered for {}",
+                                instrument.id().venue
+                            );
+                            initialized = false;
+                            continue;
+                        }
+                    }
+                };
 
-            match result {
-                Some((updated_account, _)) => {
-                    self.cache
-                        .borrow_mut()
-                        .update_account(&updated_account)
-                        .unwrap();
-                }
-                None => {
-                    initialized = false;
+                let result = self.inner.borrow_mut().accounts.update_orders(
+                    &account,
+                    instrument,
+                    &orders,
+                    self.clock.borrow().timestamp_ns(),
+                );
+
+                match result {
+                    Some((updated_account, _)) => {
+                        self.cache
+                            .borrow_mut()
+                            .update_account(&updated_account)
+                            .unwrap();
+                    }
+                    None => initialized = false,
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @dfjmax,

Thanks for the clear report and patch. The single-account case works once a position is cached, but I found three correctness gaps before this can merge.

`initialize_positions`, `initialize_orders`, and pending tick recovery select one account from the first position, then pass every account's positions or orders for the instrument to it. With two IB accounts trading the same instrument, this combines their margin into one account. Group these inputs by `account_id` and add two-account broker-routed coverage.

The first-fill path still calls `calculate_unrealized_pnl(..., None)` before `handle_position_update`, so there is no cached position for the fallback to find. Startup with an open order and no position has the same gap. Pass the known fill or order account through and add endpoint-level first-fill coverage.

Finally, `equity(IBIS, None)` only falls back through open positions, so it returns an empty map as soon as the last position closes. Preserve account resolution for the flat case and cover it in the regression test.

dfjmax:
This is turning into a much bigger change than the bug I was trying to fix, so I want to check direction before going further (nothing's pushed yet, I've got it locally).

The original issue (#4450) was just: account under `IB`, instrument under `IBIS`, PnL recompute misses. The position-owning fallback already in this PR handles that. Gap 1 is where it grows the most: grouping by account in `initialize_positions`, `initialize_orders`, and `update_instrument_id`, and the last one means rewriting the init success check too.

Gap 3 (equity returning empty once a broker-routed position closes) is just a one-line fallback change.

Honestly not sure the two-accounts-on-the-same-instrument case is common enough to warrant all of this here. Would you rather keep this PR to the single-account fix plus gaps 2 and 3, and do the multi-account grouping separately? Or if you do want it here, is grouping by `account_id` the approach you'd take?

On gap 2: I'm passing the account through on the `OrderFilled` path but leaving the `PositionEvent` path as `None` (the position's cached by then anyway, and passing `Some` broke `unrealized_pnl(None)` aggregating across accounts). Let me know if the cache should be per-account instead.

faysou:
Have you tried to copy @cjdsellers comments to your agent to see what it comes up with ?

dfjmax:
@faysou Not really, I like to think it through myself before I outsource tasks to the agent.

@cjdsellers I went ahead and addressed the gaps. Grouped the margin updates by owning account in initialize_positions, initialize_orders and the pending-tick path, and added a two-account test. Gaps 2 (first fill) and 3 (flat equity) are in too.

Would you please take another look when you get a chance?

cjdsellers:
Thanks, I took another look. Grouping by `account_id` is the right approach, and I think it belongs in this PR rather than a follow-up. Once `initialize_positions`, `initialize_orders`, and pending-tick recovery resolve broker-routed accounts, they must keep each account's orders and positions isolated. Otherwise the fix can apply one account's margin calculation to another account's state.

Passing `Some(account_id)` on the `OrderFilled` path while leaving the `PositionEvent` path as `None` also matches the current cache semantics. The fill arrives before its position is cached, while the position event recomputes the instrument-level aggregate after the cache update. I would not introduce a per-account PnL cache here because that expands into cache keys, invalidation, snapshots, and query semantics, and belongs in the later accounting refactor.

I also think this is reasonable incremental v2 work despite that planned refactor. This is the Rust v2 Portfolio path and the bug reproduces on `2.0.0rc1`, so closing the account-routing gap now gives us correctness tests that the refactor must preserve.

There is still one blocker. In `accumulate_mark_values`, `valuation_account` falls back to the position owner, but `equity_account_id` still only uses the explicit account or `cache.account_id(&venue)`. For a multi-currency cash account registered under `IB` with an `IBIS` instrument, that lookup is `None`, so the position mark is added on top of the base asset already held in the account balance. Please derive both from the same resolved account and add a broker-routed cash regression based on `test_equity_multi_currency_cash_fill_counts_credited_asset_once`.

The first-fill regression also calls the topic path rather than `portfolio_update_order`, and the multi-account test covers only `initialize_positions`. Please add endpoint, open-order-only, and pending-tick coverage for the paths changed here would help us maintain a high level of test coverage.

Thanks again for sticking with the review as well.

dfjmax:
@cjdsellers both blockers should be sorted. equity_account_id now comes from the position owner in accumulate_mark_values (added a broker-routed cash regression), and I routed the first-fill test through portfolio_update_order plus added open-order and pending-tick multi-account coverage.

cjdsellers (approved):
@dfjmax, I'll merge this now. The original broker-routed PnL issue and the first-fill, flat-equity, multi-account, and cash-account gaps are covered, while a remaining issue is limited to pending startup recovery when the same margin account has both an open order and position and its conversion rate becomes available later.

I'll follow up separately on `update_instrument_id`: `update_orders` and `update_positions` currently update separate account clones, so the second write can preserve maintenance margin while discarding newly recovered initial margin. That follow-up will apply both calculations to one working account and add a regression proving both margin components survive recovery; keeping it separate avoids expanding this PR again for a narrow edge case.

Thanks once again for the PR/report and for sticking with the review!

dfjmax:
Hi @cjdsellers , I've noticed that this one was not added to the RELEASES.md. Will you take care of it?

cjdsellers:
Hi @dfjmax, yes - there are typically periodic sweeps of this when dependencies are updated.

PR #4462, opened 2026-07-13
Propagate Hyperliquid l2Book fast subscriptions

### LINKED ISSUE #4460: [Hyperliquid] l2Book `fast` subscription parameter is dropped across Python/Rust WebSocket lifecycle
## Summary

Hyperliquid supports an `l2Book` fast mode (`{"type":"l2Book","coin":"<coin>","fast":true}`), but Nautilus Trader's current Hyperliquid adapter cannot propagate `fast: true` from a strategy subscription to the WebSocket payload.

The parameter is silently ignored in both the Python/PyO3 and native Rust data-client paths. Consequently the adapter always opens the default l2Book stream shape, even when strategy code supplies `params={"fast": True}`.

This is distinct from closed #4298: that issue addressed stale-stream observability/recovery. This issue is a missing public adapter/protocol field whose omission prevents strategies from selecting the venue's faster, shallower l2Book mode.

## Verified against

`origin/develop` at `6c97d5b00b95a50de164a62b32dd5046151ff53d`.

### Python / PyO3 path

- `nautilus_trader/adapters/hyperliquid/data.py` calls `self._ws_client.subscribe_book(pyo3_instrument_id)` from `_subscribe_order_book_deltas`, without reading `command.params`.
- `crates/adapters/hyperliquid/src/python/websocket.rs` exposes `subscribe_book(instrument_id)` with no `fast` parameter.
- The generated `python/nautilus_trader/adapters/hyperliquid/__init__.pyi` exposes the same one-argument signature.

### Native Rust data-client path

- `crates/adapters/hyperliquid/src/data.rs` reads book subscription params, but `parse_book_precision_params` only accepts `n_sig_figs` and `mantissa`; `fast` is ignored.

### WebSocket lifecycle

- `SubscriptionRequest::L2Book` in `crates/adapters/hyperliquid/src/websocket/messages.rs` has only `coin`, `nSigFigs`, and `mantissa`, so the serialized wire request cannot contain `fast`.
- `BookStreamOptions` stores only those precision fields. Therefore reconnect replay, targeted stale-stream resubscribe, and final unsubscribe also cannot preserve `fast`.

The official WebSocket docs require an unsubscribe subscription object to match the original subscribe object, so any implementation must replay the field on unsubscribe as well as subscribe/reconnect:
https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/websocket/subscriptions

## Venue behavior

The current official GitBook l2Book page omits `fast` from its optional-parameter list, but independent current SDKs expose it:

- Go: `L2BookSubscriptionParams.Fast bool`
  https://pkg.go.dev/github.com/sonirico/go-hyperliquid#L2BookSubscriptionParams
- JS: an l2Book event includes `fast: true` only when subscribed with `fast: true`
  https://jsr.io/@nktkas/hyperliquid/doc/all_symbols

A June 2026 summary of the Hyperliquid public-WebSocket change describes the practical tradeoff as 5 levels every 0.5 seconds with `fast: true`, versus a deeper but slower standard l2Book stream:
https://www.quicknode.com/blog/hyperliquid-foundation-websocket-changes

Raw A/B capture of the outbound subscription and received snapshot cadence will be attached separately.

## Expected behavior

A strategy subscription such as:

```python
self.subscribe_order_book_deltas(
    instrument_id,
    params={"fast": True},
)
```

should cause the adapter to send `fast: true` in the l2Book subscription and preserve that effective wire shape through reconnect, recovery resubscribe, and unsubscribe.

## Proposed scope

1. Parse and validate `fast: bool` from book subscription params in both Python and native Rust paths.
2. Add `fast` to `SubscriptionRequest::L2Book` and the persisted book stream options.
3. Expose the option through PyO3 and regenerate the Python stubs.
4. Preserve the effective field on reconnect, targeted resubscribe, and unsubscribe.
5. Treat `fast: true` as widened runtime intent for the shared per-coin l2Book stream. A later fast subscriber must not be silently retained on a pre-existing normal stream solely due to the current first-wins option policy.

## Regression coverage

- serialization contains `"fast":true`;
- strategy params reach the PyO3/WebSocket request;
- native Rust `SubscribeBookDeltas`/`SubscribeBookDepth10` params reach the request;
- reconnect, stale-stream resubscribe, and final unsubscribe echo `fast: true`;
- normal-then-fast shared-stream subscription upgrades to fast instead of swallowing widened intent;
- generated PyO3 stubs match the new public signature.

graceyangfan:
## Summary

Propagates Hyperliquid `l2Book` `fast` subscriptions through the full Nautilus adapter lifecycle.

- Parses `fast` from Rust and Python book subscription parameters.
- Exposes the optional flag through PyO3 and generated Python stubs.
- Serializes `fast: true` on the l2Book wire request while keeping omitted/`false` behavior unchanged.
- Persists the effective stream mode for reconnect, targeted stale-stream resubscribe, and final unsubscribe.
- Upgrades a shared normal l2Book stream when a later logical subscriber requires `fast: true`.

Fixes #4460.

## Root cause

The prior adapter only carried `nSigFigs` and `mantissa` through its request model and shared stream registry. Python ignored strategy `params`, the native Rust parser ignored `fast`, and the WebSocket request type could not serialize it.

## Validation

- `cargo test -p nautilus-hyperliquid --test websocket` (40 passed)
- `cargo test -p nautilus-hyperliquid --test data_client -- --test-threads=1` (33 passed)
- `uv run --active --no-sync pytest tests/integration_tests/adapters/hyperliquid/test_data.py -k 'subscribe_order_book_deltas or subscribe_order_book_depth' -q` (4 passed)
- `cargo clippy -p nautilus-hyperliquid --tests -- -D warnings`
- `cargo doc -p nautilus-hyperliquid --no-deps`
- `cargo fmt`, official stub generation, and pre-commit on touched files

Local-only real-mainnet adapter E2E verification (not included in this PR) received 12 decoded l2Book updates in each mode from both Rust and Python paths:

- `fast:false`: approximately 5.4 s median interval, about 39–41 deltas per message.
- `fast:true`: approximately 0.5 s median interval, about 10–11 deltas per message.


cjdsellers:
Hi @graceyangfan,

Thanks for the thorough patch, and for the issue write-up that went with it.

My concern is the premise rather than the mechanics. The PR treats `fast: true` as widening the stream, but it is a depth-for-latency trade-off. Your own E2E numbers show it: about 39-41 deltas per message on `fast:false` against about 10-11 on `fast:true`, so roughly 20 levels per side down to 5. The issue quotes the same thing ("5 levels every 0.5 seconds ... versus a deeper but slower standard l2Book stream").

That matters because one `l2Book` stream per coin is shared. If strategy A subscribes with no params and strategy B later subscribes with `params={"fast": True}`, `register` flips the stream to fast and strategy A's book silently drops to 5 levels. Nothing is logged: `options_mismatch` excludes `fast` by construction, and the `reconfigure_from` branch has no log of its own. A normal subscriber joining an existing fast stream is silent in the same way. This also cuts against the registry's own convention, where precision options are first-wins and warn on mismatch. At minimum the upgrade and the join-a-fast-stream case should warn. I'd also consider whether first-wins is simply the right policy for `fast` too, for consistency.

Second, the upgrade unsubscribes the working stream before the replacement is acked, with no rollback. If the venue rejects `fast: true`, the old stream is gone, the reject is not retried, and a later `subscribe_book` won't re-send because the registry already reports the stream as fast. Stale-stream recovery then replays `fast: true` and gets rejected again, so both subscribers lose the coin until reconnect. Since `fast` isn't in the official GitBook l2Book parameter list and your live run covered mainnet perps only, I'm not comfortable assuming every asset class accepts it. Please add coverage for a rejected upgrade.

Two smaller things. `fast: true` with depth10 can never deliver 10 real levels: `parse_ws_order_book_depth10` zero-pads 6-10, so consumers get 5 real levels and 5 zero-price ones they can't distinguish. And `docs/integrations/hyperliquid.md` needs updating: the "Order book precision controls" section still says requesting different options logs a warning and keeps the active options, which this PR makes false for `fast`, and the depth trade-off isn't documented anywhere.

Also worth a look: `fast: bool` has `skip_serializing_if` but no `#[serde(default)]`, so `SubscriptionRequest` no longer round-trips through serde when `fast` is false. Nothing deserializes it today, but it's a public type.

Happy to take another look once the shared-stream behavior is settled.

graceyangfan:
Thanks for the detailed review. After investigating the end-to-end path, we confirmed that normal and `fast: true` l2Book subscriptions are distinct physical streams, but Nautilus currently routes `OrderBookDeltas` only by instrument. Correct support for concurrent profiles, so different strategies receive only their requested mode, requires profile-aware DataEngine bookkeeping, data-event provenance, message-bus topics, and managed-book/cache ownership. That is a broader core design change and is not appropriate to partially implement in this adapter-focused PR. We will close this PR rather than merge behavior that could silently mix profiles for consumers. Thank you again for calling this out.

PR #4465, opened 2026-07-14
Fix legacy Equity round trips dropping quantity constraints

### LINKED ISSUE #4461: Equity.from_dict and legacy ParquetDataCatalog drop min_quantity and max_quantity
### Bug description

`Equity.min_quantity` and `Equity.max_quantity` are not preserved when an `Equity` is deserialized through `Equity.from_dict` or round-tripped through the legacy Python `ParquetDataCatalog` instrument serialization path.

The values are present on the original object and included in the output of `Equity.to_dict`, but both fields become `None` after deserialization.

This can silently remove execution-relevant instrument constraints.

### Environment

* NautilusTrader: `1.230.0`
* Python: `3.12.3`
* OS: Ubuntu 24.04 under WSL2
* Architecture: x86_64
* API path: legacy Python/Cython `Equity` and `ParquetDataCatalog`

### Minimal reproduction

```python
from nautilus_trader.model.currencies import USD
from nautilus_trader.model.identifiers import InstrumentId
from nautilus_trader.model.identifiers import Symbol
from nautilus_trader.model.instruments import Equity
from nautilus_trader.model.objects import Price
from nautilus_trader.model.objects import Quantity


equity = Equity(
    instrument_id=InstrumentId.from_str("TEST.XNAS"),
    raw_symbol=Symbol("TEST"),
    currency=USD,
    price_precision=2,
    price_increment=Price.from_str("0.01"),
    lot_size=Quantity.from_int(40),
    min_quantity=Quantity.from_int(1),
    max_quantity=Quantity.from_int(1000),
    ts_event=0,
    ts_init=0,
)

values = Equity.to_dict(equity)
restored = Equity.from_dict(values)

print("dict min_quantity:", values["min_quantity"])
print("dict max_quantity:", values["max_quantity"])
print("restored min_quantity:", restored.min_quantity)
print("restored max_quantity:", restored.max_quantity)
print("operator equality:", equity == restored)
```

### Actual result

```text
dict min_quantity: 1
dict max_quantity: 1000
restored min_quantity: None
restored max_quantity: None
operator equality: True
```

`Equity.__eq__` returning `True` makes the loss difficult to detect without strict field-by-field comparison.

The same fields are also lost when the instrument is written and read through the legacy Python `ParquetDataCatalog` path.

### Expected result

The following fields should survive both round-trips unchanged:

```text
min_quantity = 1
max_quantity = 1000
```

Operator equality should also not conceal the loss of execution-relevant instrument fields.

### Suspected causes

In `nautilus_trader/model/instruments/equity.pyx`:

* `Equity.to_dict_c` includes `min_quantity` and `max_quantity`;
* `Equity.from_dict_c` does not pass either field to the `Equity` constructor.

In the legacy Python Arrow schema in:

`nautilus_trader/serialization/arrow/implementations/instruments.py`

the `Equity` schema does not contain `min_quantity` or `max_quantity`, so these values are removed before reconstruction through `Equity.from_dict`.

The newer Rust Arrow `Equity` schema and decoder appear to contain both fields, so the catalog issue may be limited to the legacy Python/Cython serialization path.

### Suggested fix

1. Pass the optional quantity fields in `Equity.from_dict_c`:

```python
max_quantity=(
    Quantity.from_str(values["max_quantity"])
    if values.get("max_quantity") is not None
    else None
),
min_quantity=(
    Quantity.from_str(values["min_quantity"])
    if values.get("min_quantity") is not None
    else None
),
```

2. Add `max_quantity` and `min_quantity` to the legacy Python Arrow schema for `Equity`.

3. Add regression tests covering:

* `Equity.to_dict` to `Equity.from_dict`;
* legacy `ParquetDataCatalog` instrument round-trip;
* strict field-by-field comparison;
* both fields set individually and together.

4. Consider whether `Equity.__eq__` should include these execution-relevant fields, or document that it does not provide strict field equality.

### Additional validation

In local tests:

* `maker_fee`, `taker_fee`, `margin_init`, `margin_maint`, `isin`, `ts_event`, and `ts_init` were preserved;
* `min_quantity` and `max_quantity` were lost;
* bar data stored in the same temporary catalog remained unchanged;
* no monkey patching or custom deserialization was used.

chang-pro:

## Summary

`Equity.to_dict` serializes `max_quantity` and `min_quantity`, but they were not restored on the way back: `from_dict` never read the keys, and the legacy `ParquetDataCatalog` Arrow schema for `Equity` did not declare either field. Both paths therefore silently deserialized these execution-relevant constraints as `None`. This restores them through direct dicts and the catalog.

## Related Issues/PRs

Closes #4461.

## Type of change

- [x] Bug fix (non-breaking)

None. Older dicts/catalogs without these keys still deserialize to `None`, and the read-side schema is applied only to the built-in `Equity` class so other instruments (including legacy Betting catalogs with historical timestamp types) and `Equity` subclasses keep their existing inferred-schema path.

## Documentation

## Release notes

## Testing

Three parts, each covered by tests:
1. `Equity.from_dict` now reads `max_quantity`/`min_quantity` back (prebind idiom matching `cfd.pyx`).
2. The `Equity` Arrow schema declares both fields so the catalog persists them.
3. `_query_pyarrow` applies the registered `Equity` schema when reading, so fragments written before the fields existed don't mask them on newer fragments (pyarrow otherwise infers the schema from the first fragment only).

Added `test_legacy_equity_dict_round_trip_preserves_quantity_constraints` (round trip incl. `None` and absent-key cases), `test_catalog_equity_roundtrip_preserves_quantity_constraints`, and `test_catalog_equity_mixed_schema_fragments` (old fragment then new fragment). Full model, persistence, and serialization suites pass locally (2,886 passed).

Note: `Equity.to_dict` also emits `min_price`/`max_price`, but the legacy `Equity` constructor does not accept them; restoring those is a separate change and left out of scope here.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -76,6 +76,7 @@ releases as feedback arrives, before the final `2.0.0` release.
 - Fixed v2 multi-currency cash equity double-counting assets already credited to account balances
 - Fixed v2 quanto position notionals using quote currency instead of settlement currency
 - Fixed v2 portfolio valuations labeling and converting cost-currency amounts as settlement currency
+- Fixed legacy `Equity` round trips dropping `max_quantity` and `min_quantity` through `from_dict` and the `ParquetDataCatalog` (#4461)
 - Fixed v2 composite bar aggregation (`@` source) to deliver aggregated bars to subscribed actors and strategies
 - Fixed v2 tick, tick-imbalance, and tick-runs aggregators to emit bars with the standard bar type for composite subscriptions, matching all other aggregators
 - Fixed v2 volume-runs and value-runs aggregators dropping leftover volume when a trade spanned a bar boundary and the next trade continued the same side
--- a/nautilus_trader/model/instruments/equity.pyx
+++ b/nautilus_trader/model/instruments/equity.pyx
@@ -138,6 +138,8 @@ cdef class Equity(Instrument):
     @staticmethod
     cdef Equity from_dict_c(dict values):
         Condition.not_none(values, "values")
+        cdef str max_q = values.get("max_quantity")
+        cdef str min_q = values.get("min_quantity")
         return Equity(
             instrument_id=InstrumentId.from_str_c(values["id"]),
             raw_symbol=Symbol(values["raw_symbol"]),
@@ -145,6 +147,8 @@ cdef class Equity(Instrument):
             price_precision=values["price_precision"],
             price_increment=Price.from_str(values["price_increment"]),
             lot_size=Quantity.from_str(values["lot_size"]),
+            max_quantity=Quantity.from_str_c(max_q) if max_q is not None else None,
+            min_quantity=Quantity.from_str_c(min_q) if min_q is not None else None,
             isin=values.get("isin"),  # Can be None,
             margin_init=Decimal(values.get("margin_init", 0)) if values.get("margin_init") is not None else None,
             margin_maint=Decimal(values.get("margin_maint", 0)) if values.get("margin_maint") is not None else None,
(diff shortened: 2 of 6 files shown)


cjdsellers (approved):
Thanks for the PR @chang-pro. Whilst we're winding down focus on legacy v1, this fix is tightly scoped and clearly correct.

PR #4466, opened 2026-07-14
Fix bracket tp_post_only docstring default to True

### LINKED ISSUE #4437: OrderFactory.bracket(): tp_post_only code default is True but docstring says False (silently breaks IB brackets)
## Summary

`OrderFactory.bracket()` documents its `tp_post_only` parameter as **`default False`**, but the actual code default is **`True`**. The two disagree. Because the Interactive Brokers execution adapter categorically rejects post-only orders, a bracket built for IB *without explicitly passing `tp_post_only=False`* gets a silently post-only take-profit child that the adapter rejects — and the rejection cascades to the whole OTO list, so **none of the three legs ever reach the venue**.

This is a quiet footgun: the caller reads the docstring ("default False"), trusts it, omits the argument, and their IB bracket silently never transmits.

## The mismatch

In `nautilus_trader/common/factories.pyx`, `OrderFactory.bracket(...)`:

- **Signature** (code default is `True`):
  ```cython
  bint tp_post_only = True,
  ```
- **Docstring** (claims `False`):
  ```
  tp_post_only : bool, default False
      If the take-profit order will only provide liquidity (make a market).
  ```

The TP `LimitOrder` is then constructed with `post_only=tp_post_only`, so an omitted argument yields a post-only take-profit.

(For contrast, `entry_post_only` genuinely defaults to `False` in both code and docstring — only `tp_post_only` is inconsistent.)

## Interaction with the Interactive Brokers adapter

`InteractiveBrokersExecutionClient._transform_order_to_ib_order` (in `nautilus_trader/adapters/interactive_brokers/execution.py`) rejects post-only outright:

```python
if order.is_post_only:
    raise ValueError("`post_only` not supported by Interactive Brokers")
```

`_submit_order_list` catches that `ValueError` and rejects **every** order in the list ("declined to prevent unintended side effects"), returning before `place_order` is ever called. Net effect for an IB user who builds a bracket and omits `tp_post_only`:

- take-profit leg → `OrderRejected("`post_only` not supported by Interactive Brokers")`
- entry + stop legs → `OrderRejected("The order has been rejected due to the rejection of the order with <ClientOrderId> in the list")`
- nothing is transmitted to IB Gateway.

## Reproduction

```python
# order_factory is a nautilus OrderFactory
bracket = order_factory.bracket(
    instrument_id=instrument_id,
    order_side=OrderSide.BUY,
    quantity=qty,
    sl_trigger_price=sl_price,
    tp_price=tp_price,
    # tp_post_only omitted → inherits the code default True, contrary to the docstring
)
tp = next(o for o in bracket.orders if o.order_type == OrderType.LIMIT)
assert tp.is_post_only is True   # passes today; the docstring implies it should be False
```

Submitting that list through the IB exec client rejects all three legs on the `post_only` check.

## Suggested fix (either is fine; they should just agree)

1. **Make the code match the documented behavior** — change the default to `bint tp_post_only = False`. This is also the safer default given the IB adapter's categorical rejection, and matches `entry_post_only`'s default. (Behavior change: existing callers relying on the current `True` default would need to pass `tp_post_only=True` explicitly.)
2. **Or** correct the docstring to `default True` if the `True` default is intentional.

Option 1 seems preferable: a post-only take-profit that silently blocks IB submission is a surprising default, and the docstring already advertises `False`.

## Optionally: make the IB rejection less silent

Separately, when a single leg fails transform (e.g. the `post_only` check), `_submit_order_list` rejects the whole `OrderList` leg-by-leg but there is no single "the whole bracket was declined" signal — it can be non-obvious which leg was the root cause. Not part of this issue, but worth noting alongside.

## Environment

- `nautilus-trader==1.230.0` (line numbers above are from this release; the mismatch is present in the pinned source).
- Adapter: `nautilus-ibapi==10.45.1`.

Happy to open a PR for whichever direction the maintainers prefer.

chang-pro:

## Summary

`OrderFactory.bracket(...)` documents `tp_post_only` as `default False`, but the code defaults to `True`, as do the v2 Rust/PyO3 factory and its tests. `True` has been the intended value since the option was introduced (it preserved the previously hard-coded post-only take-profit), so the docstring was incorrect. This corrects the docstring to match the actual default; runtime behavior is unchanged.

## Related Issues/PRs

Closes #4437.

## Type of change

- [x] Documentation update
- [x] Bug fix (non-breaking)

None. This changes only docstring text and adds a test; order construction is unchanged.

## Documentation

## Release notes

## Testing

Added `test_bracket_default_post_only`, which omits `tp_post_only` and asserts the default limit take-profit is post-only (`True`) while the entry is not, matching the v2 Rust test `test_bracket_order_default_post_only`. It deterministically fails if the code default is changed to `False`. Full `test_orders.py` suite passes locally (98 passed).

Note for users hitting this on venues that reject post-only orders (e.g. Interactive Brokers): pass `tp_post_only=False` explicitly to make the take-profit a normal (taker-eligible) limit order.


cjdsellers:
Thanks for the fix @chang-pro 👌.

I'll resolve the new conflict with the release notes and merge shortly.

cjdsellers (dismissed):

cjdsellers (approved):

PR #4468, opened 2026-07-14
Add Hyperliquid persistable public trade data

### LINKED ISSUE #4467: feat(hyperliquid): expose persistable public trade custom data
## Summary

Hyperliquid public `trades` messages provide `users: [buyer, seller]` and `hash`, but the adapter currently deserializes those fields and discards them while producing `TradeTick`. The `recentTrades` response exposes the same information, while its adapter model currently retains only the generic trade fields.

- Raw WebSocket fields: https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/adapters/hyperliquid/src/websocket/messages.rs#L499
- Current `TradeTick` conversion: https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/adapters/hyperliquid/src/websocket/parse.rs#L75
- Generic `TradeTick` model: https://github.com/nautechsystems/nautilus_trader/blob/develop/crates/model/src/data/trade.rs#L44
- Hyperliquid `WsTrade` schema: https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/websocket/subscriptions

## Proposed enhancement

Expose an opt-in, self-contained, adapter-specific `HyperliquidPublicTrade` custom data type for both the public WebSocket `trades` feed and REST `recentTrades`.

```text
HyperliquidPublicTrade
  instrument_id
  price
  size
  aggressor_side
  trade_id
  buyer
  seller
  hash
  ts_event
  ts_init
```

The custom event should be sufficient for a recorder, catalog reader, or research workflow by itself. It should not require loading and joining a separate `TradeTick` dataset to recover price, size, side, or participant information. `(instrument_id, trade_id, ts_event)` remains available as an optional correlation key when both data families are consumed.

## Persistence and catalog contract

This must be Arrow-enabled custom data, not live-only `no_arrow` data. Each event should use the canonical instrument ID as its `DataType` identifier, so it records per instrument under the custom-data catalog layout and decodes back to the same concrete type with metadata and identifier intact.

A recorder should be able to subscribe to and persist this one type without recording or joining `TradeTick`.

## Subscription and request behavior

- `SubscribeCustomData(HyperliquidPublicTrade)` is opt-in.
- It shares the existing venue `trades` WebSocket subscription with `SubscribeTrades`; it must not create duplicate wire subscriptions.
- A trade-tick subscription emits only `TradeTick`; a public-trade subscription emits only `HyperliquidPublicTrade`; when both are active, both are emitted.
- Unsubscribing one logical use must keep the wire stream alive while the other remains active. Reconnect and reset must restore the correct uses.
- `RequestCustomData(HyperliquidPublicTrade)` should use `recentTrades` and return custom-data wrappers, preserving the existing `RequestTrades -> TradeTick` contract.

## Non-goals

- Do not add wallet fields or transaction hash to the generic `TradeTick` model, ABI, or Arrow schema.
- Do not require all venues or all `TradeTick` users to pay address-string storage costs.

## Acceptance criteria

- WebSocket and `recentTrades` preserve all fields above.
- `Data::Custom(HyperliquidPublicTrade)` round-trips through recorder/catalog write and query, restoring the concrete type and its `DataType` metadata/identifier.
- The custom type is usable through Rust, PyO3, Python wrapper exports, generated stubs, and catalog query.
- Shared-stream lifecycle is covered for first subscribe, partial unsubscribe, final unsubscribe, reconnect, and reset.

This enables public order-flow research, counterparty concentration analysis, maker/taker wallet studies, and market research without expanding the generic market-data schema.

graceyangfan:
## Summary

Implements #4467 by adding opt-in, persistable `HyperliquidPublicTrade` custom data. It retains Hyperliquid-specific buyer, seller, and transaction hash information without changing the generic `TradeTick` model or schema.

Closes #4467

### 1. One venue stream, two logical consumers

`SubscribeCustomData(HyperliquidPublicTrade)` and `SubscribeTrades` share the same Hyperliquid `trades` WebSocket subscription per coin. The new `TradeStreamRegistry` tracks each logical use, so:

- the first use sends one wire subscribe; adding the other does not
- each use receives only its own output type; both receive data when both are active
- a partial unsubscribe keeps the wire stream alive; the final unsubscribe closes it
- reconnect replays the active uses, and reset swaps generation-owned runtime state

Review entry points: `src/websocket/trades.rs`, `src/websocket/client.rs`, and `src/websocket/handler.rs`.

### 2. CustomData surface

`HyperliquidPublicTrade` is a self-contained custom-data event containing instrument ID, price, size, aggressor side, trade ID, buyer, seller, hash, and timestamps. The WebSocket handler emits it as `CustomData` with instrument metadata; the existing `TradeTick` path is unchanged.

Review entry points: `src/data_types.rs`, `src/websocket/parse.rs`, and `src/websocket/handler.rs`.

### 3. Historical request path

`RequestCustomData(HyperliquidPublicTrade)` calls `recentTrades` through the shared Rust HTTP client, applies the requested time window and limit, then returns `CustomDataResponse`. Rust and PyO3 therefore share parsing, ordering, filtering, and the indexer-unavailable behavior.

Review entry points: `src/http/client.rs`, `src/http/parse.rs`, `src/data.rs`, `src/python/http.rs`, and `nautilus_trader/adapters/hyperliquid/data.py`.

### 4. Python and persistence parity

The type is exported through Rust, PyO3, Python wrappers, and both generated/legacy stubs. It has Arrow serialization **and decoding**, so a recorder can persist only this custom type and a catalog query restores the concrete `HyperliquidPublicTrade` inside `CustomData`.

Review entry points: `src/python/arrow.rs`, `nautilus_trader/adapters/hyperliquid/__init__.py`, and `tests/integration_tests/adapters/hyperliquid/test_persistence.py`.

## Scope

- No generic market-data ABI/schema change.
- No second WebSocket channel for the same Hyperliquid trade feed.
- No TradingNode or end-to-end tests were added; coverage stays at adapter, binding, and local persistence layers.

## Validation

- `cargo test -p nautilus-hyperliquid logical_uses_share_one_venue_subscription -- --nocapture`
- `cargo test -p nautilus-hyperliquid test_subscribe_public_trades_emits_complete_custom_data -- --nocapture`
- `cargo test -p nautilus-hyperliquid test_data_client_request_public_trades -- --nocapture`
- `pytest tests/unit_tests/core/test_pyo3_stubs.py tests/integration_tests/adapters/hyperliquid/test_data.py tests/integration_tests/adapters/hyperliquid/test_persistence.py -q` — 639 passed
- `pre-commit run --files …`
- `cargo clippy -p nautilus-hyperliquid --all-targets --features arrow,python -- -D warnings`

Additionally, the PyO3 request path was verified against the live Hyperliquid exchange using BTC with `limit=3` (three ordered, non-empty public trades returned).


cjdsellers:
Hi @graceyangfan,

Thanks for the thorough patch. I found one v2 request-path gap before merge.

`request_data` reuses the caller's `DataType` when wrapping `recentTrades`. The documented call supplies the instrument ID only as metadata, so the returned `CustomDataResponse` and its items retain `identifier=None` instead of the canonical instrument ID required by #4467. The current test hides this by constructing its request `DataType` with the identifier already populated.

Construct the canonical response `DataType` inside the adapter and change the test to start with metadata only, then assert the response and item identifiers. The shared WebSocket path otherwise passed a read-only mainnet smoke.

graceyangfan:
Implemented in 33206be2fe after merging current `develop`.

The request path now constructs a canonical response `DataType` from the resolved instrument ID, including both `metadata["instrument_id"]` and `identifier`. The regression test begins with a metadata-only request type (`identifier is None`) and asserts the response and each `CustomData` item carry the canonical identifier.

Validated with `pytest tests/integration_tests/adapters/hyperliquid/test_data.py -q` (32 passed) and targeted pre-commit.

cjdsellers:
Thanks @graceyangfan, `33206be2` addresses the Python adapter path. The equivalent Rust `DataClient::request_data` path still needs to construct the canonical response `DataType`, but I’ll handle that
  as a focused post-merge follow-up rather than hold this PR. This is good to merge from my side.

cjdsellers (approved):

PR #4474, opened 2026-07-16
Fix Binance Spot SBE initialization timestamps

### LINKED ISSUE #4473: Binance Spot SBE stream data should preserve adapter init timestamps
### Bug description

The Binance Spot SBE stream parsers currently copy the provider event timestamp into both `ts_event` and `ts_init` for trades, quotes, depth snapshots, and depth diffs.

This collapses two different clock domains. The provider timestamp should remain `ts_event`, while `ts_init` should represent the adapter's local receive/initialization time, matching the adapter's JSON stream path.

### Expected behavior

- Capture the adapter clock once per decoded SBE message before parser dispatch.
- Preserve the provider timestamp as `ts_event`.
- Use the captured adapter timestamp as `ts_init` for every emitted trade, quote, snapshot delta, diff delta, and aggregate wrapper.
- Cover the behavior with deliberately unequal provider and adapter timestamps.

### Current source evidence

On current `develop`, the four parsers in `crates/adapters/binance/src/spot/websocket/streams/parse.rs` pass the provider timestamp into both timestamp fields, and the SBE handler in `crates/adapters/binance/src/spot/data.rs` does not capture the clock before dispatch.

seungpyoson:
## Summary

- capture one adapter initialization timestamp for each decoded Binance Spot SBE message
- preserve Binance timestamps as `ts_event` while using the adapter timestamp as `ts_init`
- propagate the same initialization timestamp through trades, quotes, snapshots, inner deltas, and aggregate wrappers
- add deliberately unequal timestamp coverage at parser and production-handler boundaries

Closes #4473.

## Rationale

The SBE path currently copies the provider timestamp into both timestamp fields. The JSON stream path already captures the adapter clock before parser dispatch. This change aligns the SBE path with that ownership boundary without changing provider event-time semantics.

The four public SBE parser functions gain a required trailing `ts_init: UnixNanos` argument. All in-repository callers are updated.

## Verification

- `cargo test -p nautilus-binance --lib` — 742 passed
- `cargo clippy -p nautilus-binance --lib --tests -- -D warnings`
- `cargo +nightly fmt --all -- --check`
- `git diff --check`

No live Binance connection test was run.

### DIFF
--- a/crates/adapters/binance/src/spot/data.rs
+++ b/crates/adapters/binance/src/spot/data.rs
@@ -338,12 +338,14 @@ impl BinanceSpotDataClient {
         http_client: &BinanceSpotHttpClient,
         clock: &'static AtomicTime,
     ) {
+        let ts_init = clock.get_time_ns();
+
         match msg {
             BinanceSpotWsMessage::Trades(ref event) => {
                 let symbol = Ustr::from(&event.symbol);
                 let cache = ws_instruments.load();
                 if let Some(instrument) = cache.get(&symbol) {
-                    let trades = parse_trades_event(event, instrument);
+                    let trades = parse_trades_event(event, instrument, ts_init);
                     for data in trades {
                         Self::send_data(data_sender, data);
                     }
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):

cjdsellers:
Thanks for the fix @seungpyoson 🙏

PR #4477, opened 2026-07-16
Fix Clock trait consistency for expired timers and failed alerts

folknor:
Two follow-ups to the timer-introspection alignment in #4400, both making the `Clock` trait surfaces behave uniformly across `TestClock` and `LiveClock`.

**1. `next_time_ns` reported expired timers**

#4400 aligned `timer_exists` with `timer_names`/`timer_count` by filtering on `!is_expired()`, but `next_time_ns` was left as a bare map read. `LiveClock` purges expired entries only lazily (on the next `set_*`/cancel), so there is an unbounded window where a fired-and-done timer answers `timer_exists == false` and `timer_count == 0` while `next_time_ns` still returns its stale scheduled time. `TestClock` removes expired timers eagerly and returns `None` for the same state, so the two impls also disagreed with each other. Both impls now filter on `!is_expired()`, and the trait / `ClockApi` docs say "active (not expired)".

Consumers are unaffected or corrected: the backtest engine's `process_next_timer` gates on `Some`, which is exactly the active-timer semantics it wants.

**2. `set_time_alert_ns` destroyed the existing timer on its error path**

`set_time_alert_ns` called `replace_existing_timer_if_needed` *before* the "No callbacks provided" predicate check, while `set_timer_ns` checks first. A failing call (e.g. re-arming an alert with `callback: None` after callbacks were released during a partial component teardown) therefore returned an error *and* silently cancelled the previously scheduled alert under that name, so the old deadline never fired. Both impls now check the predicate before touching the existing timer, matching `set_timer_ns`. The predicate reads only the callback registry, which the replace path never mutates, so the reordering changes nothing on the success path.

**Testing**

- Extended `test_live_clock_timer_exists_consistent_after_expiry` (the #4400 regression) to assert `next_time_ns` is `None` while the expired entry still physically sits in the map.
- New `test_failed_set_time_alert_ns_preserves_existing_timer` on `TestClock` and its `LiveClock` twin - both deterministic, no wall-clock waits (the live test uses a far-future alert that never fires).
- All three assertions were verified to fail against the unfixed implementations before the fix was applied.

### DIFF
--- a/crates/common/src/clock.rs
+++ b/crates/common/src/clock.rs
@@ -225,7 +225,7 @@ pub trait Clock: Debug + Any {
 
     /// Returns the time interval in which the timer `name` is triggered.
     ///
-    /// If the timer doesn't exist `None` is returned.
+    /// If no active (not expired) timer with `name` exists, `None` is returned.
     fn next_time_ns(&self, name: &str) -> Option<UnixNanos>;
 
     /// Cancels the timer with `name`.
@@ -621,7 +621,7 @@ impl<'a> ClockApi<'a> {
         }
     }
 
-    /// Returns the next trigger timestamp for the timer `name`.
+    /// Returns the next trigger timestamp for the active (not expired) timer `name`.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Thank you @folknor, good catch 👍

PR #4479, opened 2026-07-16
Fix missing-order resolution and failed-report handling in live reconciliation

folknor:
Fixes to the continuous open-order check in `ExecutionManager`, aligning it with the Python reference engine's `_resolve_order_not_found_at_venue` and hardening the failed-report guard to client identity per review. Continues the reconciliation-correctness work from #4366/#4376/#4387.

**1. Missing-order resolution emitted an event the order FSM cannot apply**

`handle_missing_order` resolved every order missing at the venue (after `open_check_missing_retries` cycles) with `OrderRejected`, regardless of status. The order FSM has no `(PartiallyFilled, Rejected)` transition, so for a partially filled resting order whose remainder disappeared from the venue's open-order response the event could never apply: the engine logged an invalid-transition error, the order stayed open in the cache, and `clear_recon_tracking` reset the retry counter - so the node emitted a doomed `Rejected` and an engine error every retry ladder, forever.

Resolution is now status-aware, matching the Python engine: `Accepted`/`Submitted` resolve to `Rejected` as before, `PartiallyFilled` resolves to `Canceled` (preserving filled quantity - a valid FSM transition), `PendingUpdate`/`PendingCancel` defer, closed orders are skipped (expected for `FILLED`), and unexpected statuses are logged and skipped.

The defer branch needs care beyond a direct port: Python's inflight checker rescans cached inflight orders every cycle, while Rust's `check_inflight_orders` walks the `inflight_checks` registration map exclusively - so the broad `clear_recon_tracking` would have permanently removed a deferred order from stuck-order recovery. The retained `InflightCheck` also carries its own `retry_count`/`last_query_at` ladder, and leaving it stale would escalate a primed order straight to a spurious stale-count cancellation on the next inflight pass. The defer branch therefore performs a narrow reset: retry ladders zeroed (both the shared map and the retained `InflightCheck`), query time stamped (so the checker first observes a full threshold delay, then retries from scratch - Python's `_clear_recon_tracking(..., drop_last_query=False)` plus `ts_last_query` stamp), registration and local-activity mark preserved.

**2. Failed bulk report queries converted into "venue has no orders"**

The open-order bulk path logged per-client `generate_order_status_reports` errors and collapsed them into the merged (possibly empty) report vector. With `open_check_open_only: false`, a transient report-endpoint outage lasting `open_check_missing_retries` check cycles marked every recently active cached order missing and walked each one up the retry ladder toward a spurious local rejection - while the venue kept working the orders (the trading channel being healthy is exactly the scenario where only the report endpoint 500s).

Per review, the guard is keyed to client identity rather than raw venue, since `failed_venues` recording `client.venue()` neither covers routing clients (IB's `handles_order_venue` returns true for exchange venues while `venue()` is the broker venue, so a failed IB query would still walk its orders toward local rejection) nor isolates two clients sharing one venue (one failure would suppress reconciliation for the healthy client). The pre-existing venue-keyed guard on the position path is consolidated onto the same mechanism.

At check-preparation time each snapshotted open order / position key is resolved to its responsible client set: orders through the cache's `client_order_id -> ClientId` index first, then an exact `account_id` match, then the set of clients whose `handles_order_venue` claims the instrument venue; position keys by exact `account_id` match, then `handles_order_venue`. The query loops carry queried and failed `ClientId` sets through reconciliation (the node's async report tasks transport only these plain sets - no venue policy remains in `node/mod.rs`). Missing-at-venue resolution runs only when the responsible set is nonempty and every member was queried successfully; otherwise the entry is skipped without advancing retry state. Unresolvable coverage (no claiming client, or a cached client ID no longer registered - the order-client index is persisted, so a client rename in config leaves open orders pointing at the old ID) logs a warning once per order rather than resolving destructively on state that was never obtained.

Consumers unaffected: with all clients healthy every responsible set is queried-and-succeeded and the flow is unchanged; a report that did arrive is always reconciled normally regardless of other clients' failures.

**3. Missing-order retry ladder counted non-consecutive and stale misses**

Two adjacent defects in the same ladder surfaced while testing the rework. First, a positive report did not reset `recon_check_retries` (the Python engine clears per report), so misses separated by successful confirmations still accumulated toward a spurious rejection; the report scan now clears the counter for every reported order, including reports carrying only a `venue_order_id`, which are mapped through the cache index and given full positive-report bookkeeping (Python shares the venue-ID-only gap). Second, `handle_missing_order` incremented the retry counter before checking for a terminal status, so an order that closed while the bulk report request was in flight re-created tracking state that nothing afterwards prunes; the closed check now precedes the increment.

Deliberately out of scope: wiring the dormant single-order verification query machinery (`check_open_order_queries`) into the loop. Python performs that targeted query as a last-chance verification before resolving; the Rust port's divergence there (and its currently dead config knobs `max_single_order_queries_per_cycle` / `single_order_query_delay_ms`) remains a tracked follow-up.

**Testing**

- `test_check_open_orders_partially_filled_missing_at_venue_generates_canceled` - a partially filled order missing at max retries produces `Canceled`, which applies cleanly to the cache with `filled_qty` preserved.
- `test_check_open_orders_failed_routing_client_does_not_resolve_exchange_order` - a failing routing client (broker `venue()`, exchange-venue order claimed via `handles_order_venue`, cached client-ID mapping) does not resolve its order.
- `test_check_open_orders_failed_client_does_not_suppress_healthy_client_same_venue` - two clients on one venue, one failing: only the healthy client's order resolves; the failed client's order keeps its retry budget.
- `test_check_open_orders_failed_routing_client_fallback_coverage_does_not_advance` - an order with no cached client-ID mapping, claimed only through `handles_order_venue`, is protected by the fallback coverage rung.
- `test_position_check_failed_routing_client_leaves_exchange_retry_untouched` / `test_position_check_failed_client_does_not_suppress_healthy_account_same_venue` - position-path twins of the two scenarios above, keyed by account.
- `test_check_open_orders_failed_client_does_not_advance_missing_retries` / `test_position_check_failed_client_query_skips_cached_position` - the original guard tests, rekeyed to client identity, retaining their multi-cycle retry assertions.
- `test_check_open_orders_positive_report_resets_missing_retry_ladder` / `test_check_open_orders_venue_id_only_report_resets_missing_retry_ladder` - miss, positive report (with and without `client_order_id`), miss: the counter returns to zero in between and a single subsequent miss does not resolve.
- `test_check_open_orders_order_closed_during_query_leaves_no_retry_state` - a mock's report callback cancels the order in the cache after the prepare-time snapshot, pinning the prepare/reconcile race: no event, no retry state.
- `test_check_open_orders_deferred_pending_order_keeps_inflight_registration` - primes one inflight retry (`inflight_max_retries: 2`) before the open-check deferral, then asserts the inflight checker issues a fresh query rather than escalating to a stale-count cancellation, proving both the surviving registration and the ladder reset.
- All paused virtual time, no wall-clock waits. Each new assertion was verified to fail against its unfixed variant: the venue-keyed guard neutralized (all five coverage tests fail) and over-suppressing (both healthy-side assertions fail), the retry-ladder clears removed, the closed-status check removed, and the deferral test against both partial variants (the broad clear loses the registration entirely; preserving registration without the ladder reset produces exactly the stale-count `Canceled`).

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -84,6 +84,13 @@ static TAG_RECONCILIATION: LazyLock<Ustr> = LazyLock::new(|| Ustr::from("RECONCI
 pub type InstrumentAccountKey = (InstrumentId, AccountId);
 type FillKey = (AccountId, InstrumentId, TradeId);
 
+/// Execution clients responsible for reporting one cached entity.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum ReportClientCoverage {
+    Resolved(IndexSet<ClientId>),
+    Unresolved,
+}
+
 /// Metadata for an external order that needs to be registered with the execution client.
 #[derive(Debug, Clone)]
 pub struct ExternalOrderMetadata {
@@ -117,6 +124,7 @@ pub struct InflightCheckResult {
 pub(crate) struct OpenOrderReportCheck {
     pub command: GenerateOrderStatusReports,
     pub filtered_orders: Vec<OrderAny>,
+    pub client_coverage: IndexMap<ClientOrderId, ReportClientCoverage>,
     pub start: Option<UnixNanos>,
 }
 
@@ -125,6 +133,7 @@ pub(crate) struct OpenOrderReportCheck {
 pub(crate) struct PositionReportCheck {
     pub command: GeneratePositionStatusReports,
     pub positions_by_key: IndexMap<InstrumentAccountKey, Vec<Position>>,
+    pub client_coverage: IndexMap<InstrumentAccountKey, ReportClientCoverage>,
 }
 
 struct RetainedFillState {
@@ -287,6 +296,8 @@ pub struct ExecutionManager {
     position_recon_retries: IndexMap<InstrumentAccountKey, u32>,
     position_reconciliation_tolerances: IndexMap<(Venue, AccountId), Decimal>,
     recent_fills_cache: RecencyMap<FillKey>,
+    missing_order_coverage_warnings: IndexSet<ClientOrderId>,
+    unresolved_order_coverage: IndexSet<ClientOrderId>,
 }
 
 impl Debug for ExecutionManager {
@@ -322,6 +333,8 @@ impl ExecutionManager {
             position_recon_retries: IndexMap::new(),
             position_reconciliation_tolerances: IndexMap::new(),
             recent_fills_cache: RecencyMap::default(),
+            missing_order_coverage_warnings: IndexSet::new(),
+            unresolved_order_coverage: IndexSet::new(),
         }
     }
 
@@ -1069,10 +1082,43 @@ impl ExecutionManager {
 
     /// Prepares a bulk open-order report request and snapshots cached open orders.
     pub(crate) fn prepare_open_order_report_check(
-        &self,
+        &mut self,
         command_id: UUID4,
+        clients: &[&dyn ExecutionClient],
     ) -> OpenOrderReportCheck {
         let filtered_orders = self.filtered_open_orders_for_reconciliation();
+        let active_order_ids: IndexSet<ClientOrderId> = filtered_orders
+            .iter()
+            .map(|order| order.client_order_id())
+            .collect();
+        self.missing_order_coverage_warnings
+            .retain(|client_order_id| active_order_ids.contains(client_order_id));
+        self.unresolved_order_coverage
+            .retain(|client_order_id| active_order_ids.contains(client_order_id));
+
+        let mut client_coverage = IndexMap::new();
+
+        for order in &filtered_orders {
+            let client_order_id = order.client_order_id();
+            let coverage = self.resolve_order_report_client_coverage(order, clients);
+
+            match &coverage {
+                ReportClientCoverage::Resolved(_) => {
+                    if self
+                        .unresolved_order_coverage
+                        .shift_remove(&client_order_id)
+                    {
+                        self.missing_order_coverage_warnings
+                            .shift_remove(&client_order_id);
+                    }
+                }
+                ReportClientCoverage::Unresolved => {
+                    self.unresolved_order_coverage.insert(client_order_id);
+                }
+            }
+
+            client_coverage.insert(client_order_id, coverage);
+        }
 
         log::debug!(
             "Found {} order{} open in cache",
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @folknor

Thank you for the PR.

The failed-report guard is keyed to the raw venue, so it does not cover supported routing clients. `failed_venues` records `client.venue()` and then compares that with `order.instrument_id().venue`; IB deliberately uses a broker venue and handles exchange venues through `handles_order_venue`, so a report failure still walks those orders toward local rejection or cancellation. The same venue-level key can also suppress reconciliation for a healthy client when another client on that venue fails.

Please carry the failed `ClientId` or resolved order-client identity through reconciliation, and cover both a routing-broker venue mismatch and two clients sharing a venue.

folknor:
Thanks @cjdsellers - confirmed on both counts. `handles_order_venue` is exactly the surface the guard ignored (the IB impl returns true unconditionally while `venue()` is the broker venue), and the venue key both misses routed orders and over-suppresses a healthy client sharing a venue with a failed one.

The guard is reworked to client-identity coverage, now consolidated across both the order and position report paths:

- At check-preparation time each cached open order / position key is resolved to its responsible client set: orders resolve through the cache's `client_order_id -> ClientId` index first, then an exact `account_id` match, then the set of clients whose `handles_order_venue` claims the instrument venue; position keys resolve by exact `account_id` match, then `handles_order_venue`. No venue comparison remains.
- The query loops carry queried and failed `ClientId` sets through reconciliation (including the node's async report tasks, which now transport no venue policy). Missing-at-venue resolution runs only when the responsible set is nonempty and every member was queried successfully; otherwise the order/position is skipped without advancing retry state. Unresolvable coverage (e.g. a cached client ID no longer registered after a config change) logs once per order rather than resolving destructively on data that was never obtained.
- Two adjacent retry-ladder hardenings surfaced while testing the rework: a positive report now resets the missing-order retry ladder - including reports carrying only a `venue_order_id`, which are mapped through the cache index - so only consecutive misses accumulate toward resolution (the Python engine clears per report but shares the venue-ID-only gap), and `handle_missing_order` now checks for a terminal status before touching the retry counter, so an order that closes while a report request is in flight leaves no stale tracking behind.

Both requested scenarios are pinned by regression tests: a failing routing client (broker `venue()`, exchange-venue order claimed via `handles_order_venue`) no longer walks its orders toward rejection, and with two clients on one venue only the failed client's orders are skipped while the healthy client's orders still reconcile. Position-path twins cover the same two cases keyed by account, plus a fallback-coverage test for an order with no cached client ID, and tests for the retry-ladder reset (both report shapes) and the closed-during-query path. Each was verified to fail against the venue-keyed implementation (or its respective unfixed variant) before finalizing.

While reviewing this we noticed the position reconciliation tolerances have the same venue-identity issue (stored under `client.venue()`, looked up under `instrument_id.venue`, so a routing client never resolves its configured tolerance) - kept out of this PR. I will file a follow-up once this is merged.

cjdsellers (approved):
Thank you for the follow-ups @folknor. The client-identity routing fix addresses the previous blocker, but the pending-order defer path still has a small restart gap. I'll address this post-merge rather than holding up the PR.

PR #4480, opened 2026-07-16
Fix nondeterministic event ordering in backtests

folknor:
Backtest runs of the same binary on the same data could produce different event streams, because randomly seeded hash-map iteration and a random tie-break leaked into behavior-bearing ordering. The Cython implementation these files port used insertion-ordered Python dicts and a stable sort, so the port introduced the nondeterminism. Three sources, closed together:

**1. Venue and matching-engine iteration order**

`BacktestEngine::venues` and `SimulatedExchange::matching_engines` were `AHashMap`s; ahash's default `runtime-rng` feature seeds every map per process, so iteration order differs between runs. These maps drive venue settle order (`settle_venues`, `run_venue_modules`, `run_venue_liquidations`), per-instrument fill iteration (`iterate_matching_engines`), expiration processing, and liquidation order - and fill callbacks can synchronously submit further commands, so the order is observable well beyond logging. Both are now insertion-ordered `IndexMap`s (registration order, matching the Python dict semantics). `IndexMap::insert` on an existing key keeps its position, so the instrument re-add path does not reorder. FX-rollover per-instrument adjustments and funding-settlement account adjustments now iterate in sorted order (`InstrumentId` / currency code - `Ustr`'s `Ord` compares string contents, so both keys are content-ordered and stable across processes).

**2. Equal-timestamp time events tie-broke on a random UUID**

`TimeEventHandler` ordering falls through `ts_event`, `name`, `ts_init` to the event UUID text, and the backtest accumulator's heap consumed that ordering directly - so same-name timers firing at the same nanosecond on different component clocks (e.g. two strategies each running a `REBALANCE` timer with the same interval) popped in random order run to run. The accumulator now assigns a monotonic FIFO sequence at push and orders by `(ts_event, name, ts_init, sequence)`, restoring the Cython stable-sort semantics. The sequence resets with `clear()` and the increment is overflow-checked. `TimeEventHandler`'s own `Ord` is unchanged - the wrapper lives entirely in the accumulator.

**3. Component-clock enumeration order (completes fix 2)**

The FIFO sequence is assigned in clock-advance order, and `Trader::get_component_clocks` enumerated `Trader.clocks` - itself an `AHashMap` - so cross-clock ties would have merely traded UUID randomness for clock-enumeration randomness. `Trader.clocks` is now a registration-ordered `IndexMap`, with `shift_remove` at every removal site so surviving order is preserved when a component is removed.

The repo already treats these orderings as significant elsewhere (`ahash::RandomState::with_seeds(0, 0, 0, 0)` in orderbook aggregation, the explicit "Sort for deterministic insertion order" in `SimulatedExchange::load_open_orders`); these were the remaining behavior-bearing sites. Non-behavior-bearing maps (instrument lookups, settlement price caches, membership sets) are deliberately left on `AHashMap`.

**Testing**

- `test_accumulator_preserves_fifo_order_for_identical_timer_keys` - two events with identical `(ts_event, name, ts_init)` and adversarially ordered UUIDs (the first pushed carries the lexically larger UUID, so the old tie-break pops it second) assert FIFO pop order.
- `test_matching_engine_iteration_order_is_stable_across_rebuilds` - registration order of matching-engine keys across 32 exchange rebuilds.
- `test_same_timestamp_fills_follow_matching_engine_registration_order` - behavioral: two instruments with resting orders and a crossing quote parked behind a market-closed status, reopened, then a single `iterate_matching_engines` pass; asserts emitted `OrderFilled` order across 32 rebuilds.
- `test_get_component_clocks_returns_registration_order` - 32 registered component clocks asserted in registration order by `Rc` pointer identity.
- Each test was verified to fail against the unfixed implementation (the fill test failed with fills in hash order; the clock test failed with pointer-order mismatches).
- `test_list_venues_multiple` previously sorted its result to work around the nondeterminism; it now asserts insertion order directly.

### DIFF
--- a/crates/backtest/src/accumulator.rs
+++ b/crates/backtest/src/accumulator.rs
@@ -15,7 +15,10 @@
 
 //! Time event accumulation and scheduling for the backtest engine.
 
-use std::{cmp::Reverse, collections::BinaryHeap};
+use std::{
+    cmp::{Ordering, Reverse},
+    collections::BinaryHeap,
+};
 
 use nautilus_common::{clock::TestClock, timer::TimeEventHandler};
 use nautilus_core::UnixNanos;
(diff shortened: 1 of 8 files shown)


folknor:
Hmm @cjdsellers I will request a review here, the CI is failing on two python builds but the cause is:

```
error[1m: linking with `cc` failed: exit status: 1
...
##[error]No space left on device : '/home/runner/actions-runner/cached/2.335.1/_diag/pages/...'
```

Seems to me that failure is unrelated to my changes.

cjdsellers (approved):
Thanks @folknor, I reviewed this and found no blockers. The Python 3.12 and 3.14 failures were runner disk-space issues during Rust adapter linking, and both jobs passed on an unchanged rerun. I also ran the four new focused regressions locally and they all pass.

PR #4483, opened 2026-07-17
Fix underflow panics on out-of-order fill events

folknor:
Four guards against unsigned-arithmetic panics in the order and position
event paths, all reachable from out-of-order venue event streams that the
surrounding code already tolerates by design (warn-and-continue on early
timestamps, validate-and-store on corrections that arrive before their
fills). Same class of defensive alignment as the reconciliation fixes in
#4479.

**Position duration_ns underflow on an out-of-order closing fill**

`Position::apply_fill` deliberately tolerates `fill.ts_event < ts_opened`
(it logs a warning and proceeds), but the flat-close branch then computes
`ts_closed.as_u64() - self.ts_opened.as_u64()` as a plain `u64`
subtraction. A closing fill stamped earlier than the position open panics
in debug builds and wraps `duration_ns` to a value near `u64::MAX` in
release. The existing out-of-order regression
(`test_position_apply_fill_with_earlier_timestamp_adjusts_ts_opened`) only
covers an increasing fill, which never reaches the duration branch.

Fixed with `saturating_sub`, clamping the duration to zero. The legacy
Cython implementation has the same wrap (`uint64_t` fields in
`position.pxd`); this PR is scoped to the Rust engine.

**Quantity underflow when an `OrderFillVoided` precedes its fill**

`validate_fill_void` has a dedicated branch for a void whose fill has not
arrived yet: it bounds `voided_qty` only against the order quantity, since
there is no fill to compare with, and stores the event. When the fill
later arrives with `last_qty < voided_qty`, the next recompute
(`OrderCore::recompute_fill_state`, triggered by any subsequent
`OrderFillVoided`) matches the stored correction against the fill and
computes `fill.last_qty - removed`, which lands in `Quantity::sub`'s
`checked_sub(...).expect(...)` - a process panic from an ordinary
out-of-order message pair.

Fixed by clamping the matched correction to the fill's quantity
(`event.voided_qty.min(fill.last_qty)`) inside the recompute, consistent
with the saturating arithmetic already used throughout that accumulator.

**The same underflow in snapshot reconciliation**

`create_reconciliation_fill_voids` in nautilus-execution reads the same
stored corrections when projecting fill voids from an authoritative venue
snapshot, and computed `fill.last_qty - prior_qty` with the raw stored
`voided_qty`. Reconciling an order that carries a pre-fill oversized void
therefore panicked on exactly the event data the model-side fix tolerates.
Fixed with the identical clamp.

**Currency-mismatch panic on a zero-commission pre-fill void**

The same no-fill validation branch accepts `commission_voided` when it is
zero - in any currency, since there is no fill commission to compare
against. `recompute_fill_state` then subtracts it from the fill's
commission, and `Money::sub` asserts currency equality: a zero-USD void
meeting a fill commissioned in ETH aborts the process. Fixed by filtering
zero voided commissions out of the subtraction.

**Testing**

- `test_position_close_before_open_clamps_duration` pins the flat-close
  state for an out-of-order close: `duration_ns == 0`, `ts_closed` set
  from the fill, `ts_opened` and `closing_order_id` intact (proving the
  clamp is not achieved by rewriting the open timestamp).
- `test_fill_void_before_smaller_fill_clamps_recomputed_quantity` stages
  void(60,000) before fill(40,000) and triggers the recompute; it pins the
  clamped totals (`voided_qty == 50,000` proves the matched correction was
  clamped to 40,000 while an unmatched 10,000 correction stayed counted)
  plus status, filled, and leaves.
- `test_fill_void_stored_before_smaller_fill_does_not_underflow_snapshot_reconciliation`
  drives the same stored state through the public
  `generate_reconciliation_order_snapshot_events` with a filled=0 voided
  snapshot and pins the projected terminal state.
- `test_zero_commission_fill_void_before_fill_in_other_currency` stages a
  zero-USD pre-fill void against a 2-ETH-commission fill and pins the
  surviving commission.

Each was verified to FAIL against the unfixed code with its specific fix
hunk reverted: the two quantity tests and the reconciliation test panic at
the respective subtraction sites, and the commission test hits the
`Money::sub` currency assertion.

### DIFF
--- a/crates/execution/src/reconciliation/orders.rs
+++ b/crates/execution/src/reconciliation/orders.rs
@@ -553,7 +553,7 @@ fn create_reconciliation_fill_voids(
         });
         let prior_qty = previous.map_or_else(
             || Quantity::zero(fill.last_qty.precision),
-            |voided| voided.voided_qty,
+            |voided| voided.voided_qty.min(fill.last_qty),
         );
         let effective = fill.last_qty - prior_qty;
         if effective.is_zero() {
--- a/crates/execution/src/reconciliation/tests.rs
+++ b/crates/execution/src/reconciliation/tests.rs
@@ -23,8 +23,8 @@ use nautilus_model::{
         OrderAccepted, OrderEventAny, OrderFilled, OrderPendingCancel, OrderPendingUpdate,
         OrderSubmitted,
         order::spec::{
-            OrderAcceptedSpec, OrderFilledSpec, OrderPendingCancelSpec, OrderPendingUpdateSpec,
-            OrderSubmittedSpec, OrderUpdatedSpec,
+            OrderAcceptedSpec, OrderFillVoidedSpec, OrderFilledSpec, OrderPendingCancelSpec,
+            OrderPendingUpdateSpec, OrderSubmittedSpec, OrderUpdatedSpec,
         },
     },
     identifiers::{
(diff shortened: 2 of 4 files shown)


cjdsellers:
Hi @folknor,

Thanks again for another high-quality PR.

Just one finding to address before merge. The new clamp only recomputes `OrderCore`; it does not replay a pre-fill correction through the execution-owned position and portfolio path. With void(T1, 60, reopened) -> fill(T1, 40) -> unrelated void(T2, 10), the late fill is applied in full to the position, then T2 makes the order report `filled_qty == 0` while the position retains 40.

Please handle the stored T1 correction when its fill arrives, or reject/buffer that fill, and add an `ExecutionEngine` regression that asserts the order, position, and portfolio exposure remain aligned.

folknor:
Thank you for the kind words!

I've investigated it now, and you're obviously correct. But the fix covers a large area of the codebase including execution + event_store + portfolio-aware tests.

I would propose to land this panic-hardening PR as-is, and I do an immediate follow-up with the larger changes required to properly fix your concerns.

To be clear: I am happy to accomodate either way.

cjdsellers (approved):

PR #4489, opened 2026-07-17
Improve Python v2 Interactive Brokers type annotations

### LINKED ISSUE #4476: Improve type annotations in Interactive Brokers web module
# Feature Request

## Problem Statement

The `Product` NamedTuple inside the Interactive Brokers (IB) web adapter module (`nautilus_trader/adapters/interactive_brokers/web.py`) currently uses `Any` for its fields, accompanied by `# TODO: More specific type` comments. 

Additionally, the generator functions `_parse_products` and `load_product_list` return a generic `Generator` type without specifying type arguments. This lacks type safety and limits IDE autocomplete features for developers working with these modules.

### Proposed Solution

1. Update the `Product` NamedTuple fields to use the precise `str` type, as these values are parsed from table elements and stripped as strings.
2. Specify the generic type parameters for the generator functions to return `Generator[Product, None, None]` instead of a raw `Generator`.
3. Remove the stale `# TODO` comments.

### Example Usage

```python
class Product(NamedTuple):
    """
    Interactive Brokers Web Product.
    """

    ib_symbol: str
    description: str
    native_symbol: str
    currency: str


def _parse_products(table: _Element) -> Generator[Product, None, None]:
    ...


def load_product_list(
    exchange: Exchange,
    product_class: ProductClass,
    limit: int = 500,
    debug: bool = False,
) -> Generator[Product, None, None]:
    ...

decllxn:
Closes #4476


faysou:
You could move the refactored file to https://github.com/nautechsystems/nautilus_trader/tree/develop/nautilus_trader/adapters/interactive_brokers_pyo3 as the python adapter will soon be retired. The idea of nautilus v2 is to be mostly implemented in rust, but I think in this case we could make an exception.

decllxn:
> You could move the refactored file to https://github.com/nautechsystems/nautilus_trader/tree/develop/nautilus_trader/adapters/interactive_brokers_pyo3 as the python adapter will soon be retired. The idea of nautilus v2 is to be mostly implemented in rust, but I think in this case we could make an exception.

Thanks for the suggestion! That makes sense, especially given the plan to retire the Python adapter and move toward the Rust-based architecture in v2. I'll move the refactored implementation to `nautilus_trader/adapters/interactive_brokers_pyo3` and update the PR accordingly. I appreciate the guidance.

decllxn:
Done. I have moved the refactored \web.py\ file to \
nautilus_trader/adapters/interactive_brokers_pyo3/web.py\.

cjdsellers:
Hi @decllxn 

Thanks for the PR. This change only modifies the legacy v1 package under `nautilus_trader/`. With the v2 cutover imminent we are only accepting this type of improvement for Python v2, where type annotations are generated from Rust binding metadata using `make py-stubs-v2`; so handwritten annotations are not a source of truth.

`Product`, `load_product_list`, and this web module do not currently exist in the v2 API. Please rework this as a Rust/PyO3 v2 change and regenerate the stubs if this functionality is still needed.

cjdsellers:
Hi @decllxn,

There hasn't been a v2 rework, and this PR still only changes the legacy v1 Python/Cython package. v1 is now out of scope for this improvement, and the CLA check also remains incomplete because one commit is not linked to a signing GitHub account.

I'm going to close this PR and #4476. If this functionality is added to the Rust/PyO3 v2 API later, its annotations should come from the generated v2 stubs rather than a handwritten Python module.

Thanks again for taking the time to propose this.

PR #4490, opened 2026-07-17
Fix position reconciliation tolerance lookup for routing clients

folknor:
A follow-up to the routing-client identity work in #4479.

**Position reconciliation tolerance never resolved for routing clients**

The per-account reconciliation tolerance (`ExecutionClient::position_reconciliation_tolerance`) was stored in `ExecutionManager` under `(client.venue(), account_id)` (`set_position_reconciliation_tolerance`) but looked up under `(instrument_id.venue, account_id)` (`position_reconciliation_tolerance`). For a single-venue client the two venues are identical and the lookup resolved. For a routing broker they are not: IB uses a broker `venue()` while claiming the instrument's exchange venue through `handles_order_venue` (the exact surface #4479 keyed its guard to), so `instrument_id.venue != client.venue()`, the lookup missed, and the tolerance silently fell back to `DEFAULT_POSITION_RECONCILIATION_TOLERANCE` (`1e-8`). A routing client's configured tolerance therefore never applied, and any sub-tolerance position drift was "reconciled" with a spurious market order/fill - on both the continuous check (`check_position_discrepancy`) and the startup mass-status netting path (`reconcile_position_report_netting`).

The fix keys the tolerance map on `AccountId` alone. The `ExecutionClient` trait binds each client to exactly one `account_id` and one tolerance, and `account_id` is already the component both call sites share and agree on - only the venue component ever mismatched, and only for routing clients. Keying by account resolves routing clients correctly and needs nothing threaded into the mass-status path, which already carries the account id.

**Testing**

- `test_position_check_uses_routing_client_tolerance` - a routing client (broker `venue()` `IB`, an exchange-venue instrument claimed via `handles_order_venue`, configured tolerance `0.01`) with a `0.005` cached-vs-venue drift generates no reconciliation event on the continuous `check_position_discrepancy` path.
- `test_mass_status_netting_uses_routing_client_tolerance` - the same routing setup exercised through `reconcile_execution_mass_status` -> `reconcile_position_report_netting`, after seeding tolerance state the way the builder does in production (the seeding report is asserted to generate no events).
- `test_routing_clients_on_same_venue_use_account_tolerances_independently` - two IB accounts on one venue with tolerances `0.01` and `0.001` and an identical `0.005` drift: only the strict account produces a fill, proving per-account resolution and no cross-account collision.

All tests use paused/virtual time with no wall-clock waits. Each assertion was verified to FAIL against the unfixed venue-keyed code: the routing lookup misses and falls to the `1e-8` default, so the `0.005` drift exceeds tolerance and a spurious reconciliation order/fill is emitted (and the two-account test yields fills for both accounts instead of only the strict one).

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -57,7 +57,7 @@ use nautilus_model::{
     events::{OrderCanceled, OrderEventAny, OrderFilled, OrderInitialized},
     identifiers::{
         AccountId, ClientId, ClientOrderId, InstrumentId, PositionId, StrategyId, TradeId,
-        TraderId, Venue, VenueOrderId,
+        TraderId, VenueOrderId,
     },
     instruments::{Instrument, InstrumentAny},
     orders::{Order, OrderAny, TRIGGERABLE_ORDER_TYPES},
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
Thanks @folknor, solid focused changes 👍

PR #4491, opened 2026-07-17
Reconcile out-of-order fill voids through positions, portfolio, and replay

folknor:
A follow-up to the fill-void panic-hardening in #4483: when an `OrderFillVoided` is recorded before its matching `OrderFilled` arrives, only `OrderCore` recomputed on the late fill - the execution-owned position and portfolio kept the full fill, and event-store replay diverged from the live engine on flips, reopens, and spreads.

**Apply the stored correction when the late fill arrives**

`OrderCore::apply` recomputes the order when a fill lands on a trade that already carries a stored void, but `ExecutionEngine` applied the full fill to the position and never replayed the correction, so a later unrelated void drove the order to `filled_qty == 0` while the position and portfolio still held the full quantity. The trace `void(T1,60,reopened) -> fill(T1,40) -> void(T2,10)` ended with the order at 0 and the position at 40. After `handle_order_fill`, the engine now replays the matched, normalized correction (`voided_qty` clamped to `last_qty`, zero or currency/sign-incompatible `commission_voided` dropped) through `prepare_order_fill_void_positions` and refreshes the portfolio via the published position events. It deliberately does NOT re-dispatch the stored `OrderFillVoided`: that event was already captured, and a re-dispatch is persisted a second time once its identity ages out of the capture dedup window, which makes an otherwise-valid live run unreplayable. The portfolio's `events.position.*` handler is a superset of the fill-void endpoint refresh, so the position events alone keep net position and PnL correct.

**Gate contingents on surviving filled quantity**

A fully-voided catch-up fill is published like any other, and the OTO/OCO handlers acted on every fill callback - a zero-effect fill submitted the OTO child (at quantity zero) and cancelled OCO siblings. `handle_order_filled` now computes the spawn-aware surviving filled quantity and returns no actions when it is zero; the leaves-gated OUO path is unchanged.

**Accept a same-trade fill on a voided order**

A non-reopened pre-fill void puts the order in `Voided`, and the FSM had no `(Voided, Filled)` transition, so the late fill was rejected outright (live) and hard-failed the whole run (replay) - the surviving exposure was lost. The transition is added but gated in `OrderCore::apply` on the fill's trade having a stored void at all; unrelated fills on a `Voided` order are still rejected. A matched void applies as the catch-up correction. A mismatched void is rejected as superseded and the fill applies uncorrected, followed by an immediate projection rebuild - without it the ordinary fill handler re-voids the order off the stale `voided_qty`. The rebuild keeps the order terminally `Voided` when a different trade's non-reopened void is still awaiting its own fill, so the gate stays armed against later unrelated fills.

**One classification for stored voids, shared by every consumer**

`OrderCore` selected the latest void for a trade then filtered it by match, while the engine and replay scanned for the latest matching void - so a superseding mismatched void disabled the catch-up on the order but resurrected an earlier correction on the position. Other consumers paired stored voids to fills by `trade_id` alone: `recompute_fill_state` applied a void the fill-time selector had rejected the next time any unrelated void triggered a recompute (diverging the order from position and portfolio), the `validate_fill_void` baseline let a rejected 10k void fail a genuine later 5k correction as stale, reconciliation's `create_reconciliation_fill_voids` counted a rejected prior into its cumulative quantity, and Betfair cache resync seeded fill lots with it.

`OrderFillVoided::classify_catch_up_for_fill` now classifies the latest stored void for a fill's trade - matched (apply `normalized_for_fill`), mismatched (rejected), or absent - and all of the above consume it (`find_catch_up_for_fill` remains as its matched-only view). A rejected void stays in the event history for audit but contributes nothing to `voided_qty`, commissions, leaves, status derivation, validation baselines, or reconciliation. The duplicate-void comparison canonicalizes zero commissions on both sides, since the normalized baseline drops `Some(zero)` while a re-sent duplicate void still carries it.

**Live/replay parity for flips, reopens, and spreads**

Event-store replay applied the raw unsplit fill to positions, so it diverged from the live engine three ways: it flipped closed positions that live reopens, never reconstructed live's one-to-many flip fragment topology (losing hedging positions and fragment identities), and created phantom positions for spread fills that live skips.

The flip-split arithmetic is extracted into `plan_position_flip` (in `nautilus-execution`), called by both live's `flip_position` and replay, with a deterministic domain-separated fragment-2 `event_id` so `Position.events` and `replay_events` match field-for-field.

Replay now mirrors live's dispatch: it reopens closed non-reduce-only positions via `Position::new` + `merge_reopened_position_history` (keeping the raw fill `event_id`), reconstructs hedging flips by stashing the opening fragment and claiming the live-generated position id from the captured `PositionOpened`, mirrors the reduce-only rejection, and skips spread positions.

Two boundaries are stated rather than hidden. A durable log truncated between an `OrderFilled` and its captured `PositionOpened` is fail-closed on replay, not reconstructed from a partial prefix. And the in-order over-void asymmetry is unchanged: a post-fill `void(60)` against a `fill(40)` is still rejected, while the out-of-order `void(60) -> fill(40)` clamps to 40, matching #4483.

**Testing**

The headline is a live-capture to replay round-trip differential, `live_capture_replay_round_trip_preserves_order_position_and_portfolio_state`. It runs a real `ExecutionEngine` with the bus capture tap through the maintainer trace, an open hedging flip with a void split across fragments, a closed to reopen-opposite lifecycle, and a spread fill, then replays the captured store into a fresh cache and asserts orders (event-for-event), positions (field-wise, including generated hedging ids, fragment topology, and realized PnL), and portfolio `net_position` are identical. It was verified to FAIL against the unfixed replay: neutralizing the closed-position reopen makes the captured run unreplayable.

The catch-up alignment is pinned directly. `test_pre_fill_void_catch_up_keeps_order_position_and_portfolio_aligned` holds the full maintainer trace to order/position/portfolio all zero; `test_partial_pre_fill_void_catch_up_keeps_real_exposure_aligned` holds `void(10) -> fill(40)` to a surviving 30 across all three. Both fail against develop, where the position and portfolio stay at 40. Non-reopened full and partial catch-up are covered on both the live and replay paths.

Selection and gating follow. `test_superseded_mismatched_void_disables_catch_up_on_order_and_position` pins that a superseding mismatched void disables the catch-up on both sides (verified to fail - reverting the shared selector yields order `filled_qty` 30k against the asserted 40k). `test_pre_fill_void_catch_up_dispatches_no_fill_voided` is a bus-tap spy on the no-re-dispatch path, and OTO-zero and OCO-zero manager tests pin the contingent gating.

The mismatch semantics are pinned end to end, each test verified to fail against the code without its fix hunk: the non-reopened mismatch tests (parameterized partial and full fill) fail with `InvalidStateTransition` on the old gate; the recompute test fails at `filled_qty` 30k vs 40k when the match filter is removed; the terminal-hold test fails as `PartiallyFilled` vs `Voided` with the rebuild guard disabled; the zero-commission duplicate test fails by appending the redundant correction under direct commission equality; the rejected-baseline test proves a later genuine same-trade correction applies; the reconciliation test proves a rejected prior is excluded from the emitted cumulative quantity and commission; the Betfair resync test proves a rejected void is not seeded into the fill tracker; and replay-parity scenario F holds the live and replayed order, position, and portfolio to the full uncorrected fill.

Replay parity is pinned by `replay_commission_only_flip_void_correction_applies_no_corrections` (verified to fail - neutralizing the commission-only bail applies a correction live rejects), the netting and hedging flip replay tests, `replay_hedging_flip_without_opened_event_fails_closed`, `replay_spread_fill_does_not_create_a_position`, and the reduce-only reopen-skip test. All tests use `#[rstest]` with paused or virtual time and no wall-clock waits.


cjdsellers:
Hi @folknor,

Thanks again for the PR.

The matched catch-up path looks good, but mismatch handling still has two state-loss cases.

A non-reopened pre-fill void puts the order into `Voided`. If the same-trade fill later mismatches the void metadata, `find_catch_up_for_fill` returns `None` and the new gate rejects the fill, despite the selector contract saying a mismatch leaves the fill applied uncorrected.

For the reopened case, the fill applies initially, but a later unrelated `OrderFillVoided` makes `recompute_fill_state` apply the rejected void using only `trade_id`. Execution corrects no position for that unrelated void, so the order diverges from the position and portfolio.

Please distinguish “no same-trade void” from “same-trade mismatch” in the `Voided + Filled` gate, and use the same match and normalization rule during recomputation. Regression tests should cover both sequences.

folknor:
Thanks @cjdsellers - both cases were real, and chasing them surfaced two more consumers with the same trade-id-only pairing. Reworked and force-pushed.

The gate now distinguishes the two `None` cases: `classify_catch_up_for_fill` returns matched / mismatched / no-same-trade-void, the `Voided + Filled` gate rejects only the last, and a mismatched same-trade fill applies uncorrected followed by an immediate projection rebuild (without it the ordinary fill handler re-voids the order off the stale `voided_qty`). The rebuild keeps the order terminally `Voided` when a different trade's non-reopened void is still awaiting its own fill, so the gate stays armed against unrelated fills.

Recomputation now uses the same match and normalization rule: `recompute_fill_state` pairs a fill only with a matched void via `matches_fill` + `normalized_for_fill`, and a rejected void contributes nothing anywhere - not to `voided_qty`, leaves, status, the `validate_fill_void` stale/duplicate baseline (a rejected 10k void no longer fails a genuine later 5k correction as stale), reconciliation's prior-quantity baseline in `create_reconciliation_fill_voids`, or Betfair cache-resync seeding. The event stays in history for audit only.

Regression tests cover both of your sequences (the non-reopened mismatch previously failed with `InvalidStateTransition`; the recompute sequence previously dropped `filled_qty` from 40k to 30k) plus the rejected-baseline, terminal-hold, zero-commission duplicate, reconciliation, Betfair, and live/replay-parity cases - each verified to fail against the code without its fix hunk.

cjdsellers:
Hey @folknor, 

I revisited this after your latest update and decided to tighten the `OrderFillVoided` contract instead: reopened corrections require the referenced fill to exist locally, while unapplied non-reopened corrections terminally void the order. That removes the need for the catch-up and replay machinery and greatly simplifies the implementation, so we're replacing this with a smaller contract-focused change.

Thanks again for all the work, especially after you already addressed the earlier review prior to the change in direction here.

folknor:
Fair enough, I do have a tendency to "go big", and your solution is much less invasive.

I just want to note there is a latent edge here: the moment an adapter sets `is_reopened=true`, or emits a partial non-reopened void, the contract drops/over-voids it, and the continous reconciliation path won't self-heal a bare reopen report.

However, my experience with the codebase and its history is limited, so I have no idea whether that was worth noting or not :-)

cjdsellers:
Thanks, that edge _was_ worth noting. I've clarified the `OrderFillVoided` docs to make the adapter contract explicit: a reopened correction, or a partial correction that leaves the order executable, must follow its referenced fill; without that fill, a non-reopened correction is an authoritative terminal void.

For now we're relying on venue adapters to map upstream reports into that contract correctly, which keeps the model and replay paths simpler; the docs also state that continuous reconciliation won't repair an ordering violation.

PR #4494, opened 2026-07-18
Fix persistence session runtime ownership

### LINKED ISSUE #4493: ParquetDataCatalog deallocation can livelock at 100% CPU (per-session owned Tokio runtime dropped inside PyO3 `tp_dealloc`)
# Bug Report

## Confirmation

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates (nearest are #3847 and #3889 — same subsystem, different defects; nothing matches this one).
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue using a recent pre-release or development wheel — reproduced on a v2 (`_libnautilus`) build from a recent `develop` checkout (`nautilus-persistence` v0.61.0).

### Expected Behavior

Releasing a `ParquetDataCatalog` (explicitly, or implicitly via Python garbage collection) should free its resources promptly. Catalog teardown should never block the interpreter indefinitely, because deallocation can be triggered by cyclic GC at arbitrary points in a long-lived process — including inside live/production processes at shutdown.

### Actual Behavior

When a `ParquetDataCatalog` is collected by Python's cyclic GC mid-process, the interpreter can spin at 100% CPU **indefinitely** (observed twice, 30+ minutes each before being killed) inside the catalog's PyO3 destructor. It is a busy-spin livelock, not a parked deadlock: one core stays pinned and the process never progresses.

The mechanism is structural:

1. Every `DataBackendSession` constructs and **owns** a dedicated multi-thread Tokio runtime (`crates/persistence/src/backend/session.rs` — `DataBackendSession::new` builds `tokio::runtime::Builder::new_multi_thread()` per instance; `EagerStream` in `kmerge_batch.rs` holds the same `Arc<Runtime>`).
2. `ParquetDataCatalog` owns the session; the Python `ParquetDataCatalog` owns the catalog. There is no close/shutdown API anywhere in that chain, so the runtime is destroyed implicitly whenever the Python object is deallocated.
3. Tokio explicitly documents that plain `Runtime::drop` performs an unbounded synchronous shutdown: *"The thread initiating the shutdown blocks until all spawned work has been stopped. This can take an indefinite amount of time. The `Drop` implementation waits forever for this."*
4. That unbounded drop therefore runs inside PyO3 `tp_dealloc` (via the `trampoline_unraisable` dealloc trampoline), on whatever thread GC happens to run on, at whatever moment GC fires. In our observed hangs the spin is source-consistent with multi-thread scheduler shutdown (`worker.rs` `close()` path) rather than the parked blocking-pool wait.

Even when the drop completes normally, one heavyweight multi-thread runtime per catalog instance contradicts the project's own single-global-runtime rationale (`crates/common/src/live/runtime.rs`: a single long-lived runtime avoids repeated startup/shutdown overhead).

The livelock itself is timing-dependent — it needs a session with real write/query history whose final reference is dropped by **cyclic GC mid-process** (not at interpreter shutdown). We hit it twice reliably in a large pytest suite where catalogs created early were kept alive by reference cycles and collected during later, unrelated tests; the hang position moved when the test order changed, confirming GC-timing dependence. Isolated create/drop, `gc.collect()`, and interpreter-exit reproductions complete normally.

The defective ownership design, however, reproduces deterministically:

1. Create N `DataBackendSession` instances (directly, or via N `ParquetDataCatalog` objects).
2. Observe N distinct multi-thread Tokio runtimes exist simultaneously (e.g. via `tokio::runtime::Handle::id()` inside the crate, or OS thread counts: each instance spawns its own worker/blocking threads).
3. Each of those runtimes is shut down synchronously and unboundedly from whatever thread drops the last reference — in Python usage, inside PyO3 `tp_dealloc` at GC time.

Observed hang (twice, independently, macOS/Apple Silicon): full-venv pytest run over a suite that creates ~10 catalogs with real Parquet write/query traffic; process pins one core at 100% for 30+ minutes with zero progress; `sample` snapshots ~25 minutes apart show the identical stack (excerpt below); the Python test "running" at the time passes in 0.02 s in isolation — the time is entirely inside the dealloc of an earlier test's catalog.

### Code Snippets or Logs

macOS `sample` hot stack (identical across two snapshots 25 minutes apart, 100% CPU throughout):

```text
subtype_dealloc  (in Python)
PyObject_ClearManagedDict  (in Python)
pyo3::impl_::pyclass::tp_dealloc<... PyParquetDataCatalog>  (in _libnautilus)
pyo3::impl_::trampoline::trampoline_unraisable / catch_unwind  (in _libnautilus)
<nautilus_persistence ... PyParquetDataCatalog as drop>  (in _libnautilus)
core::ptr::drop_glue<... backend::catalog::ParquetDataCatalog>
core::ptr::drop_glue<... backend::session::DataBackendSession>
alloc::sync::Arc<tokio::runtime::runtime::Runtime>::drop_slow   <-- spins here, never returns
```

Ownership chain (paths as of a recent `develop`):

- `crates/persistence/src/python/catalog.rs` — `PyParquetDataCatalog { inner: ParquetDataCatalog }`
- `crates/persistence/src/backend/catalog.rs` — `ParquetDataCatalog { session: DataBackendSession, .. }`
- `crates/persistence/src/backend/session.rs` — `DataBackendSession::new`: `Builder::new_multi_thread().enable_all().build()` per instance, stored as `Arc<Runtime>`; no close API.
- `crates/persistence/src/backend/kmerge_batch.rs` — `EagerStream` holds the same `Arc<Runtime>`.

Proposed fix (we have a working patch and are happy to open a PR): store a clone of the global runtime `Handle` (per `crates/common/src/live/runtime.rs`) in `DataBackendSession` and `EagerStream` instead of an owned `Runtime`. `Handle` does not own or shut down the runtime, so PyO3 deallocation becomes trivially cheap; all sessions share the single global runtime, consistent with the documented design intent. Our regression test asserts all sessions report the same `Handle::id()` — it fails on current `develop` (20 sessions → 20 distinct runtime IDs) and passes with the patch.

### Specifications

- OS platform: macOS (Apple Silicon, Darwin 25.5.0); production target Linux Ubuntu 24.04 aarch64 (not yet reproduced there, but the mechanism is platform-independent)
- Python version: CPython 3.14.6 (PyArrow 25.0.0 installed)
- `nautilus_trader` version: v2 (`nautilus_trader._libnautilus`) built from source at a recent `develop` checkout; `nautilus-persistence` v0.61.0

faysou:
# Pull Request

## Summary

- Fixes [#4493](https://github.com/nautechsystems/nautilus_trader/issues/4493), where Python garbage collection can synchronously drop a catalog-owned Tokio runtime and block indefinitely inside PyO3 deallocation.
- Replaces the runtime owned by each `DataBackendSession` with a cloned handle to the global Nautilus runtime.
- Makes `EagerStream` retain the same lightweight runtime handle instead of extending ownership through `Arc<Runtime>`.
- Adds a regression test proving independently created backend sessions use the same global runtime.

### Design decisions

- Reuses `nautilus_common::live::get_runtime()` to match the existing single-runtime design and avoid repeated worker startup and shutdown.
- Keeps the runtime handle on `DataBackendSession` because synchronous DataFusion operations still need `block_on`, `spawn`, and runtime context access.
- Keeps the runtime handle on `EagerStream` so its blocking iterator can receive from the Tokio channel without owning runtime shutdown.
- Uses runtime identity as the regression contract because the reported livelock is timing-dependent, while distinct per-session runtimes reproduce deterministically.

### Runtime ownership

```mermaid
flowchart LR
    Global[Global Nautilus runtime] -->|clone Handle| Session[DataBackendSession]
    Session -->|clone Handle| Stream[EagerStream]
    Catalog[ParquetDataCatalog] -->|owns| Session
    Python[Python garbage collection] -->|drops| Catalog
    Session -. no shutdown ownership .-> Global
    Stream -. no shutdown ownership .-> Global
```

### Relevant files

- `crates/persistence/src/backend/session.rs`: uses the global runtime handle and verifies shared runtime identity.
- `crates/persistence/src/backend/kmerge_batch.rs`: stores a runtime handle in `EagerStream`.

### Verification

- `cargo test -p nautilus-persistence data_backend_sessions_share_global_runtime --lib`
- `cargo test -p nautilus-persistence backend::kmerge_batch::tests --lib`
- Repository pre-commit hooks, including Clippy and documentation checks.

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/4493

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/persistence/src/backend/kmerge_batch.rs
+++ b/crates/persistence/src/backend/kmerge_batch.rs
@@ -13,11 +13,10 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
 
-use std::{sync::Arc, vec::IntoIter};
+use std::vec::IntoIter;
 
 use futures::{Stream, StreamExt};
 use tokio::{
-    runtime::Runtime,
     sync::mpsc::{self, Receiver},
     task::JoinHandle,
 };
@@ -30,11 +29,11 @@ use super::{
 pub struct EagerStream<T> {
     rx: Receiver<T>,
     task: JoinHandle<()>,
-    runtime: Arc<Runtime>,
+    runtime: tokio::runtime::Handle,
 }
 
 impl<T> EagerStream<T> {
-    pub fn from_stream_with_runtime<S>(stream: S, runtime: Arc<Runtime>) -> Self
+    pub fn from_stream_with_runtime<S>(stream: S, runtime: tokio::runtime::Handle) -> Self
     where
         S: Stream<Item = T> + Send + 'static,
         T: Send + 'static,
--- a/crates/persistence/src/backend/session.rs
+++ b/crates/persistence/src/backend/session.rs
@@ -21,6 +21,7 @@ use datafusion::{
     physical_plan::SendableRecordBatchStream, prelude::*,
 };
 use futures::StreamExt;
+use nautilus_common::live::get_runtime;
 use nautilus_core::{UnixNanos, ffi::cvec::CVec};
 use nautilus_model::data::{Data, HasTsInit};
 use nautilus_serialization::arrow::{
@@ -68,7 +69,7 @@ pub type QueryResult = KMerge<EagerStream<std::vec::IntoIter<Data>>, Data, TsIni
 )]
 pub struct DataBackendSession {
     pub chunk_size: usize,
-    pub runtime: Arc<tokio::runtime::Runtime>,
+    pub runtime: tokio::runtime::Handle,
     session_ctx: SessionContext,
     batch_streams: Vec<EagerStream<IntoIter<Data>>>,
     registered_tables: AHashSet<String>,
@@ -76,16 +77,8 @@ pub struct DataBackendSession {
 
 impl DataBackendSession {
     /// Creates a new [`DataBackendSession`] instance.
-    ///
-    /// # Panics
-    ///
-    /// Panics if Tokio cannot create the worker runtime.
     #[must_use]
     pub fn new(chunk_size: usize) -> Self {
-        let runtime = tokio::runtime::Builder::new_multi_thread()
-            .enable_all()
-            .build()
-            .unwrap();
         let session_cfg = SessionConfig::new()
             .set_str("datafusion.optimizer.repartition_file_scans", "false")
             .set_str("datafusion.optimizer.prefer_existing_sort", "true");
@@ -94,7 +87,7 @@ impl DataBackendSession {
             session_ctx,
             batch_streams: Vec::default(),
             chunk_size,
-            runtime: Arc::new(runtime),
+            runtime: get_runtime().handle().clone(),
             registered_tables: AHashSet::new(),
         }
     }
@@ -434,5 +427,22 @@ impl Drop for DataQueryResult {
     fn drop(&mut self) {
         self.drop_chunk();
         self.result.clear();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use nautilus_common::live::get_runtime;
+    use rstest::rstest;
+
+    use super::*;
+
+    #[rstest]
+    fn data_backend_sessions_share_global_runtime() {
+        let first = DataBackendSession::new(10);
+        let second = DataBackendSession::new(10);
+
+        assert_eq!(first.runtime.id(), second.runtime.id());
+        assert_eq!(first.runtime.id(), get_runtime().handle().id());
     }
 }


cjdsellers (approved):
Thanks @faysou, this is a solid, focused change. iirc we planned to experiment with a separate runtime for catalog streaming, but this resolves the issue cleanly and we may no longer need to revisit that design.

PR #4496, opened 2026-07-18
Remove unsound Send/Sync impls from TimeEventCallback

folknor:
Removes the unsound `unsafe impl Send`/`Sync` on `TimeEventCallback` and gives the live timer channel a send-safe payload, implementing the registry dispatch the TODO in `LiveTimer::start` calls for.

**1. `RustLocal` callbacks crossed threads through the unsafe impls**

`TimeEventCallback` carries `RustLocal(Rc<dyn Fn(TimeEvent)>)`, yet `timer.rs` declared the whole enum `Send`/`Sync`. `LiveTimer::start` moved a clone into its tokio task and cloned it again on a worker thread on every fire (`TimeEventHandler::new(event, callback.clone())`) before sending the handler to the event-loop drain.

For a `RustLocal` callback that is an unsynchronized `Rc` refcount mutation across threads: a data race on the count, with premature free or double free as the failure mode. The path is reachable in production - the bar-aggregation timers register `RustLocal` callbacks on `LiveClock` in the live data engine.

Both unsafe impls are removed. `TimeEventCallback` and `TimeEventHandler` are now `!Send`, so moving a `RustLocal` callback across threads no longer compiles. `TestClock`, the backtest accumulator, and the FFI/Python handler wrappers keep the callback-bearing `TimeEventHandler` unchanged.

**2. The live channel now carries a send-safe `TimeEventMessage`**

The live channels carry a new opaque `TimeEventMessage` (compile-time asserted `Send + Sync`) instead of `TimeEventHandler`, with two private dispatch modes fixed per timer.

`Python` and `Rust` callbacks travel as Direct messages holding only their `Arc`-backed forms; any drain thread rebuilds a local `TimeEventHandler` and runs it, so behavior is unchanged - including cancel-with-queued-events and spawned runners draining on another thread.

`RustLocal` callbacks never enter the channel. They live in a thread-local registry keyed by a per-timer-instance token; the task sends only the token, and the drain resolves it after an owner-`ThreadId` check. A message drained on the wrong thread is dropped with an error log instead of touching the foreign `Rc`.

Token lifecycle is one atomic word (closed flag + in-flight lease count). Each fire acquires a lease before sending, so cancellation still lets already-queued events dispatch; a stop before the first fire sends a cleanup message that unregisters without invoking; same-name replacement mints a distinct token, so a queued old event can neither call nor unregister the re-armed callback.

**3. `start()` restart semantics under the token design**

`start()` resets `canceled` and previously re-cloned the stored callback, so restart-after-cancel is a public contract. The owner side now retains the `RustLocal` callback next to its token and `start()` registers a fresh token when the current one is closed - otherwise a restarted timer would acquire no leases and silently stop dispatching.

Starting a still-active timer stops the old task first, closing its token before the abort: `abort()` does not join, and a task already past its tick could otherwise close the reused token under the new one. The worker also publishes the advanced `next_time_ns` before a channel send, so a restart snapshotting the schedule mid-fire cannot re-emit the in-flight event (senderless dispatch keeps its original invoke-then-advance ordering).

**Migration notes**

FFI/Python wrapper layouts and clock method signatures are unchanged. For Rust consumers this is source-breaking where it must be: `TimeEventSender::send` and the live channel types now use `TimeEventMessage`, and types embedding `TimeEventCallback` (including `TestClock`, `LiveClock`, `LiveTimer`) lose their erroneous `Send`/`Sync`.

**Testing**

- `test_time_event_message_is_send_and_sync` pins the channel-payload bound at compile time; on the fixed tree, moving a `RustLocal` callback into `std::thread::spawn` fails with E0277.
- Registry lifecycle tests cover same-thread dispatch, wrong-thread drop, close/lease interactions (immediate removal, queued leases surviving close, off-owner drop reclaimed by purge), same-name replacement, one-shot re-arm from inside a callback, and cleanup messages not counting as dispatched.
- Live timer tests cover `RustLocal` dispatch through the registry, cancel preserving a queued lease, stop-before-first-fire cleanup, restart-after-cancel re-registering, and start-while-active restarting and continuing to dispatch.
- All `#[rstest]`, no wall-clock waits beyond bounded 1s channel receives.
- Differential verification: the wrong-thread and cancel-with-queued-lease tests fail against the unfixed implementation, and the restart test fails (silent non-fire) with the re-registration removed.

### DIFF
--- a/crates/common/src/live/clock.rs
+++ b/crates/common/src/live/clock.rs
@@ -28,7 +28,7 @@ use crate::{
         CallbackRegistry, Clock, replace_existing_timer, validate_and_prepare_time_alert,
         validate_and_prepare_timer,
     },
-    runner::{TimeEventSender, try_get_time_event_sender},
+    runner::{TimeEventSender, purge_closed_time_event_callbacks, try_get_time_event_sender},
     timer::{TimeEvent, TimeEventCallback, TimeEventHandler, create_valid_interval},
 };
 
@@ -66,6 +66,7 @@ impl LiveClock {
 
     fn clear_expired_timers(&mut self) {
         self.timers.retain(|_, timer| !timer.is_expired());
+        purge_closed_time_event_callbacks();
     }
 
     fn replace_existing_timer_if_needed(&mut self, name: &Ustr) {
@@ -298,9 +299,9 @@ mod tests {
     use super::*;
     use crate::{
         clock::Clock,
-        runner::TimeEventSender,
+        runner::{TimeEventMessage, TimeEventSender},
         testing::wait_until,
-        timer::{TimeEvent, TimeEventCallback, TimeEventHandler},
+        timer::{TimeEvent, TimeEventCallback},
     };
 
     #[derive(Debug)]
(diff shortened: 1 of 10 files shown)


cjdsellers (approved):
Thanks @folknor, this removes the unsafe cross-thread `Rc` path cleanly. I'll follow up separately on a narrow active-restart race in `LiveTimer::start`.

PR #4499, opened 2026-07-18
Fix CVec FFI soundness with checked reconstruction and unique ownership

folknor:
Removes the undefined behavior around `CVec`: the misaligned empty sentinel reconstructed by every typed drop helper, shallow `Copy`/`Clone` on an owning type, and safe `extern "C"` entry points rebuilding vectors and slices from unchecked public fields.

**1. Typed drops reconstructed `Vec<T>` from the u8-aligned empty sentinel**

`CVec::empty()` returns a `NonNull::<u8>::dangling()` pointer - address 1. Every typed drop helper (`vec_drop_book_levels`, `vec_drop_book_orders`, `vec_drop_fills`, `orderbook_deltas_vec_drop`, `vec_time_event_handlers_drop`, and the persistence `drop_chunk`) called `Vec::<T>::from_raw_parts` on it with no zero-capacity guard, violating the alignment precondition for any `T` with alignment above 1. Miri reports it as hard UB: constructing an unaligned reference (required 8, found 1).

Reconstruction is now centralized in two unsafe `CVec` primitives. `into_vec<T>` returns `Vec::new()` when `cap == 0` and `as_slice<T>` returns `&[]` when `len == 0`, in both cases without inspecting the pointer; the non-empty paths keep the `len <= cap` and null checks as hard asserts and add debug asserts for alignment and the `isize::MAX` byte bound. All drop helpers route through `into_vec`, which also gives the two helpers that previously checked nothing but null the same validation as the rest. No sentinel change: `From<Vec<T>>` is public and generic, so no fixed sentinel alignment could be correct for every downstream element type.

**2. `CVec` derived `Copy` and `Clone` while owning its allocation**

Any implicit copy of a `CVec` produced two values claiming the same allocation - a double free reachable from safe Rust through any two consumers of the same value. Both derives are removed; borrowing goes through `as_slice` and consumption through `into_vec`, so ownership transfer is a move the compiler enforces. The `repr(C)` layout and generated C declarations are unchanged - C can still copy the struct, so exactly-once remains an FFI contract on that side, now documented in `docs/developer_guide/ffi.md`.

The fallout sites are part of this diff: `drop_cvec_pycapsule` transfers ownership out of the capsule with `ptr::replace` (leaving the empty sentinel, so repeated drops stay harmless), and `DataQueryResult.chunk` becomes private with `set_chunk` returning a borrow.

**3. Safe entry points rebuilt vectors from unchecked foreign metadata**

`level_new`, `orderbook_deltas_new`, and `synthetic_instrument_calculate` were safe `extern "C"` functions reconstructing from whatever `ptr`/`len`/`cap` the caller supplied - a forged or corrupted `CVec` was UB inside a safe function. All three, plus the five typed drop exports, are now `unsafe extern "C"` with `# Safety` contracts, matching the convention already used across the model FFI. cbindgen emits identical declarations; the generated header diffs are comment-only.

`orderbook_deltas_new` also had an allocator-provenance violation: Cython passes a `PyMem_Malloc` buffer, and the function constructed a Rust `Vec` over it before cloning and forgetting. A `Vec` must never be built over memory the Rust allocator did not produce, forgotten or not. It now borrows the buffer with `as_slice` and clones with `to_vec`; the caller keeps ownership, as before.

**Testing**

- Miri regression tests cover the empty typed-drop path for a plain element type and a `repr(align(64))` element type (proving the zero-capacity path never reconstructs regardless of payload alignment), plus a non-empty round-trip with a drop counter asserting exactly-once destruction. A new `cargo-miri-core-ffi` slice runs them, wired into `cargo-miri-core` so the nightly Miri workflow reaches it.
- Behavior tests: empty typed drops return without panic across all helpers, `orderbook_deltas_new` clones rather than consumes its caller-owned buffer, `synthetic_instrument_calculate` leaves its input intact, `level_new` behavior is unchanged for valid input, and `as_slice` borrows without consuming.
- Differentials: the empty typed-drop test fails under Miri against the unfixed implementation with the unaligned-reference error above; with only the zero-capacity short-circuit removed, the new debug alignment assert catches the same class. The ownership and safety changes are compiler-enforced and were each demonstrated once pre/post: reuse after move fails E0382, `.clone()` fails E0599, and calling a reconstructor without `unsafe` fails E0133.


cjdsellers (approved):
Thanks @folknor, the FFI changes and focused Miri checks look solid. I'll note the CVec ownership changes and safety fixes in the release notes.

PR #4500, opened 2026-07-18
Fix Bybit missing due_post_only flag on post-only rejections

dxwil:
# Pull Request

## Summary

Bybit never set `due_post_only` on rejections, so a post-only order that crossed the book (`EC_PostOnlyWillTakeLiquidity`) was emitted as an ordinary `OrderRejected` with `due_post_only = false`, unlike every other adapter (OKX, Coinbase, Derive, etc.). This adds a `bybit_rejection_due_post_only` helper and wires it into the three reachable rejection sites (WS order-status stream, plus the single and batch HTTP submit paths).

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/bybit/src/common/parse.rs
+++ b/crates/adapters/bybit/src/common/parse.rs
@@ -172,6 +172,15 @@ use crate::{
 
 const BYBIT_HOUR_INTERVALS: &[u64] = &[1, 2, 4, 6, 12];
 
+const BYBIT_POST_ONLY_REJECT_REASON: &str = "EC_PostOnlyWillTakeLiquidity";
+
+/// Returns whether a Bybit rejection reason indicates a post-only order that
+/// would have taken liquidity (crossed the book).
+#[must_use]
+pub fn bybit_rejection_due_post_only(reason: &str) -> bool {
+    reason.contains(BYBIT_POST_ONLY_REJECT_REASON)
+}
+
 /// Extracts the raw symbol from a Bybit symbol by removing the product type suffix.
 #[must_use]
 pub fn extract_raw_symbol(symbol: &str) -> &str {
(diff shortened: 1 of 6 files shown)


cjdsellers:
Hi @dxwil

Thanks, the classifier itself is correct, but this does not cover the live Bybit response. Bybit documents post-only crossings as `orderStatus="Cancelled"` with `rejectReason="EC_PostOnlyWillTakeLiquidity"`, while this only sets the flag in the `Rejected` branch. The canceled WebSocket and status-report paths still produce `OrderCanceled`, and the updated test forces `Rejected`.

Please handle the canceled-plus-post-only combination in both paths, update the test to use the documented venue payload.

dxwil:
Hi @cjdsellers,

You're right. Added the post only cancel to rejection conversion where possible. I left the check inside the reject case if Bybit decides to change it in the future (it's only one line). If you want me to remove it then that's also fine.

Note on pre-commit: if I add the empty lines above the if statement then `cargo fmt` fails.

cjdsellers:
Hey @dxwil

Thanks, happy to keep that check in the `Rejected` branch: reasonable future-proofing in case Bybit ever flips these to `Rejected`, and it's cheap.

On the pre-commit note: no blank line is needed there. When the `if` is the first line of a block, `rustfmt` correctly strips a leading blank line, so `cargo fmt` reverting it is expected, not a conflict. Our formatting rules complement `rustfmt` rather than fight it, so just leave it as `cargo fmt` wants and pre-commit should be green.

dxwil:
`cargo fmt` currently passed, the code is formatted the way it wants. What doesn't pass is the `if` blank line check, it wants a blank line above the two single-pattern match guards which wrap `if` into a new line since they're too long. But there cannot be a blank line there because the `if` is part of the same match leg. The hook flags it incorrectly, it should allow it. Perhaps you could take a look since this probably needs a hook fix which affects the whole project (apparently this pattern has not been used in the whole project yet or at least not flagged :).

cjdsellers:
Hey @dxwil

Thanks for catching this, you were right: the formatting hook was incorrectly treating wrapped single-pattern match guards as ordinary `if` blocks. I've fixed the hook and added regression coverage on `develop`, so you can rebase onto the latest `develop` and try the checks again when you get the chance.

dxwil:
Done.

cjdsellers (approved):

Issue #4442, opened 2026-07-11
Enhancement: SimulatedExchange.purge_instrument() Rust-side + documented parameter-sweep pattern for long-window backtests


gtalknitin:
### Environment

- `nautilus_trader` version: `1.231-dev` (git SHA `8b63cc915e7910b9a0cad9a52160911e148f2c92` on `develop` branch)
- Python: 3.12 (base image `python:3.12-slim` via `nautilus-base` custom Docker image)
- OS: Docker on Ubuntu 26.04 host (AWS r7i.large, 16 GB RAM, x86_64)
- Backtest workload: NIFTY options (weekly + monthly expiries), multi-year windows

### Problem

We are running event-driven options backtests on Nautilus and hitting hard memory ceilings during **long-window runs (18+ months)** despite following canonical Nautilus patterns and being on 1.231-dev.

**Observed failure mode**:
- 20-month straddle-NIFTY backtest (window 2024-01-01 → 2025-08-28) hits **>10 GB RSS during the `LazyLoadBacktestNode` preparation / instrument-catalog scan phase** — BEFORE the first bar is processed
- Container cgroup OOM at 12 GB limit; host at 15+ GB / 15.7 GB total RAM
- Reproduced 3 times: 2026-07-10 23:04 UTC (crash — required AWS Console EC2 Reboot), 2026-07-11 07:44 IST (voluntary cancel at 96.64% cap to prevent second crash), 2026-07-11 15:11 IST (voluntary cancel at 65% mem / t+19min still in init phase — mem-watchdog data confirms linear ~300 MiB/min growth, no plateau, no `run_id` assigned, no sim-date crossed)
- Backtest never reached compute phase — problem is instrument metadata + engine state pinning, not runtime accumulation

**Memory ownership analysis** (per Nautilus 1.231-dev source inspection at `8b63cc9`):

| Memory surface | Nautilus type / location | Bytes/instrument (approx) | Primitive to release? |
|---|---|---|---|
| Cache dicts (11 collections) | `Cache._instruments`, `_quote_ticks`, `_trade_ticks`, `_bars`, etc. (cache.pyx) | ~50-100 KB | ✅ `Cache.purge_instrument(id)` (cache.pyx:1104, added 1.230+) |
| Cache indices | `_index_instrument_*` (cache.pyx:1180-1191) | ~10-20 KB | ✅ same call clears indices |
[...]

PR #4503, opened 2026-07-19
Add SimulatedExchange.purge_instrument for long-window backtest memory release

gtalknitin:
Refs #4442

## Summary

Adds `SimulatedExchange.purge_instrument(instrument_id)` on both the Cython production surface (`nautilus_trader/backtest/engine.pyx`) and the parallel Rust venue in `crates/backtest/src/exchange.rs`. Mirrors the existing `Cache.purge_instrument` shape and semantics so an actor implementing an expiry-rollover lifecycle can release both cache-owned and venue-owned per-instrument state in the same event.

This is the primitive requested in issue #4442 (part 1 of the three-part enhancement — the documentation + CI RSS regression parts remain open and can be addressed separately). It is complementary to `Cache.purge_instrument` and does **not** modify any existing purge logic.

## Motivation

Long-window multi-instrument backtests (options universes with weekly / monthly expiry rollover in particular) accumulate venue-owned state in the `SimulatedExchange.instruments` and `_matching_engines` maps for the full run. Framework code has no existing primitive to release that state, so an actor that already invokes `Cache.purge_instrument` on rollover cannot reach the venue-owned side.

A representative failure mode: an options backtest running 18–24 months against a ~22k-strike universe hits container memory ceilings (>10 GiB RSS) before the first bar processes, dominated by accumulated per-instrument matching-engine state. Adding this primitive lets long-window backtests bound memory at `O(active_instruments)` instead of `O(all_touched_instruments_over_window)`, which is essential for regime-change / walk-forward analysis over multi-year windows on standard cloud VM tiers.

## Design

Both implementations follow the same shape as `Cache.purge_instrument`:

1. Return quietly with a warning if the instrument is not registered on the venue.
2. Refuse and log a warning if any associated order is open, or any associated position is open (via `Cache.has_orders_open` / `has_positions_open` on the venue's wired cache). This mirrors the guard on the cache-side purge and stays defensive by default — callers whose lifecycle logic has already invalidated order/position state cache-side can safely call this after; callers who have not will get a no-op and a log message rather than a torn matching engine.
3. Drop the instrument entry, drop the matching engine, and drop any per-instrument settlement / funding state.
4. On the Cython side, recompute `_has_next_instrument_expiration` and `_next_instrument_expiration_ns` from the remaining engines so the venue does not continue timing off a purged instrument's expiration.
5. On the Rust side, use `IndexMap::shift_remove` on `matching_engines` to preserve the registration-order invariant covered by `test_matching_engine_iteration_order_is_stable_across_rebuilds`.

Active data or execution engine subscriptions are intentionally not touched here — those belong to the data and execution engines, matching the boundary drawn by `Cache.purge_instrument`.

## Usage example

```python
class MemoryHousekeepingActor(Actor):
    def on_weekly_expiry(self, expired_instrument_ids: list[InstrumentId]) -> None:
        for iid in expired_instrument_ids:
            self.cache.purge_instrument(iid)  # existing primitive (cache-side)
            self.exchange.purge_instrument(iid)  # new primitive (venue-side)
```

The guard means a mistimed call is a no-op with a log message rather than a silent tear-down of active state.

## Tests

Rust integration tests (`crates/backtest/tests/exchange.rs`):
- `test_purge_instrument_removes_entry_and_matching_engine` — happy path
- `test_purge_instrument_when_not_present_is_noop` — noop when not registered
- `test_purge_instrument_preserves_matching_engine_registration_order` — guards the `IndexMap` iteration-order invariant when removing a middle instrument from a 3-instrument venue
- `test_purge_instrument_refuses_when_orders_open` — refusal guard for open orders

Python unit tests (`tests/unit_tests/backtest/test_engine.py::TestBacktestEnginePurgeInstrument`):
- `test_purge_instrument_removes_from_venue_state` — asserts the instrument leaves both `instruments` and `_matching_engines` while a sibling instrument on the same venue remains
- `test_purge_instrument_when_not_present_is_noop`

Ran locally on macOS aarch64 with rustc 1.97.1:

```
cargo test -p nautilus-backtest --test exchange
# 56 passed (52 pre-existing + 4 new)

cargo clippy -p nautilus-backtest --tests --no-deps
# clean
```

## Not in scope for this PR

- The docs addition (`docs/how_to/multi_config_backtest.md`) and CI-gated RSS acceptance test from parts 2 and 3 of issue #4442 are left for follow-ups — they are independent of this primitive and touching them here would over-scope the review.
- `Cache.purge_instrument` and `Cache.purge_instrument_skip_order_guard` are not modified.
- No changes to `BacktestNode.dispose()` semantics.

## Willing to iterate

Happy to adjust the guard shape, method placement, error semantics, or docstring per maintainer preference. Also willing to open the follow-up docs / CI PRs referenced above once this primitive lands.

### DIFF
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -432,6 +432,68 @@ impl SimulatedExchange {
         Ok(())
     }
 
+    /// Purges the instrument with the given `instrument_id` from the venue (if found).
+    ///
+    /// Drops the entry from both `instruments` and the `matching_engines` map,
+    /// releasing the per-instrument order matching engine and any book state it owned.
+    /// Callers should typically also purge the corresponding cache-side state via
+    /// [`Cache::purge_instrument`] in the same lifecycle event so that both venue-owned
+    /// and cache-owned per-instrument state are released together.
+    ///
+    /// For safety, an instrument is prevented from being purged while any associated
+    /// order is open or any associated position is open on this venue. When such state
+    /// exists the call is a no-op and a warning is logged, mirroring the behavior of
+    /// [`Cache::purge_instrument`].
+    ///
+    /// Active data or execution engine subscriptions are not touched here; those belong
+    /// to the data and execution engines.
+    ///
+    /// # Warning
+    ///
+    /// Intended for actors and strategies that have their own lifecycle logic for
+    /// deciding when an instrument is no longer needed (for example on weekly or monthly
+    /// option expiry rollover during a long-window backtest). Purging an instrument that
+    /// any other actor, strategy, or engine still relies on may cause incorrect behavior
+    /// (missing instrument lookups, matching-engine misses on subsequent market data).
+    /// The caller is responsible for ensuring the instrument is no longer in use before
+    /// purging.
+    pub fn purge_instrument(&mut self, instrument_id: InstrumentId) {
+        let found = self.instruments.contains_key(&instrument_id)
+            || self.matching_engines.contains_key(&instrument_id);
+
+        if !found {
+            log::warn!("Instrument {instrument_id} not found when purging");
+            return;
+        }
+
+        let cache = self.cache.borrow();
+        if cache.has_orders_open(None, Some(&instrument_id), None, None, None) {
+            log::warn!(
+                "Instrument {instrument_id} has open orders when purging, skipping purge"
+            );
+            return;
+        }
+
+        if cache.has_positions_open(None, Some(&instrument_id), None, None, None) {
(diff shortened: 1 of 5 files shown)


cjdsellers:
Hi @gtalknitin,

Thanks for the PR and the work you put into it. As noted in #4442, this changes the Cython implementation, and v1 work is now out of scope. The CLA is also still unsigned, and the implementation needs further work.

I'll close this PR and track the v2 implementation separately in a ticket owned by the maintainers. So no further changes are needed here.

PR #4504, opened 2026-07-19
Normalize typographic Unicode in comments and docs

folknor:
Comment and documentation prose across the tree carries typographic Unicode - em and en dashes and curly quotes - which breaks grep and diff hygiene and steadily leaks into new patches by copy-adjacency; this normalizes those characters to ASCII.

@cjdsellers told me in an earlier PR that these kinds of gremlins were unwanted in the codebase, so I figured I'd help clean it up.

## Scope and mechanism

Em/en dashes become hyphens and curly quotes become straight quotes in comments, doc comments, markdown, and issue-template prose. After this change, no em/en dash or typographic quote remains in tracked `.rs`/`.md`/`.toml`/`.sh` sources outside the deliberate exclusions listed below. Three glyph cleanups ride along: two developer-guide examples that used emoji cross/check marks (`docs/developer_guide/coding_standards.md`, `releases.md`) now use ASCII `Bad:`/`Good:` labels, and one test comment in `indicators/average/dema.rs` that consisted of an orphaned non-ASCII numbered marker is removed. Where an em dash sat directly between words, spacing is adjusted so the result reads as prose (e.g. the `InstrumentClass::Warrant` doc). One fuzz-target module doc (`fuzz_compute_tx_hash`) is reflowed while its em dash is normalized - which also resolves `clippy::doc_lazy_continuation` under `--all-features` - and its stale 80-byte input claim is corrected to the actual 84 (`INPUT_LEN` and the field layout both say 84; there is no tail padding).

Five log/assertion format strings are included in the normalization (two `log::warn!` messages in the Kraken L3 parser, one in `common/src/clients/mod.rs`, one `log::info!` in a Polymarket example, one test assertion message in the MACD indicator): message text only - no serialized value, protocol string, identifier, or code token changes anywhere in the diff.

## Deliberately preserved

The U+2713/U+2717 support-table and log-output glyphs in README, docs, and scripts; the non-ASCII test fixtures in `core/src/correctness.rs` (they exercise `check_valid_string_ascii` rejection); U+2011 non-breaking hyphens in docs tables (required by `check_docs_conventions.sh`); `RELEASES.md` and `CLA.md` untouched; vendored `patches/` untouched; `.pre-commit-hooks/` scripts untouched (they carry their own glyph output).

## Generated headers

`nautilus_trader/core/includes/core.h`, `includes/model.h`, `core/rust/core.pxd`, and `core/rust/model.pxd` are regenerated so the emitted doc comments match the normalized Rust sources. The diffs are comment-only: declaration tokens, enum values, and the high-precision mode lines are byte-identical.

## Testing

Every hunk is inside a comment, doc comment, markdown/template prose, a generated header's comment block, or one of the five message strings listed above - the change is invisible to compilation except through doc comments. The `check_hidden_chars`, `check_non_latin_text`, and `check_docs_conventions` hooks are clean, workspace clippy reports no diagnostics that are not already present on develop, and the workspace test sweeps (default, ffi, live feature shapes) pass unchanged with this change applied.


folknor:
Just to note, I have an app that rewrites these for me, so adding/removing gremlins from the scope and redoing the PR takes me 2 minutes.

cjdsellers (approved):
Thank you, @folknor. Great cleanup.

Tomorrow I'll strengthen the pre-commit so that these can't creep back in.

PR #4505, opened 2026-07-19
Fix clippy lints that fire outside the pinned-stable gate

folknor:
Fixes the clippy lints that never fire on CI's pinned-stable configuration but do on current nightly and under `--all-features` / feature-unified workspace builds, where they fail any run carrying `-D warnings`.

## Fuzz-target lints (`--all-features` builds)

Two fuzz targets only compile with fuzzing features enabled, so the workspace gate never sees them. `fuzz_decimal_decode` passed `serde_json::Value` by value into `decode_value` (`clippy::needless_pass_by_value`); it now borrows. `fuzz_nonce_sequence` used `chunks_exact(CHUNK_LEN)` with a constant chunk size (`clippy::chunks_exact_to_as_chunks`); it now uses `as_chunks::<CHUNK_LEN>()`.

## Model types test lints (local nightly)

In `types/fixed.rs`, 15 `#[case]` literals gain digit separators (`clippy::unreadable_literal`). Three `as` casts in property-test code become `From`/`u128::from` conversions (`clippy::cast_lossless`): two in `quantity.rs` under `cfg(not(feature = "high-precision"))` - a path the default high-precision build never compiles - and one in `money.rs` under `cfg(not(feature = "defi"))`. In `price.rs`, `#[expect(clippy::useless_conversion)]` becomes `#[allow]` with the reason expanded: whether the conversion is useless is width-dependent, so the expectation is unfulfilled (and itself denied) in exactly one of the two configurations.

## Feature-unification conversions in event_store

`markers/extractor.rs`'s `price_raw_at_precision` / `quantity_raw_at_precision` convert `PriceRaw` / `QuantityRaw` into `i128` / `u128`. Under workspace feature unification the raw types are already 128-bit and `clippy::useless_conversion` fires; in a standalone 64-bit build the conversion is required. Reasoned `#[allow]`s document the width dependence instead of removing conversions standalone builds need.

## Testing

Mechanical lint fixes with no intended behavior change; every touched line is test, fuzz, or instrumentation code. `cargo clippy --all-targets --all-features` on `nautilus-derive` was verified to FAIL against the unfixed targets with exactly the named lints (`needless_pass_by_value`, `chunks_exact_to_as_chunks`) and is clean with the fixes. The model and event_store lints are the long-standing local-nightly set; with the fixes, nightly `cargo clippy --all-targets` on `nautilus-model` and `nautilus-event-store` reports zero diagnostics on this branch, and the model types and event_store test suites pass unchanged under the workspace's default, ffi, and live feature shapes.


cjdsellers (approved):
Hey @folknor

Thanks for cleaning these up. I checked the patch against the latest `develop` and ran the targeted nightly Clippy paths for the all-feature fuzz targets, default-width model, and feature-unified event store. Everything passes and this looks good to merge.

folknor:
Thank you, I'll set up commit signing, sign the commit, and merge it a bit later - I am working on a followup to PR 4496 right now.

PR #4506, opened 2026-07-19
Fix three dead pre-commit checks and the violations they hid

folknor:
Three pre-commit checks have been silently dead - each disabled by a small mechanical defect in its own script - and the tree accumulated violations behind two of them; this fixes the scripts and the violations they hid.

## DST rule 5 guards a moved file path (check_dst_conventions.sh)

`RULE5_FILES` listed `crates/live/src/manager.rs`, but the reconciliation manager lives at `crates/live/src/execution/manager.rs`, and the `[[ -f "$rule5_file" ]] || continue` silently skipped missing entries - so the AHashMap/AHashSet determinism guard on the manager has not run since the file moved. The path is corrected, and a missing `RULE5_FILES` entry is now reported as a violation, so a future file move cannot silently disarm the rule again. The manager itself is currently clean.

## The Python error-variable filter never fires (check_error_conventions.sh)

The loop reads rg output with `IFS=: read -r file line_num line_content`, and bash strips the trailing delimiter from the last field - `except X as exc:` arrives in `line_content` as `except X as exc`, while the filter regex required the trailing colon. No line has ever matched, so the Python half of the check has never flagged anything. The colon requirement is dropped from the extraction regex (the rg pattern feeding the loop already anchors the full `as xxx:` shape). The five `as exc:` handlers this hid are renamed to `as e:` (with their body uses): `examples/live/interactive_brokers_v2/option_greeks.py`, `python/examples/polymarket/updown_smoke_tester.py`, `python/tests/strategies/backtest_surface.py`, `python/tests/unit/analysis/test_tearsheet.py`, `python/tests/unit/common/actor.py`.

## The adapter-dependency-group extraction is always empty (check_cargo_conventions.sh)

Check 9's awk resets `in_section` on any `# -----` line, and the manifest's section banners are dash-line / title / dash-line - so the banner's own closing dash cleared the flag one line after the header set it, and the extracted adapter-dependency list was always empty. The awk now tracks the banner state so only a subsequent dash line closes the section. The two violations this hid are fixed: `itoa` (used by `nautilus-common`) and `arc-swap` (used by `nautilus-core`) move from the "Adapter dependencies" group to Core dependencies, placed alphabetically.

## Testing

Each revived check was differentially verified. With the script fixes applied but the tree fixes reverted, `check_error_conventions.sh` fails on exactly the five `as exc:` sites, and `check_cargo_conventions.sh` fails on exactly `itoa` and `arc-swap`. `check_dst_conventions.sh` fails when the manager file is moved aside (the new missing-file arm) and when an `AHashMap` alias is temporarily introduced into it (the revived scan itself). With both halves applied all three hooks pass, as does the rest of the Rust-relevant hook battery. The dependency regroup changes no versions or features (comment-group membership only; `Cargo.lock` is untouched), and the five Python renames are mechanical and compile-checked.


cjdsellers (dismissed):
Hi @folknor

Thanks for finding and fixing these, changes look good.

Just a merge conflict on the `Cargo.toml` to resolve when you get the chance.

folknor:
Thank you. All future commits from me will be signed now.

cjdsellers (approved):

Issue #4508, opened 2026-07-19
BitMEX: sparse order updates lose explicit-null vs missing-field semantics on merge


4px4d9cdby-star:
# BitMEX: sparse order updates lose explicit-null vs missing-field semantics on merge

### Problem

BitMEX `order:update` rows are sparse table updates. A legal update can contain only the table key and the fields that changed. Handling each update as a standalone row loses the symbol and client-order identity needed for routing, and a terminal update such as `{"orderID":"…","ordStatus":"Canceled"}` may never produce the terminal lifecycle event.

A correct merge must also distinguish a missing field from an explicit JSON `null`:

- missing `price` or `text`: retain the cached value;
- explicit `"price": null` or `"text": null`: clear the cached value;
- present non-null value: replace the cached value.

Finally, terminal rows must be evicted before any later processing can return early. Inserting a full terminal row and evicting only after instrument lookup leaks cache entries whenever that lookup misses.

### Evidence

The pre-fix Python WebSocket route decoded updates independently, required update-local symbol/client identity, and emitted only a non-terminal amendment event. It did not merge against the preceding partial/insert row or route a merged terminal row through the lifecycle parser.

Focused fixture replays establish three separate contracts:

1. A full New row followed by a key-only Canceled update must emit the accepted and canceled lifecycle transitions.
2. Omission must preserve cached nullable values, while explicit null must clear them.
3. A terminal full row must not remain in the order-row cache, including when later processing would return early.

The cache should be scoped to one WebSocket task, keyed by venue order UUID, cleared on a fresh `partial` snapshot and reconnection, populated only by non-terminal full rows, and updated by sparse rows. A sparse cache miss should warn and drop rather than fabricate identity or symbol.

### Extracted patch diff

Source patch: `0009-P9-merge-sparse-BitMEX-order-updates.patch` (trimmed to the upstream-relevant cache/merge path):

```diff
+let mut order_row_cache: AHashMap<Uuid, BitmexOrderMsg> = AHashMap::new();

-BitmexTableMessage::Order { data, .. } => {
+BitmexTableMessage::Order { action, data } => {
+    if action == BitmexAction::Partial {
+        order_row_cache.clear();
+    }
     handle_order_messages(
         data,
         &instruments,
         order_type_cache,
         order_symbol_cache,
+        order_row_cache,
         dispatch_state,
         trader_id,
         account_id,
@@
-OrderData::Update(msg) => {
+OrderData::Update(mut update) => {
[...]

PR #4511, opened 2026-07-19
Fix BitMEX sparse order update merge semantics

4px4d9cdby-star:
Implements the fix proposed in #4508.

BitMEX WebSocket `order` partial/update rows are sparse: absent keys mean "unchanged", but a present `null` was previously merged as a value overwrite, evicting known state (e.g. `price: null` on terminal updates clearing the resting price). This change merges only non-null incoming fields over retained state, matching the venue's documented update semantics.

Tested with unit coverage over the merge path (absent vs explicit-null vs value keys); verified against recorded BitMEX stream captures on our fork.

### DIFF
--- a/crates/adapters/bitmex/src/python/websocket.rs
+++ b/crates/adapters/bitmex/src/python/websocket.rs
@@ -50,6 +50,7 @@ use nautilus_model::{
 use nautilus_network::websocket::TransportBackend;
 use pyo3::{conversion::IntoPyObjectExt, prelude::*};
 use ustr::Ustr;
+use uuid::Uuid;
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @4px4d9cdby-star

Thanks for reporting this and putting together a patch. The CLA still needs to be signed before we can accept contributed code, and since this was a small fix it was quicker for me to implement it directly across both the Rust execution client and Python WebSocket wrapper. I've credited you for the report in the release notes, so I'm closing this PR in favor of the maintainer implementation once it lands shortly.

Issue #4509, opened 2026-07-19
BitMEX: binary WebSocket frames carrying UTF-8 JSON are dropped instead of routed to the JSON handler


4px4d9cdby-star:
# BitMEX: binary WebSocket frames carrying UTF-8 JSON are dropped instead of routed to the JSON handler

### Problem

The BitMEX WebSocket feed handler parses JSON carried by `Message::Text`, but the binary branch only logs and discards its payload. The generic transport delivers both text and binary data messages to this handler, so a valid UTF-8 JSON table frame carried as `Message::Binary` is silently lost before table routing.

The two WebSocket data opcodes should share the same JSON/control-message parser after UTF-8 validation. Invalid UTF-8 should be logged and ignored without panicking, and Ping/Pong/Close behavior should remain unchanged.

### Evidence

A direct regression frame containing:

```json
{"table":"order","action":"partial","data":[]}
```

parses to an order-table message when supplied as text. Before the change, the same bytes supplied as a binary frame return no routed message. After the change, both opcodes produce the same `BitmexWsFrame::Table` variant.

This is a protocol-hardening defect established by deterministic fixture replay. It is not a claim that binary opcode delivery caused any particular production incident: in one instrumented window all captured data frames were text, and an earlier window did not retain opcodes. No conclusion about the behavior of an uninstrumented pinned runtime follows from this patch.

### Extracted patch diff

Source patch: `0010-P10-parse-UTF-8-BitMEX-binary-frames.patch` (trimmed to the shared parse path and regression):

```diff
-    fn parse_raw_message(msg: Message) -> Option<BitmexWsFrame> {
-        match msg {
-            Message::Text(text) => {
-                if text == RECONNECTED {
-                    log::info!("Received WebSocket reconnected signal");
-                    return Some(BitmexWsFrame::Reconnected);
-                }
+    fn parse_text_payload(text: &str) -> Option<BitmexWsFrame> {
+        if text == RECONNECTED {
+            log::info!("Received WebSocket reconnected signal");
+            return Some(BitmexWsFrame::Reconnected);
+        }

-                log::trace!("Raw websocket message: {text}");
+        log::trace!("Raw websocket message: {text}");
+        if Self::is_heartbeat_message(text) {
+            log::trace!("Ignoring heartbeat control message: {text}");
+            return None;
+        }

-                if Self::is_heartbeat_message(&text) {
-                    log::trace!("Ignoring heartbeat control message: {text}");
-                    return None;
[...]

PR #4512, opened 2026-07-19
Fix BitMEX binary WebSocket frame routing

4px4d9cdby-star:
Implements the fix proposed in #4509.

The BitMEX WebSocket client assumed text frames; binary frames (observed under gzip/permessage-deflate negotiation and via certain proxies) were dropped before JSON decode, silently losing execution stream rows. Binary frames are now decoded through the same routing path as text frames.

Tested with an RFC 6455 frame observer harness (text and binary variants of the same payload route identically) on our fork.

### DIFF
--- a/crates/adapters/bitmex/src/websocket/handler.rs
+++ b/crates/adapters/bitmex/src/websocket/handler.rs
@@ -228,59 +228,63 @@ impl BitmexWsFeedHandler {
         }
     }
 
-    fn parse_raw_message(msg: Message) -> Option<BitmexWsFrame> {
-        match msg {
-            Message::Text(text) => {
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @4px4d9cdby-star

Thanks for reporting this and putting together a patch. The CLA still needs to be signed and the `pre-commit` check is failing, so we can't merge this as-is. Since this was a small, focused change, it was quicker for me to implement it directly. I've kept the fix scoped to UTF-8 JSON carried in binary WebSocket frames and credited you for the report in the release notes, so I'm closing this PR in favor of the maintainer implementation once it lands.

Issue #4510, opened 2026-07-19
Execution engine: order lifecycle events are DEBUG-only; proposal for explicit observability


4px4d9cdby-star:
# Execution engine: order lifecycle events are DEBUG-only; proposal for explicit observability

### Problem

`ExecutionEngine._handle_event` logs a received order event only when `ExecEngineConfig.debug` is enabled, and then emits the line at DEBUG. An otherwise healthy node running at INFO therefore has no execution-engine log record for normal lifecycle transitions such as Accepted, Canceled, or Filled.

That behavior is easy to misread operationally: absence of an event name in INFO logs can be mistaken for absence of delivery or processing. It also creates a two-part enablement requirement (`debug=True` and a logger threshold that admits DEBUG) that is not evident from the execution configuration documentation.

The event bus and cache remain the authoritative programmatic state surfaces. The issue is that the human observation channel does not state its filtering contract and offers no narrow way to expose lifecycle transitions without enabling all execution-engine debug logging.

### Evidence

The current handler begins with:

```cython
cpdef void _handle_event(self, OrderEvent event):
    if self.debug:
        self._log.debug(f"{RECV}{EVT} {event}", LogColor.MAGENTA)

    self.event_count += 1
```

`ExecEngineConfig.debug` defaults to `False` and is documented only as providing "extra debug logging." In a controlled diagnosis, INFO logs contained no Accepted event lines even though an independent event-stream subscriber received and correlated the lifecycle events. Enabling transport and event-stream observation showed that the missing log lines were a filtered observation channel, not proof of missing event processing.

This finding does not establish whether any separately pinned runtime did or did not have a WebSocket push defect. It establishes only the logging behavior above and the risk of using INFO-log absence as a delivery oracle.

### Proposal

Provide one explicit, documented order-event logging control rather than coupling lifecycle visibility to the engine-wide debug flag. Two acceptable designs are:

1. Add an `ExecEngineConfig.order_event_log_level` option whose default preserves DEBUG behavior, but which can be set to INFO without enabling every engine debug line.
2. If configurable levels are not desirable, promote only state-changing lifecycle transitions (`OrderAccepted`, `OrderRejected`, `OrderCanceled`, `OrderExpired`, `OrderDenied`, and `OrderFilled`) to INFO while retaining high-frequency amendment/update detail at DEBUG.

The configurable-level design has the smaller compatibility and volume impact. Its documentation should state:

[...]

PR #4513, opened 2026-07-19
Make execution order-event log level configurable

4px4d9cdby-star:
Implements the proposal in #4510.

High-frequency order lifecycle events (submit/accept/cancel churn from quoting strategies) currently log at a fixed level, which either floods INFO or forces global suppression. This adds an execution-engine config knob for the order-event log level, defaulting to current behavior (no change unless set).

Tested with config round-trip and level-emission assertions on our fork.

### DIFF
--- a/nautilus_trader/execution/config.py
+++ b/nautilus_trader/execution/config.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 from typing import Any
+from typing import Literal
 
 import msgspec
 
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @4px4d9cdby-star

Thanks for the detailed issue and the PR.

We're keeping the v1 (Cython) core in maintenance while we finish the transition to the v2 Rust + PyO3 core, so I'm not taking new config surface into v1 at this stage and will close this one.

For what you're after now, `log_component_levels` should cover the flood-vs-suppression tradeoff without a new knob: it sets a per-component minimum level, so you can dial a noisy strategy down to `WARNING` while leaving everything else at `INFO`, or pair `ExecEngineConfig(debug=True)` with `log_component_levels={"ExecEngine": "DEBUG"}` to surface the exec-engine lifecycle lines specifically. `log_components_only` is the complementary switch if you'd rather allowlist just a few components.

Once we're on v2 I'm open to additional per-event log-level control if the component filter isn't granular enough. Appreciate the write-up.

PR #4516, opened 2026-07-20
Prevent time-event callback destructors from aborting on thread-local teardown

folknor:
A follow-up to the time-event callback registry added in #4496, whose thread-local callback map is read infallibly from two `Drop`-reachable paths.

## Infallible registry access from a destructor aborts on thread-local teardown

`TimeEventCallbackToken::remove_on_owner_thread` and `TimeEventCallbackLease::drop` (both `crates/common/src/runner.rs`) reach `TIME_EVENT_CALLBACKS.with(...)`, which panics with `AccessError` once that thread-local has been destroyed. Both are reachable from a destructor: `LiveTimer::drop` -> `close_registered_callback` -> `TimeEventCallbackToken::close` -> `remove_on_owner_thread`, and the final-lease arm of `TimeEventCallbackLease::drop`. The `owner == thread::current().id()` guard does not defend against this - the destructor runs on the owner thread, so the guard passes and the call then touches an already-freed thread-local. A panic inside a TLS destructor aborts the process.

The window is a `LiveTimer` (or a still-queued `TimeEventCallbackLease`) outliving the registry thread-local at thread exit. It is observable when a node is dropped without a graceful stop: a `RustLocal` equity/aggregation timer armed on a `LiveClock` is kept alive by a strong `Rc` held in `MESSAGE_BUS`-owned handlers, so the clock - and its timers - drop during thread teardown, after `TIME_EVENT_CALLBACKS` has been destroyed. The failure surfaces as `cannot access a Thread Local Storage value during or after destruction: AccessError` -> `thread local panicked on drop` -> SIGABRT.

Both sites now use `try_with` and discard the `AccessError`: if the thread-local is already gone the entry cannot be removed and need not be, since the map is being torn down. The removed entry is returned out of the closure so it drops after the `RefCell` borrow is released, and the teardown path does not log (the logging thread-local may itself be in teardown). On a live owner thread `try_with` resolves the same instance and performs the identical removal, so the non-teardown path is unchanged; `RefCell` borrow panics still propagate.

## Testing

Two `#[rstest]` cases in `crates/common/src/runner.rs`, each reproducing the teardown ordering by initializing a holder thread-local before `TIME_EVENT_CALLBACKS`, so the registry is torn down first and the held value's destructor runs against gone TLS:

- `test_final_lease_drop_survives_registry_tls_teardown` pins the `TimeEventCallbackLease::drop` path: a final, closed lease is parked in the holder and drops after the registry.
- `test_owner_close_survives_registry_tls_teardown` pins the `remove_on_owner_thread` path via a `CloseOnDrop` wrapper that calls `token.close()` from its own `Drop`, standing in for `LiveTimer::drop` without dragging in the global runtime, a task, or a sender.

Each runs the drop inside a spawned thread and asserts the thread joins - an abort would take the whole test binary down with SIGABRT, not fail an assertion. Both were verified to abort with `AccessError` / SIGABRT against the unfixed `.with` code across the default, ffi, and live sweeps, and to pass with `try_with`. No wall-clock waits.

The reproduction relies on the platform's thread-local destruction order (holder initialized before the registry, destroyed after it), which matches the reported Linux abort; the fix itself makes the destructor order-independent.

## Restore the blank line before control-flow constructs

The repository style requires a blank line above a control-flow construct (`if`, `match`, ...) unless the preceding line opens a block, is a comment or attribute, or shares an identifier with the condition. Thirty-one sites across twenty files had drifted from it. Each gets a single blank line inserted - whitespace-only, no behavior change, no tests - bundled here so the branch is clean under the style gate.

### DIFF
--- a/crates/adapters/architect_ax/src/data.rs
+++ b/crates/adapters/architect_ax/src/data.rs
@@ -1183,6 +1183,7 @@ fn handle_md_message(
             let quotes_subscribed = sdt_snap
                 .get(symbol.as_str())
                 .is_some_and(|entry| entry.quotes);
+
             if quotes_subscribed {
                 match parse_book_l2_quote(&book, instrument, ts_init()) {
                     Ok(quote) => {
@@ -1214,6 +1215,7 @@ fn handle_md_message(
             let quotes_subscribed = sdt_snap
                 .get(symbol.as_str())
                 .is_some_and(|entry| entry.quotes);
+
             if quotes_subscribed {
                 match parse_book_l3_quote(&book, instrument, ts_init()) {
                     Ok(quote) => {
@@ -1238,6 +1240,7 @@ fn handle_md_message(
             let mark_prices_subscribed = sdt_snap
                 .get(ticker.s.as_str())
                 .is_some_and(|e| e.mark_prices);
+
             if mark_prices_subscribed && let Some(mark_price) = ticker.m {
                 match Price::from_decimal_dp(mark_price, price_precision) {
                     Ok(price) => {
@@ -1254,6 +1257,7 @@ fn handle_md_message(
                 let status_subscribed = sdt_snap
                     .get(ticker.s.as_str())
                     .is_some_and(|e| e.instrument_status);
+
                 if status_subscribed {
                     let prev = instrument_states.insert(ticker.s, state);
                     if prev != Some(state) {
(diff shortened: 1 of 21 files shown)


cjdsellers (approved):
Hey @folknor

Thanks for tracking this down and adding focused regression tests. The `try_with` changes look correct, and this is ready to merge as-is. I'll handle the release-note entry and consider a focused re-entrant-removal regression test as follow-ups.

The formatting changes are also appreciated and follow the codebase convention. It looks like the Rust formatting pre-commit hook scanned the full tree once this PR touched `runner.rs`; for future changes, we'd prefer repo-wide formatting cleanup in a separate PR so the behavioral change stays clear.

PR #4517, opened 2026-07-20
Fix stale live position reconciliation snapshots

folknor:
A follow-up to the position-reconciliation tolerance fix in #4490, on the same live reconciliation path.

## Position reconciliation acted on a prepare-time snapshot

The continuous position-consistency check captures open positions at prepare time and then reconciles them against venue reports that arrive after an asynchronous round-trip. `start_position_report_check` calls `open_positions_by_key_for_reconciliation`, cloning each open `Position` into `PositionReportCheck.positions_by_key`, and only then awaits `request_position_reports` (sequential over clients). When that future resolves, `reconcile_position_reports` -> `check_position_discrepancy` derives both `cached_signed_qty` and `cached_avg_px` from the frozen vector. The runner keeps draining execution events during the await, so a fill that closes or reduces the position mutates the cache but not the snapshot; the only guard is the `position_local_activity` grace window (`position_check_threshold_ns`, `5s` by default, `0` disables), and retries do not debounce - the first discrepancy acts. A position that goes flat during the request against an empty venue response yields stale `cached_signed_qty=5, venue=0` and a synthetic market sell of `5`, which applied to the now-flat cache opens a real short. The venue-only loop is the same defect in reverse: reports for keys absent at prepare are reconciled against `&[]`, so a position that opened during the request is treated as absent and its full venue quantity is re-synthesized.

The mass-status netting path (`reconcile_position_report_netting`) already reads `cache.positions_open` live inside a single borrow with no intervening await, and is unaffected. The fix brings the continuous path into line: `check_position_discrepancy` re-reads the current open-`Position` vector for the key from the cache at reconciliation time and drives both quantity and average price from it, in the prepare-key loop and the venue-only loop alike, and retry pruning keys on the reconcile-time active set rather than the prepare-time keys. The prepared `client_coverage` snapshot is retained deliberately - it records which clients were actually queried, which is what lets an absent report be read as flat - so a newly opened key that carries no coverage entry is deferred to the next cycle rather than reconciled against a report that was never requested for it.

## Testing

- `test_position_check_rereads_position_closed_during_request` - a position long `5` at prepare time is closed to flat while the report request is pending; an empty venue response generates no reconciliation event and leaves no open position (the stale long would otherwise synthesize a sell that opens a real short).
- `test_position_check_rereads_position_opened_during_request` - a position opened during the request and matched by the venue report generates no event; the venue-only path defers the key that carried no prepare-time coverage.
- `test_position_check_uses_current_avg_px_after_request` - a residual discrepancy prices its synthetic fill from the current position (`3.0` at `avg_px_open` `3100`, reconciliation price `3300.00`), not the stale snapshot (`5.0` at `3000`).
- `test_position_check_preserves_retry_for_uncovered_position_opened_during_request` - a deferred uncovered key stays in the reconcile-time active set, so its retry counter is not pruned.

All tests use paused/virtual time (the in-flight cache mutation is driven by a virtual `dst::time::sleep`, no wall-clock waits). Each assertion was verified to FAIL against the unfixed snapshot-keyed code.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -151,11 +151,10 @@ pub(crate) struct OpenOrderReportCheck {
     pub start: Option<UnixNanos>,
 }
 
-/// Snapshot and command for one continuous position reconciliation check.
+/// Prepare-time client coverage and command for one continuous position reconciliation check.
 #[derive(Debug, Clone)]
 pub(crate) struct PositionReportCheck {
     pub command: GeneratePositionStatusReports,
-    pub positions_by_key: IndexMap<InstrumentAccountKey, Vec<Position>>,
     pub client_coverage: IndexMap<InstrumentAccountKey, ReportClientCoverage>,
 }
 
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the detailed fix and regression coverage. I'm happy to merge this as is. I want to investigate one additional multi-client timing case locally, but I don't think it should hold this PR up and I'll handle any follow-up separately.

cjdsellers:
Also, thanks for signing your commits so that the checks gate can go green now :ok_hand:

PR #4518, opened 2026-07-20
Commit reconciliation fill dedup after canonical apply, and bound its retention

folknor:
A follow-up to the live-reconciliation series (#4479, #4490, #4517), on the startup mass-status fill path.

## Mark a reconciliation fill deduplicated only after it applies

`ExecutionManager::create_order_fill` performed the dedup read-guard (`processed_fills.contains_key`) and the commit (`insert`) in one step, then returned the `OrderFilled` event for the caller to dispatch. All three mass-status callers - the orphan-fill path, `reconcile_order_with_fills`, and `handle_external_order` - collect their events and dispatch them later, after a global sort, through `exec_engine.process` / `project_reconciliation_fill`, none of which reports success. So the key was committed before the fill reached the canonical order cache. `reconcile_order_with_fills` projects the fill onto a `working` clone, and if `working.apply` fails it returns without emitting the event - but `processed_fills` already holds the key, so the next reconciliation cycle's `contains_key` short-circuits the fill to `None` and it is never retried, suppressed until the manager is dropped. A fill the canonical engine rejects (unknown order, duplicate, overfill, invalid transition) is lost the same way.

`create_order_fill` now read-guards against the committed `processed_fills` and a per-pass `pending_fill_keys` reservation set, builds the event, and inserts nothing. A real report fill is reserved in `pending_fill_keys` only when its event is actually queued (a failed `working.apply` returns first and reserves nothing), and the dedup key is committed to `processed_fills` inside the dispatch loop, per fill, immediately after that fill's synchronous `process` / `project_reconciliation_fill` returns - gated on a canonical-cache postcondition: the order resolved by `client_order_id` (venue-order-id fallback) must match the `FillKey` account and instrument and already carry the fill's `trade_id`. A trade id already present counts as applied and commits (correct replay); an absent one does not commit and the fill is left for the next cycle. Inferred top-up fills (`create_inferred_fill_for_qty`) keep their plain `order_events.push` and never enter the dedup map. The doc comment on `observe_execution_report` claimed the `Fill` report arm marks the fill processed; it only records activity, and is corrected to say so.

## Bound processed_fills retention to the reconciliation lookback

`processed_fills` was an `IndexMap` with no prune path - it grew for the lifetime of the manager, unlike the sibling `recent_fills_cache`, which `LiveNode` prunes each minute. It becomes a `RecencyMap<FillKey>` pruned in that same per-minute pass, at horizon `max(60s, reconciliation_lookback_mins)` - the startup mass-status lookback (`ExecutionManagerConfig::lookback_mins`), which is what actually bounds how far back a replayed fill report can arrive. When no finite startup lookback is configured (`reconciliation_lookback_mins: None`, the default), pruning is skipped and the map is retained for the manager's lifetime: without a finite lookback there is no horizon past which a replayed fill is provably stale, so no finite TTL is replay-safe. This bounds retention for deployments with a finite startup lookback; it deliberately does not bound the default `None` configuration, which keeps the replay guarantee at the cost of the memory bound (the map is non-persistent and resets on restart).

## Testing

All tests use `#[rstest]` with paused/virtual time and no wall-clock waits.

- `test_rejected_reconciliation_fill_is_retried` - a fill whose duplicate-trade projection fails on the `working` clone is not committed to `processed_fills`, so a later cycle re-emits it and it lands on the canonical order.
- `test_reconciliation_fill_dispatch_rejection_does_not_commit` - a fill whose event reaches the dispatch loop but is rejected by the canonical `process` (applied to an `Initialized` order) is not committed; after the order is promoted to `Accepted` the same fill applies on retry. This pins the canonical-cache commit gate itself, which the case above does not reach (it stops at the clone, before dispatch).
- `test_applied_reconciliation_fill_is_committed` - an applied fill commits its key and a duplicate report in a later cycle is deduped.
- `test_same_cycle_cross_order_fill_is_queued_once` - two orders reporting the same `FillKey` in one mass status emit a single fill; the pending set dedups across orders, which per-order projection alone does not.
- `test_inferred_fill_is_not_committed_as_reported_fill` - an inferred top-up fill does not enter `processed_fills`.
- `test_processed_fill_retention` - a committed key is pruned only past `max(60s, reconciliation_lookback_mins)`, and a `None` lookback retains it.

The two retry assertions - `test_rejected_reconciliation_fill_is_retried` and `test_reconciliation_fill_dispatch_rejection_does_not_commit` - were verified to fail against the unfixed premature-commit implementation; the remaining tests characterize the surrounding behavior.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -166,6 +166,24 @@ struct RetainedFillState {
     netting_lifecycle_starts: IndexMap<(AccountId, InstrumentId, StrategyId), UnixNanos>,
 }
 
+#[derive(Default)]
+struct ReconciliationFillQueue {
+    pending_fill_keys: IndexSet<FillKey>,
+    event_fill_keys: IndexMap<UUID4, FillKey>,
+}
+
+impl ReconciliationFillQueue {
+    fn push(&mut self, events: &mut Vec<OrderEventAny>, event: OrderEventAny, fill_key: FillKey) {
+        let OrderEventAny::Filled(fill) = &event else {
+            unreachable!("reported fills always create filled events");
+        };
+
+        self.pending_fill_keys.insert(fill_key);
+        self.event_fill_keys.insert(fill.event_id, fill_key);
+        events.push(event);
+    }
+}
+
 /// Configuration for execution manager.
 #[expect(
     clippy::struct_excessive_bools,
@@ -310,7 +328,7 @@ pub struct ExecutionManager {
     config: ExecutionManagerConfig,
     inflight_checks: IndexMap<ClientOrderId, InflightCheck>,
     external_order_claims: IndexMap<InstrumentId, StrategyId>,
-    processed_fills: IndexMap<FillKey, ClientOrderId>,
+    processed_fills: RecencyMap<FillKey>,
     recon_check_retries: IndexMap<ClientOrderId, u32>,
     order_query_recency: RecencyMap<ClientOrderId>,
     order_local_activity: RecencyMap<ClientOrderId>,
@@ -349,7 +367,7 @@ impl ExecutionManager {
             config,
             inflight_checks: IndexMap::new(),
             external_order_claims: IndexMap::new(),
-            processed_fills: IndexMap::new(),
+            processed_fills: RecencyMap::default(),
             recon_check_retries: IndexMap::new(),
             order_query_recency: RecencyMap::default(),
             order_local_activity: RecencyMap::default(),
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
Hey @folknor

Thanks for tracking this down and adding focused regression coverage. I'm happy to merge this as is.

I found one edge case where an existing canonical `DuplicateFill` returns before `fill_queue.push`, allowing the same fill key to reach another order on a later pass. Rather than hold this PR, I'll fix that locally after merge and update the retry regression and release notes.

PR #4519, opened 2026-07-20
Fix Interactive Brokers startup instrument loading

### LINKED ISSUE #4455: [Interactive Brokers] A failed load_contracts resolution at startup is silent and never retried — node runs "connected" but cannot trade the instrument
### Confirmation
- [x] I've searched existing issues and discussions to avoid duplicates. (Closest are #3752 — spread/combo instruments on restart — and #3237 — on-demand load on subscribe; neither covers a *failed `load_contracts` resolution at startup*.)
- [x] I've reviewed the source to confirm the behavior is not by design.
- [x] I've tested with a recent development wheel and can still reproduce it. *(Observed live on the `1.231.0a20260709` develop nightly; the code path below is confirmed unchanged on current `develop`. A controlled re-repro on the very latest wheel is a reasonable triage step.)*

### Expected Behavior
When the InteractiveBrokers adapter is configured with `load_contracts` (e.g. a `CONTFUT` with
`build_futures_chain=True`) and that resolution **fails at startup because of a transient connection
error** (a socket disconnect during the daily Gateway restart window, or the node starting before the
Gateway is ready), the adapter should not silently proceed into a state where the node is `RUNNING`
and IB-connected but has **no tradeable instrument in the cache**. It should either retry the
contract-details request, surface the failed required load (rather than marking the provider
`_loaded = True`), or re-attempt the load once the connection is restored.

### Actual Behavior
The failure is swallowed and never recovered:

1. `InstrumentProvider.initialize()` runs `load_all_async()` → `load_ids_with_return_async()` →
   `load_with_return_async()`. When `get_contract_details()` returns empty (the pending
   `reqContractDetails` future was failed with `ConnectionError('Socket disconnected.')`), it logs
   `Unable to resolve contract details for IBContract(...)` and `return None` — **no retry**
   (`providers.py:337-341`). The per-contract loop swallows it.
2. `initialize()` then sets **`self._loaded = True` unconditionally** (`providers.py:109`), so the
   node proceeds to `TradingNode: RUNNING` with an **empty/partial instrument cache** and only an
   ERROR line to show for it.
3. On reconnect, the client's `_resume()` → `_resubscribe_all()` re-subscribes **market data only**
   (it iterates `self._subscriptions`); it does **not** re-run the instrument load. So a
   `load_contracts` entry that failed at startup stays missing until a full process restart.

Net effect: the node looks healthy (IB link up, farms OK, `RUNNING`, reconciliation runs), but every
order for the configured instrument is silently dropped because the instrument can't be resolved
(`_resolve_instrument`/cache lookup returns nothing). On a production box this is silent
order-dropping with no self-recovery.

The runtime escape hatch works — calling `request_instruments(venue, params={"ib_contracts": [...]})`
re-runs `load_ids_with_return_async` and populates the cache successfully — but **nothing triggers it
automatically**, so the adapter can't recover on its own.

### Steps to Reproduce
1. Configure IB data + exec clients with `load_contracts` (e.g.
   `IBContract(secType="CONTFUT", symbol="MES", exchange="CME", build_futures_chain=True)`).
2. Start the node while the IB Gateway is **unreachable / mid-restart**, so the initial
   `reqContractDetails` fails (socket disconnect during the startup load). Equivalently, bounce the
   Gateway during `initialize()`.
3. Bring the Gateway up. The node reconnects — IB link restored, market-data farms `OK`,
   `TradingNode: RUNNING`.
4. Observe: the configured instrument is **never** in the cache; the log shows `Unable to resolve
   contract details for IBContract(...)`; `initialize()` set `_loaded = True`; subscribing to /
   ordering that instrument fails or is silently dropped. No recovery until a full process restart.

(Real-world trigger: our supervisor restarted the node during IBKR's daily Gateway restart; the
startup CONTFUT load raced the disconnecting socket. IB reconnected ~45 min later — link green — but
`MESU6.CME` was never loaded, so entries would have been silently dropped until we manually restarted.)

### Suggested Fix (any one helps; ideally 1+3)
1. **Retry `get_contract_details` on a transient `ConnectionError`** during the startup load (bounded
   retries / backoff), rather than a single attempt that `return None`s.
2. **Don't set `_loaded = True` when a requested `load_contracts` entry failed to resolve** — track /
   expose the partial-load failure so the caller (or the node) can react instead of proceeding
   silently.
3. **Re-attempt the instrument load on reconnect** for `load_contracts` entries that never loaded
   (the reconnect path already re-subscribes market data; extend it to reload missing required
   instruments).

### Environment
- `nautilus_trader`: observed on `1.231.0a20260709` (develop nightly); code path confirmed unchanged on current `develop`.
- Adapter: Interactive Brokers, headless `ghcr.io/gnzsnz/ib-gateway:stable` (IBC + Xvfb + socat), IBKR **paper**.
- Python 3.13, Linux (Docker, EC2).

### Code references (line numbers from the inspected source; equivalents on current `develop`)
- `adapters/interactive_brokers/providers.py:337-341` — "Unable to resolve contract details" + `return None` (no retry).
- `adapters/interactive_brokers/providers.py:109` — `self._loaded = True` set unconditionally after the load loop.
- `adapters/interactive_brokers/client/client.py` — `_resume()` → `_resubscribe_all()` re-subscribes market data only (no instrument reload) on reconnect.
- `adapters/interactive_brokers/client/client.py` (`_await_request`) — `ConnectionError` caught, request ended, returns the default (the empty result the load then sees).
- Runtime recovery that works but isn't auto-triggered: `data.py` `_request_instruments` with `params={"ib_contracts": [...]}` → `load_ids_with_return_async`.

mahimn01:
## Summary

- load every configured instrument ID and contract during startup in both the v1 and v2 IB adapters
- fail startup when any configured instrument cannot be resolved
- keep failed initialization retryable and preserve timeout and connection errors
- avoid registering or marking clients connected until startup loading succeeds
- add coverage for partial failures, retries, and concurrent initialization

Closes #4455

## Tests

- 496 passed with 2 existing flaky-mock skips in the Python IB integration suite
- 374 passed in the native Rust suite
- 386 passed with all Rust features enabled
- strict all-feature Clippy passed
- full pre-commit suite passed
- verified the v1 success and forced socket-disconnect paths against IB Gateway without placing or cancelling orders

The running Gateway reports protocol 203, while the current Rust client requires 213 or newer, so the v2 live connection could not be exercised against it.

### DIFF
--- a/crates/adapters/interactive_brokers/src/data/core.rs
+++ b/crates/adapters/interactive_brokers/src/data/core.rs
@@ -643,7 +643,7 @@ impl DataClient for InteractiveBrokersDataClient {
         .await
         .context("Failed to connect to IB Gateway/TWS")?;
 
-        let client = handle.as_arc();
+        let client = Arc::clone(handle.as_arc());
 
         tracing::info!(
             "Connected to IB Gateway/TWS at {}:{} (client_id: {})",
@@ -663,29 +663,12 @@ impl DataClient for InteractiveBrokersDataClient {
             tracing::info!("Set market data type to {:?}", self.config.market_data_type);
         }
 
-        let notice_client = Arc::clone(client);
-        self.ib_client = Some(handle);
-        self.is_connected.store(true, Ordering::Relaxed);
-
-        let data_farm_state = Arc::clone(&self.data_farm_state);
-        let cancellation_token = self.cancellation_token.child_token();
-        let clock = self.clock;
-
-        self.tasks.push(get_runtime().spawn(async move {
-            if let Err(e) =
-                monitor_data_farm_notices(notice_client, data_farm_state, clock, cancellation_token)
-                    .await
-            {
-                tracing::warn!("IB data farm notice monitor stopped: {e:?}");
-            }
-        }));
-
         // Initialize provider and load instruments from cache/config if configured
         tracing::debug!("Initializing IB data instrument provider");
 
         if let Err(e) = self
             .instrument_provider
-            .initialize_with_client(self.ib_client.as_ref().unwrap().as_arc().as_ref())
+            .initialize_with_client(client.as_ref())
             .await
         {
             if !self.config.instrument_provider.load_ids.is_empty()
@@ -696,6 +679,21 @@ impl DataClient for InteractiveBrokersDataClient {
 
             tracing::warn!("Failed to load instruments on startup: {}", e);
         }
+
+        self.ib_client = Some(handle);
+
+        let data_farm_state = Arc::clone(&self.data_farm_state);
+        let cancellation_token = self.cancellation_token.child_token();
+        let clock = self.clock;
+
+        self.tasks.push(get_runtime().spawn(async move {
+            if let Err(e) =
+                monitor_data_farm_notices(client, data_farm_state, clock, cancellation_token).await
+            {
+                tracing::warn!("IB data farm notice monitor stopped: {e:?}");
+            }
+        }));
+        self.is_connected.store(true, Ordering::Relaxed);
 
         let instrument_count = self.instrument_provider.count();
         if instrument_count > 0 {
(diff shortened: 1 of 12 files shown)


faysou:
@cjdsellers the CI needs to be rerun here and the PR seems ready to be reviewed.

cjdsellers:
Hey @mahimn01,

Thanks for tracking this down, fixing it across both IB adapters, and covering the failure and retry paths :pray:

PR #4520, opened 2026-07-20
Serialize the global logger lifecycle and keep its worker process-lifetime

folknor:
The global logger's initialization and teardown were coordinated by unsynchronized atomic flags, allowing two races: a concurrent initializer could observe a half-published registration, and a re-initialization after teardown could return a guard bound to a dead worker.

## Concurrent initialization could observe a partial registration

`Logger::init_with_config` checked `LOGGING_INITIALIZED`, then registered the logger with `set_boxed_logger`, published `LOGGER_TX`, and set the flag - none of it serialized. When two threads initialized at once (any process that builds more than one kernel, e.g. parallel tests), the loser of `set_boxed_logger` fell into the kernel's `SetLoggerError` recovery, which called `LogGuard::new()` and read `LOGGER_TX` before the winner had published it - returning `None` and failing with "A non-Nautilus logger is already registered". The window is the few lines between `set_boxed_logger` and `LOGGER_TX.set`.

## Re-initialization after teardown returned a guard on a dead worker

The last `LogGuard` drop sent `Close`, joined the worker thread, and reset `LOGGING_INITIALIZED` - but `log::set_boxed_logger` is permanent for the process and `LOGGER_TX` is a `OnceLock` that is never cleared. A later `init_with_config` therefore passed the fast-path check, hit `SetLoggerError`, and recovered a `LogGuard` whose sender pointed at the joined, dead worker - logging silently broken while bypass stayed set.

## Fix: a mutex-serialized lifecycle with a process-lifetime worker

A process-global `Mutex<LoggerLifecycle>` (`Uninitialized -> Running -> Terminated`, with no `Running -> Uninitialized` edge) serializes initialization, guard acquisition, the last-guard release, and shutdown; `LOGGING_INITIALIZED` is demoted to an informational mirror. Initialization spawns the writer thread before the irreversible `set_boxed_logger`, so a spawn failure cannot leave a registered logger with no receiver, and it publishes `LOGGER_TX`/`LOGGER_HANDLE` and flips to `Running` under the lock, so a concurrent acquirer either performs the one initialization or observes a fully published `Running` state. A normal last-guard drop now synchronously flushes and syncs (via the existing acknowledgement path, preserving drop-then-read-output tests) and leaves the worker running; `logging_shutdown` is the sole terminal operation, after which a further init returns a clear error instead of a dead-sender guard. The kernel's `SetLoggerError` recovery now fires only for a genuinely foreign logger. `Logger::new_for_benchmark` and the per-kernel shutdown-on-error machinery are unaffected. Under the `simulation`/`madsim` build, which spawns no worker and drops the receiver, the sync and channel sends are compiled out so the path never touches the dead channel.

## History

This surfaced as an intermittent `A non-Nautilus logger is already registered; cannot initialize Nautilus logging` failure in tests that construct more than one kernel in a single process. Two narrower approaches were considered and rejected. Reclassifying or serializing the affected tests does not fix the race: under `cargo test` a crate's tests share one process regardless of their module, and `nextest` already runs each test in its own process, so neither changes the in-process initialization ordering that triggers it - both only mask it. Fixing the concurrent-initialization window alone was also rejected, since the teardown-then-reinitialization dead-worker bug is the same lifecycle seam and would survive it.

## Testing

All three regressions run in a fresh child process, because `log::set_boxed_logger` is irreversible within a test binary.

- `test_init_publish_boundary_serializes_guard_acquisition` - initialization is parked at the publish boundary holding the lifecycle mutex, and a concurrent acquisition observes the lock held rather than a partial state. No timing dependency; it fails if initialization stops holding the mutex across publication.
- `test_concurrent_init_high_contention` - 64 threads barrier-start and each initialize a logger; every call returns a valid guard and a subsequent sync succeeds.
- `test_sequential_init_reuses_live_worker_and_original_file` - init, log marker A, drop the guard, assert logging is still `Running`, init again and receive a live guard, log marker B, and confirm both markers reached the original file; a final shutdown makes a further init error.

The contention and sequential tests were verified to fail against the pre-fix implementation, and the boundary test to fail if the lifecycle mutex is removed.

### DIFF
--- a/crates/common/src/logging/logger.rs
+++ b/crates/common/src/logging/logger.rs
@@ -17,7 +17,7 @@ use std::{
     cell::RefCell,
     fmt::{Display, Write as _},
     sync::{
-        Mutex, OnceLock,
+        Mutex, OnceLock, PoisonError,
         atomic::{AtomicBool, Ordering},
         mpsc::SendError,
     },
@@ -105,6 +105,31 @@ static LOGGER_TX: OnceLock<std::sync::mpsc::Sender<LogEvent>> = OnceLock::new();
 
 /// Global handle to the logging thread - only one thread exists per process.
 static LOGGER_HANDLE: Mutex<Option<std::thread::JoinHandle<()>>> = Mutex::new(None);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum LoggerLifecycle {
+    Uninitialized,
+    Running,
+    Terminated,
+}
+
+/// Process-global logger lifecycle serialization.
+static LOGGER_LIFECYCLE: Mutex<LoggerLifecycle> = Mutex::new(LoggerLifecycle::Uninitialized);
+
+#[cfg(test)]
+struct InitPublishHook {
+    reached: std::sync::mpsc::Sender<()>,
+    resume: std::sync::mpsc::Receiver<()>,
+}
+
+#[cfg(test)]
+static INIT_PUBLISH_HOOK: Mutex<Option<InitPublishHook>> = Mutex::new(None);
+
+#[cfg(test)]
+enum TestGuardAcquire {
+    LifecycleBusy,
+    Acquired(Option<LogGuard>),
+}
 
 static SHUTDOWN_ON_ERROR: OnceLock<ShutdownOnError> = OnceLock::new();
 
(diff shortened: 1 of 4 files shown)


folknor:
Force-pushed to fold two regenerated cbindgen headers into the commit: `nautilus_trader/core/includes/common.h` and `nautilus_trader/core/rust/common.pxd`.

The `cargo clippy` pre-commit hook was failing on CI, but not on a lint - clippy was clean. The failure was the hook's "files were modified by this hook" check. This commit rewrites the `LogGuard` doc comment (the new process-lifetime thread / `logging_shutdown`-only-terminates wording), and cbindgen copies that doc comment verbatim into the two generated headers. The original push updated the Rust source but not the headers, so when CI compiled `nautilus-common` with the `ffi` feature the build regenerated them, git saw two modified tracked files, and the hook exited non-zero.

The diff to both headers is purely the doc-comment text - no ABI, signature, or layout change - so the update is a documentation regeneration only.

cjdsellers (approved):
Hey @folknor

Thanks for tracing both logger lifecycle failures and covering them with tests. I reviewed this against the v2 kernel path, no findings from me, this is ready to merge!

PR #4521, opened 2026-07-21
Prepare both cross-zero reconciliation legs before the first cache mutation

folknor:
A follow-up to the live reconciliation series (#4479, #4490, #4517), closing a path where a fallible open-leg construction could strand a synthetic order.

## Cross-zero reconciliation mutates the close leg before the open leg is known-constructible

`ExecutionManager::reconcile_cross_zero_position` splits a sign-flipping position discrepancy into two synthetic fills: close the cached side, then open the venue side. It ran the close leg through `handle_external_order` before constructing the open leg.

`handle_external_order` is not a pure event builder - before returning events it calls `Cache::add_order` (inserting the synthetic order, marking it active-local, updating the order/venue/instrument/strategy indexes, and persisting to the cache database), registers both directions of the client-order-id/venue-order-id index, and publishes `OrderInitialized` on the message bus.

The open leg's fallible conversions ran only after that: `Quantity::from_decimal_dp(...).ok()?` for the open quantity and `OrderStatusReport::new(...).with_avg_px(...).ok()?` for the open report. When either returned `None` the function early-returned and dropped the accumulated events, so no close fill ever reached the position - but the synthetic close order was already cached, indexed, and published. Because the synthetic `client_order_id` is derived deterministically from the venue order id, the next reconciliation cycle regenerates the same id, `Cache::add_order` collides, the "exists in cache in non-terminal state; fill not regenerated" branch returns no events, and the cross-zero reconciliation can never complete - a permanently stranded non-terminal order with the discrepancy left unresolved.

The fix builds and validates both leg reports before the first `handle_external_order` call: when `venue_avg_px` is `Some`, the complete open `OrderStatusReport` is constructed up front, and if it cannot be built the function returns `None` with nothing cached, indexed, or published. The close leg is then applied, followed by the open leg, preserving close-before-open ordering. The `venue_avg_px == None` close-only path (deliberately closes the stale cached side when the venue reports no average price) is unchanged. A small pure `build_cross_zero_leg_report` helper removes the duplicated report construction.

## Testing

`test_cross_zero_unbuildable_open_leg_has_no_side_effects`: a cross-zero discrepancy whose open leg cannot be built (venue signed quantity forced past `QUANTITY_MAX`) leaves the order count, both client/venue order-id index directions, and the cached position quantity unchanged, and publishes no order event. Verified to fail against the unfixed code, which caches and publishes the synthetic close order.

`test_cross_zero_unbuildable_open_leg_retries_without_poisoning_order_id`: the first reconciliation fails the open-leg conversion and increments (does not reset) the position-discrepancy retry counter; a subsequent cycle with a buildable open leg produces both fills in close-then-open order and drives the cache to the venue quantity. Verified to fail against the unfixed code, where the stranded close order collides on the deterministic id and only one leg is regenerated.

Both use paused virtual time with no wall-clock waits. The existing `test_cross_zero_with_missing_venue_avg_px_closes_only` remains green, pinning the unchanged close-only contract.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -87,6 +87,52 @@ static TAG_RECONCILIATION: LazyLock<Ustr> = LazyLock::new(|| Ustr::from("RECONCI
 pub type InstrumentAccountKey = (InstrumentId, AccountId);
 type FillKey = (AccountId, InstrumentId, TradeId);
 
+#[expect(clippy::too_many_arguments)]
+fn build_cross_zero_leg_report(
+    instrument: &InstrumentAny,
+    account_id: AccountId,
+    instrument_id: InstrumentId,
+    order_side: OrderSide,
+    quantity: Decimal,
+    avg_px: Decimal,
+    tag: &str,
+    ts_now: UnixNanos,
+    venue_ts_last: UnixNanos,
+) -> Option<OrderStatusReport> {
+    let order_qty = Quantity::from_decimal_dp(quantity, instrument.size_precision()).ok()?;
+    let fill_price = Price::from_decimal_dp(avg_px, instrument.price_precision()).ok();
+    let venue_order_id = create_position_reconciliation_venue_order_id(
+        account_id,
+        instrument_id,
+        order_side,
+        OrderType::Market,
+        order_qty,
+        fill_price,
+        None,
+        Some(tag),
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for catching this. This looks good to merge as well.

PR #4522, opened 2026-07-21
Defer recent-fill dedup marking until the fill applies to its order

folknor:
A follow-up to the reconciliation fill-dedup work in #4518, which closed a premature-commit window on the mass-status `processed_fills` map; the same pattern remained on the normal-live `recent_fills_cache`.

## recent_fills_cache is marked before the fill applies to its canonical order

The live node dedups repeated venue fills through `recent_fills_cache`, keyed by `(account_id, instrument_id, trade_id)`: a direct `OrderFilled` marks the key, and a later matching `ExecutionReport::Fill` carrying the same key is skipped (`observe_exec_event_before_dispatch`). The mark was committed before the event reached the canonical order cache - `observe_order_event` marked every direct `OrderFilled` while the node called it in the pre-dispatch hook, and `process_reconciliation_events` marked before `ExecutionEngine::process`. Both dispatch paths are synchronous and neither reports application success, so a fill the engine rejected (unknown order, invalid transition) was still marked, and its later report - the venue's authoritative record - was suppressed for the recency window (~60-120s). This is the same class #4518 fixed on the mass-status map.

The fix defers the mark to after dispatch, gated on the canonical order. `commit_recent_fill_if_applied` resolves the order (by `client_order_id`, venue-order-id fallback), requires the account, instrument, and a `trade_ids` containment match, and only then marks - reusing the `is_fill_applied` postcondition introduced in #4518. An already-present trade id counts as applied (correct replay); an unresolved or unapplied fill is not marked, so its later report stays eligible. The mark is removed from the pre-dispatch `observe_order_event` (its position and local activity recording stay); the normal-ingress path captures the `OrderFilled` before the event is moved into dispatch and commits after; `process_reconciliation_events` processes then commits. The skip and prune sites are unchanged.

One intentional behavior change: an orderless leg fill (one the engine applies to a position with no canonical order) is no longer marked, since it has no order to satisfy the postcondition. It stays eligible for a later report - the honest result under an order-cache postcondition; a position-cache postcondition would be separate scope.

## Testing

- `test_rejected_direct_fill_stays_eligible_for_later_report`: a direct fill whose order is absent from the cache is not marked (before or after dispatch) and its later `Fill` report stays eligible.
- `test_applied_direct_fill_commits_and_skips_later_report`: a fill applied to its canonical order is marked only after dispatch, and its later report is skipped. Drives the production dispatch-and-commit path.
- `test_canonical_duplicate_fill_counts_as_applied`: a fill already present on the order (resolved via the venue-order-id fallback) commits.
- `test_continuous_reconciliation_commits_only_applied_fill` (applied/rejected): the reconciliation path marks only when the fill applied.
- `test_recent_fill_commit_requires_account_and_instrument_match`: an account or instrument mismatch does not commit.
- `test_applied_inferred_fill_remains_recently_processed`: an applied inferred reconciliation fill retains the existing recent-cache behavior.

The correctness assertions were verified to fail against the pre-change premature-mark code (a rejected fill was marked before dispatch and its report suppressed). Tests use paused virtual time with no wall-clock waits; the existing recency-TTL tests remain green.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -2063,7 +2063,7 @@ impl ExecutionManager {
     /// Observes a local order event and updates tracking state.
     ///
     /// This is the `LiveNode` dispatch path for order events: acknowledgement
-    /// events clear reconciliation tracking, fills record fill/position
+    /// events clear reconciliation tracking, fills record position
     /// activity, and every event stamps local activity. The stamp must come
     /// AFTER any [`Self::clear_recon_tracking`] call - that call drops the
     /// local-activity mark, which is the sole grace gate protecting a
@@ -2073,7 +2073,6 @@ impl ExecutionManager {
         match event {
             OrderEventAny::Filled(fill) => {
                 self.record_position_activity(fill.instrument_id, fill.account_id);
-                self.mark_fill_processed(fill.account_id, fill.instrument_id, fill.trade_id);
             }
             OrderEventAny::Accepted(_)
             | OrderEventAny::Rejected(_)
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor

Thanks for the detailed fix and regression coverage here. The post-apply gating for canonical orders looks good.

The orderless leg-fill case is worth a separate look, particularly the reliance on `-LEG-` IDs and what happens if a later `FillReport` arrives. I don't think that should block this PR, so I'll investigate that path locally and follow up separately if needed.

PR #4523, opened 2026-07-21
Fix inflight coverage and bookkeeping-map retention in live reconciliation

folknor:
A follow-up to the live reconciliation series (#4479, #4490, #4517, #4518, #4521), covering three defects in the manager's inflight-check and local-activity bookkeeping maps: a coverage gap and two unbounded-growth leaks.

## Batch cancels never register for inflight timeout coverage

The exec-command dispatch registers orders for inflight tracking as each `TradingCommand` is handled: `SubmitOrder`, `SubmitOrderList`, `ModifyOrder`, `ModifyOrders`, and single `CancelOrder` all call `register_inflight`, but `TradingCommand::CancelOrders` (the batch) fell through the wildcard arm. The execution engine forwards a batch cancel to the client directly without emitting per-order `CancelOrder` commands, and the strategy has already moved each included order to `PendingCancel`, so without registration `check_inflight_orders` never observes the children. A batch cancel the venue silently drops therefore leaves each child stuck in `PendingCancel` indefinitely with no timeout to resolve it.

The fix adds a `CancelOrders` arm that iterates `cancels` and registers every child `client_order_id`, mirroring the existing `ModifyOrders` arm. `CancelAllOrders` is deliberately left out: it carries no per-order ids and does not move individual orders to `PendingCancel`, so per-order timeout coverage there would need the command to carry a captured target set - a separate change.

## Filtered inflight orders are never retired from the check map

`check_inflight_orders` first snapshots every past-threshold `inflight_checks` key, then for any id in `filtered_client_order_ids` performs a bare `continue` that mutates nothing. The entry's `submitted_at` stays beyond the threshold, so the id is re-selected and re-skipped on every subsequent cycle - it is never retired, and `inflight_checks` grows without bound for the process lifetime.

The fix retires a filtered entry via `clear_recon_tracking` when the checker encounters it, and adds an early guard in `register_inflight` so a filtered id is never inserted in the first place. The checker-path cleanup remains as the retirement route for an id that is filtered after it was already registered.

## Closed-order reports re-mark local activity that nothing prunes

`observe_execution_report` records `order_local_activity` as the real-time settling window that defers reconciliation while a local order interaction is still in flight. For a non-pending order report it called `clear_recon_tracking` (which removes the activity entry) and then unconditionally re-marked `order_local_activity`. For a terminal report the re-mark re-inserts an entry that nothing subsequently removes: report ingress contributes no id to the post-dispatch closed-order cleanup, and unlike the fill recency maps `order_local_activity` was never age-pruned. Every closed order thus left a permanent entry, and the map grew unbounded.

The `Order` and `OrderWithFills` arms now share one `observe_order_status_report` helper implementing the status matrix: `PendingUpdate`/`PendingCancel` retain inflight tracking and mark activity, an open non-pending status clears inflight bookkeeping and marks activity for the settling grace, and a closed status clears all tracking and does not re-mark. This also removes an inconsistency where a pending `OrderWithFills` report previously recorded no activity while a pending `Order` report did. Independently, `prune_order_local_activity` runs on the existing periodic maintenance tick (beside the fill-cache prunes) and drops entries older than `open_check_threshold_ns`. That window is exactly what all four `order_local_activity` consumers already gate on, so an entry older than it is already treated as no activity - pruning it changes no decision - and the periodic bound also covers the fill-report-close and orphaned-order paths that the per-report matrix does not reach.

## Testing

`test_batch_cancel_command_registers_each_child_for_inflight_timeout`: a `CancelOrders` batch whose children are transitioned `Accepted -> PendingCancel` via cache-applied events; after the inflight threshold elapses, `check_inflight_orders` emits one `OrderCanceled` event per child and nothing else. Verified to fail against the unfixed code, where the batch falls through the wildcard arm, no child is registered, and the timeout set is empty.

`test_register_inflight_skips_filtered_order` and `test_inflight_check_retires_order_filtered_after_registration`: a filtered id is never inserted into `inflight_checks`, and an id filtered after registration is removed on the next check and stays absent across a second cycle. The retirement test fails against the unfixed code, which leaves the entry to be revisited every cycle.

`test_observe_order_status_report_tracking_matrix`: parametrized over `Order` and `OrderWithFills` crossed with pending, accepted, and terminal statuses, asserting the resulting `inflight_checks`, retry-count, `order_local_activity`, query-recency, and coverage-set state. The terminal cases fail against the unfixed code, which re-marks local activity immediately after clearing it.

`test_prune_order_local_activity_uses_open_check_threshold`: an entry older than `open_check_threshold_ns` is pruned while a fresher one is retained.

All four use paused virtual time with no wall-clock waits.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -1059,6 +1059,7 @@ impl ExecutionManager {
                 .filtered_client_order_ids
                 .contains(&client_order_id)
             {
+                self.clear_recon_tracking(&client_order_id, true);
                 continue;
             }
 
@@ -1995,6 +1996,14 @@ impl ExecutionManager {
 
     /// Registers an order as inflight for tracking.
     pub fn register_inflight(&mut self, client_order_id: ClientOrderId) {
+        if self
+            .config
+            .filtered_client_order_ids
+            .contains(&client_order_id)
+        {
+            return;
+        }
+
         self.inflight_checks.insert(
             client_order_id,
             InflightCheck {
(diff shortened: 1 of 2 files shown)


folknor:
@cjdsellers I realise this PR has failed CI on one of the python gates, but I don't think that's related to this patch. So I requested review anyway.

I am unsure how to force a re-run of CI.

cjdsellers (approved):
Hey @folknor

Thanks for this, solid changes once more. 

The earlier Python 3.12 CI failure was an infrastructure flake rather than a test failure: the runner stopped during the Rust adapter tests without reporting a failure. We can expect these flakes to stabilize into v2.

I found one very narrow cancel-replace edge around stale terminal reports, but I don't think it should hold up this PR and I'll follow it up separately after merge if necessary.

PR #4527, opened 2026-07-21
Parse min_notional for Bybit spot instruments

dxwil:
# Pull Request

## Summary

Parse `min_notional` for Bybit spot instruments from the `minOrderAmt` lot-size filter field. Also add the newer lot-size fields returned by the API (`maxLimitOrderQty`, `maxMarketOrderQty`, `postOnlyMaxLimitOrderSize`, and the deprecated `maxOrderAmt`) as optional, following the linear instrument convention.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Release notes

## Testing

### DIFF
--- a/crates/adapters/bybit/src/common/parse.rs
+++ b/crates/adapters/bybit/src/common/parse.rs
@@ -306,6 +306,13 @@ pub fn parse_spot_instrument(
         &definition.lot_size_filter.min_order_qty,
         "lotSizeFilter.minOrderQty",
     )?);
+    let min_notional = Some(Money::from_decimal(
+        parse_decimal(
+            &definition.lot_size_filter.min_order_amt,
+            "lotSizeFilter.minOrderAmt",
+        )?,
+        quote_currency,
+    )?);
 
     let maker_fee = parse_decimal(&fee_rate.maker_fee_rate, "makerFeeRate")?;
     let taker_fee = parse_decimal(&fee_rate.taker_fee_rate, "takerFeeRate")?;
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hey @dxwil

Thanks for adding the Bybit spot minimum notional mapping. The implementation matches the current API contract, and the focused regression passes, so this looks good to merge.

The only non-blocking follow-up I see is to update the spot instrument fixture to cover the three new optional quantity fields and the missing `maxOrderAmt` path with exact assertions, when you get the chance.

dxwil:
Hi @cjdsellers, modified the test fixtures with the new fields.

cjdsellers:
Hey @dxwil

Thanks for updating the fixtures. I missed two issues in my first pass that need fixing before merge.

`SpotLotSizeFilter` is part of the published `nautilus-bybit` API, so changing `max_order_amt` from `String` to `Option<String>` and adding fields to the exhaustive public struct is source-breaking. The parser does not use the new quantity fields, and Serde already ignores unknown response fields, so we can leave that DTO shape unchanged here.

Also, `minOrderAmt` is an exact wire string, but `parse_optional_notional` sends it through `f64` before constructing `Money` (a separate audit it about to fix/improve this use of f64). Best to just parse it as `Decimal`, use `Money::from_decimal`, and add an exact assertion that would catch float rounding.

dxwil:
Makes sense to not change the public API if not needed. Now this PR is just the added parsing of min_notional.

cjdsellers (approved):
Thanks for the follow-ups @dxwil 👌

PR #4528, opened 2026-07-21
Bound live node startup and shutdown by the lifecycle timeouts

folknor:
A follow-up to the live-node work behind the reconciliation series (#4479, #4517, #4518, #4521, #4522, #4523), hardening the node's startup and shutdown lifecycle against hung adapters and a false-success timeout path.

## Client connect and disconnect are awaited outside the lifecycle timeouts

The kernel `connect_data_clients`, `connect_exec_clients`, and `disconnect_clients` await the underlying engine future with no bound. Only the `check_connected`/`check_disconnected` poll loops (`await_engines_connected`, `await_engines_disconnected`) were bounded, by `timeout_connection` and `timeout_disconnection`. In `start()` the connect calls were bare awaits ahead of that poll; in `run()` they were driven through `drive_with_event_buffering`, which has no timeout arm (an in-code `TODO` acknowledged the gap); `finalize_stop` awaited `disconnect_clients` bare. A pending adapter connect or disconnect future therefore hangs startup or shutdown indefinitely - the poll meant to bound it never runs.

The fix treats `timeout_connection` as a single deadline spanning data connect, instrument flush, execution connect, and the engine-connected poll: each connect future is wrapped in `dst::time::timeout` against the remaining budget, and the readiness poll shares that same deadline rather than restarting a fresh one. `finalize_stop` bounds `disconnect_clients` and the disconnected-readiness poll under one shared `timeout_disconnection` deadline and always finalizes the kernel locally. `disconnect_clients` now disconnects the data and execution engines concurrently, so a hung data disconnect no longer prevents execution teardown from being attempted. This bounds how long the node waits and how it reports the outcome; it does not retrofit cancellation-safety onto the adapter `connect`/`disconnect` contract, so an adapter that records transport state only after setting its connected flag can still retain that state if its connect future is dropped mid-flight - a separate follow-up.

## A connection timeout reports the node as running

On a readiness timeout `await_engines_connected` returns `TimedOut`. `start()` logged an error, set `NodeState::Running`, and returned `Ok(())`; `run()` logged "Not starting trader" and then unconditionally set `Running`. The trader never started, yet the node's externally observable state was `Running` and the call reported success. `EngineConnectionStatus::abort_reason()` also excluded `TimedOut`, so the `run()` path treated the timeout as a benign non-abort.

A connect or readiness timeout now fails closed: it aborts startup, disconnects and finalizes the kernel, sets the local state to `Stopped`, and returns an error naming the stage that timed out (data connect, execution connect, or readiness). The node never publishes `Running` on a timeout. Intentional stop and shutdown requests observed during startup keep their existing behavior (abort and return `Ok`).

## LiveNode::build validates the environment after overwriting it

`LiveNode::build` assigned `config.environment = Environment::Live` before the match that validates it. A caller-supplied `Sandbox` was silently promoted to `Live`, and the `Backtest` rejection branch was unreachable. The builder entry points (`LiveNodeBuilder::new`, `from_config`) already validate the incoming environment without overwriting; only `build` had the ordering wrong.

`build` no longer overwrites the environment; it validates the incoming value through a shared `validate_live_environment` helper that the builder paths also call, so the two cannot drift apart. `LiveNodeConfig::default()` is already `Live`, so a `None` config still yields `Live`, a supplied `Sandbox` is preserved, and a supplied `Backtest` is rejected.

## Testing

`test_start_hung_data_connect_times_out_fail_closed`, `test_run_hung_data_connect_times_out_fail_closed`, `test_start_hung_exec_connect_times_out_fail_closed`, `test_run_hung_exec_connect_times_out_fail_closed`: a data or execution client whose `connect()` never resolves; `start()` and `run()` return an error naming the connect stage within `timeout_connection`, the node ends `Stopped` and never `Running`, and disconnect is attempted. These fail against the unfixed code, where the bare connect await hangs forever.

`test_start_readiness_timeout_fails_closed`, `test_run_readiness_timeout_fails_closed`: `connect()` returns but the engine never reports connected; the readiness poll times out and both entry points fail closed to `Stopped`. These fail against the unfixed code, which set `Running` and returned success.

`test_start_readiness_timeout_uses_shared_connection_budget`: a connect that consumes part of `timeout_connection` followed by a never-ready engine; the total virtual elapsed time stays within the single connection budget rather than the connect delay plus a restarted readiness timeout. This pins the shared-deadline property.

`test_hung_data_disconnect_still_attempts_execution_disconnect`: a data client whose `disconnect()` never resolves; execution disconnect is still attempted, the shutdown is bounded by `timeout_disconnection`, and the node finalizes to `Stopped`. This fails against a sequential disconnect, where the hung data disconnect blocks execution teardown.

`test_stop_fails_when_disconnect_readiness_poll_times_out`: a client whose `disconnect()` returns `Ok` but stays connected; the disconnected-readiness poll times out and `stop()` returns an error while the node still finalizes to `Stopped`. This fails against the unfixed code, where the poll logged a timeout but returned success.

`test_live_node_build_preserves_sandbox_environment`, `test_live_node_build_rejects_backtest_environment`: `build` keeps a `Sandbox` config and rejects `Backtest`; the former replaces an earlier test that asserted the overwrite to `Live`.

All use paused virtual time with no wall-clock waits, and each timeout assertion was verified to fail against the unfixed implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6911,6 +6911,7 @@ dependencies = [
  "async-trait",
  "bon",
  "bytes",
+ "futures",
  "indexmap 2.14.0",
  "log",
  "nautilus-common",
--- a/crates/live/src/node/builder.rs
+++ b/crates/live/src/node/builder.rs
@@ -50,7 +50,7 @@ use super::{
     LiveNode,
     config::{
         LiveDataEngineConfig, LiveExecEngineConfig, LiveNodeConfig, LiveRiskEngineConfig,
-        RoutingConfig,
+        RoutingConfig, validate_live_environment,
     },
 };
 use crate::{
@@ -135,12 +135,7 @@ impl LiveNodeBuilder {
     ///
     /// Returns an error if `environment` is invalid (BACKTEST).
     pub fn new(trader_id: TraderId, environment: Environment) -> anyhow::Result<Self> {
-        match environment {
-            Environment::Sandbox | Environment::Live => {}
-            Environment::Backtest => {
-                anyhow::bail!("LiveNode cannot be used with Backtest environment");
-            }
-        }
+        validate_live_environment(environment)?;
 
         let config = LiveNodeConfig {
             environment,
@@ -171,12 +166,7 @@ impl LiveNodeBuilder {
     ///
     /// Returns an error if the config's environment is invalid (BACKTEST).
     pub fn from_config(config: LiveNodeConfig) -> anyhow::Result<Self> {
-        match config.environment {
-            Environment::Sandbox | Environment::Live => {}
-            Environment::Backtest => {
-                anyhow::bail!("LiveNode cannot be used with Backtest environment");
-            }
-        }
+        validate_live_environment(config.environment)?;
 
         Ok(Self {
             name: "LiveNode".to_string(),
--- a/crates/live/src/node/config.rs
+++ b/crates/live/src/node/config.rs
@@ -59,6 +59,15 @@ const DEFAULT_ORDER_RATE_LIMIT: &str = "100/00:00:01";
 const RUST_RUNTIME_UNSUPPORTED: &str = "not supported by the Rust live runtime yet";
 const RATE_LIMIT_FORMAT: &str = "expected 'limit/HH:MM:SS'";
 
+pub(crate) fn validate_live_environment(environment: Environment) -> anyhow::Result<()> {
+    match environment {
+        Environment::Sandbox | Environment::Live => Ok(()),
+        Environment::Backtest => {
+            anyhow::bail!("LiveNode cannot be used with Backtest environment")
+        }
+    }
+}
+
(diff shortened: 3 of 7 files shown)


folknor:
Same story as PR 4529: CI was green so I requested review, but then I noticed I had to rebase this on top of your commit 08cb51102f. No conflicts on the merge, so it should be green again :-)

cjdsellers:
Thanks, @folknor.

No need to rebase in future unless there are merge conflicts or you suspect logic changes that may fail CI :ok_hand:

cjdsellers (approved):
Hey @folknor,

Thanks for the PR, these are good changes. The shared deadlines cover the actual connect and disconnect futures, timeout paths now fail closed to `Stopped`, and the regression tests cover the risky paths well. The adapter cancellation-safety follow-up doesn't block this change, so this looks good to merge.

PR #4529, opened 2026-07-21
Time out hung reconciliation report tasks

folknor:
The companion to the cross-zero atomicity fix in #4521, on the same continuous-reconciliation path.

## A hung client stalls all periodic reconciliation indefinitely

Continuous reconciliation runs one report task at a time: the open-order, targeted-order, and position checks share a single mutually-exclusive slot, and `maybe_start_reconciliation_check` refuses to start any check while another is in progress. A task holds the slot until it completes, and the bulk request functions (`request_open_order_reports`, `request_position_reports`) and the targeted path await each client's report generation sequentially with no deadline. A single client whose `generate_*_reports` never resolves therefore occupies the slot for the lifetime of the process: open, targeted, and position reconciliation all stop, silently, while the node keeps running and trading. `timeout_reconciliation` bounded only startup mass-status collection; it had no continuous-reconciliation caller.

Each spawned report task now carries its own deadline. `start_open_order_report_check`, `start_targeted_order_report_check`, and `start_position_report_check` capture `now + timeout_reconciliation` when the task is created and store it inside the pinned future, which wraps the client requests in a single `dst::time::timeout` and resolves to `Completed` or `TimedOut`. The deadline lives in the future, not the `select!` arm: the arm is reconstructed whenever another branch wins, so a relative timeout placed there would restart on every runner iteration and never fire. On `TimedOut` the task is dropped first (releasing the execution-client borrow held across the request), deferred instrument updates are then flushed to every client (the per-client report facade only flushes after its await returns, which cancellation skips), a timed-out targeted task removes exactly its planned client-order IDs from `targeted_order_queries` (a narrow removal that leaves inflight retry counts and query recency intact, so those orders are reconsidered fairly on a later cycle), the expiry is logged, and no partial batch is reconciled. `last_open_check` and `last_position_check` are left untouched, so the blocked check class - already overdue - wins the next scheduling pass once the slot frees. Ordinary per-client errors keep their existing partial-success behavior; only expiry discards the accumulated batch, which is conservative because reports gathered across a long stall may be stale and reconciliation acts on positive reports without requiring full client coverage. The same cleanup runs at shutdown, where the tasks were previously dropped without flushing deferred instruments or clearing targeted markers.

The bound reuses the existing `timeout_reconciliation` field (30s default); its documentation now states that it covers both startup reconciliation and each continuous report-collection task.

## Testing

`test_hung_open_report_task_times_out_and_position_check_starts` and `test_hung_position_report_task_times_out_and_open_check_starts`: a client whose bulk report generation never resolves occupies the slot; the task times out, releases the slot, and the other check class subsequently runs. The deadline is set above the runner's maintenance cadence so several maintenance ticks (and other `select!` arms) fire during the task's lifetime, distinguishing the persistent stored deadline from a select-local timer that would restart each iteration.

`test_hung_targeted_report_task_cleans_markers_and_checks_resume`: a hung targeted task times out; a planned order that is also inflight-checked is queried by the inflight checker afterwards - which only happens once its targeted marker is removed, since the checker skips an inflight order still present in `targeted_order_queries` - while its reconciliation retry count is preserved, and position/inflight checking resumes.

`test_timed_out_open_report_task_discards_earlier_client_reports`: reports collected from an earlier client before a later client hangs are not applied when the task expires.

`test_timed_out_report_task_flushes_deferred_instrument_update`: an instrument update deferred while a client is borrowed during a report is delivered after the timeout cancellation.

All use paused virtual time with no wall-clock waits, and each assertion was verified to fail against the unfixed code, where the hung task holds the slot indefinitely and no cleanup runs.

### DIFF
--- a/crates/live/src/execution/client.rs
+++ b/crates/live/src/execution/client.rs
@@ -145,7 +145,7 @@ impl LiveExecutionClient {
         result
     }
 
-    fn flush_pending_instruments(&self) {
+    pub(crate) fn flush_pending_instruments(&self) {
         let mut pending = self.pending_instruments.borrow_mut();
         if pending.is_empty() {
             return;
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -181,6 +181,12 @@ pub(crate) struct TargetedOrderQuery {
     command: GenerateOrderStatusReport,
 }
 
+impl TargetedOrderQuery {
+    pub(crate) const fn client_order_id(&self) -> ClientOrderId {
+        self.client_order_id
+    }
+}
+
 #[derive(Debug)]
 pub(crate) struct TargetedOrderReportResult {
     client_order_id: ClientOrderId,
@@ -2041,6 +2047,12 @@ impl ExecutionManager {
             self.order_query_recency.remove(client_order_id);
         }
         self.order_local_activity.remove(client_order_id);
+    }
+
+    pub(crate) fn remove_targeted_order_queries(&mut self, client_order_ids: &[ClientOrderId]) {
+        for client_order_id in client_order_ids {
+            self.targeted_order_queries.shift_remove(client_order_id);
+        }
     }
 
     /// Returns any external order claim for the given instrument ID.
--- a/crates/live/src/node/config.rs
+++ b/crates/live/src/node/config.rs
@@ -749,7 +749,7 @@ pub struct LiveNodeConfig {
     /// The timeout for all clients to connect and initialize.
     #[builder(default = Duration::from_mins(1))]
     pub timeout_connection: Duration,
-    /// The timeout for execution state to reconcile.
+    /// The timeout for startup reconciliation and each continuous report-collection task.
     #[builder(default = Duration::from_secs(30))]
     pub timeout_reconciliation: Duration,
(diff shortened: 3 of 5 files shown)


folknor:
CI was green so I requested review, but then I noticed I had to rebase this on top of your commit 08cb51102f. No conflicts on the merge, so it should be green again :-)

cjdsellers:
Hey @folknor

Thanks, the timeout handling itself looks good, and the five focused `report_task` regressions pass for me on `815b6fea2d`. The branch now conflicts with current `develop` in `crates/live/tests/node.rs` after #4528 landed, so when you get the chance - please rebase and retain both the lifecycle-timeout coverage from #4528 and the report-task coverage here. I'll recheck the resolved head once CI reruns.

folknor:
@cjdsellers #4500 broke the CI (blank line above control flow). I will fix and tag that along in this PR, because my local tooling now doesn't allow me to proceed unless it's green.

Or if you want I can commit such CI style fixes directly (do I have commit access? I'm not sure), and keep all other commits to PRs like now.

It feels silly opening another PR for two blank lines :-P

folknor:
Nevermind, I see now that my tooling is stricter than your CI.

cjdsellers (approved):

PR #4530, opened 2026-07-22
Fix non-atomic order event application

folknor:
Two defects in the order event-application path, both the same class as the defensive alignment in #4483: an ordinary venue event stream reaching code that mutates before it finishes validating.

## Order updates are validated after the state transition commits

`OrderCore::apply` writes `previous_status`, transitions `status`, runs the per-event handler, sets `ts_last`, and pushes the event onto the history, then returns `Ok`. Only afterwards does each concrete `apply` call `self.update(event)`, and six order types validate the update there with `assert!`: `MarketOrder` rejects both `price` and `trigger_price`, `StopMarketOrder`, `MarketIfTouchedOrder` and `TrailingStopMarketOrder` reject `price`, `LimitOrder` and `MarketToLimitOrder` reject `trigger_price`. An `OrderUpdated` carrying a field the type cannot hold therefore panics with the transition already committed and the event already in history, and never returns `Err(OrderError::InvalidOrderEvent)`. The release profile sets `panic = "abort"`, so this terminates the process; under unwind the order is left with a status and a history entry for an update that was never applied to its subtype fields. The producers are venue adapters, the order emulator, and reconciliation - `reconciliation/orders.rs` already filters spurious venue trigger prices before building the event, handling one call site rather than the boundary that owns the invariant.

The fix moves the field check into `OrderCore::apply`, where the invariant belongs. `OrderCore` already carries `order_type`, so an exhaustive `match` rejects an `OrderUpdated` whose `price`/`trigger_price` the type cannot hold and returns `Err(OrderError::InvalidOrderEvent)` before any field is written. It runs after the existing client-order-id/strategy-id identity checks and the `OrderStatus::transition` check and before the first mutation, so an update that is both mis-addressed and field-invalid still reports the identity mismatch, and one that is field-invalid in a status that rejects updates still reports `InvalidStateTransition` - the field check never masks a more fundamental rejection. The six restrictive `update()` methods keep their `assert!` as a defensive guard on the low-level mutation path, so a direct `update()` call that bypasses `apply` still trips rather than silently mutating.

## Rejected events mutate core state before the FSM rejects them

The same function has the same ordering problem for its own checks, without any panic involved. `previous_status` is assigned before the duplicate-fill check, the triggerable-type check, and `OrderStatus::transition`; `status` is assigned before the `Initialized` rejection and before `modify_rejected` and `cancel_rejected` can fail with `NoPreviousState`. An event the FSM rejects thus returns `Err` but leaves `previous_status` advanced to the current status - a rejected event is supposed to be a no-op on the order. `previous_status` feeds the `OrderModifyRejected`/`OrderCancelRejected` restore path, so this is core state a rejected event has no business touching; the fix restores the invariant that a rejected event leaves order state unchanged rather than relying on a later transition to overwrite the damage.

The fix hoists every fallible check above the first mutation: duplicate fill, triggerable type, duplicate initialization, the transition itself, and the `previous_status` precondition that the two rejection handlers need, which is now read into a local before anything is written. Only once all of them pass does the function assign `previous_status` and `status`, run the handler, and record the event. The existing conditions governing which transitions save a previous status are unchanged. `modify_rejected` and `cancel_rejected` no longer return `Result`, since their only failure mode is now checked up front.

One observable consequence: applying a second `OrderInitialized` to an existing order previously reached `OrderStatus::transition` first and returned `InvalidStateTransition`, leaving the `AlreadyInitialized` arm dead. The pre-commit check makes that variant reachable, so `OrderAny::from_events` now reports `AlreadyInitialized` for a replay stream carrying a duplicate initialization. No production path discriminates on either variant here.

## Testing

`test_market_order_rejects_invalid_update_atomically` (parametrized over `price` and `trigger_price`) and its five siblings on `StopMarketOrder`, `MarketIfTouchedOrder`, `TrailingStopMarketOrder`, `LimitOrder` and `MarketToLimitOrder` cover every condition that was previously an assertion: each asserts `Err(OrderError::InvalidOrderEvent)` and that `status`, `previous_status`, `ts_last` and the event count are all unchanged. Against the unfixed implementation these panic rather than return.

`test_update_identity_mismatch_precedes_field_validation` (parametrized over a wrong client-order-id and a wrong strategy-id) sends an update that is both mis-addressed and carries a `MarketOrder`-invalid `price`, and asserts the surfaced error is the identity `PredicateViolation` whose message names the mismatched field, not `InvalidOrderEvent`. `test_update_invalid_transition_precedes_field_validation` applies such an update to a `Denied` order and asserts `InvalidStateTransition`. Both fail against a validate-first ordering, which reports `InvalidOrderEvent` in each case. `test_direct_update_panics_on_invalid_field_before_mutation` calls `MarketOrder::update` directly, under `catch_unwind`, with an invalid `price` and a quantity distinct from the order's, and asserts the call panics and the quantity is unchanged - the defensive `assert!` fires ahead of the field writes. It fails with the guard removed: the call returns and the quantity moves.

`test_rejected_event_preserves_status_restoration` and `test_rejected_event_preserves_cancel_rejected_restoration` drive an order to `Accepted`, apply an event the FSM rejects, and assert `previous_status` is untouched, then confirm `OrderModifyRejected` and `OrderCancelRejected` restore `Accepted` through `PendingUpdate` and `PendingCancel` respectively. Both were verified to fail against the unfixed ordering, where the rejected event advances `previous_status` from `Submitted` to `Accepted`.

All of these are deterministic unit tests over constructed events with no clock or wall-clock dependency.

### DIFF
--- a/crates/model/src/orders/any.rs
+++ b/crates/model/src/orders/any.rs
@@ -777,11 +777,11 @@ mod tests {
 
         match &err {
             OrderReplayError::ApplyFailed { source } => {
-                assert!(matches!(source, OrderError::InvalidStateTransition));
+                assert!(matches!(source, OrderError::AlreadyInitialized));
             }
             _ => panic!("expected ApplyFailed, was {err:?}"),
         }
-        assert_eq!(err.to_string(), "Invalid order state transition");
+        assert_eq!(err.to_string(), "Order was already initialized");
     }
 
     #[rstest]
--- a/crates/model/src/orders/limit.rs
+++ b/crates/model/src/orders/limit.rs
@@ -610,7 +610,7 @@ mod tests {
         events::{OrderEventAny, OrderUpdated},
         identifiers::InstrumentId,
         instruments::{CurrencyPair, stubs::*},
-        orders::{Order, OrderTestBuilder, stubs::TestOrderStubs},
+        orders::{Order, OrderError, OrderTestBuilder, stubs::TestOrderStubs},
         types::{Price, Quantity},
     };
 
@@ -768,6 +768,36 @@ mod tests {
 
         assert_eq!(accepted_order.quantity(), updated_quantity);
         assert_eq!(accepted_order.price(), Some(updated_price));
+    }
+
+    #[rstest]
+    fn test_limit_order_rejects_invalid_update_atomically() {
+        let order = OrderTestBuilder::new(OrderType::Limit)
+            .instrument_id(InstrumentId::from("BTC-USDT.BINANCE"))
+            .quantity(Quantity::from(10))
+            .price(Price::new(100.0, 2))
+            .build();
+        let mut accepted_order = TestOrderStubs::make_accepted_order(&order);
+        let state = (
+            accepted_order.status(),
+            accepted_order.previous_status(),
+            accepted_order.ts_last(),
+            accepted_order.events().len(),
+        );
+        let event = OrderUpdated {
+            client_order_id: accepted_order.client_order_id(),
+            strategy_id: accepted_order.strategy_id(),
+            trigger_price: Some(Price::new(95.0, 2)),
+            ..Default::default()
+        };
+
+        let result = accepted_order.apply(OrderEventAny::Updated(event));
+
+        assert!(matches!(result, Err(OrderError::InvalidOrderEvent)));
+        assert_eq!(accepted_order.status(), state.0);
+        assert_eq!(accepted_order.previous_status(), state.1);
+        assert_eq!(accepted_order.ts_last(), state.2);
+        assert_eq!(accepted_order.events().len(), state.3);
     }
 
     #[rstest]
--- a/crates/model/src/orders/market.rs
+++ b/crates/model/src/orders/market.rs
@@ -565,7 +565,9 @@ mod tests {
         enums::{OrderSide, OrderType, TimeInForce},
         events::{OrderEventAny, OrderUpdated, order::spec::OrderInitializedSpec},
         instruments::{CurrencyPair, stubs::*},
-        orders::{MarketOrder, Order, builder::OrderTestBuilder, stubs::TestOrderStubs},
+        orders::{
+            MarketOrder, Order, OrderError, builder::OrderTestBuilder, stubs::TestOrderStubs,
+        },
         types::{Price, Quantity},
     };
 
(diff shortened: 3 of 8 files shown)


cjdsellers:
Hey @folknor

Thanks for tracking these failures down. The individual fixes look sound, but I don't think the validation boundary is safe as written.

`validate_update` runs before `OrderCore::apply`, so an update with a wrong client or strategy ID plus an unsupported field reports `InvalidOrderEvent` instead of the identity mismatch. The change also makes validation optional through a new public defaulted trait method while removing the existing guards from public `update` methods, allowing direct invalid updates to mutate subtype state silently.

Could you move the update-field validation into the core preflight, after identity checks and before any mutation, and retain a defensive guard on the low-level mutation path? `OrderCore` already knows `order_type`, so this should also remove the repeated validation call from all nine concrete `apply` methods.

I'd also prefer the `PendingUpdate`/`Submitted` parity and stop-market `protection_price` change split into follow-ups. Both look correct, but they are independent of atomic application.

folknor:
Thanks, that is the right call on the boundary. Reworked:

- Deleted the `validate_update` trait method. The update-field check now lives in `OrderCore::apply` as an exhaustive `match` on `order_type`, after the client-order-id/strategy-id identity checks and the `OrderStatus::transition` check and before the first mutation. An update that is both mis-addressed and field-invalid now reports the identity mismatch, and one that is field-invalid in a status that rejects updates reports `InvalidStateTransition` - the field check never masks a more fundamental rejection. The match has no wildcard arm, so a future `OrderType` is a compile error rather than a silently-accepted one.
- Dropped the validation call from all nine concrete `apply` methods; the three permissive types are back to unchanged.
- Restored the hard `assert!` on the six restrictive `update()` methods as the defensive guard on the low-level mutation path, so a direct `update()` that bypasses `apply` still trips.
- Added tests: identity mismatch (wrong client / wrong strategy) plus an invalid field surfaces the identity `PredicateViolation` naming the mismatched field; invalid transition plus an invalid field surfaces `InvalidStateTransition`; a direct `update()` with an invalid field panics before the quantity write.

Split the `(PendingUpdate, Submitted)` parity and the stop-market `protection_price` change out into separate follow-ups as suggested; this PR is now just the atomic-application boundary. I'll file those one by one after this lands.

cjdsellers (approved):
Hey @folknor

I reviewed the rework and found no blocking issues, solid improvement. The validation now runs at the right boundary without masking identity or transition errors, and the low-level guards remain in place. This looks good to merge.

PR #4531, opened 2026-07-22
Replace timer-driven mock stream teardown in Betfair tests

folknor:
The Betfair client test suite spends most of its runtime asleep: 256 of its 307 seconds are mock stream servers waiting out fixed delays that nothing is waiting for.

## Mock stream teardown is driven by a timer instead of by the test

Each test in `tests/exec_client.rs` and `tests/data_client.rs` spawns a mock stream server whose last act is a fixed sleep followed by closing the socket:

```rust
let server = tokio::spawn(async move {
    let (_reader, write_half) = accept_and_auth(&listener).await;
    tokio::time::sleep(Duration::from_secs(5)).await;
    drop(write_half);
});
```

and ends with `let _ = server.await`. The test body performs its work and assertions in well under a second, then blocks on that timer. Nothing observes the socket close: no assertion depends on it, and the four tests that exercise reconnection drive it with a synthetic second `Connection` frame over the still-open socket, which `BetfairExecutionClient` treats as a reconnect regardless of transport state. The delay exists only to keep the peer alive long enough for the test to finish, which is a signal, not a duration.

Because it is a duration, it is also a latent flake. Every test disconnects its client before joining the server, so today the socket closes after the client is gone. A test body that runs longer than its keepalive - under a loaded CI runner, say - instead gets transport EOF mid-test, which the client reads as a dropped connection and can turn into a spurious reconnect. The failure would appear as an unrelated assertion failing intermittently.

`test_cancel_all_orders_sends_request` measures 5.20s unchanged and 0.37s with its keepalive shortened, so the timer accounts for essentially the whole runtime.

## Replacing the timer with an explicit completion signal

The mock server now waits on a `tokio::sync::oneshot` instead of the clock:

```rust
let (server_done_tx, server_done_rx) = tokio::sync::oneshot::channel();

let server = tokio::spawn(async move {
    let (_reader, write_half) = accept_and_auth(&listener).await;
    let _ = server_done_rx.await;
    drop(write_half);
});

// ... test body and assertions ...

client.disconnect().await.unwrap();
let _ = server_done_tx.send(());
server.await.unwrap();
```

The signal is sent after the client's own cleanup and after the final assertion, so the socket still closes with the client already gone - the ordering the timer was approximating - but now by construction rather than by racing a deadline. Sender drop on a failing test releases the receiver, so a panicking test does not leave the mock parked.

The join also becomes `server.await.unwrap()`. Previously `let _ = server.await` discarded the mock task's result, so a panic inside a mock server - an `unwrap` on a malformed frame, a failed write - was swallowed and the test passed anyway. That is now a test failure.

## What is converted, and what is deliberately not

The rule is positional, not by duration: a sleep immediately preceding `drop(write_half)` as the final act of the spawned task is keepalive and is converted, whatever its constant. 75 sites qualify - 61 in `tests/exec_client.rs` (21 at 2s, one at 3s, 37 at 5s, one at 10s, one at 15s), 13 in `tests/data_client.rs`, plus the accept-loop timeout in `test_data_client_connect_is_idempotent`, which is stopped by the same signal.

Sleeps that sit mid-task are observation windows and are untouched: the 150ms gaps between the two synthetic `Connection` frames in the reconnect-queue tests, the fixture-send delays, the three inline one-second negative-event waits, the 200ms second-connect window in the idempotence test, and the `betting_response_delays` entries that make an HTTP method slow on purpose so a reconciliation can be aborted mid-flight. `test_data_client_deduplicates_same_market_subscription` keeps its sleep as well, because its server returns an observed value to the test after the drop and so is not a terminal task; its absence check is the preceding 500ms read timeout.

No production code is touched.

## Testing

The change is to test scaffolding only; every assertion is unchanged, so the existing suite is the regression test and the differential is runtime.

The Betfair suite drops from 307.32s to 51.62s, a reduction of 255.70s. The two slowest tests in the suite, both of which set a floor on their test lane, drop furthest: `test_disconnect_during_reconciliation_clears_halt` from 15.09s to 0.32s, and `test_queued_reconnect_re_asserts_halt` from 10.28s to 1.90s. A full workspace test run is about 45s faster; the gap between that and the 255.70s is parallel overlap, since sleeping tests consume no CPU and interleave - the saving is proportionally larger the narrower the lane, which is the CI case.

### DIFF
--- a/crates/adapters/betfair/tests/data_client.rs
+++ b/crates/adapters/betfair/tests/data_client.rs
@@ -76,9 +76,11 @@ async fn test_data_client_connect_disconnect() {
     let (stream_port, listener) = start_mock_stream().await;
     let (mut client, _rx) = create_test_data_client(addr, stream_port);
 
+    let (server_done_tx, server_done_rx) = tokio::sync::oneshot::channel();
+
     let server = tokio::spawn(async move {
         let (_reader, write_half) = accept_and_auth(&listener).await;
-        tokio::time::sleep(Duration::from_secs(3)).await;
+        let _ = server_done_rx.await;
         drop(write_half);
     });
 
@@ -89,7 +91,8 @@ async fn test_data_client_connect_disconnect() {
     client.disconnect().await.unwrap();
     assert!(client.is_disconnected());
 
-    let _ = server.await;
+    let _ = server_done_tx.send(());
+    server.await.unwrap();
 }
 
 #[rstest]
@@ -99,9 +102,11 @@ async fn test_data_client_emits_instruments_on_connect() {
     let (stream_port, listener) = start_mock_stream().await;
     let (mut client, mut rx) = create_test_data_client(addr, stream_port);
 
+    let (server_done_tx, server_done_rx) = tokio::sync::oneshot::channel();
+
     let server = tokio::spawn(async move {
         let (_reader, write_half) = accept_and_auth(&listener).await;
-        tokio::time::sleep(Duration::from_secs(3)).await;
+        let _ = server_done_rx.await;
         drop(write_half);
     });
 
@@ -121,7 +126,8 @@ async fn test_data_client_emits_instruments_on_connect() {
     );
 
     client.disconnect().await.unwrap();
-    let _ = server.await;
+    let _ = server_done_tx.send(());
+    server.await.unwrap();
 }
 
 #[rstest]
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Nice work @folknor, this is a clean change. The positional rule for the conversion reads well, and making the joins `server.await.unwrap()` so mock panics actually surface is a good catch on top of the speedup.

There is a small join with `let _ = server.await;` remaining I'll tidy up locally after merge!

PR #4536, opened 2026-07-22
Parse Bybit commision currency from WebSocket stream

dxwil:
# Pull Request

## Summary

Fixes the hardcoded commission currency (`instrument.quote_currency()`) in the two Bybit WS execution paths (`dispatch.rs` tracked-order OrderFilled and `websocket/parse.rs` fill-report). These now read `feeCurrency` adaptively from the payload, following the already-correct HTTP reconciliation path in `common/parse.rs`. This matters because the fee currency can genuinely differ from the quote currency, e.g. spot fills charge the fee in the received asset (base on a buy, quote on a sell), so the quote-only assumption mislabeled commissions.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Testing

Also verified live that Bybit returns this field.

### DIFF
--- a/crates/adapters/bybit/src/websocket/dispatch.rs
+++ b/crates/adapters/bybit/src/websocket/dispatch.rs
@@ -55,7 +55,7 @@ use crate::{
     common::{
         enums::BybitOrderStatus,
         parse::{
-            bybit_rejection_due_post_only, make_bybit_symbol, parse_millis_timestamp,
+            bybit_rejection_due_post_only, get_currency, make_bybit_symbol, parse_millis_timestamp,
             parse_price_with_precision, parse_quantity_with_precision,
         },
     },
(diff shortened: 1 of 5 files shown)


dxwil:
Build failed because it ran out of storage, rerunning it now.

cjdsellers:
Hey @dxwil,

Thanks for the PR. The core change looks right: it matches the HTTP reconciliation path in `common/parse.rs` (which already reads `feeCurrency` via `get_currency`), and the required-field treatment mirrors `BybitExecution` in `http/models.rs`.

**Some improvements we can make before merge:**

The two new assertions don't actually exercise the bug being fixed. Both fixtures are BTCUSDT linear, where `feeCurrency=USDT` equals `quote_currency=USDT`, so the assertions pass with the old `instrument.quote_currency()` code as well. The divergent case this PR exists for (fee currency != quote, e.g. spot buys charged in base) is not covered, and the `dispatch.rs::parse_order_filled` path has no commission-currency assertion at all.

Could you add a case where `feeCurrency` differs from the quote currency (e.g. set `feeCurrency=BTC` on a fixture copy, or a spot buy) and assert the resulting currency on both `parse_ws_fill_report` and `parse_order_filled`? That makes the fix provable and catches future regressions.

Minor (non-blocking): `parse_ws_fill_report_fast` still uses `instrument.quote_currency()` for the zero commission. That's expected since the `execution.fast` payload omits fee data, but a one-line comment there would stop a future copy-paste from "fixing" it.

cjdsellers (approved):
Thank you for the follow-ups @dxwil 👌

PR #4429, opened 2026-07-09
Fix VWAP day rollover for nanosecond timestamps

akashchakrabortymsc-cmd:
Fix : #4428.

SECONDS_PER_DAY (86_400.0) was compared against ts_init which is in nanoseconds, causing spurious resets every ~86 µs instead of once per day.

Fix: replace with NANOSECONDS_PER_DAY = 86_400.0 * 1_000_000_000.0

Also update all test constants and parametric case timestamps to nanosecond scale, and add regression test 

# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [x] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Type of change

- [x] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [ ] Improvement (non-breaking)
- [ ] Breaking change (impacts existing behavior)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

- [x] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)

## Release notes

- [ ] I added a concise entry to `RELEASES.md` that follows the existing conventions (when applicable)

## Testing

**Ensure new or changed logic is covered by tests.**

- [ ] Affected code paths are already covered by the test suite
- [ ] I added/updated tests to cover new or changed logic

PR #4537, opened 2026-07-22
fix(polymarket): replace Price::from with fallible path to prevent panic on excess tick precision

### LINKED ISSUE #4534: [Polymarket] Gamma tick precision can panic before validation
# Bug Report

## Confirmation

- [x] I've tested this issue using a recent pre-release or development wheel and it still reproduces.

Source-level reproduction was run against NautilusTrader revision
`47a4498cc649b6466eca4294a023af552771d352`. Reproduction through an official
development wheel remains to be completed.

### Expected Behavior

`Gamma` `tick_size` is external venue data and should be handled as untrusted
input. If its decimal precision cannot be represented by NautilusTrader's
configured fixed precision, the Polymarket adapter should return an error that
reaches the existing fail-soft boundary, log the invalid instrument, and skip
that instrument while continuing to process valid instruments.

In particular, `create_instrument_from_def` and
`rebuild_instrument_with_tick_size` should not panic on an unsupported tick
precision.

### Actual Behavior

In the tested NautilusTrader source revision, both affected paths call the
infallible conversion `Price::from(...)`. That conversion calls `expect(...)`
internally, so it can panic before the newer checked-bounds code runs:

- `crates/adapters/polymarket/src/http/parse.rs`: the instrument creation path
  constructs `Price` at line 189 before calling `tick_relative_price_bounds`
  at line 206.
- `crates/adapters/polymarket/src/http/parse.rs`: the tick-size rebuild path
  constructs `Price` at line 279 before applying the checked bounds.
- `crates/model/src/types/price.rs`: `Price::from(...)` calls `expect(...)` at
  line 596.

With the default Polymarket high-precision configuration,
`FIXED_PRECISION` is 16. A synthetic Gamma tick size with 19 decimal places
therefore produces this panic instead of an adapter error or skipped
instrument:

```text
thread '...' panicked at crates/model/src/types/price.rs:596:40:
Condition failed: "`precision` exceeded maximum `FIXED_PRECISION` (16), was 19"
```

The checked `tick_relative_price_bounds(...)` helper does not prevent the
panic because it is reached after the infallible conversion. In a production
build configured to abort on panic, this can terminate the process rather than
contain the malformed instrument.

1. Check out NautilusTrader revision
   `47a4498cc649b6466eca4294a023af552771d352`.
2. In the Polymarket parser tests, load a valid Gamma market fixture.
3. Set its `order_price_min_tick_size` to a value with more than 16 decimal
   places, for example:

   ```rust
   invalid_market.order_price_min_tick_size =
       Some(0.000_000_000_000_000_000_1);
   ```

4. Parse the market and pass the resulting definitions to
   `instruments_from_defs(&defs, ts_init)`.
5. Observe a panic while converting the tick size, instead of an error and a
   warning/skip for the invalid instrument.
6. The same behavior can be reproduced by constructing a valid instrument and
   calling:

   ```rust
   rebuild_instrument_with_tick_size(
       &instrument,
       "0.0000000000000000001",
       ts_init,
       ts_init,
   );
   ```

The reproduction uses a synthetic malformed tick size to exercise the
conversion boundary. It does not claim that a live Gamma response has emitted
this exact 19-decimal value.

### Code Snippets or Logs

The targeted upstream tests reproduced the same panic in both the creation and
rebuild paths:

```text
Condition failed: "`precision` exceeded maximum `FIXED_PRECISION` (16), was 19"
```

The targeted test command was run with `RUSTFLAGS='-A dead_code'` only to work
around unrelated upstream dead-code errors in `crates/live/src/runner.rs` that
otherwise prevented the test binary from building. The panic itself is from
the Polymarket parser's normal `Price::from(...)` execution path.

### Specifications

- OS platform: macOS on Apple Silicon (`aarch64-apple-darwin`)
- Python version: N/A; this is the Rust Polymarket adapter path
- `nautilus_trader` version: `2.0.0rc1`, source revision
  `47a4498cc649b6466eca4294a023af552771d352`
- Build configuration: Polymarket default `high-precision`, with
  `FIXED_PRECISION = 16`

akashchakrabortymsc-cmd:
Price::from calls expect() internally and panics when the decimal precision of the tick size exceeds FIXED_PRECISION (16). Both create_instrument_from_def and rebuild_instrument_with_tick_size called Price::from before tick_relative_price_bounds, so a malformed Gamma tick size with >16 decimal places could terminate the process.

Fix: move tick_relative_price_bounds before price_increment construction and replace Price::from with Price::from_decimal which returns a Result. The existing fail-soft boundary in instruments_from_defs now correctly logs and skips the invalid instrument.

Fixes #4534

# Pull Request

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Testing

### DIFF
--- a/crates/adapters/polymarket/src/http/parse.rs
+++ b/crates/adapters/polymarket/src/http/parse.rs
@@ -186,7 +186,6 @@ pub fn create_instrument_from_def(
     let raw_symbol = Symbol::new(def.token_id);
     let currency = get_currency(PUSD);
 
-    let price_increment = Price::from(def.tick_size.to_string());
     let size_increment = Quantity::from("0.000001");
 
     let activation_ns = def
@@ -204,6 +203,8 @@ pub fn create_instrument_from_def(
     // these bounds land inside the venue's `[tick, 1 - tick]` range; execution-side validation in
     // `PolymarketOrderBuilder::validate_limit_price` remains the source of truth.
     let (min_price, max_price) = tick_relative_price_bounds(def.tick_size)?;
+    let price_increment = Price::from_decimal(def.tick_size)
+        .map_err(|e| anyhow::anyhow!("Invalid tick_size '{}': {e}", def.tick_size))?;
     // Polymarket exposes `orderMinSize` (limit-order minimum shares) and a separate
     // $1 market-order minimum amount; the instrument model can only carry one
     // `min_quantity`, so leave it unset and let the venue reject out-of-bounds orders.
@@ -276,8 +277,9 @@ pub fn rebuild_instrument_with_tick_size(
         .parse()
         .map_err(|e| anyhow::anyhow!("Failed to parse tick size '{new_tick_size}': {e}"))?;
     let price_precision = tick_size.scale() as u8;
-    let price_increment = Price::from(tick_size.to_string());
     let (min_price, max_price) = tick_relative_price_bounds(tick_size)?;
+    let price_increment = Price::from_decimal(tick_size)
+        .map_err(|e| anyhow::anyhow!("Invalid tick_size '{tick_size}': {e}"))?;
 
     let rebuilt = BinaryOption::new_checked(
         bo.id,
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @akashchakrabortymsc-cmd, thanks for the PR and the correct root-cause read on this one.

I'd already fixed #4534 in 440cb43 before this opened, using a slightly different approach, so I'm going to close this rather than merge it.

A few things worth flagging for next time:

- The fix is right, but `tick_relative_price_bounds` already converts the tick to a `Price` and returns it as `min_price` (the lower bound is the tick itself), so a second `Price::from_decimal` is redundant. Reusing `min_price` as `price_increment` drops the duplicate conversion and its error path.
- `assert!(result.is_err())` doesn't tell you *how* it failed. The behavior that actually matters is the fail-soft path in `instruments_from_defs`, where the bad def is skipped and the valid ones still come through. I'd drive that end to end and assert exact error strings on the rebuild path.
- The pre-commit check failed on this PR. As requested on #4429, running `make pre-commit` (or `prek run --all-files`) before pushing catches this, and `prek install` sets the hook up once so it runs automatically before each commit.
- On commit titles, the repo uses plain sentence-case imperatives (`Fix Polymarket unsupported tick precision`) rather than conventional commits. Across ~20K commits there's essentially no `fix(scope):` style conventional commits, so best to conform to the established commit title style.

PR #4538, opened 2026-07-22
Make the Binance Spot WS trading setup timeout configurable

folknor:
A follow-up to the test-timing work in #4532, on the same class of production timing constants that no consumer can reach.

## The WS trading setup deadlines are hardcoded

`BinanceSpotExecutionClient::connect` bounds each WebSocket trading setup phase with a literal `Duration::from_secs(10)`: one for the `session.logon` response, one for `userDataStream.subscribe`. `BinanceSpotWsTradingClient::session_logon` only enqueues the command on the handler channel and returns, so the response arrives asynchronously and is signalled through `ws_authenticated`, `ws_user_data_subscribed`, or the setup-error channel, which `wait_for_ws_setup_response` races against the deadline. A venue that accepts the socket and then never answers holds `connect()` for the full 10 seconds per phase - up to 20 seconds across both - before the client falls back to HTTP-only execution, and no operator can shorten that for a fast-failover deployment or lengthen it for a proxied or high-latency route.

This adds `BinanceExecClientConfig::ws_trading_setup_timeout_ms`, defaulting to 10000 so current behavior is unchanged. `connect()` derives one `Duration` and passes it to both waits independently, preserving the existing per-phase budget rather than collapsing the two into a single deadline. `validate()` rejects zero, which would make setup a scheduler race that almost always enters HTTP-only mode for the process lifetime while logging a timeout. No upper bound is imposed: the `recv_window_ms` ceiling is a Binance protocol constraint rather than a general timeout precedent, and a setup deadline may legitimately exceed a minute on an unusual route.

## The user data subscribe deadline was never exercised

The second deadline had no test that reached it. `WsSetupBehavior` offered `RejectUserDataSubscribe`, which sends an error the receiver wins immediately, but no behavior that leaves the subscribe unanswered, so every existing test left that wait through the error path and the timeout branch was dead in coverage. `IgnoreUserDataSubscribe` records the method and then does nothing, driving the second wait to expiry the way `IgnoreSessionLogon` already did for the first.

## Testing

`test_ws_trading_auth_timeout_uses_http_only_mode` and the new `test_ws_trading_user_data_subscribe_timeout_uses_http_only_mode` pin the same property at each deadline: an unanswered setup phase expires, the client enters HTTP-only mode, and a subsequent order is accepted over HTTP. Both configure `ws_trading_setup_timeout_ms` through the new field, so they exercise the knob rather than asserting around it. Both were verified to fail against the unfixed code - with the call sites restored to `Duration::from_secs(10)`, each exceeds its guard and fails with `Elapsed`. Together they take about 0.7s where the auth test alone previously took 10.1s.

The two rejection tests are unchanged and keep the production default, so the error paths still prove they return without waiting on a deadline.

`test_exec_config_rejects_zero_ws_trading_setup_timeout` pins the validation message, and the TOML and PyO3 construction tests pin both the 10000 default and an explicit override.

The setup tests use a 250 ms timeout rather than a smaller value deliberately. The deadline starts when the command is queued on the handler channel, not when the frame reaches the socket, so the budget must still cover signing, the WebSocket send, and the mock server recording the method; a tighter value can expire before that record under parallel load and fail the received-method assertion. No test waits on wall-clock time beyond its own configured deadline.

### DIFF
--- a/crates/adapters/binance/src/config.rs
+++ b/crates/adapters/binance/src/config.rs
@@ -317,6 +317,9 @@ pub struct BinanceExecClientConfig {
     /// Whether to use the WebSocket trading API for order operations (Spot and USD-M Futures).
     #[builder(default = true)]
     pub use_ws_trading: bool,
+    /// Timeout in milliseconds for each Binance Spot WS trading setup response.
+    #[builder(default = 10_000)]
+    pub ws_trading_setup_timeout_ms: u64,
     /// Instrument loading and fee configuration.
     #[builder(default)]
     pub instrument_provider: BinanceInstrumentProviderConfig,
@@ -396,6 +399,7 @@ nautilus_core::impl_pyo3_config_getters!(BinanceExecClientConfig {
     base_url_ws: Option<String>,
     base_url_ws_trading: Option<String>,
     use_ws_trading: bool,
+    ws_trading_setup_timeout_ms: u64,
     instrument_provider: BinanceInstrumentProviderConfig,
     instrument_refresh_interval_secs: u64,
     use_gtd: bool,
@@ -423,9 +427,15 @@ impl BinanceExecClientConfig {
     ///
     /// # Errors
     ///
-    /// Returns an error for invalid receive-window, provider, or Binance US settings.
+    /// Returns an error for invalid receive-window, WS trading setup timeout, provider, or
+    /// Binance US settings.
     pub fn validate(&self) -> anyhow::Result<()> {
         validate_recv_window(self.recv_window_ms)?;
+        anyhow::ensure!(
+            self.ws_trading_setup_timeout_ms > 0,
+            "ws_trading_setup_timeout_ms must be greater than 0, was {}",
+            self.ws_trading_setup_timeout_ms
+        );
         self.instrument_provider.validate(self.product_type)?;
 
         if self.us {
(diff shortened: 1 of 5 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR, the timeout change itself looks good.

When you get a chance could you run `make py-stubs-v2` and commit the regenerated files? The new `ws_trading_setup_timeout_ms` constructor argument and getter come from the Rust/PyO3 metadata, but the committed Binance `.pyi` is still stale.

Stub regeneration isn't suitable for `pre-commit` because it synchronizes the full v2 Python environment and compiles and runs the Rust stub generator, which would make every hook run too heavy. The PR pre-commit job therefore won't catch this, but the `build-v2` workflow on `develop` regenerates the stubs and checks for drift, so it will fail after merge if the generated files aren't committed.

As a tip, `make pre-flight-v2` is the best local QA gate for v2 changes before pushing. It includes stub regeneration and the generated-drift check, along with the broader v2 checks (I'll be updating the contributor guide shortly).

folknor:
@cjdsellers Thank you. A new `python/nautilus_trader/adapters/binance/__init__.pyi` has been pushed onto the branch.

I'll stay away from touching `crates/pyo3/**` or `crates/**/src/python/**/*.rs` from now on.

cjdsellers:
Hey @folknor,

The stub update looks good, thanks for pushing that.

On the bindings, don't hold back from editing `crates/**/src/python/**/*.rs` - those are "hand-written", so anything that adds or changes a Python-facing field, getter, or signature needs a matching edit there. The way you wired `ws_trading_setup_timeout_ms` through `src/python/config.rs` in this PR is exactly right. For purely internal Rust changes that don't reach Python, you're fine to leave them alone.

The only step that's easy to miss is regenerating the `.pyi` afterward so the stubs stay in sync. `make pre-flight-v2` before pushing covers both that and the drift check; if you just need the stubs on their own, `make py-stubs-v2` is the lighter command.

cjdsellers (approved):

PR #4539, opened 2026-07-22
Collect Polymarket data events through the correlated response

folknor:
A follow-up to the test-timing work in #4532 and #4538, on the same class of fixed waits in adapter tests.

## The event drain cannot return early

`drain_data_events` in `crates/adapters/polymarket/tests/data_client.rs` collects events with `tokio::time::timeout_at(deadline, rx.recv())` in a loop, so it only returns once the deadline expires. Every call therefore costs its full window no matter when the events actually arrive. There are thirteen call sites, twelve at five seconds and one at 500 ms, which is 60.5 seconds of serial waiting built into the file, plus a `tokio::time::sleep(Duration::from_millis(300))` used as an absence check in `test_reset_reconnect_does_not_replay_stale_market_subscriptions`.

The correlated `DataResponse` is already a deterministic terminal marker for these requests, so the tests can collect through it instead of through a quiet period. `collect_data_events_until_response` receives until it sees a `DataEvent::Response` carrying the request's correlation ID, then sweeps anything already queued with `try_recv()`. Callers that assert on the events keep the returned vector unchanged; callers that only prime the instrument cache discard it. The timeout argument survives as a failure deadline that a passing run never reaches.

This does not relax the exact-count assertions. `request_instrument` sends `DataEvent::Instrument` and then `DataEvent::Response` with no await between the two sends, and the unbounded channel preserves that order, so the correlated response guarantees any publish is already queued and the sweep observes it. The other request paths emit only their response. The assertions do narrow in one respect worth stating plainly: `exactly one` now means exactly one through the terminal response boundary rather than none during a subsequent five second window, since `try_recv` is not a general quiescence proof. No producer emits anything after its correlated response, and a duplicate publish occurs before it.

`test_request_instrument_not_found_emits_no_publish` keeps its 500 ms window and gains a comment saying why: the missing-instrument path logs and returns without sending a response, so it has no terminal marker to collect through. It is the only remaining wall-clock wait in the file, and it is deliberate.

## The reconnect test slept instead of fencing

`test_reset_reconnect_does_not_replay_stale_market_subscriptions` asserted that reset plus reconnect sends no stale subscription, but it established that by sleeping 300 ms and then counting payloads - which proves nothing about whether a replay was still in flight.

`connect_client` performs the WebSocket connect, which queues any connect-time replay, and only then sends the new-market discovery subscribe, an empty `assets_ids` subscription issued under the pool wire mutex. Handler commands and frames are ordered, so a stale replay must reach the mock before that discovery payload. The test now enables `subscribe_new_markets`, waits until the mock has observed the second discovery payload, and asserts the non-empty subscription count is still exactly one. A stale replay cannot hide among the discovery payloads because it can never be empty: subscription state records only supplied asset IDs, fresh-connect replay is guarded on a non-empty topic set, and reconnect replay returns early when the ID list is empty.

## Testing

No test cases were added; the existing assertions are the regression, and every one is preserved with its original strength.

The risk in a change like this is that the tests keep passing while quietly ceasing to test anything, so the exactness was verified directly: with a second `DataEvent::Instrument` send added to `request_instrument`, `collect_data_events_until_response` returns `[Instrument, Instrument, Response]` and `assert_eq!(publish_count, 1)` fails with 2. It fails in 0.013 s rather than after a five second window.

The package goes from 2m16s to 1m26s over its 1108 tests. `test_reset_reconnect_does_not_replay_stale_market_subscriptions` goes from 10.63 s to 0.43 s, and the four other former ten-second members no longer appear among the slowest tests. `test_request_instrument_not_found_emits_no_publish` remains at roughly 0.5 s, which is its deliberate window.

### DIFF
--- a/crates/adapters/polymarket/tests/data_client.rs
+++ b/crates/adapters/polymarket/tests/data_client.rs
@@ -213,6 +213,16 @@ fn create_test_data_client(
     PolymarketDataClient,
     tokio::sync::mpsc::UnboundedReceiver<DataEvent>,
 ) {
+    create_test_data_client_with_new_markets(addr, false)
+}
+
+fn create_test_data_client_with_new_markets(
+    addr: SocketAddr,
+    subscribe_new_markets: bool,
+) -> (
+    PolymarketDataClient,
+    tokio::sync::mpsc::UnboundedReceiver<DataEvent>,
+) {
     let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
     // Use replace_ rather than set_ so this test can run on a thread that
     // already had a sender installed by another test in the same harness.
@@ -225,7 +235,7 @@ fn create_test_data_client(
     let data_api = PolymarketDataApiHttpClient::new(Some(base_url.clone()), 5).expect("data_api");
     let ws = PolymarketMarketConnectionPool::new(
         Some(format!("ws://{addr}/ws/market")),
-        false,
+        subscribe_new_markets,
         TransportBackend::default(),
         WS_DEFAULT_SUBSCRIPTIONS,
     );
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor

Thanks for the PR. I checked the event and WebSocket ordering and found no blockers. All required checks pass. Will merge shortly.

PR #4541, opened 2026-07-22
Fail BitMEX WebSocket authentication on rejection instead of waiting out the timeout

folknor:
A follow-up to the WebSocket setup-timeout work in #4538, on the same authentication seam in another adapter.

## An authentication rejection is dropped instead of failing the login

BitMEX answers a failed `authKeyExpires` with an error frame that echoes the originating request. `BitmexWsFeedHandler` matches that as `BitmexWsFrame::Error`, logs it, and falls through without returning the message, so it never reaches the event loop. `auth_tracker.fail` is reachable only from `handle_subscription_message`, which requires a subscription-shaped frame whose request `op` is `authKeyExpires`. The rejection therefore never resolves the receiver created in `authenticate`, and `connect` blocks until the authentication timeout expires before returning a generic timeout error. An operator with a revoked or mistyped key sees a ten second hang and no reason, and the client cannot distinguish a rejected key from a venue that has gone silent.

`BitmexWsFrame::Error` already carries `request: BitmexHttpRequest`; the arm was discarding it. It is now bound, and when its `op` is `authKeyExpires` the tracker is failed with the venue's own error text, so `connect` returns that message immediately. The frame is still dropped afterwards: it has been consumed as a control message and should not also surface as data.

This covers the echoed-request rejection. `BitmexWsFrame::Error` requires `status`, `meta` and `request`, so a bare `{"error": ...}` payload, an envelope without `meta`, or an outright connection close does not take this branch and still resolves through the timeout or the reconnect path. The timeout remains the necessary fallback rather than a redundant one.

## The authentication timeout cannot be bounded by the caller

The wait used the shared `AUTHENTICATION_TIMEOUT_SECS` from `nautilus-network` directly, so no consumer could shorten it for a fast-failover deployment or extend it for a slow route, and the genuinely silent-venue case had no way to be exercised quickly in a test.

`BitmexWebSocketClient` now takes `auth_timeout_secs: Option<u64>` and stores `auth_timeout_secs.unwrap_or(AUTHENTICATION_TIMEOUT_SECS)`, matching how `OKXWebSocketClient` already parameterises the same constant. The option is threaded through `new`, `new_with_env` and the PyO3 constructor; existing callers pass `None` and the shared constant is unchanged, so default behaviour is identical.

## Testing

`test_login_failure_emits_error` previously discarded the result of `connect` and asserted only that the session was not authenticated, so it did not verify what its name claimed and passed while the rejection was being dropped. It now asserts that `connect` returns the venue's `Authentication failed` message, and it passes on the fixed code without any shortened timeout. Verified to fail against the unfixed handler, where it falls through to the timeout path and takes ten seconds.

`test_login_times_out_when_server_is_silent` is new and covers the remaining slow path: the server ignores the auth request entirely, the client is constructed with a one second `auth_timeout_secs`, and the timeout error is asserted. It passes against both the fixed and unfixed handler, which is the point, since it exercises a path the rejection fix does not touch.

`test_reconnection_waits_for_delayed_auth_ack` withheld the auth acknowledgement with a fixed three second delay. The mock server now gates the acknowledgement on a semaphore, and because the server previously blocked inside its only receive loop while delaying, its assertion that no subscription had resumed could not have observed a premature one: the frames would have queued unread in the socket. The socket is now split so the receive loop keeps reading while the acknowledgement is held, and any premature subscription is recorded and fails the assertion. Confirmed by making the client subscribe without waiting for the acknowledgement, which the repaired test detects and the original did not.

No test waits on wall-clock time except the deliberate one second silent-server timeout.

### DIFF
--- a/crates/adapters/bitmex/bin/ws_data.rs
+++ b/crates/adapters/bitmex/bin/ws_data.rs
@@ -80,6 +80,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         None,   // No API secret
         None,   // Account ID
         5,      // 5 second heartbeat
+        None,   // Default authentication timeout
         TransportBackend::default(),
         None,
     )
--- a/crates/adapters/bitmex/bin/ws_exec.rs
+++ b/crates/adapters/bitmex/bin/ws_exec.rs
@@ -59,6 +59,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         None,
         None,
         5, // 5 second heartbeat
+        None,
         TransportBackend::default(),
         None,
     )
--- a/crates/adapters/bitmex/src/data.rs
+++ b/crates/adapters/bitmex/src/data.rs
@@ -650,6 +650,7 @@ impl DataClient for BitmexDataClient {
                 self.config.api_secret.clone(),
                 None,
                 self.config.heartbeat_interval_secs.unwrap_or(5),
+                None,
                 self.config.environment,
                 self.config.transport_backend,
                 self.config.proxy_url.clone(),
--- a/crates/adapters/bitmex/src/execution.rs
+++ b/crates/adapters/bitmex/src/execution.rs
@@ -152,6 +152,7 @@ impl BitmexExecutionClient {
             config.api_secret.clone(),
             Some(account_id),
             config.heartbeat_interval_secs,
+            None,
             config.environment,
             config.transport_backend,
             config.proxy_url.clone(),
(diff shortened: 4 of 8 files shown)


cjdsellers (approved):
Hey @folknor,

This looks good to merge, nice catch on the dropped rejection.

One thing I'll follow up on locally: `auth_timeout_secs` doesn't reach the data and execution client configs yet, only the low-level client. It mirrors OKX so I'm not holding this up, but I'll thread it through on my side.

PR #4543, opened 2026-07-22
Auto-repay Bybit SPOT margin borrows

dxwil:
# Pull Request

## Summary

Adds optional automatic repayment of SPOT margin borrows for the (Rust/v2) Bybit adapter. When a SPOT **BUY** fully fills, i.e. a short opened by borrowing the base coin has been closed; the adapter repays the outstanding borrow so interest stops accruing, without the strategy having to manage it.

### How it works
- New `auto_repay_spot_borrows` config flag on `BybitExecClientConfig` (**default `false`**).
- When enabled, the execution client spawns a background consumer on connect and tears it down on disconnect/stop.
- The WS dispatch path enqueues a repay request whenever a fully-filled (`leaves_qty == 0`) SPOT BUY is seen (`RepayRequest { coin, quantity }` over an unbounded channel).
- The consumer, for each request: queries the current borrow liability, caps the amount at `min(outstanding, bought)`, and repays. The `outstanding == 0` early-out skips buys that aren't closing a borrow.

### Relation to the existing Python adapter
This mirrors the auto-repay the Python adapter already had, with two corrections:
- **Correct blackout timing.** Bybit blocks repayment during a recurring window at **mm:04:00-mm:05:30 every hour**, not 04:00–05:30 once per day (this is probably something Bybit changed since then and now). The Python check used the daily interpretation; this uses the correct per-hour window.
- **Waits it out instead of failing.** When a repay lands inside the blackout, the Python path skipped/failed the repay. Here the consumer computes the remaining window and **sleeps until it clears**, then repays - so the borrow is still paid down rather than left open until the next fill.

### Why the *converting* repay (not no-convert)
Bybit charges the trading fee for a SPOT BUY **in the base coin**, the coin you just bought. So the amount actually credited is slightly less than the amount borrowed. The no-convert repay endpoint only draws on that coin's own spot balance, so repaying the full liability fails with an insufficient-balance error on exactly that fee-sized shortfall. The converting repay (`POST /v5/account/repay`) covers the small gap from other assets, so the repayment succeeds. Combined with the `min(outstanding, bought)` cap, it never converts more than needed: the amount is bounded by the actual liability, so unrelated holdings aren't touched beyond the fee difference.

### Also included
- HTTP layer: the `POST /v5/account/repay` endpoint (`repay` raw call + `repay_spot_borrow_with_conversion` wrapper), request/response models, and a dedicated rate-limit quota mirroring the existing no-convert repay.
- Python bindings for the new wrapper (+ regenerated `.pyi` stubs).
- Tests: repay-endpoint HTTP tests (mock handler + route) and blackout-window unit tests.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Testing


cjdsellers:
Hey @dxwil,

Thanks for the PR, this is a well-structured addition.

I went through the whole thing and the main thing I'd want before merging is a test for `repay_coin` itself. The HTTP wrapper and the blackout window are covered, but the orchestration that actually moves money, the `outstanding.min(bought.as_decimal())` cap and the zero-outstanding early exit, is only covered by reading. Since this autonomously converts assets, I'd like that invariant pinned down with at least one test against the existing mock server.

**A couple of smaller follow-ups that can wait:**
- A doc comment on `auto_repay_spot_borrows` noting it only fires on tracked, standard-channel, fully-filled SPOT BUYs, so the fast channel and untracked orders don't trigger it.
- `BybitRepayParams.repayment_type` is declared but never set by any caller, so we keep for completeness for now?

Everything else looks good. The blackout timing fix and the converting-repay rationale both make sense to me.

dxwil:
Hi @cjdsellers,

Thanks for the review. I added tests to target `repay_coin` specifically and improved the doc comment. `BybitRepayParams.repayment_type` is kept for completness with the API for now, yes.

cjdsellers:
Hey @dxwil,

Thanks for adding the `repay_coin` tests, the cap and the zero-outstanding exit are pinned down now.

One thing the new tests surfaced by omission though. In `repay_coin` the zero check at repay.rs:105 runs on the raw `Decimal`, before `from_decimal_dp` rounds down to `bought.precision` at line 109. So a sub-tick residual gets through: borrow 1 ETH, sell, buy 1 ETH back, the first repay is capped at 1.0 and leaves the accrued interest outstanding, say 0.000000012. The next fully-filled ETH BUY sees that as non-zero, passes the guard, then rounds to zero at precision 5 and we POST `/v5/account/repay` with a zero `amount`.

Bybit doesn't say what an explicit zero does there. Best case it's a params error and we log `error!` on every subsequent fill; worst case it's treated as an omitted amount and repays the full liability with conversion, which is the case we most want to avoid. Could you move the zero check to after the `from_decimal_dp` conversion and add a case to the parametrized test with an outstanding below one tick?

A couple of non-blocking things I could pick up separately:

- For a SPOT market BUY submitted with quote quantity, `spot_market_unit` sets `marketUnit=quoteCoin`, so `exec.order_qty` is a quote amount and the `min(outstanding, bought)` cap stops binding. v1 has the same shape so it's not a regression, just worth knowing about.
- `docs/integrations/bybit.md` already lists `auto_repay_spot_borrows` with default `True` from the v1 config, so that table needs to reflect the new default.

Everything else looks good to me.

dxwil:
Good catch, I moved the check to where it should be and added a test case as you requested. 

The quote quantity case could be handled by either reading `execQty` (but this would need to be added up, since we only repay on fully filled) or looking at the `order` stream which has a `cumExecQty` field (although that path in the code might not be reached as the `execution` stream cleans up the order identities). But I'd say all of this is outside the scope of this PR.

It would be great if you could take care of the docs, those need some more changes than just the config default switch: we're now using the converting endpoint (to account for fees); the repay blackout timing and behaviour differ.

I also found one more issue that you could perhaps look at post-merge. Inside http `client.rs:2407`, the position side is derived **before** converting to `Quantity`. So a small `wallet_balance` of e.g. interest could round and produce a `PositionStatusReport` with `side = Short` and `quantity = 0`.

cjdsellers (approved):
Thank you for the follow-ups @dxwil 👌

PR #4544, opened 2026-07-23
Fix TWAP child-order sizing and interval validation

folknor:
Two defects in the Rust TWAP execution algorithm's schedule construction, where the child-order split and the interval timer diverge from the `TWAPExecAlgorithm` reference and can, respectively, emit off-grid child orders and panic mid-sequence.

## Child quantities are split at the internal fixed-point scale instead of the instrument size precision

`on_order` computes `qty_per_interval_raw = total_raw / num_intervals` on the raw fixed-point value and rebuilds each child with `total_qty.precision`, so a `1.0` order over three intervals yields children of raw `333333333` (nine-decimal `0.333333333`) and a final remainder slice of raw `1` (`0.000000001`). The children are off the instrument size grid, the dust remainder is orders of magnitude below any size increment, and the child precision is the primary order's rather than the instrument's - a quantity whose raw value carries precision the `precision` field says is unavailable. The reference `TWAPExecAlgorithm` instead divides in `Decimal`, floors the quotient to `instrument.size_precision`, and builds each child through `make_qty`.

The fix mirrors the reference exactly: the quotient is taken in `Decimal`, floored to `instrument.size_precision()` with round-toward-zero, and both the per-interval quantity and the remainder are constructed through `instrument.make_qty`, with the remainder computed from the pre-`make_qty` floored value. The N-plus-remainder slice layout is retained, and a runtime check now rejects the order before any state mutation if the assembled schedule does not sum exactly to the order quantity. For a `1.0` order over three intervals on a three-decimal instrument the schedule becomes `0.333, 0.333, 0.333, 0.001`, on-grid and summing exactly.

## A large interval panics after the first child is already submitted

The interval timer is created with `Duration::from_secs_f64(interval_secs)` only after the first child has been spawned, the primary reduced, and the child submitted. `from_secs_f64` panics on a non-representable value, so a finite but huge `interval_secs` (validated only as finite and positive) terminates the algorithm with a child order already live and no timer installed - a stranded sequence. `Clock::set_timer` additionally narrows the interval's nanoseconds to `u64` with an `as` cast, so intervals above the `u64` nanosecond range would silently install the wrong timer, and the clock's `start + interval` can overflow.

The fix validates the interval before any mutation, immediately after the submit-whole guards (mirroring the reference, which only builds a timer on the multi-slice path): `Duration::try_from_secs_f64`, a non-zero check, a checked `u64` nanosecond conversion, and a checked add against the current clock timestamp. Each failure logs and returns like the existing finite/positive guards, and the validated `Duration` is reused for `set_timer`.

## Parity scope

This aligns the Rust algorithm with the Python `TWAPExecAlgorithm` reference; it does not change the reference. The reference's coarser behavior - a single lopsided remainder slice, no per-slice size-increment alignment for non-power-of-ten increments, and no per-slice maximum - is shared by both implementations and is deliberately left as-is here, since redistributing or capping the remainder in Rust alone would make the two implementations disagree on child count, timing, and per-child quantity. That belongs in a separate, coordinated change across both, which I am not planning to take on; I am flagging the shared limitation here rather than committing to address it.

## Testing

`test_twap_calculates_size_schedule_with_remainder` now pins the on-grid schedule (`0.333, 0.333, 0.001` remaining after the first spawn) and asserts every slice carries `instrument.size_precision()`, lies on the size increment, and that the full schedule sums exactly to the order quantity; against the unfixed split it renders `0.3, 0.3, 0.0` at the wrong precision. `test_twap_children_use_instrument_size_precision` submits a nine-decimal `1.000000000` order and asserts the children carry the instrument's three-decimal precision, not the primary's nine; against the unfixed code they carry precision nine.

`test_twap_rejects_huge_finite_interval_before_submission`, `test_twap_rejects_subnanosecond_interval_before_submission`, and `test_twap_rejects_interval_exceeding_timestamp_headroom_before_submission` each assert that the order is rejected with no submitted command, no scheduled sizes, and no timer installed. Against the unfixed code the first panics in `Duration::from_secs_f64` with the first child already submitted, the second returns an error only after reaching timer setup, and the third overflows the clock timestamp - all after the first child has gone out.

All tests are deterministic over constructed events and a paused test clock with no wall-clock waits.

N.B.: `crates/adapters/polymarket/tests/exec_client.rs` carries a pre-existing deprecation unrelated to this change - `Atomic::fetch_update` was renamed to `try_update` and warns under a deny-warnings clippy on newer toolchains. It is outside the scope of this PR and left untouched.

### DIFF
--- a/crates/trading/src/algorithm/twap.rs
+++ b/crates/trading/src/algorithm/twap.rs
@@ -42,8 +42,9 @@ use nautilus_model::{
     identifiers::ClientOrderId,
     instruments::Instrument,
     orders::{Order, OrderAny},
-    types::{Quantity, quantity::QuantityRaw},
+    types::Quantity,
 };
+use rust_decimal::{Decimal, RoundingStrategy, prelude::ToPrimitive};
 use ustr::Ustr;
 
 use super::{
@@ -193,11 +194,24 @@ nautilus_execution_algorithm!(TwapAlgorithm, {
         }
 
         let total_qty = order.quantity();
-        let total_raw = total_qty.raw;
-        let precision = total_qty.precision;
-
-        let qty_per_interval_raw = total_raw / (num_intervals as QuantityRaw);
-        let qty_per_interval = Quantity::from_raw(qty_per_interval_raw, precision);
+        let interval_count = Decimal::from(num_intervals);
+        let quotient = total_qty.as_decimal() / interval_count;
+        let floored = quotient.round_dp_with_strategy(
+            u32::from(instrument.size_precision()),
+            RoundingStrategy::ToZero,
+        );
+        let Some(floored_f64) = floored.to_f64() else {
+            log::error!("Cannot execute order: qty_per_interval={floored} is not representable");
+            return Ok(());
+        };
+        let qty_per_interval = match instrument.try_make_qty(floored_f64, None) {
+            Ok(quantity) => quantity,
+            Err(e) => {
+                log::error!("Cannot execute order: invalid qty_per_interval={floored}: {e}");
+                return Ok(());
+            }
+        };
+        let remainder = total_qty.as_decimal() - floored * interval_count;
 
         if qty_per_interval == total_qty || qty_per_interval < instrument.size_increment() {
             log::warn!(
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. The fixes look right to me: child sizing now divides in `Decimal`, floors to the instrument's `size_precision`, and rebuilds each slice through `make_qty`, and moving the interval validation ahead of any state mutation closes the stranded-order panic path cleanly.

One small thing, not a blocker: the floored `Decimal` still round-trips through `f64` because `make_qty` only takes an `f64`, where the Python reference passes the `Decimal` straight through. It's exact for realistic sizes and your sum check rejects safely if it ever diverges. I'll add a `Decimal`-accepting overload to the instrument `make_qty`/`make_price` functions so this kind of math can skip the `f64` hop entirely.

PR #4545, opened 2026-07-23
Stop backtest order and account queries panicking the simulated exchange

folknor:
A `QueryOrder` or `QueryAccount` reaching the backtest simulated exchange panics rather than being handled as the unsupported operation it is.

## The simulated exchange panics on an order or account status query

`SimulatedExchange::send` dispatches a `TradingCommand` three ways: synchronously via `process_trading_command` when the message queue is disabled, into the plain `message_queue` when there is no latency model, else into the latency `inflight_queue` via `generate_inflight_command`. A query breaks two of those paths. `generate_inflight_command` matches only the submit/modify/cancel variants and ends `_ => panic!("Cannot handle command")`, so any query under a latency model panics. `process_trading_command` opens with `command.instrument_id()`, which is a hard `panic!("No instrument ID for command")` for the account-level `QueryAccount`. So a strategy that polls order or account status aborts the backtest instead of learning the venue does not answer queries.

The backtest venue does not support these queries - the reference execution client leaves `query_order`/`query_account` unimplemented and synthesizes no report. `BacktestExecutionClient::query_order` and `query_account` now log that the query is unsupported and return without forwarding to the exchange, so the normal path no longer reaches the panicking code. As defense for the public exchange entry point, `SimulatedExchange::send` also intercepts both query variants as a logged no-op ahead of the latency and instrument-id dispatch. `TradingCommand::instrument_id()` is left unchanged: its panic for an account-level command is correct, and the exchange was the invalid caller.

## Testing

`test_query_order_is_not_forwarded_to_exchange` and `test_query_account_is_not_forwarded_to_exchange` drive the client methods against a latency-configured exchange while holding an immutable borrow of that exchange across the call, so any forwarding attempt would panic in `send`'s `borrow_mut`; each asserts the call returns and no inflight command was scheduled. `test_send_query_order_is_no_op` and `test_send_query_account_is_no_op` drive `SimulatedExchange::send` with each query variant across all three dispatch configurations - latency, queued, and synchronous - and assert nothing is queued or scheduled. Against the unfixed code every case panics: the latency path in `generate_inflight_command`, the synchronous `QueryAccount` at `instrument_id()`, and the client cases by forwarding into the same paths.

### DIFF
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -693,6 +693,14 @@ impl SimulatedExchange {
 
     /// Sends a trading command to the exchange for processing.
     pub fn send(&mut self, command: TradingCommand) {
+        if matches!(
+            &command,
+            TradingCommand::QueryOrder(_) | TradingCommand::QueryAccount(_)
+        ) {
+            log::warn!("Simulated exchange does not support queries: {command}");
+            return;
+        }
+
         if !self.use_message_queue {
             self.process_trading_command(command);
         } else if self.latency_model.is_none() {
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor

Thank you for the PR. I traced both panic paths independently and they line up with your write-up: `instrument_id()` on `QueryAccount` in `process_trading_command`, and the `_` arm in `generate_inflight_command`. Returning `Ok` with a warn matches the `ExecutionClient` trait default, so the client-side change is consistent with how unsupported queries are already handled.

The held-borrow trick in the forwarding tests is a nice touch: it makes a client-only revert panic in `borrow_mut()` instead of being masked by the exchange-side guard.

No blockers from me, happy for this to merge.

PR #4546, opened 2026-07-23
Fix O(n^2) NETTING backtest cost (snapshot replay state, account event-log clones)

nah294:
Fixes quadratic (O(n²) in total fills) CPU and memory growth on the NETTING backtest path, which currently makes high order-count tick backtests impractical and eventually OOM.

## Problem

Running a simple EMA-cross strategy over raw trade ticks (BacktestEngine, SimulatedExchange, NETTING OMS, MARGIN account, single instrument, standard-precision build) the per-order cost grows roughly linearly with the number of prior orders, so total cost is O(n²):

| Orders in run | Avg order-path cost per order |
|---|---|
| 950 | ~1.0 ms |
| 8,000 | ~9.4 ms |
| 58,459 | did not complete (retained snapshot memory also grows quadratically → OOM) |

The same 58k-order workload completes in 140 s on nautilus_trader v1 (1.230.0), which does not exhibit the quadratic.

## Root cause

Three composing mechanisms:

1. **`Position.replay_events` is carried across every NETTING close/reopen cycle and never trimmed.** Each fill appends one entry (`Position::apply_fill`), and `ExecutionEngine::open_position` re-attaches the prior position's entire log to the reopened position. After n fills the reused-ID position is O(n)-sized, so every prior-position fetch (`position_owned`), `add_position` cache clone, and snapshot clone is O(n).
2. **`Cache::snapshot_position` serializes and retains the full position on every NETTING flip/reopen.** Each blob embeds the ever-growing `replay_events`, so each flip serializes O(n) bytes and all blobs are retained ⇒ O(n²) retained memory (this is what OOMs).
3. **The portfolio's per-fill handlers deep-clone the whole account up to four times per fill.** `BaseAccount.events` grows by one `AccountState` per fill and is never purged during a run, so each clone (`Cache::account_owned`, the `AccountsManager::update_positions` wrapper, `Cache::update_account` into the cell, and the defensive rollback clone in `update_balances`) is O(n). This was the dominant residual cost after fixing (1)–(2).

## Changes (three commits)

1. **Strip replay-only state from position snapshot blobs.** `Cache::snapshot_position` clears `replay_events` and `fill_voids` on the throwaway snapshot copy before serializing. No snapshot consumer reads these vectors back: the portfolio's realized-PnL reconstruction reads `realized_pnl`/`account_id`, the analyzer reads summary fields, and live fill-void fragment discovery reads the *live* cached position, never a snapshot blob. Both fields are `#[serde(default)]`, so existing persisted blobs and `position_snapshots()` round-trips remain compatible. Snapshot blob size becomes independent of run length (regression test added).
2. **Add `carry_replay_events_on_reopen` to `ExecutionEngineConfig` (default `true` — behavior unchanged).** The replay log exists to service venue `OrderFillVoided` corrections that may reference fills from prior cycles; that cannot occur against purely simulated venues. When disabled, the engine skips the prior-position fetch and replay carry on reopen, bounding position state to the current cycle. `LiveExecEngineConfig` conversion pins the option to `true` so live trading always carries replay state. Python bindings expose the new option.
3. **Stop cloning the account event log on the per-fill portfolio path.** `update_position` now peeks at the account under a scoped borrow and, only for the margin recompute path, moves the account out with the existing `take_account`, recalculates with the existing `update_positions_in_place`, and moves it back (`update_account_owned` on success; `cache_account_owned` — no database write, as before — when the recompute returns `None`, which leaves the account untouched). Cash/betting accounts and margin accounts without `calculate_account_state` no longer write the account back at all, matching the previous read-only behavior. `update_balances` snapshots only the state it can mutate (balances and commissions maps) for its rollback path instead of cloning the account.

## Results

Same tick workload as above (BTC perp trade ticks, i64/9dp build, single core). "Before" measured at develop `0cfa1310` (the touched files are unchanged through `ec58b2a`); "after" is this branch with `carry_replay_events_on_reopen: false`:

| Workload | Before | After |
|---|---|---|
| 1 week (1.43M ticks), 8,000 orders / 9,000 fills | 18.7k ticks/s (76.3 s) | **1.04M ticks/s (1.4 s)** |
| 45 days (10.6M ticks), 6,847 orders / 7,728 fills | 168k ticks/s (63.1 s, 16 GiB RSS) | **1.26M ticks/s (8.4 s, 1.4 GB)** |
| 45 days (10.6M ticks), 58,459 orders / 65,466 fills | did not complete (OOM) | **1.02M ticks/s (10.4 s, 1.6 GB)** — ~13× faster than v1's 140 s on the identical workload |
| Null-strategy floor (no orders) | 1.30M ticks/s | unchanged (1.28M ticks/s) |

Per-fill cost is now flat (~30 µs/fill at 9k and at 65k total fills).

Correctness evidence:

- Fills are event-identical (`side`, `px`, `qty`, `ts_event`) with `carry_replay_events_on_reopen` **on vs off** on the same window.
- Fills are event-identical to **nautilus_trader v1 (1.230.0)** running the same mirrored backtest (same venue semantics, hand-rolled EMA mirrored op-for-op on both engines): **7,728/7,728** fills on the 45-day window (EMA 500/2000) and **65,466/65,466** fills (EMA 50/200) — the latter run previously could not complete on v2 at all. Final balances match v1 to ≤2e-5 USDT (currency-precision rounding between the 9dp raw build and the Python wheel, present before and after this change).
- Full test suites pass: `nautilus-execution` (211 + 50), `nautilus-portfolio` (115), `nautilus-common` cache (244), including `test_portfolio_realized_pnl_with_position_snapshots_netting_oms`; two regression tests added (snapshot blob size independent of replay-log length; replay carry bounded when the option is disabled).

For context, v1's Cython engine has none of these costs: it creates a fresh `Position` per NETTING reopen (per-cycle event list), pickles only the bounded just-closed position on flip, and mutates the account in place. These changes bring v2's default cost model in line while preserving the replay-carry capability for live use.

## Out of scope (can file separately)

- HEDGING OMS has a separate quadratic: `Portfolio::update_position` re-scans all open positions per fill (net-position, unrealized-PnL and margin recompute), degrading with thousands of open positions (measured 25.5k ticks/s with 8,000 open positions).
- `snapshot_position` stores each blob twice (`position_snapshots` map and the generic `cache://position-snapshots/...` keys via `Cache::add`), and `purge_position` removes only the former.
- `BaseAccount.events` still grows O(n) in memory over a run (unchanged behavior; only the per-fill time cost is removed here).

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -4748,6 +4748,12 @@ impl Cache {
     /// Creates a snapshot of the `position` by cloning it, assigning a new ID,
     /// serializing it, and storing it in the position snapshots.
     ///
+    /// The snapshot copy excludes replay-only state (`replay_events` and `fill_voids`):
+    /// these vectors exist to service fill-void corrections on the live cached position
+    /// and are never read back from snapshot blobs, while growing with every fill
+    /// applied to the position ID (which would make each blob O(total fills) in both
+    /// serialization time and retained memory).
+    ///
     /// # Errors
     ///
     /// Returns an error if serializing or storing the position snapshot fails.
@@ -4757,6 +4763,8 @@ impl Cache {
         let mut copied_position = position.clone();
         let new_id = format!("{}-{}", position_id.as_str(), UUID4::new());
         copied_position.id = PositionId::new(new_id);
+        copied_position.replay_events.clear();
+        copied_position.fill_voids.clear();
 
         // Serialize the position (TODO: temporarily just to JSON to remove a dependency)
         let position_serialized = serde_json::to_vec(&copied_position)?;
--- a/crates/common/src/cache/tests.rs
+++ b/crates/common/src/cache/tests.rs
@@ -59,7 +59,7 @@ use nautilus_model::{
         builder::OrderTestBuilder,
         stubs::{TestOrderEventStubs, TestOrdersGenerator},
     },
-    position::Position,
+    position::{Position, PositionReplayEvent},
     stubs::TestDefault,
     types::{AccountBalance, Currency, Money, Price, Quantity},
 };
@@ -5737,6 +5737,58 @@ fn test_position_snapshots_round_trip(mut cache: Cache) {
             .position_snapshots(None, Some(&AccountId::new("OTHER-000")))
             .is_empty(),
     );
+}
+
+#[rstest]
+fn test_position_snapshot_blobs_exclude_replay_state(mut cache: Cache) {
+    let audusd_sim = InstrumentAny::CurrencyPair(audusd_sim());
+
+    let order = OrderTestBuilder::new(OrderType::Market)
+        .instrument_id(audusd_sim.id())
+        .side(OrderSide::Buy)
+        .quantity(Quantity::from(100_000))
+        .build();
+    let fill = TestOrderEventStubs::filled(
+        &order,
+        &audusd_sim,
+        Some(TradeId::new("T-1")),
+        Some(PositionId::new("P-1")),
+        Some(Price::from("1.00000")),
+        None,
+        None,
+        None,
+        Some(UnixNanos::from(1_000_000_000)),
+        None,
+    );
+    let fill: OrderFilled = fill.into();
+    let mut position = Position::new(&audusd_sim, fill.clone());
+    let position_id = position.id;
+
+    cache.snapshot_position(&position).unwrap();
+
+    // Grow the replay log as NETTING close/reopen cycles would
+    for _ in 0..1000 {
+        position
+            .replay_events
+            .push(PositionReplayEvent::Filled(fill.clone()));
+    }
+    cache.snapshot_position(&position).unwrap();
+
+    // Blob size is independent of the replay-log length
+    let frames = cache.position_snapshot_bytes(&position_id).unwrap();
+    assert_eq!(frames.len(), 2);
+    assert_eq!(frames[0].len(), frames[1].len());
+
+    // Replay-only state is stripped; summary fields round-trip intact
+    let snapshots = cache.position_snapshots(Some(&position_id), None);
+    assert_eq!(snapshots.len(), 2);
+    for snapshot in &snapshots {
+        assert!(snapshot.replay_events.is_empty());
+        assert!(snapshot.fill_voids.is_empty());
+        assert_eq!(snapshot.realized_pnl, position.realized_pnl);
+        assert_eq!(snapshot.account_id, position.account_id);
+        assert_eq!(snapshot.quantity, position.quantity);
+    }
 }
 
 fn snapshot_test_position() -> Position {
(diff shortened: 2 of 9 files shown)


cjdsellers:
Hi @HungNgo4444,

Thank you for the PR, this is a thorough piece of work and the correctness evidence (event-identical fills against v1, plus the on/off comparison) is exactly what I want to see for a change on this path. The snapshot stripping and the account event-log clone removal both check out as behavior-preserving, and the diagnosis of the three composing mechanisms is spot on.

Worth mentioning: I ran into the same quadratic during backtest engine performance work here yesterday and landed on the same root cause independently, so your diagnosis is well corroborated from the other direction. That work also turned up one other optimization in the same area, which I'll push separately after we merge this.

**Two things before merge:**

1. `carry_replay_events_on_reopen` should default to `false`. The `From<LiveExecEngineConfig>` conversion already pins it `true` and always returns `Some(..)`, so live is unaffected by the default, and the simulated exchange never emits `OrderFillVoided`. As it stands the OOM you reported is still the out-of-the-box behavior and only users who find the flag escape it, which I'd rather not ship. Could you also reword the doc comment while you're there: setting it `true` doesn't simulate voids (nothing in the sim generates one yet), it preserves cross-cycle void reversibility if a void arrives from a custom or external execution client.

2. The generated stub wasn't regenerated after adding the pyo3 getter and constructor arg, so `python/nautilus_trader/execution/__init__.pyi` has neither. The `check-v2-generated-drift` job in `build-v2.yml` will fail on that. Please run `make py-stubs-v2` and commit the result (it may pick up a docstring diff for the new field too).

The out-of-scope items you listed all sound good to file separately, the HEDGING re-scan one especially.

Thanks again.

cjdsellers:
@HungNgo4444 Looks like the pre-commit failed, but this is easy to fix.

Just run the following and push any changes to the same PR:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

cjdsellers:
Hi @HungNgo4444

I've landed this on `develop` myself rather than wait for the pre-commit fix. The quadratic is one of the worst bottlenecks in the backtest path right now and it was holding up further performance work in the same area, so I didn't want to leave it sitting.

Your approach went in as designed, all three mechanisms, with the two review items applied: `carry_replay_events_on_reopen` now defaults to `false` with the doc comment reworded, and the generated stub and docstrings are regenerated. I added test coverage on top, including a NETTING close and reopen cycle with a prior-cycle `OrderFillVoided`, which is the case the flag exists for.

You're credited in the release notes under #4546 for both the fix and the new config option.

I'll close this as superseded. Your out-of-scope list is still worth filing, and I'd be glad to see more from you.

nah294:
Thanks @cjdsellers, and no problem at all about landing it on develop.

This hit me directly while developing a high-frequency strategy on NautilusTrader, where backtest runtime was scaling badly with fill count under NETTING. Since it was blocking my own work, I had Claude Fable 5 dig into the backtest path and put the fix together as a contribution back upstream. Glad it turned out to be useful beyond my case.

I'm not planning to pick up the out-of-scope items myself, but happy to have them on record for whoever does.

PR #4547, opened 2026-07-23
Stop okx data-client tests waiting out the event-drain deadline

folknor:
The okx data-client instrument-request tests each wait out a fixed five-second event-drain deadline, costing five seconds apiece regardless of when the response arrives.

## The event drain cannot return early

`drain_data_events` in `crates/adapters/okx/tests/data_client.rs` collects events with `tokio::time::timeout_at(deadline, rx.recv())` in a loop, so it only returns once the deadline expires. The `OKXDataClient` holds the `UnboundedSender` for the life of the test, so `rx.recv()` never sees a closed channel and blocks the full window after the last event. The three instrument-request tests each pass a five-second window and so cost five seconds apiece even though the response arrives in milliseconds.

The correlated `DataResponse` is a deterministic terminal marker for these requests. `collect_data_events_until_response` receives until it sees the `DataEvent::Response` whose `correlation_id` equals the request's id, then sweeps anything already queued with `try_recv`. The three positive tests capture the request id before submitting and collect through that response; the five-second argument survives only as a failure deadline a passing run never reaches. Every existing assertion is unchanged - the exact spread ids, spread and spot counts, response variants, instrument id and type, and the server-side query counts and parameters - and the response is emitted only after the relevant HTTP queries complete, so returning at the correlated response does not race those checks.

No test can now pass vacuously: a missing correlated response times out and panics, a closed channel panics, and a correlated response of the wrong variant still trips the `instruments_response`/`instrument_response` variant assertions.

`test_request_instrument_emits_no_spread_when_disabled` keeps its one-second window: with spreads disabled that path issues the query but emits no response, so it has no terminal marker to collect through, and the bounded wait is a deliberate absence check.

## Testing

No test cases were added; the existing assertions are the regression and every one is preserved. The three converted tests drop from ~5.005s each to a few milliseconds, so the former five-second members no longer appear among the package's slowest tests, and their five-second arguments are now failure ceilings. The correlation was verified: a request id that never receives its response makes the helper time out and panic rather than pass.

### DIFF
--- a/crates/adapters/okx/tests/data_client.rs
+++ b/crates/adapters/okx/tests/data_client.rs
@@ -166,6 +166,35 @@ async fn drain_data_events(
     events
 }
 
+async fn collect_data_events_until_response(
+    rx: &mut tokio::sync::mpsc::UnboundedReceiver<DataEvent>,
+    request_id: UUID4,
+    timeout: Duration,
+) -> Vec<DataEvent> {
+    let mut events = Vec::new();
+    tokio::time::timeout(timeout, async {
+        loop {
+            let event = rx.recv().await.expect("data event channel closed");
+            let is_correlated_response = matches!(
+                &event,
+                DataEvent::Response(response) if response.correlation_id() == &request_id
+            );
+            events.push(event);
+
+            if is_correlated_response {
+                break;
+            }
+        }
+    })
+    .await
+    .unwrap_or_else(|_| panic!("timed out waiting for data response {request_id}"));
+
+    while let Ok(event) = rx.try_recv() {
+        events.push(event);
+    }
+    events
+}
+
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor

Thanks for the PR. I traced the ordering in `data.rs` and the response is only sent once both the instruments and spreads calls have returned, so collecting through the correlated response can't race the query-count assertions. Everything the tests check still holds, and the 1s absence window on the disabled-spreads path is the right call since that path never emits.

One follow-up on my side: `collect_data_events_until_response` now exists verbatim in both the okx and polymarket test crates. Both already dev-depend on `nautilus-testkit`, so I'll lift it there once a third adapter wants it rather than churn two crates now.

PR #4549, opened 2026-07-23
Add pending-update to submitted order status transition

folknor:
A follow-up to the order-apply atomicity work in #4530, split out at your request on that thread.

## Missing `(PendingUpdate, Submitted)` state-machine transition

`OrderStatus::transition` in `crates/model/src/orders/mod.rs` omits one arm that the Cython `FiniteStateMachine` table carries (`nautilus_trader/model/orders/base.pyx:132`, `(PENDING_UPDATE, SUBMITTED) -> PENDING_UPDATE`). When a submission acknowledgement arrives after an order has already entered `PendingUpdate` - a venue re-emitting, or a submit ack racing an in-flight modify request - the event matches no arm, falls through to the catch-all, and `OrderCore::apply` returns `InvalidStateTransition`. The event is dropped, diverging from the Python engine, which tolerates it and holds `PENDING_UPDATE`.

The fix adds the single arm `(PendingUpdate, Submitted) -> PendingUpdate`. No handler change is needed: `apply` does not overwrite `previous_status` while the current status is `PendingUpdate` or `PendingCancel`, so holding `PendingUpdate` preserves the pre-update status, and a subsequent `OrderUpdated` or `OrderModifyRejected` still restores it. The `submitted` handler only records `account_id` and `ts_submitted`, which is harmless here.

## Testing

`test_pending_update_accepts_submitted_and_restores_status_on_updated` (an extension of the prior `test_pending_update_order_restores_status_on_updated`) drives an order `Initialized -> Submitted -> Accepted -> PendingUpdate`, applies a delayed `OrderSubmitted`, and asserts the order holds `PendingUpdate` with `previous_status` still `Some(Accepted)` on both sides of the event, then applies `OrderUpdated` and asserts restoration to `Accepted` with the updated quantity. The delayed-submission assertions were verified to FAIL against the unfixed table: `apply` returns `InvalidStateTransition` at that step.

### DIFF
--- a/crates/model/src/orders/mod.rs
+++ b/crates/model/src/orders/mod.rs
@@ -242,6 +242,7 @@ impl OrderStatus {
             (Self::PendingUpdate, OrderEventAny::Canceled(_)) => Self::Canceled,
             (Self::PendingUpdate, OrderEventAny::Expired(_)) => Self::Expired,
             (Self::PendingUpdate, OrderEventAny::Triggered(_)) => Self::Triggered,
+            (Self::PendingUpdate, OrderEventAny::Submitted(_)) => Self::PendingUpdate,  // Real world possibility
             (Self::PendingUpdate, OrderEventAny::PendingUpdate(_)) => Self::PendingUpdate,  // Allow multiple requests
             (Self::PendingUpdate, OrderEventAny::PendingCancel(_)) => Self::PendingCancel,
             (Self::PendingUpdate, OrderEventAny::ModifyRejected(_)) => Self::PendingUpdate,  // Handled by modify_rejected to restore previous_status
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hey @folknor

Thanks for the PR, the new arm matches the Cython table and the `previous_status` handling holds up.

One request on the test: folding the delayed ack into `test_pending_update_order_restores_status_on_updated` leaves the plain `PendingUpdate -> Updated` restore path without standalone coverage, so a regression in the new arm now breaks both behaviors at once. Could you leave that test as it was and add a separate one for the delayed submission? Worth giving the delayed `OrderSubmitted` a distinct `account_id` and `ts_event` too, since both specs currently build identical events and the assertions can't catch the `submitted` handler overwriting either field.

folknor:
Both done, thank you - the second point turned out to matter more than a test change.

`test_pending_update_order_restores_status_on_updated` is back exactly as it was, and the delayed ack now lives in `test_pending_update_accepts_delayed_submitted`. Verified they fail independently: with the new arm removed, the new test fails at the delayed `OrderSubmitted` with `InvalidStateTransition` while the restore test still passes, so a regression in the arm no longer takes the plain `PendingUpdate -> Updated` path down with it.

On the distinct fields: the submission specs now differ, and the assertions pin what `submitted` actually does. The first `OrderSubmitted` carries `ts_event` 1000 and the default `SIM-001`; the delayed one carries `SIM-002` and `ts_event` 3000. `OrderCore::submitted` assigns `account_id` and `ts_submitted` unconditionally, so the order ends up on `SIM-002` with `ts_submitted` 3000 - the delayed event's values, not the original submission's. That mirrors `_submitted` in `base.pyx`, which assigns both the same way, so the arm plus this handler is Cython-equivalent end to end and the test now documents it rather than leaving it unobservable.

My original description called the handler "harmless here" and that was too casual: a re-emitted ack does move `ts_submitted` forward, so it records the latest ack rather than the first submission. The Rust and Cython engines agree on this, so I have left the behavior alone and only pinned it. If you would rather `ts_submitted` hold the first submission, that is a parity change on both engines and should probably be a separate PR.

Rebased onto current develop.

folknor:
@cjdsellers The CI failures here are all again `couldn't create a temp dir: No space left on device (os error 28)`.

I restarted the python 3.14 CI now, but up to you how to proceed here.

folknor:
The CI failures here are runner disk exhaustion, not the diff. Both run attempts fail the same way across all three ubuntu build jobs (3.12/3.13/3.14): the "Run Rust doctests" step dies on the same ten `crates/persistence/src/backend/catalog.rs` doctests with `couldn't create a temp dir: No space left on device (os error 28)` under `/tmp/rustdoctest*`, plus `rust-lld` crash dumps from the same cause (a full disk turns lld's mmap'd output writes into SIGBUS mid-link). The job enters the test phase with ~120G free; the core and adapters test builds consume nearly all of it over ~40 minutes, and the persistence doctest binaries - each statically linking datafusion, arrow, and aws-lc - tip it over.

This is not specific to this branch: [another PR's build](https://github.com/nautechsystems/nautilus_trader/actions/runs/30149052138/job/89657280764) failed this morning at the same step, on the same ten doctests, with the same lld crash signature, while other runs squeak past - so the ubuntu jobs are sitting right at the runner's disk margin (the workflow's own disk-space guard already warns "the Rust build may run out of space, re-run if it fails" below 40G). Nothing in this branch to change; will re-run once the disk margin improves, or rebase if a CI-side mitigation lands first.

cjdsellers (approved):

cjdsellers:
@folknor I'm currently working on the GH hosted runner disk space issue. Once that's fixed I'll rebase and re-run this.

folknor:
@cjdsellers Fantastic, thank you.

cjdsellers:
@folknor, now pushed the CI disk use improvements - feel free to rebase one of the PRs when you get the chance.

PR #4557, opened 2026-07-24
Fail closed on incomplete live reconciliation

seungpyoson:
## Summary

Live reconciliation could silently omit or repair venue state that NautilusTrader could not represent, and continuous reconciliation failures could leave a running node backed by an unverified cache.

This change closes both paths while preserving compatibility:

- Polymarket mass-status reconciliation requires every relevant open order, confirmed fill, and current position to resolve through the loaded NT instrument map.
- Invalid or unrepresentable quantities, prices, commissions, timestamps, expirations, aggregates, asset correlations, and position sizes return errors instead of being omitted or repaired.
- Raw instrument/order scope is applied before strict validation, so scoped requests ignore unrelated venue records while mass reconciliation remains exhaustive.
- `LiveExecEngineConfig.reconciliation_fail_closed` adds an opt-in strict runtime profile, defaulting to `false`.
- In strict mode, any continuous open-order or position report error, timeout, or incomplete targeted-order coverage stops the trader and live node before partial reports are applied.

The resulting boundary is adapter-owned: reconciliation either returns complete typed NT reports for the requested scope or fails. Downstream systems do not need a second venue parser, reconciliation universe, or order-lifecycle authority.

## Related Issues/PRs

Related to seungpyoson/bolt-v2#1505.

## Type of change

- [x] Bug fix (non-breaking)

## Breaking change details

Not applicable. Existing users retain best-effort continuous reconciliation unless they explicitly enable `reconciliation_fail_closed`.

## Documentation

- [x] Live runtime and configuration documentation updated for the strict profile.

## Testing

- [x] Polymarket adapter unit/integration tests and clippy.
- [x] Live-node integration tests covering open-order, position, and targeted-query errors and timeouts.
- [x] Live library tests, production/test clippy, Python-enabled Rust compilation, and Rust formatting.


cjdsellers:
Hi @seungpyoson

Thank you for the PR.

Before the code, two process things.

This branch has kicked off 9 CI runs in about 21 hours, three of them cancelled mid-run by the next push arriving. Each run is pre-commit plus the Rust and Python build jobs, so that adds up quickly on shared runners. Draft status won't help here, since the workflow triggers on every PR push regardless. Could you batch commits locally and push once when you have something you want looked at, rather than pushing each commit as you write it?

Related to that: two of the commits in this PR exist only to fix formatting that CI caught (`style(polymarket): satisfy reconciliation formatting` and `Fix reconciliation formatting policy`). Running `make pre-commit` locally before you push will catch those in seconds and save a round trip.

Second, scope. This started as a Polymarket open-order reconciliation fix and has grown to include a new `ExecutionClient` trait method, a public `ExecutionReconciliationCapabilities` type, an irreversible authority-revocation path in `ExecutionEngine`, and a `reconciliation_fail_closed` option on both `LiveExecEngineConfig` and the v1 Python config. That's several separable changes across 21 files, and the title has been renamed twice as it grew.

The Polymarket work stands on its own and I'm happy to review it as-is. The execution engine and live node half is a different matter: it adds public API surface and a runtime mode that can halt a live node, and I'd want to agree on the design before reviewing an implementation. Could you split the adapter fix into its own PR, and open an issue for the fail-closed reconciliation proposal so we can settle the shape first? That way the adapter fix can land without waiting on the larger discussion.

Going forward, please keep each PR to one defined fix or feature, and only a couple open at a time. It's in CONTRIBUTING under Tips, and it makes review turnaround much faster for both of us.

One small thing for when you re-split the commits: our commit titles are plain verb-led sentence case naming the affected surface, for example `Fix Polymarket open-order reconciliation`, rather than the `fix(polymarket):` conventional-commit form.

seungpyoson:
Thank you. I agree with the process and scope feedback.

I closed this PR and split out only the adapter fix as draft PR #4566: one commit, six adapter files, with make pre-commit and the targeted Polymarket nextest suite run locally before the single push. The execution-engine fail-closed mode is not being pursued in this contribution; I will not open another implementation PR or issue for it.

PR #4566 then exposed an unrelated repository-wide CI regression from the new doctest gate: doctests ran after the full nextest build and exhausted the 146 GB runner filesystem. I opened draft PR #4570 as the second and only other open PR: one commit changing two workflow files, moving the unchanged doctest command before nextest without reducing coverage or adding retries. Local pre-commit passed before that single push. Its current run has passed doctests, core tests, and adapter tests on all three Linux matrices and is finishing the unchanged wheel stages.

The old #4566 run still has two jobs stuck in Post Common setup after the disk failure. I attempted to cancel run 30149052138, but GitHub requires upstream admin rights.

I will keep at most these two focused drafts open, make no incremental pushes, and present only locally verified, batched heads for review.

PR #4567, opened 2026-07-25
Convert serialization decode against the model's resolved raw types

folknor:
A follow-up to #4550 implementing the root fix you asked for on that thread.

## Decode narrowing keyed on the serializer's own feature

`nautilus-model` resolves `PriceRaw`, `QuantityRaw` and `MoneyRaw` to 128-bit or 64-bit integers from its `high-precision` feature. The five `nautilus-serialization` decode sites - `decode_price` and `decode_quantity` in `sbe/market/common.rs`, and `Price`/`Quantity`/`Money` `from_capnp` in `capnp/conversions.rs` - instead branch on `nautilus-serialization`'s own `high-precision`, which only forwards to the model. The two are not the same predicate: a build can activate `nautilus-model/high-precision` without `nautilus-serialization/high-precision`, and then decode compiles its `#[cfg(not(feature = "high-precision"))]` path against 128-bit aliases. Any raw beyond the 64-bit range fails with `SbeDecodeError::NumericOverflow` or the capnp overflow string, even though `Price::from_raw` would accept it. Encode is unaffected: all five encoders widen to a fixed 128-bit wire representation with no feature branch.

`nautilus-common` is a concrete in-tree route - its `capnp` and `sbe` features activate `dep:nautilus-serialization` independently, while the `nautilus-serialization?/high-precision` forward hangs off `nautilus-common/high-precision`, so `nautilus-common/capnp` plus a direct `nautilus-model/high-precision` reproduces the desync. `nautilus-common/defi` is another, since model `defi` implies model high precision. Any external crate can do the same.

The fix adds one crate-private `numeric.rs` with `wire_to_raw` and `raw_to_wire`, generic over the resolved raw type, and routes all ten codec sites through it. Decode now converts into `PriceRaw`/`QuantityRaw`/`MoneyRaw` directly, so it tracks the model's actual width and `nautilus-serialization/high-precision` becomes a convenience alias rather than a correctness input. Error variants and message strings are unchanged.

## Feature-conditional lint attributes left only one configuration buildable

The same asymmetry had a second consequence. Five `#[expect(clippy::useless_conversion)]` attributes on the encoders are fulfilled only when the resolved raw is 128-bit, and two on the capnp decoders only when it is 64-bit. An unfulfilled `#[expect]` is itself an error under `-D warnings`, so each width errored on the attributes belonging to the other. Nothing in CI observes this: `make clippy` runs `--all-features`, which always resolves high precision, and `make check-features` is `cargo hack --each-feature`, which is `cargo check` and never evaluates `clippy::` expectations. Separately, `Money::from_capnp`'s standard-precision branch did `from_raw(raw.into())` with `raw` already `i64` - a real `useless_conversion`, not an expectation artifact.

Because clippy checks a generic body before monomorphization, an identity conversion is invisible through the helpers' type parameter, so no suppression is needed in any width. All seven attributes are deleted and none added.

## Testing

`cargo clippy -p nautilus-serialization --all-targets -- -D warnings` and `cargo test -p nautilus-serialization`, across the three width configurations:

| features | before | after |
|---|---|---|
| `capnp,sbe` | 6 clippy errors: 5 unfulfilled expectations, 1 `useless_conversion` | clean, tests pass |
| `capnp,sbe,high-precision` | clean, tests pass | clean, tests pass |
| `capnp,sbe,nautilus-model/high-precision` | 2 clippy errors; clippy aborts before any test runs | clean, tests pass |

`test_capnp_resolved_raw_width_roundtrip` round-trips `Price`, `Quantity` and `Money` through the crate's `serialize_*`/`deserialize_*` helpers; `test_sbe_resolved_raw_width_roundtrip` round-trips a `BookOrder` carrying the same price and quantity. Both use ordinary economic values whose fixed-point raw crosses 64 bits only when the model resolves to high precision - `Price::from("50000.00")`, `Quantity::from("20000.00")`, `Money::new(20_000.0, Currency::USD())` - so they carry no `cfg` and no width-dependent assertions; the dependency graph supplies the differential.

Both tests were verified to FAIL against the unfixed decode in the third configuration, and each decode site was then reverted individually to confirm the assertions bite separately rather than all tripping on the first one: reverting only capnp `Money` fails at the money assertion with `"Money value overflows i64 in standard precision mode"`, only capnp `Quantity` fails at the quantity assertion with its own message, and only SBE `decode_quantity` fails inside `from_sbe` with `NumericOverflow { type_name: "Quantity" }`. The existing direct Money capnp test used raw `100_000_000`, which fits both widths and so could not catch this.

## Guarding against a recurrence

As you noted, a compile-only feature sweep cannot catch this class, and no Rust test can select the dependency graph it is compiled in - so the committed tests and the third invocation above are jointly the regression. What the change enforces structurally is that codec code no longer inspects its own precision feature: after this there is no `cfg(feature = "high-precision")` left in the SBE or capnp paths, so the desync cannot reappear by omission. If you want a mechanical guard, a `.pre-commit-hooks` check forbidding that `cfg` under `crates/serialization/src/{sbe,capnp}` would do it.

### DIFF
--- a/crates/serialization/src/capnp/conversions.rs
+++ b/crates/serialization/src/capnp/conversions.rs
@@ -46,7 +46,10 @@ use nautilus_model::{
         AccountId, ActorId, ClientId, ClientOrderId, ComponentId, ExecAlgorithmId, InstrumentId,
         OrderListId, PositionId, StrategyId, Symbol, TradeId, TraderId, Venue, VenueOrderId,
     },
-    types::{AccountBalance, Currency, MarginBalance, Money, Price, Quantity},
+    types::{
+        AccountBalance, Currency, MarginBalance, Money, Price, Quantity, money::MoneyRaw,
+        price::PriceRaw, quantity::QuantityRaw,
+    },
 };
 use rust_decimal::Decimal;
 use ustr::Ustr;
@@ -54,8 +57,9 @@ use uuid::Uuid;
 
 use super::{FromCapnp, ToCapnp};
 use crate::{
-    base_capnp, enums_capnp, identifiers_capnp, market_capnp, order_capnp, position_capnp,
-    types_capnp,
+    base_capnp, enums_capnp, identifiers_capnp, market_capnp,
+    numeric::{raw_to_wire, wire_to_raw},
+    order_capnp, position_capnp, types_capnp,
 };
 
 trait CapnpWriteExt<'a, T>
@@ -442,9 +446,8 @@ impl<'a> FromCapnp<'a> for InstrumentId {
 impl<'a> ToCapnp<'a> for Price {
     type Builder = types_capnp::price::Builder<'a>;
 
-    #[expect(clippy::useless_conversion)] // Needed for non-high-precision builds
     fn to_capnp(&self, mut builder: Self::Builder) {
-        let raw_i128: i128 = self.raw.into();
+        let raw_i128: i128 = raw_to_wire(self.raw);
         let lo = raw_i128 as u64;
         let hi = (raw_i128 >> 64) as u64;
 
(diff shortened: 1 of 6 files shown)


folknor:
Hey @cjdsellers

Merged develop into the branch to resolve an overlap with `Fix standard-precision make targets and add nightly job` (`f963c98f83`), which landed about two hours after this PR was opened and touches the same lines.

That commit converted the two SBE encoder `#[expect(clippy::useless_conversion)]` attributes in `crates/serialization/src/sbe/market/common.rs` to `#[allow]`. This PR removes them instead: both encoders now go through `numeric::raw_to_wire`, and because clippy checks a generic body before monomorphization, the conversion is invisible to that lint in either width, so no suppression is needed at all. I resolved in favour of this branch's form - the attributes are deleted rather than converted - and did the merge as its own commit rather than a rebase, so the original commit is unchanged.

Two statements in the description above are now out of date, and I have left the description as filed rather than editing it after the fact:

- It says `make clippy` runs `--all-features` and therefore always resolves high precision. Still true of that target, but `check-code-standard-precision` now exists.
- It says nothing in CI observes the standard-precision configuration. The new nightly job does, and its comment describes the same gap this PR was written against - lints that only fire when the raw types are `u64`/`i64` reaching develop unnoticed.

That job is what makes the change here checkable rather than merely argued: under `check-code-standard-precision` the codec paths carry no width-conditional suppression, so the sweep passes without any `#[allow]` in them. The rest of the PR is unaffected - the decode sites, the `Money` fix, and the tests are unchanged by the merge, and the three width configurations in the Testing section are still green (401 tests each).

One thing worth mentioning from your Makefile comment: the verification trick you documented - delete an `#[allow(clippy::useless_conversion)]` in `quantity.rs` and confirm clippy reports it under that selection - is how the tests here were validated too, by reverting each decode site individually rather than only comparing against develop.

cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I checked the five decode paths against the resolved `PriceRaw`, `QuantityRaw`, and `MoneyRaw` aliases, including the feature graph where `nautilus-model/high-precision` is enabled without `nautilus-serialization/high-precision`. Both new round-trip tests pass in that configuration, standard-precision Clippy is clean, and the branch merges cleanly with the current `develop`.

The generic conversion boundary removes the serializer-local width decision while preserving the wire representation and existing overflow errors. This looks good to merge :ok_hand:

PR #4568, opened 2026-07-25
Fix reverse exchange-rate spread side and make route selection deterministic

folknor:
`get_exchange_rate` in `crates/common/src/xrate.rs` is the only exchange-rate implementation in the codebase - reached from `Cache::get_xrate` for portfolio valuation and PnL, and from the Cython cache through the pyo3 binding - and it has two defects in how it builds its conversion graph.

## Reverse edges invert the requested side instead of the opposite side

The function collapsed `quotes_bid` and `quotes_ask` into a single `effective_quotes` map according to `price_type`, then built the graph from that one map, adding `BASE -> QUOTE` at rate `r` and `QUOTE -> BASE` at `1/r`. Since `r` is the already-selected side, a `Bid` request produced a reverse edge of `1/bid`. That is the wrong side of the spread: selling QUOTE to acquire BASE crosses the original ask, so the reverse bid is `1/ask`, and symmetrically the reverse ask is `1/bid`. The collapse had already discarded the other side by the time the reverse edge was built, so the error was structural rather than arithmetic. A reverse-bid rate came out too high, flattering any position valued through an inverted pair, and the error compounds once per inverted leg on a multi-hop route.

The fix pairs each bid with its ask by key before constructing the graph and derives both directional rates per pair: `Bid -> (bid, 1/ask)`, `Ask -> (ask, 1/bid)`, `Mid -> (mid, 1/mid)`. `Mid` is deliberately unchanged: the inverse of the midpoint preserves reciprocity, `mid(A/B) * mid(B/A) = 1`, which the arithmetic midpoint of the inverted market does not. The public signature is unchanged, so `Cache::get_xrate` and the pyo3 wrapper are untouched.

## Route selection depends on hash iteration order

The graph was populated by iterating an `AHashMap`, whose order varies per process, and the DFS marks a node visited when pushing it. When several indirect routes exist, whichever route the arbitrary adjacency order happened to reach first won, so two runs over identical quotes could return rates differing by the compounded spread of a different path. A direct one-hop edge could not lose - it is pushed with the direct rate and returns it when popped - so the nondeterminism was confined to the choice among indirect routes.

The fix sorts every adjacency list by `(neighbor, rate)` before the search, which removes all map-iteration influence while leaving the DFS itself unchanged. The ordering is total rather than by neighbor alone because parallel edges - a caller supplying both `EUR/USD` and `USD/EUR` - would otherwise retain their randomized relative order. Among parallel edges to the same neighbour the higher rate wins the visited claim; that is a mechanical tie-break, not a best-route policy, since the rate never affects ordering between distinct neighbours. Callers whose ordering happened to select a different indirect route may see a changed rate, which is unavoidable when removing nondeterminism; the function does not promise a shorter or better route, only a consistent one.

## Missing and non-positive quote sides

Pairing by key also makes the documented "missing side" error real for `Bid` and `Ask`, which previously never consulted the other map at all - only `Mid` did. The `quotes_bid.len() != quotes_ask.len()` check proves equal counts, not equal key sets; equal counts plus a successful lookup for every bid key now does prove it. The positivity guard also only examined the selected side, so a zero or negative value on the unused side could still become a denominator; both sides are now validated before either is inverted.

## Testing

`test_inverse_pair_uses_opposite_spread_side` covers all three price types against `1/ask`, `1/bid` and `1/mid`. `test_indirect_route_is_deterministic_across_hash_seeds` builds the same logical quote table under four explicit `ahash::RandomState::with_seeds` states, asserts the raw iteration orders actually differ, and requires all four to return the one canonically ordered route - a loop over a cloned map would prove nothing about hash order. `test_non_positive_opposite_side_is_skipped` and `test_equal_length_quotes_with_different_keys` cover the validation changes, and `test_cycle_handling` now asserts an exact canonical rate where it previously tolerated a range because the edge order was nondeterministic.

Both defects were verified to FAIL against the unfixed code, separately: restoring the same-side inversion fails only the `Bid` and `Ask` cases while `Mid` continues to pass, and removing the adjacency sort fails only the determinism test, on its rate assertion rather than its iteration-order assertion. Two further tests, `test_same_currency_shortcut_precedes_all_validation` and `test_quote_map_errors_precede_invalid_price_type`, pin the validation order so the identical-currency shortcut keeps returning a rate of one before any quote or price-type inspection.

No existing Rust expectation changed: every prior test in the file used `Mid`, or `Bid`/`Ask` on a direct edge only, which is how the reverse-side error survived. Three Python expectations in `tests/unit_tests/accounting/test_calculators.py` did encode it and are updated - `test_get_rate_for_inverse2` and `result1` of `test_calculate_exchange_rate_by_inference` to their recomputed values, and `test_get_rate_for_inverse1`'s fixture, which quoted `BTC/USD` with a bid above its ask, is corrected to a valid spread; its expected value is unchanged but now follows from `1/ask`. The forward-`Ask` and `Mid` expectations are untouched.

### DIFF
--- a/crates/common/src/python/xrate.rs
+++ b/crates/common/src/python/xrate.rs
@@ -31,10 +31,11 @@ use crate::xrate::get_exchange_rate;
 ///
 /// # Errors
 ///
-/// Returns an error if:
-/// - `price_type` is equal to `Last` or `Mark` (cannot calculate from quotes).
+/// For conversions between distinct currencies (an identical `from_currency` and `to_currency`
+/// returns a rate of one without inspecting the quotes), returns an error if:
 /// - `quotes_bid` or `quotes_ask` is empty.
 /// - `quotes_bid` and `quotes_ask` lengths are not equal.
+/// - `price_type` is equal to `Last` or `Mark` (cannot calculate from quotes).
 /// - The bid or ask side of a pair is missing.
 #[pyfunction]
 #[pyo3_stub_gen::derive::gen_stub_pyfunction(module = "nautilus_trader.common")]
--- a/crates/common/src/xrate.rs
+++ b/crates/common/src/xrate.rs
@@ -22,6 +22,18 @@ use nautilus_model::enums::PriceType;
 use rust_decimal::Decimal;
 use ustr::Ustr;
 
+fn directional_rates(bid: Decimal, ask: Decimal, price_type: PriceType) -> (Decimal, Decimal) {
+    match price_type {
+        PriceType::Bid => (bid, Decimal::ONE / ask),
+        PriceType::Ask => (ask, Decimal::ONE / bid),
+        PriceType::Mid => {
+            let mid = (bid + ask) / Decimal::TWO;
+            (mid, Decimal::ONE / mid)
+        }
+        _ => unreachable!("Price type was validated before graph construction"),
+    }
+}
+
 /// Calculates the exchange rate between two currencies using provided bid and ask quotes.
 ///
 /// This function builds a graph of direct conversion rates from the quotes and uses a DFS to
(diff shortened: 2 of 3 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. This looks good to me: the focused xrate tests pass, and the remaining CI failures are infrastructure-related. There's one small function-order follow-up for `directional_rates`, which I'll handle locally.

PR #4569, opened 2026-07-25
Clean all index entries when purging an order

folknor:
`Cache::purge_order` leaves several index entries behind, one of which makes a later query panic outright, and the cleanup it does perform cannot see state the order object never learned.

## A purged order can remain in `position_orders`, and `orders_for_position` then panics

`add_position_id` writes both the forward `index.order_position` and the reverse `index.position_orders` from a *supplied* position ID - the path taken when a position ID arrives on a command, including `add_order(.., Some(position_id), ..)` and the matching engine's OTO handling. An order's own `position_id` field, by contrast, is assigned only when a fill applies. So a fill-less order with a command-supplied position ID is present in both indexes while `order.position_id()` is `None`.

`purge_order` cleaned the reverse index only under `if let Some(position_id) = order.position_id()`, and discarded the value returned by `index.order_position.remove(..)` - the one piece of state that identified the entry. The stale `ClientOrderId` then reaches `get_orders_for_position` -> `get_orders_for_ids`, which does `.unwrap_or_else(|| panic!("Order {client_order_id} not found"))`. Any `orders_for_position` call for that position panics.

The fix cleans the union of both sources - the order object's position ID and the value returned from the forward index - rather than choosing one as authoritative, since neither covers the other's case.

## Historical venue order IDs survive the purge

An order accumulates a history of venue order IDs across cancel/replace, kept in `venue_order_ids` and exposed by `Order::venue_order_ids()`. `purge_order` removed only the current one, so every earlier ID stayed in `index.venue_order_ids` pointing at an order that no longer exists. `check_integrity` iterates that map and reports a failure when the client order is absent, and the map grows without bound across a long session of replaces.

The fix collects the history, the current ID and the value returned from the forward `client_order_ids` map - the current ID must be included separately because a fill assigns `venue_order_id` without appending to the history - and removes each entry only if the reverse map still names the purged order as its owner. An unconditional removal would be unsafe: a collision can already have rebound a historical ID to another live order, and deleting that mapping would turn an index leak into index corruption. A cancel/replace chain is not a cross-order transfer in this model, so a different owner always means a collision rather than a legitimate handover.

## Speculative position metadata and empty registries

`add_position_id` also seeds `position_strategy`, `strategy_positions` and `venue_positions`, so a fill-less order could leave an entire position family behind after its last association was purged. `check_integrity` rejects stale `position_strategy` and `strategy_positions` entries, and a position query can panic when a reverse bucket names a missing position. These are now retired when no cached `Position` exists for the ID and no other order references it - determined from the reverse bucket, or by consulting the remaining forward mappings when that bucket is itself absent.

The `strategies` and `exec_algorithms` registries were never retired at all, which violates the invariant that every registry entry has a corresponding orders bucket. They are now retired when their last order goes, and an absent bucket is deliberately distinguished from an empty one throughout: an absent bucket is left alone so a pre-existing defect stays visible to `check_integrity` rather than being silently papered over.

Where a real position or another position for the same instrument still exists, the companion `position_orders`, `strategy_orders` and `instrument_orders` buckets are preserved but emptied, because `check_integrity` requires those keys to exist. `purge_position` now completes that lifecycle, retiring a companion bucket when its last position disappears and the bucket is present and empty.

## Deliberately deferred to a follow-up

Two related defects in the same area are not in this PR, because both hinge on a trade-off worth your call rather than mine.

`add_venue_order_id` validates only the client-to-venue direction. It rejects a different venue ID for the same order, but never checks whether the incoming venue order ID is already owned by a different client order before rebinding `index.venue_order_ids`, so a second claim silently steals the mapping and venue-ID lookups then resolve to the wrong order. Rejecting the claim is the obvious fix, but it converts today's silent misrouting into a fail-fast error: `IdsGenerator::get_venue_order_id` propagates the result and its production callers `unwrap`, so a duplicate generated ID would abort a backtest rather than quietly misroute. Silent wrong-order routing seems clearly worse to me, but that is a behavioural change I would rather raise than bundle.

`Cache::refresh_order` compounds it by skipping `add_venue_order_id` whenever the reverse key merely exists, without checking who owns it, so cross-owner collisions are never reported through that path. It only becomes meaningful once the validation above exists, so the two travel together.

One test in this PR seeds a rebound venue-ID mapping directly through the index maps rather than through `add_venue_order_id`, specifically so it keeps passing once that validation lands.

## Testing

Eleven tests in `crates/common/src/cache/tests.rs` cover: the command-supplied position link, historical venue IDs, a rebound historical ID surviving the purge of its former owner, speculative position metadata cleanup, an empty bucket preserved for a real position with `check_integrity` asserted, registry retirement for the last strategy and exec-algorithm order, the `purge_order` -> `purge_position` companion-bucket sequence asserted directly on the index maps, and one test per union source in isolation - the object's position ID with the forward entry absent, the current venue ID present in neither history nor forward map, and an index-only forward venue ID with no object venue ID.

Verified against mutants rather than only against the unfixed baseline: replacing the ownership predicate with an unconditional removal fails exactly the rebound test; dropping the forward half of the position union fails three tests, and the command-position case fails with the real panic rather than a soft assertion; and the companion-bucket test fails without the `purge_position` change.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -3243,29 +3243,26 @@ impl Cache {
                 }
             }
 
-            if let Some(instrument_orders) =
-                self.index.instrument_orders.get_mut(&details.instrument_id)
-            {
-                instrument_orders.remove(&client_order_id);
-            }
+            // As with the strategy buckets below, an absent bucket is left absent: recreating
+            // it would suppress the `index.instrument_positions` integrity check.
+            // As with the strategy buckets below, an absent bucket is left absent: recreating
+            // it would suppress the `index.instrument_positions` integrity check.
+            let instrument_orders_became_empty = self
+                .index
+                .instrument_orders
+                .get_mut(&details.instrument_id)
+                .is_some_and(|instrument_orders| {
+                    instrument_orders.remove(&client_order_id);
+                    instrument_orders.is_empty()
+                });
 
             let has_instrument_positions = self
                 .index
                 .instrument_positions
                 .get(&details.instrument_id)
                 .is_some_and(|positions| !positions.is_empty());
-            let has_instrument_orders = self
-                .index
-                .instrument_orders
-                .get(&details.instrument_id)
-                .is_some_and(|orders| !orders.is_empty());
 
-            if has_instrument_positions {
-                self.index
-                    .instrument_orders
-                    .entry(details.instrument_id)
-                    .or_default();
-            } else if !has_instrument_orders {
+            if instrument_orders_became_empty && !has_instrument_positions {
                 self.index.instrument_orders.remove(&details.instrument_id);
             }
 
@@ -3377,24 +3374,27 @@ impl Cache {
         }
 
         for strategy_id in strategy_ids {
-            if let Some(strategy_orders) = self.index.strategy_orders.get_mut(&strategy_id) {
-                strategy_orders.remove(&client_order_id);
-            }
-
-            let has_orders = self
+            // An absent reverse bucket is not an empty one: it means the index is already
+            // inconsistent, possibly while another cached order still uses this strategy.
+            // Retiring the registry entry here would both drop a live strategy from
+            // `strategy_ids` and stop `check_integrity` reporting the missing bucket, so the
+            // absent case is left exactly as found.
+            let strategy_orders_became_empty = self
                 .index
                 .strategy_orders
-                .get(&strategy_id)
-                .is_some_and(|strategy_orders| !strategy_orders.is_empty());
+                .get_mut(&strategy_id)
+                .is_some_and(|strategy_orders| {
+                    strategy_orders.remove(&client_order_id);
+                    strategy_orders.is_empty()
+                });
+
             let has_positions = self
                 .index
                 .strategy_positions
                 .get(&strategy_id)
                 .is_some_and(|strategy_positions| !strategy_positions.is_empty());
 
-            if has_positions {
-                self.index.strategy_orders.entry(strategy_id).or_default();
-            } else if !has_orders {
+            if strategy_orders_became_empty && !has_positions {
                 self.index.strategy_orders.remove(&strategy_id);
                 self.index.strategies.remove(&strategy_id);
             }
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -3184,82 +3184,108 @@ impl Cache {
     ///
     /// For safety, an order is prevented from being purged if it's open.
     pub fn purge_order(&mut self, client_order_id: ClientOrderId) {
-        // Check if order exists and is safe to purge before removing
-        let order_cell = self.orders.get(&client_order_id).cloned();
+        struct OrderDetails {
+            is_open: bool,
+            instrument_id: InstrumentId,
+            strategy_id: StrategyId,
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @folknor

Thank you for the PR. There is one missing-bucket case in the strategy cleanup that still needs handling.

`has_orders` treats an absent `strategy_orders` bucket as empty. If another cached order uses the same strategy but that reverse bucket is missing, purging this order reaches `else if !has_orders` and removes the strategy from `index.strategies`. `strategy_ids()` then omits a strategy still used by a cached order, and `check_integrity()` no longer reports the missing reverse bucket. The `has_positions` branch has the complementary issue: `entry(...).or_default()` recreates a missing bucket and hides the defect.

Could you preserve the absent state, only retire the registry when the reverse bucket existed and became empty, and add the matching regression test with two orders sharing a strategy?

folknor:
Done, thank you - the absent case was doing more damage than it first looks.

You are right on both halves. `has_orders` was `is_some_and(|orders| !orders.is_empty())`, which reads false for an absent bucket exactly as it does for an empty one, so purging an order whose reverse bucket was already missing fell into `else if !has_orders` and retired the strategy even with another cached order still using it. The part I had not seen until your comment is the second-order effect: `check_integrity` detects a missing reverse bucket by iterating `index.strategies`, so removing the strategy there does not merely drop it from `strategy_ids()`, it removes the only route by which that inconsistency would ever be reported. The purge was erasing the evidence of the defect it was walking over. The `entry(..).or_default()` in the `has_positions` branch arrived at the same place from the other side, by manufacturing the bucket.

`purge_order` now mirrors what `purge_position` already does - `get_mut(..).is_some_and(|orders| { orders.remove(..); orders.is_empty() })` for "existed and became empty" - so the registry is retired only in that case and an absent bucket is left exactly as found. That puts both purge paths on one rule rather than introducing a second one. The cases are now:

- bucket exists, still non-empty: nothing retired.
- bucket exists, became empty, positions remain: the empty bucket is kept, so the `index.strategies` invariant still holds.
- bucket exists, became empty, no positions: bucket removed and the strategy retired.
- bucket absent: nothing touched, so `check_integrity` still reports it.

`test_purge_order_preserves_absent_strategy_orders_bucket` pins this with two orders sharing a strategy and the reverse bucket removed. Purging one leaves the strategy in `index.strategies`, leaves the bucket absent rather than fabricated, and `check_integrity()` still returns false. It was verified to fail against the previous code, on the assertion that the strategy must not be retired.

The existing assertion covering an emptied bucket with a live position is unchanged and still passes: that case was previously held up by the `or_default()` and is now held up by simply not removing the bucket.

Reviewing for the same pattern elsewhere turned up one more instance of it in the same function, which I have fixed in this revision rather than leave for a second round. The `instrument_orders` cleanup did the identical `entry(..).or_default()` when positions remained, which suppresses the `index.instrument_positions` check in `check_integrity` the same way. It carries no registry retirement, so it cannot drop a live instrument the way the strategy case dropped a live strategy, but it hides the inconsistency just as effectively. Same shape applied, and `test_purge_order_preserves_absent_instrument_orders_bucket` covers it with a live position and a missing bucket; also verified to fail against the previous code. `purge_position`, `exec_algorithm_orders` and both `position_orders` paths already preserved absence correctly, so those are untouched.

cjdsellers (approved):
Thank you for the follow-ups @folknor 🙏

PR #4570, opened 2026-07-25
Run Rust doctests before nextest

seungpyoson:
## Summary

- run Rust doctests before the core and adapter nextest suites in both Linux x86 workflows
- keep the existing workspace, feature, profile, and cleanup commands unchanged

## Root cause

The new doctest gate ran after the full nextest build had accumulated the complete test-binary set. On unrelated pull requests, `nautilus-persistence` doctest linking then exhausted the 146 GB GitHub runner filesystem, causing `lld` to terminate with `SIGBUS` followed by `No space left on device` errors in `/tmp`.

Running doctests first releases their transient rustdoc/linker files before nextest adds the full test-binary set. This preserves the gate and avoids a retry, suppression, reduced feature set, or separate compatibility path.

## Verification

- `make pre-commit`
- `git diff --check`

### DIFF
--- a/.github/workflows/build-v2.yml
+++ b/.github/workflows/build-v2.yml
@@ -198,6 +198,12 @@ jobs:
         if: needs.plan.outputs.run-rust-tests == 'true'
         uses: ./.github/actions/common-test-data
 
+      # Doctests run before nextest so their transient linker artifacts are
+      # released before the full test-binary set accumulates on disk.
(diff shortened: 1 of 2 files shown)


cjdsellers:
Thanks @seungpyoson,

Have now landed this to unblock CI for PRs and credited you in the release notes.

PR #4572, opened 2026-07-25
Fix reconnect-path liveness and session fencing in network clients

folknor:
Three defects on the reconnect path in `crates/network`, one of which can wedge a client permanently.

## A stalled write can wedge reconnect indefinitely

The socket writer task writes with `active_writer.write_all(&write_buf).await` and no deadline. If the peer accepts the connection and then stops reading, the socket buffer fills and that await never returns, so the writer never loops back to `writer_rx.recv()`. The reconnect path queues `WriterCommand::Update` on that same channel and then waits on a oneshot for the drain confirmation, so it waits for a reply the wedged writer cannot send. Reconnect stalls forever, and because the controller is inside `inner.reconnect()`, the attempt counter never advances either, so `reconnect_max_attempts` cannot terminate it. The websocket client has the same shape in two places: an unbounded active `send`, and an unbounded reconnect-buffer drain (the socket client already bounds its drain).

The fix bounds the writes with a new internal `WRITE_TIMEOUT_SECS` deadline - not the configurable dial timeout - and treats a timeout exactly as the existing I/O-error path does: requeue the whole logical message, CAS `Active -> Reconnect`, and abandon the old writer. The controller's confirmation awaits are deliberately left untimed. Cancelling them would risk orphaning an already-swapped writer, which is why the reconnect timeout covers dialing only; with the writes bounded there is no longer unbounded I/O ahead of those awaits.

Cancellation does not roll back bytes, so a stalled write may leave a prefix in the kernel send buffer. Closing the old socket makes that prefix end at EOF rather than allowing a replay to be appended behind it, and the complete message is replayed on the new connection - the same at-least-once property the existing failed-send requeue already had.

## `post_disconnection` never fired when reconnects were exhausted

The owner's `post_disconnection` handler was invoked only from the controller's disconnect branch. On reconnect exhaustion the controller stored `Closed` and exited silently, so the one case an owner most needs to observe - the client giving up permanently - was the one case it was never told about. `close()` was also silent, while `disconnect()` notified.

The fix introduces a terminal finalizer holding the callback behind an atomic once-guard, reached from every controller exit and from `close()`. The handler runs after the `Closed` transition, so it observes the true terminal state, and before `notify_waiters`, so a woken waiter cannot outrun it. Exhaustion transitions with `compare_exchange(Reconnect, Closed)` and yields to a concurrent disconnect rather than overwriting it. `close()` waits for finalization rather than for `is_closed()`, and on timeout forces finalization through the same guard instead of waiting on an abort that cooperative cancellation may never deliver. A panicking callback still publishes completion, so a panic cannot strand later callers. Dropping the client aborts the controller without notifying, which is deliberate: a drop is not a connection lifecycle event.

## One stale frame could be dispatched after the session ended

Both read loops checked the connection mode before awaiting a read and dispatched whatever arrived without re-checking, so a frame that arrived while the mode moved to `Reconnect` was delivered from the abandoned session. In the socket client a single read can also contain several delimited frames, and a handler can itself end the session, so one check is not enough.

The fix re-checks after the read returns and again immediately before each handler call, and on either observation discards the entire task-local buffer - new bytes, undispatched frames, and any partial frame accumulated by earlier reads. Carrying a partial frame across would splice an old fragment onto bytes from a new connection and manufacture a frame that neither session sent. The websocket loop checks before both the message and ping handlers. Each drop logs its byte or message count at debug: this is an expected reconnect race rather than an error, but silent discards would make future investigation opaque.

## Deliberately not in this PR

The fourth member of this class is the zero-backoff reconnect storm: both controllers reset the backoff *and* the attempt count on handshake success, so an accept-then-close server is retried forever and `reconnect_max_attempts` never trips. Fixing it needs a decision about when a connection counts as proven stable, which is a contract question rather than a mechanical change, so it is left alone here.

The Python `close()` path also still stores `Closed` directly rather than going through the finalizer, so `post_disconnection` does not fire when that path times out and forces closure. The fix is a few lines, but it edits `crates/network/src/python/socket.rs`, which is part of the v2 python-generation surface. The change is body-only - no signature, attribute or docstring changes, so nothing the stub generator reads - but I could not verify that locally, so I have kept it out rather than guess. I'd rather you deal with the python side of things.

## Testing

Fourteen tests in `crates/network`. Liveness: a scripted `AsyncWrite` that accepts a prefix, signals backpressure and then stays pending, driven under `#[tokio::test(start_paused = true)]`, asserting reconnect is requested, the complete message is retained, the queued `Update` is consumed and the message is replayed through a recording writer; `tokio::io::duplex(1)` models "accepted, then stopped reading" deterministically, since kernel buffer sizes are platform-dependent. Equivalents cover the websocket active send and the websocket replay drain. Notification: exhaustion and graceful close each assert exactly one callback, with further tests for concurrent finalizers, both CAS interleavings, a controller aborted before finalization, and a panicking callback. Fencing: scripted readers deliver a partial frame that completes only after the mode flips, plus a two-frames-in-one-read case where the first handler ends the session, and a websocket case asserting neither the message nor the ping handler runs.

Ordering comes from channels, `Notify` signals and virtual time; no test uses a sleep as an ordering primitive, and the one real-time timeout is a deadlock watchdog. Each defect was verified by mutation rather than only against the unfixed baseline: removing a write bound fails the corresponding liveness test with "writer update should not remain queued behind a stalled write"; removing the notification calls fails the exhaustion test on the callback assertion; removing the fences fails both stale-frame tests. Note the guards are intentionally redundant - the controller epilogue also finalizes on exit, and the two socket fences cover different windows - so falsifying a defect requires removing all of its guards, which is how these were checked.

### DIFF
--- a/crates/adapters/lighter/src/websocket/handler.rs
+++ b/crates/adapters/lighter/src/websocket/handler.rs
@@ -1696,10 +1696,15 @@ pub(crate) fn should_retry_lighter_ws_error(error: &LighterWsError) -> bool {
     match error {
         LighterWsError::Network(_) => true,
         // Closed and BrokenPipe are terminal on this client; only Timeout
-        // (wait_for_active) can recover if the connection comes up.
+        // (wait_for_active) can recover if the connection comes up. WriteTimeout
+        // is not retryable: the write was cancelled after it began, so the peer
+        // may already have the message and a retry could duplicate it.
         LighterWsError::Transport(send_error) => match send_error {
             SendError::Timeout => true,
-            SendError::Closed | SendError::ConnectionChanged | SendError::BrokenPipe(_) => false,
+            SendError::Closed
+            | SendError::ConnectionChanged
+            | SendError::BrokenPipe(_)
+            | SendError::WriteTimeout => false,
         },
         LighterWsError::Authentication(_)
         | LighterWsError::Parse(_)
@@ -3130,6 +3135,10 @@ mod tests {
     #[case::client_does_not_retry(LighterWsError::Client("no active WebSocket client".into()), false)]
     #[case::transport_closed_does_not_retry(LighterWsError::Transport(SendError::Closed), false)]
     #[case::transport_timeout_retries(LighterWsError::Transport(SendError::Timeout), true)]
+    #[case::transport_write_timeout_does_not_retry(
+        LighterWsError::Transport(SendError::WriteTimeout),
+        false
+    )]
     #[case::transport_broken_pipe_does_not_retry(
         LighterWsError::Transport(SendError::BrokenPipe(
             "writer closed".into(),
@@ -3144,6 +3153,7 @@ mod tests {
     #[rstest]
     #[case::closed(SendError::Closed)]
     #[case::timeout(SendError::Timeout)]
+    #[case::write_timeout(SendError::WriteTimeout)]
     #[case::broken_pipe(SendError::BrokenPipe("writer dropped".into()))]
     fn send_error_converts_into_transport_variant(#[case] send_error: SendError) {
         let err: LighterWsError = send_error.into();
--- a/crates/network/src/error.rs
+++ b/crates/network/src/error.rs
@@ -28,6 +28,13 @@ pub enum SendError {
     /// Timed out waiting for the client to become active.
     #[error("send failed: timeout waiting for active state")]
     Timeout,
+    /// Timed out while writing to the transport, so delivery is undetermined.
+    ///
+    /// Unlike [`SendError::Timeout`], which reports that a send never started, the write was
+    /// cancelled after it began: the peer may or may not have received the message. Callers must
+    /// not treat this as a plain retry.
+    #[error("send failed: timed out writing to transport, delivery undetermined")]
+    WriteTimeout,
     /// The connection changed before an ownership-bound message reached the writer.
     #[error("send failed: connection changed before write")]
     ConnectionChanged,
--- a/crates/network/src/mode.rs
+++ b/crates/network/src/mode.rs
@@ -15,9 +15,39 @@
 
 //! Connection mode enumeration for socket clients.
 
-use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::{
+    Arc,
+    atomic::{AtomicBool, AtomicU8, Ordering},
+};
 
 use strum::{AsRefStr, Display, EnumString};
+
+/// Irreversible validity token for a single connection's read task.
+#[derive(Clone, Debug)]
+pub(crate) struct ReadSessionFence {
+    valid: Arc<AtomicBool>,
+}
+
+impl ReadSessionFence {
+    /// Creates a valid fence for a newly spawned read task.
+    #[must_use]
+    pub(crate) fn new() -> Self {
+        Self {
+            valid: Arc::new(AtomicBool::new(true)),
+        }
+    }
+
+    /// Invalidates the associated read session.
+    pub(crate) fn invalidate(&self) {
+        self.valid.store(false, Ordering::SeqCst);
+    }
+
+    /// Returns whether the associated read session is still current.
+    #[must_use]
+    pub(crate) fn is_valid(&self) -> bool {
+        self.valid.load(Ordering::SeqCst)
+    }
+}
 
 /// Connection mode for a socket client.
 ///
--- a/crates/network/src/socket/client.rs
+++ b/crates/network/src/socket/client.rs
@@ -36,7 +36,7 @@ use std::{
     pin::pin,
     sync::{
         Arc,
-        atomic::{AtomicU8, Ordering},
+        atomic::{AtomicBool, AtomicU8, Ordering},
     },
     time::Duration,
 };
@@ -44,7 +44,7 @@ use std::{
 use bytes::Bytes;
 use nautilus_core::CleanDrop;
 use nautilus_cryptography::providers::install_cryptographic_provider;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tokio_tungstenite::tungstenite::{Error, client::IntoClientRequest, stream::Mode};
 
 use super::{SocketConfig, TcpMessageHandler, TcpReader, TcpWriter, WriterCommand};
@@ -53,7 +53,7 @@ use crate::{
     dst,
     error::{SendError, is_connection_drop_io_error},
     logging::{log_task_aborted, log_task_started, log_task_stopped},
-    mode::ConnectionMode,
+    mode::{ConnectionMode, ReadSessionFence},
     net::TcpStream,
     tls::{Connector, create_tls_config_from_certs_dir, tcp_tls},
 };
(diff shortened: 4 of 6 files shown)


cjdsellers:
Hi @folknor

Thank you for the PR. The bounded write paths look good, but I found two remaining races in the stated guarantees.

- The read fence uses the shared `ConnectionMode`, which returns to `Active` for the replacement session. Since the old read task is aborted without awaiting termination, it can resume after reactivation and dispatch another buffered old-session frame. Could you bind the reader to a session identity or ensure it has terminated before restoring `Active`, and add a test that cycles `Reconnect` back to `Active` while the old handler is still in flight?

- The forced finalization path also runs `post_disconnection` inline after `close_with_timeout()` has already timed out. If the controller has not claimed the finalizer and that callback blocks, `close()` remains blocked indefinitely. The current blocking-callback test covers only the case where the controller claimed it first, so this needs the inverse interleaving too.

folknor:
Thanks - both are real, and the first one shows the fence I added was answering a weaker question than the one that matters.

## The read fence could not tell an old session from a new one

Every check I added reads the shared `ConnectionMode`, and reconnect restores that atomic to `Active` for the replacement session. The old reader is disposed of with `read_task.abort()` and nothing awaits its termination, and `abort` is cooperative: it takes effect only when the task next reaches an await point. In `SocketClientInner::run_read_loop` the drain over delimited frames contains no await at all, so once `read_buf` has returned the task runs the whole drain straight through. Frame 1 dispatches while the mode is `Active`; reconnect stores `Reconnect`, calls `abort()` (no effect, since no await is reached), dials, and compare-exchanges back to `Active`; the old task then reaches the per-handler check before frame 2, reads `Active`, and dispatches an abandoned session's frame into the live one. The websocket loop has the same shape with a one-message window, between `reader.next()` returning and the task being polled again.

I looked at awaiting reader termination before restoring `Active` and decided against it. Both handlers are arbitrary synchronous callbacks, so `abort` cannot preempt one, and a handler that blocks - or that waits for `Active` or for `post_reconnection` - closes a cycle in which reconnect waits for the reader while the reader waits for reconnect. That is the same unbounded-wait class the rest of this PR removes, and it would sit after the dial and writer-swap bounds. A timeout around the join does not rescue it either: on expiry you must restore `Active` and take the race back, remain in `Reconnect` and wedge reconnection, or fall back to an identity check, in which case the identity check was the fix.

So the reader is now bound to a session identity. `ReadSessionFence` is an irreversible token over an `Arc<AtomicBool>`: each spawned reader gets a fresh one, the inner retains a clone for the current reader, and reconnect invalidates the old token before both the `abort()` and the `Reconnect -> Active` compare-exchange, with the replacement reader receiving a new one. Every existing mode check now also tests its own token, mode first and token second, so an old reader cannot observe the replacement's `Active` while relying on a validity result read before invalidation. The mode guards stay: the token distinguishes sessions, while the mode still carries disconnect and close for the writer, the bindings and lifecycle control. Invalidation also happens on the disconnect and drop paths, where it is merely tidy since the mode is terminal there.

What this does not do is preempt a callback that had already entered before invalidation - nothing can, for a synchronous callback. The guarantee is that an invalidated reader cannot begin another dispatch. The read buffer remains task-local, so the partial-frame no-splice property is unchanged.

## Forced finalization ran the callback on the caller

`close_with_timeout` bounded only the wait. On expiry it called `transition_and_finalize()`, and when the controller had not claimed the once-guard that call won the compare-exchange and ran `post_disconnection` inline on the caller's task, unbounded - after the timeout had already fired and with no way to preempt synchronous code. The timeout was guaranteeing the interleaving where somebody else owned the callback and not the one where the caller did.

I fixed this in the finalizer rather than in `close()`, because a controller that claims the guard first has the same exposure: it would pin a runtime worker on an arbitrary blocking callback. `finalize_closed` now claims and dispatches instead of claiming and running. The compare-exchange still admits exactly one winner, and that winner hands the callback and the completion guard to a job executed off the claimant - via `Handle::try_current()` to `spawn_blocking` where there is a runtime, and a detached thread otherwise, which keeps the finalizer callable from a synchronous context. `Closed` is still stored before the callback is invoked, the job owns the completion guard so a discarded job, a panicking callback and the no-callback branch all publish completion exactly once, and `notification_completed` cannot be observed true while the callback is still running.

One residual is worth stating rather than engineering around: a permanently blocked callback occupies a blocking-pool worker, and since these jobs are not mandatory, runtime shutdown can discard one that is still queued. The completion guard moving into the job means neither case can strand a waiter.

## Testing

`test_read_loop_drops_remaining_frames_after_session_replaced` is the cycle you asked for. One read returns two frames, the first handler signals entry and blocks on a condvar, and another thread stores `Reconnect`, invalidates the token, aborts the handle and restores `Active` before releasing the handler. Only the first frame is dispatched. `test_read_loop_drops_partial_old_session_frame` was strengthened the same way: it now cycles back to `Active` after invalidation and asserts that the old partial frame cannot combine with bytes that arrive afterwards, while a fresh reader still delivers them. The websocket case uses a transport that blocks inside `poll_next` immediately before returning `Ready`, which is what makes the abort non-preemptive, and is parameterized over text, binary and ping.

`test_close_dispatches_unclaimed_blocking_callback` covers the inverse interleaving. The controller is left pending so it never claims the guard, `close_with_timeout` runs as its own task with a short internal timeout, and the callback signals entry then blocks. It asserts that close returns while the callback is still blocked, that completion is still false at that point, and after release that exactly one invocation occurred even across a further close and finalize. The existing test is kept as the opposite interleaving.

All of these were verified to fail against the unfixed code. The guards are deliberately redundant, so it is worth being precise about what each one needed: the multi-frame test fails on removing only the per-handler identity check, the partial-frame test and the three websocket cases require identity removed from both the post-read and the handler fences, and the close test fails when the callback is put back inline. Ordering throughout comes from channels and condvars rather than sleeps, and every rendezvous and completion wait is bounded with a named failure, so a future regression fails rather than hangs.

folknor:
@cjdsellers Unrelated to this PR, but you will want to know that `develop` does not currently compile under `--all-targets`.

1006789062 ("Fix Polymarket non-power tick handling") added a `price_increment: Price` parameter to `parse_quote_from_snapshot` and `parse_quote_from_price_change` in `crates/adapters/polymarket/src/websocket/parse.rs`, and updated the in-file tests, but `crates/adapters/polymarket/benches/data.rs` still calls both with the old argument list. So `cargo clippy --all-targets` and `cargo bench -p nautilus-polymarket` both fail with E0061 at `data.rs:108` and `data.rs:133`. Passing the instrument's price increment at those two call sites is all it needs.

It is easy to miss because benches are only compiled under `--all-targets`, so the ordinary test job stays green.

This branch is rebased onto 80a653ce83 and therefore inherits the failure. It touches only `crates/network` and nothing in the Polymarket adapter, so if you see that error against this PR it is not from these changes.

folknor:
@cjdsellers I'm reworking this PR now to accomodate your latest websocket changes.

folknor:
Rebased onto the connection-epoch work and squashed to a single commit. The obvious question is whether epochs subsume this PR, and for one of its parts they nearly do - so it is worth being precise about which.

## The fences stay, because an epoch is advisory and a fence is not

The epoch supplies exactly the durable per-session identity `ReadSessionFence` was introduced for: the writer advances it when it installs the replacement sink, and each reader captures its value, so an old reader's identity can no longer be laundered by the `Reconnect -> Active` restore. That is the same property arrived at more cheaply, and if it were enforced in the transport I would have dropped the fence.

It is not enforced there. `spawn_message_handler_task` tags the message and dispatches it; nothing compares the reader's captured epoch against the current one, so the decision lands on the handler. A handler built on `EpochMessageHandler` can make it, though it has to consult the shared atomic rather than infer ownership from event order, because a stale frame can arrive before the `RECONNECTED` notification that would tell it the epoch moved. A handler built on `MessageHandler` cannot make it at all, since it receives no epoch - and every production adapter is still on that constructor; only the new turmoil test uses the epoch-aware one. `PingHandler` receives no epoch either.

So the fence is kept as the enforcing half of the mechanism the epoch identifies: checked after the read returns and again immediately before each handler invocation, including the ping handler, so a superseded reader cannot begin another dispatch whichever handler type is configured. Where the two overlap the fence is redundant, and I am happy to drop it for epoch-aware handlers if you would rather push that decision out to consumers - but as it stands the untagged paths are the ones the adapters actually use.

The read loop now checks ownership once per frame ahead of both dispatches rather than inside the legacy branch, so the epoch handler is covered by the same guard.

## The ownership-bound send had the same unbounded await

`WriterCommand::SendOnConnection` awaited `active_writer.send(message)` with no deadline, which is the wedge this PR exists to remove: a peer that accepts bytes without progressing stalls the writer task, and the reconnect handshake then queues behind it. It is now bounded by the same `WRITE_TIMEOUT_SECS` as the ordinary send.

The expiry path deliberately differs from the ordinary one. An ordinary send buffers the message for replay; an ownership-bound send must not, so on expiry it reports the failure to the caller and requests a reconnect without touching the reconnect buffer. The message moves into the send rather than being cloned, so non-replay is structural rather than a convention a later edit could quietly break.

## A cancelled write needs its own error

The expiry reports a new `SendError::WriteTimeout` rather than the existing `Timeout`. `Timeout` means the send never started; `WriteTimeout` means the write was cancelled after it began, so the peer may or may not have received the message. `should_retry_lighter_ws_error` classifies `Timeout` as retryable - correctly, with a comment noting `wait_for_active` is its only source - so reusing it would have made a delivery-undetermined transaction send retryable, which can duplicate it. Lighter now classifies `WriteTimeout` as terminal.

Adding a variant is a source-compatibility break for external code matching `SendError` exhaustively, since the enum is not `#[non_exhaustive]`. In-tree, Lighter holds the only exhaustive match and is updated; Bybit and dYdX convert generically to strings. Marking the enum `#[non_exhaustive]` would be a separate API policy decision and would not spare existing exhaustive matches anyway, so I have left it to you.

## Testing

`test_stalled_ownership_bound_send_times_out_without_replay` stalls the transport inside `send`, asserts the caller receives `WriteTimeout` rather than hanging, asserts the writer requested a reconnect, and then asserts the message was not replayed onto the replacement sink. That last assertion needs a barrier rather than a bare check for absence: the writer can accept a command from a `recv()` that was already pending when the mode flipped, so the test sends two sentinels bound to the new epoch, each awaited to its response before the next, and only then asserts the recorded sink holds exactly those two. With one sentinel the assertion could run before the writer reached its drain and pass while a regression was buffering.

Differentials, each verified to fail against the unfixed code: removing the deadline fails on the wedge with the named expectation rather than hanging; buffering the timed-out message for replay fails the non-replay assertion; and classifying `WriteTimeout` as retryable fails the Lighter retry-matrix case, which is pinned directly so that arm cannot be flipped back silently.

The three `Update` assertions that previously only proved an update had happened now assert the exact epoch the replacement sink installs, so they pin the epoch-advance semantics as well as the original claim that a writer update does not queue behind a stalled send.

cjdsellers (approved):
Hey @folknor,

I’ve gone through the current head. Both earlier concerns are resolved, and this looks good to merge.

Let’s keep `ReadSessionFence` for both handler modes. The epoch identifies the session, while the fence enforces ownership before legacy message and ping handlers can observe a stale frame. I also agree with keeping `WriteTimeout` distinct from `Timeout` and leaving `SendError` exhaustive, since the Lighter retry path should classify any future send error explicitly.

I’ll take the Python forced‑close finalization gap and the reconnect stability contract as separate follow‑ups. This PR is also a prerequisite for some additional Lighter connection nonce/epoch assurance machinery I have coming shortly (and will likely start rolling out as a standard later).

cjdsellers:
Hey @folknor,

I've followed up on the reconnect stability contract we left out here. Socket and WebSocket reconnects now keep consuming attempts and backoff until a connection stays active for 10 seconds, with Turmoil coverage for rapid-drop exhaustion and stable reset. **I left the Python forced-close path alone since those handlers are going away with the v2 cutover.**

PR #4573, opened 2026-07-25
Add PyO3 bindings for position sizers

dfjmax:
## Summary
Exposes the existing fixed-risk position sizing in `crates/risk/src/sizing.rs` to Python as `PositionSizer` and `FixedRiskSizer`, registered in the `risk` PyO3 submodule. The Rust implementation was already present but not bound to Python, leaving position sizing unreachable from Python.

## Changes
- `crates/risk/src/python/sizing.rs` — `PositionSizer` + `FixedRiskSizer` PyO3 wrappers over `calculate_fixed_risk_position_size`.
- `crates/risk/src/python/mod.rs` — register the classes in the `risk` submodule.
- `python/nautilus_trader/risk/__init__.pyi` — regenerated stub.
- `python/tests/unit/risk/test_sizing.py` — unit tests.

## Validation
- `cargo fmt`, `cargo clippy -p nautilus-risk --features python` clean (workspace `-Dwarnings`).
- `ruff` / `ruff-format` clean.
- `pytest tests/unit/risk/test_sizing.py` — 10 passed.
- Stubs regenerated via `make py-stubs-v2`.
- Full pre-commit suite passed locally.


cjdsellers:
Hi @dfjmax,

Thanks for the PR. I found two blockers at the new Python boundary:

- `calculate` validates only `risk`. A negative commission rate currently increases the returned size instead of raising, and `units=0` reaches a Rust divide-by-zero panic. Please preserve the existing numeric preconditions before calling the Rust sizing function.
- `FixedRiskSizer` is not a `PositionSizer` subclass, so it loses the inherited `update_instrument` API and the existing base-class contract.

Could you address those two points and add exact regression tests for the invalid inputs and inheritance behavior?

dfjmax:
@cjdsellers I made the changes, please let me know if something else is missing.

cjdsellers:
Hey @dfjmax

The original cases are fixed: the reported negative inputs and `units=0` now raise `ValueError`, and `FixedRiskSizer` inherits `PositionSizer.update_instrument`. All 21 sizing tests pass.

I found two remaining blockers:

- Representable positive `Decimal` inputs can still overflow the Rust arithmetic and raise `PanicException`, including large `risk` or `commission_rate` values and very small positive `exchange_rate` values. Release builds use `panic = "abort"`, so this can terminate Python instead of returning an error.
- `PositionSizer` still lacks the base `calculate` method from the existing Python API, which should preserve the polymorphic contract and raise `NotImplementedError` when called directly.

The next update needs to make arithmetic overflow return a Python exception, restore the base `calculate` contract, and add regression tests for both.

You’ve been consistently engaged across these PRs and followed through on the feedback well. One thing that would likely reduce review cycles on future v2 bindings is to start with a short contract checklist from the existing Python API: inherited methods, input invariants, panic paths, generated stubs, and exact failure probes. That kind of pass would surface issues like these earlier. This isn’t about using a particular tool, but about making the verification steps explicit before handoff. I’m happy to share the workflow I use if that would be useful.

dfjmax:
Hi @cjdsellers, thanks for the detailed review. I've pushed the changes to address the remaining points. 

I think it'd be useful if you can share your workflow so hopefully I can avoid back and forth in future PRs.

cjdsellers:
Hey @dfjmax,

Yeah, definitely. Thanks for sticking with the review. We're raising the bar for code review across the project, including contributor PRs, because improving reliability means catching more edge cases before merge. I know this makes reviews more demanding, and I appreciate you continuing to work through the feedback.

On the current update, the overflow handling and base `calculate` contract are fixed. The exact-head build passes, along with 19 Rust and 25 Python sizing tests.

**I found two remaining compatibility issues at the Python boundary:**

- `PositionSizer` and `FixedRiskSizer` accept `None`, strings, and other invalid values as instruments during construction. The existing API requires a non-null `Instrument`, so both constructors need to validate the value before storing it.
- The numeric conversion accepts strings, integers, and floats, while explicit `None` for `commission_rate`, `exchange_rate`, and `unit_batch_size` silently selects defaults. The existing API requires `Decimal` for these parameters and permits `None` only for `hard_limit`.

The next update needs to preserve those constructor and numeric type contracts, regenerate the stubs with concrete types and defaults, and add exact regression tests for the rejected values.

**For a v2 binding, I use a contract-first loop:**

1. Map the existing Python contract before editing anything: public methods, properties, inheritance, defaults, accepted types, validation rules, exceptions, and generated stubs.
2. Read the Rust implementation and nearby PyO3 bindings to identify what can be reused and where the Python boundary needs extra validation.
3. Turn the contract into a short checklist of success cases, invalid inputs, boundary values, and failure modes.
4. Implement the smallest binding that satisfies that checklist.
5. Build and run the focused tests, then probe the boundary directly with values chosen to expose omissions, swaps, overflow, and panics.
6. Review the exact diff again against the original checklist and regenerate any derived files before handoff.

**For this PR, the initial checklist would have included:**

- `FixedRiskSizer` is a `PositionSizer`.
- The inherited `instrument` and `update_instrument` APIs behave like v1.
- The base `calculate` contract remains present.
- Defaults and numeric preconditions match v1.
- Invalid and extreme values return Python exceptions rather than Rust panics.
- The generated stub matches the runtime API.

The focused proof would then be `make build-debug-v2`, the exact sizing test file, direct runtime probes for each boundary, and `make check-v2-generated-drift`.

The tools can vary. The important part is making the contract and evidence explicit before considering the work complete.

I hope that helps!

dfjmax:
@cjdsellers I have made the changes for both points. Constructors now validate the instrument and reject None/str/anything that isn't an Instrument.

The numeric params are strict Decimal now, and reject an explicit None (only hard_limit accepts None).

The one part that might need explaining is the decimal shim struct in nautilus_core::python.

For the fields commission_rate, exchange_rate and unit_batch_size need real Decimal defaults but also have to reject an explicit None. An Option with a None default doesn't work for that, since pyo3 treats "not passed" and "passed None" the same, so both just take the default.
To reject an explicit None you need a real non-None default (a Decimal), same as v1's = Decimal(0).

The problem is the stub. pyo3-stub-gen writes each signature default token straight into the .pyi as-is. If I build the default the normal way (a helper like Python::attach) the stub ends up with invalid Python and fails ruff and the drift check. And pyo3 defaults have to be non-capturing fn() expressions, so there's no expression that is both a valid `Py<PyAny>` default and valid Python.

The shim gets around it. decimal.Decimal("0") builds a real decimal.Decimal at runtime, and since the token is valid Python it comes out as a proper default in the stub, e.g. commission_rate: decimal.Decimal = decimal.Decimal(0).

I put it in nautilus_core::python next to py_to_decimal so it's not tied to sizing. Any other v2 binding that needs a Decimal default can reuse it instead of redoing the same workaround.

If you'd rather not have the shim, let me know and I'll change it, or feel free to adjust it directly if that's easier.

cjdsellers:
Hey @dfjmax,

Thanks again for sticking with what became a long review. Your point about needing concrete non-`None` defaults was right. The part I changed was where that is handled: the runtime signature now uses `Decimal::ZERO` and `Decimal::ONE`, while `python/generate_stubs.py` translates those defaults for the generated `.pyi`. Decimal extraction uses PyO3's `rust_decimal` support after a strict `decimal.Decimal` type check, so `py_to_decimal` and the reusable `decimal` shim aren't needed.

I also removed `check_non_negative_decimal` and placed the input checks in `calculate_fixed_risk_position_size` using the existing correctness functions, so Rust and Python callers share the same validation and overflow behavior.

The PR was very close, but rather than ask you for another review round or merge it and immediately rewrite those parts, I finished the remaining changes locally. This has now landed in fcb6d76d5594530f258a842a171b28de2518c691 with the bindings, generated stubs, and regression coverage. I've credited you in `RELEASES.md` and linked #4573.

I'm going to close this PR as superseded by that commit.

PR #3065, opened 2025-10-11
Add price protection support for market orders

Antifrajz:
# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [ x] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

In this PR, the concept of Market and Stop Market orders with protection is introduced. It basically converts any Market or Stop Market order, once triggered, into a Limit xor Stop Limit order using the protection points from the market configuration and the best opposite limit on the order book.

## Related Issues/PRs

Resolves #1082

## Type of change

- [ ] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [ ] Breaking change (impacts existing behavior)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

**Ensure new or changed logic is covered by tests.**

- [ ] Affected code paths are already covered by the test suite
- [x] I added/updated tests to cover new or changed logic

I added Unit tests for calucation and all affected code paths

PR #4578, opened 2026-07-26
Preserve stop-market protection price on updates that omit it

folknor:
The second half of the #4530 split, following #4549.

## `StopMarketOrder::update` clears `protection_price` on every update that omits it

`StopMarketOrder::update` in `crates/model/src/orders/stop_market.rs` assigns `self.protection_price = event.protection_price` unconditionally, so an `OrderUpdated` that carries no protection price - the ordinary shape of a quantity or trigger-price modify - overwrites a stored value with `None`. `MarketOrder::update` gates the same field on `Some` (`crates/model/src/orders/market.rs:434`), and `StopMarketOrder::update` itself gates `trigger_price` two lines above the defect, so the unconditional assignment is the odd one out. `OrderUpdated` is a patch rather than a full-state snapshot throughout the codebase: every other optional price and trigger field in every Rust order updater is gated, the Cython updaters gate theirs, and the matching engine deliberately passes `None` for prices on quantity-only updates. There is also no way to express "clear this field" as distinct from "field omitted" without an `Option<Option<Price>>` or an update mask.

The asymmetry is a leftover. `ff1be34471` (#3065) introduced `protection_price` on both order types with the unconditional assignment in both `update()` bodies, alongside a submission-time producer, `OrderMatchingEngine::update_protection_price`, which emitted an `OrderUpdated` carrying `Some`. `fa3733861e` moved protection to a fill-time calculation and deleted that producer. `3dfb7e0858` then gated the field in `MarketOrder::update` while resequencing `apply()` in both files; the stop-market assignment was not given the same treatment. This change applies the same gate there.

Because the producer is gone, no in-tree path currently emits `Some(protection_price)`: every `generate_order_updated` call in the matching engine, the order emulator, reconciliation, the backtest exchange, and each venue adapter passes `None`, and the Cython `OrderUpdated` has no such parameter. The defect is therefore dormant for a run driven only by current in-tree producers, and reachable through replay of event streams recorded before `fa3733861e`, the pyo3 `OrderUpdated` constructor, serde or Cap'n Proto reconstruction, external adapters, and any future producer.

The stored field is not the fill-protection input, so nothing about fill safety changes either way: `fill_market_order` computes its boundary locally from `price_protection_points` and the current bid/ask, and `protection_price_calculate` reads only the order's type and side. What the field does feed is `Order::price()` and `Order::has_price()` for stop-market orders. `AccountManager` prefers `order.price()` over `order.trigger_price()` when computing locked balance and initial margin (`crates/portfolio/src/manager.rs`), so clearing it silently moves a stop-market order's margin basis to the trigger price; `should_handle_own_book_order` and `Cache::update_own_order_book` return early once `has_price()` is false, leaving an already-added entry stale and skipping its delete on close; and `OrderStatusReport::is_order_updated` gates its price comparison on `has_price()`.

## Testing

`test_stop_market_order_update_preserves_protection_price_when_omitted` establishes a protection price through one `OrderUpdated`, applies a second carrying `protection_price: None` together with a different quantity and trigger price, and asserts that the quantity and trigger changes landed while `price()` and `has_price()` still report the original protection price. The changed quantity and trigger are what make the assertion non-vacuous: a rejected event would fail at `unwrap()` and a no-op would fail those two assertions, so the preservation assertion can only pass by exercising the gate. A third update carrying `Some` then pins that an explicit value still overwrites. The preservation assertions were verified to FAIL against the unfixed assignment, and only that test fails - the existing `test_stop_market_order_protection_price_update` continues to pass, keeping standalone coverage of the set path.

### DIFF
--- a/crates/model/src/orders/stop_market.rs
+++ b/crates/model/src/orders/stop_market.rs
@@ -494,7 +494,9 @@ impl Order for StopMarketOrder {
             self.trigger_price = trigger_price;
         }
 
-        self.protection_price = event.protection_price;
+        if let Some(protection_price) = event.protection_price {
+            self.protection_price = Some(protection_price);
+        }
         self.quantity = event.quantity;
         self.leaves_qty = self.quantity.saturating_sub(self.filled_qty);
     }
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. The `StopMarketOrder::update` change matches the existing `MarketOrder` semantics, and the regression test covers both omitted and explicit `protection_price` updates. I also checked the synthetic merge against current `develop`; the focused test passes there too.

This looks good to merge 👌

PR #4580, opened 2026-07-26
Assign backtest matching engine raw IDs from a monotonic counter

folknor:
Re-registering an instrument on a simulated venue currently hands the replacement matching engine a `raw_id` that another live engine already holds.

## Instrument re-registration mints a colliding matching-engine `raw_id`

`SimulatedExchange::add_instrument` in `crates/backtest/src/exchange.rs` inserts into `self.instruments` and only then derives `let raw_id = u32::try_from(self.instruments.len())`. Because the insert precedes the length read, the value is the count of distinct instruments registered so far, which is not monotonic across a re-registration: adding A yields 1, adding B yields 2, and re-adding A replaces the map entry without changing the length, so A's replacement engine is constructed with 2 - the ID B is already using.

`raw_id` is an identity discriminator rather than a label. `IdsGenerator` in `crates/execution/src/matching_engine/ids_generator.rs` formats venue order IDs as `{venue}-{raw_id}-{order_count}` and position IDs as `{venue}-{raw_id}-{position_count}`, and feeds `raw_id` into `fnv1a_trade_id_hash`. Two live matching engines on one venue sharing a `raw_id` therefore mint colliding venue order IDs and, under a hedging OMS, colliding position IDs; trade IDs collide when both engines fill at the same timestamp with the same fill ordinal. Venue order and position IDs are unaffected when `use_random_ids` is enabled, but the trade-ID hash is always deterministic.

Re-registration cannot simply be rejected: it is supported and covered by `test_instrument_update_cancels_previous_expiry_timer`, which calls `add_instrument` twice for one ID with a changed expiration, and `BacktestEngine::add_instrument` reads the previous expiration specifically to reschedule that timer.

The fix allocates from a lifetime-monotonic `last_raw_id` counter on the exchange, incremented for every matching engine constructed, including a replacement. Reproducing the current values on the ordinary append-only path matters, since `raw_id` appears verbatim in generated ID strings: the counter starts at zero and allocates with `checked_add(1)`, so first-registered instruments still receive 1, 2, 3 and no existing backtest's IDs shift. Only the re-registration case differs, where the replacement now receives a fresh number instead of a live one. Reusing the retired engine's `raw_id` would not be sufficient - the replacement engine starts its order and position sequences at zero, so it would regenerate IDs the retired engine had already emitted.

The allocation also moves ahead of the instrument insert, so the exhaustion error at `u32::MAX` now leaves both `instruments` and `matching_engines` unmodified. Previously the instrument was inserted before the checked conversion, and a failure left the map mutated with no corresponding matching engine.

## The same defect exists in the Cython engine

`nautilus_trader/backtest/engine.pyx` performs the same `self.instruments[instrument.id] = instrument` followed by `raw_id=len(self.instruments)` in its own `add_instrument`, so both engines are wrong in the same way and this is not a Rust-side parity gap. This change is Rust-only; I have not touched the Cython path.

## Whether re-registration should preserve matching state

Separate from the collision, re-registration silently discards the previous matching engine along with its resting orders, book state and ID sequences. `OrderMatchingEngine::update_instrument` already exists and is built for exactly this - it updates the definition, resets only precision-dependent state, and drops incompatible orders while preserving the engine - and the Cython engine keeps `update_instrument` as a separate method reached from the instrument-data path. Routing an existing instrument ID through it would change whether resting orders survive a re-registration, whether book state is retained when precision is unchanged, and whether ID sequences continue, so it is a behavioural decision rather than a correctness one and is left out of this change. Happy to follow up with it, or with documentation of the destructive behaviour as intended, whichever you prefer.

## Testing

`test_readded_instrument_does_not_collide_generated_fill_ids` pins the observable consequence rather than the counter value. It registers A and B, fills A once, re-adds A, then fills the replacement A and B at one shared timestamp, and asserts that all three fills carry distinct venue order IDs and trade IDs and that the two live engines produce distinct position IDs under a hedging OMS. The pre-replacement fill is what distinguishes a monotonic counter from a reused retired ID. Both assertions were verified to FAIL against the corresponding unfixed implementations: restoring the insert-before-length derivation fails on the A-versus-B venue order ID, and modelling a reused retired `raw_id` fails on the pre-versus-post-replacement venue order ID.

`test_append_only_matching_engine_raw_ids_start_at_one_and_increment` pins that ordinary registration still yields 1 and 2, so a future change to the allocator cannot silently renumber existing backtests. `test_add_instrument_raw_id_overflow_does_not_mutate_maps` drives the counter to `u32::MAX` and asserts the registration is rejected with both maps and the counter unchanged.

All three tests are deterministic and use no wall-clock waits.

### DIFF
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -136,6 +136,7 @@ pub struct SimulatedExchange {
     latency_model: Option<Box<dyn LatencyModel>>,
     instruments: AHashMap<InstrumentId, InstrumentAny>,
     matching_engines: IndexMap<InstrumentId, OrderMatchingEngine>,
+    last_raw_id: u32,
     settlement_prices: AHashMap<InstrumentId, Price>,
     pending_funding_rates: AHashMap<InstrumentId, FundingRateUpdate>,
     funding_settled_through: AHashMap<InstrumentId, UnixNanos>,
@@ -221,6 +222,7 @@ impl SimulatedExchange {
             latency_model: config.latency_model,
             instruments: AHashMap::new(),
             matching_engines: IndexMap::new(),
+            last_raw_id: 0,
             settlement_prices: config.settlement_prices,
             pending_funding_rates: AHashMap::new(),
             funding_settled_through: AHashMap::new(),
@@ -365,7 +367,9 @@ impl SimulatedExchange {
     ///
     /// # Errors
     ///
-    /// Returns an error if the exchange account type is `Cash` and the instrument is a `CryptoPerpetual` or `CryptoFuture`.
+    /// Returns an error if:
+    /// - The exchange account type is `Cash` and the instrument is a `CryptoPerpetual` or `CryptoFuture`.
+    /// - The matching engine raw ID is exhausted.
     ///
     /// # Panics
     ///
@@ -386,8 +390,6 @@ impl SimulatedExchange {
         {
             anyhow::bail!("Cash account cannot trade futures or perpetuals")
         }
-
-        self.instruments.insert(instrument.id(), instrument.clone());
 
         let price_protection = if self.price_protection_points == 0 {
             None
@@ -412,10 +414,13 @@ impl SimulatedExchange {
             .maybe_price_protection_points(price_protection)
             .build();
         let instrument_id = instrument.id();
-        let raw_id = u32::try_from(self.instruments.len())
-            .map_err(|e| anyhow::anyhow!("number of instruments exceeds u32::MAX: {e}"))?;
+        let raw_id = self
+            .last_raw_id
+            .checked_add(1)
+            .ok_or_else(|| anyhow::anyhow!("matching engine raw ID exhausted at u32::MAX"))?;
+        self.last_raw_id = raw_id;
         let matching_engine = OrderMatchingEngine::new(
-            instrument,
+            instrument.clone(),
             raw_id,
             self.fill_model.clone(),
             self.fee_model.clone(),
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. This looks good, and we'll merge it shortly 🙏

folknor:
Thanks for merging. One question from the body is still open, and I would rather ask it than pick a direction myself: what should re-registering an instrument do to the existing matching engine?

This PR fixed only the `raw_id` collision. The discard is untouched - `add_instrument` still replaces the matching engine outright, so a re-registration drops the previous engine's resting orders, book state and ID sequences. `OrderMatchingEngine::update_instrument` already exists for this case: it updates the definition, resets only precision-dependent state, and drops incompatible orders while keeping the engine. Routing an existing instrument ID through it would change whether resting orders survive, whether book state is retained when precision is unchanged, and whether the order and position sequences continue - all observable to a running backtest, which is why it seemed like your call rather than a correctness fix I should fold in.

There is a parity dimension too. The Cython engine keeps `update_instrument` as a separate `cpdef` on the exchange and the matching engine, reached from the instrument-data path at `engine.pyx:1698` rather than from `add_instrument`, so making the Rust `add_instrument` update in place would diverge the two, while leaving it destructive keeps them aligned on the current shape.

To be clear about what I would be picking up: I am keeping to the Rust side. `py_add_instrument` is a pass-through, so the Rust change touches no binding surface and needs no stub regeneration, but I am not proposing to touch `engine.pyx`, the pyo3 wrappers or any generated interface. If you decide the two engines should move together, the Cython half is yours - I would do the Rust side only, or leave the whole thing alone, whichever fits better.

So, either: route re-registration through `update_instrument` on the Rust side and accept the divergence until the Cython engine follows, or document the destructive behaviour as intended so the next reader does not mistake it for an oversight. I am happy with either, including leaving it exactly as it is. Say which you prefer.

cjdsellers:
Hey @folknor,

Apologies, I missed this question in the PR body. Good call leaving it out of the change.

Let's leave the current behavior as‑is for now. I'll ticket this separately to define the instrument re‑registration contract for v2 before changing anything. Legacy v1 will remain out of scope for this. Nothing further needed from you on this PR.

PR #4583, opened 2026-07-27
Consume response handlers after data response dispatch

xxxxxx-oss:
# Pull Request

## Summary

`send_response` currently clones correlation handlers and leaves them registered after a completed data response, so duplicate or late responses can invoke callbacks again and completed handlers accumulate without bound. This change atomically takes the handler before dispatch while preserving bus taps and response-count semantics, with coverage for every typed historical response plus join and reset behavior.

## Related issues/PRs

- Related to #3038
- Related to #3310
- No issue was filed for this focused v2 follow-up.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

N/A. The public API is unchanged; correlation handlers now follow the request/response completion lifecycle and are consumed on the first response.

## Documentation

- [ ] For PyO3 binding or wrapped Rust doc changes, I ran `make py-stubs-v2` and committed the generated output

No documentation files or Python stubs changed.

## Testing

**Ensure new or changed logic is covered by tests.** Check at least one:

Validated with:

- `make format`
- `cargo test -p nautilus-common response_handler`
- `cargo test -p nautilus-common send_response`
- `cargo test -p nautilus-data --test engine` (256 tests)
- Targeted `cargo clippy` runs for `nautilus-common` and `nautilus-data`
- `SKIP=uv-lock make pre-commit`
- `git diff --check`

The response tests cover all 11 `DataResponse` variants, duplicate delivery, bus taps and response counts, same-ID re-registration, join leg and parent completion, single- and multi-request joins, and late responses after reset. The `uv-lock` hook was excluded because it rewrites pre-existing root `bitarray` exclusion metadata on current `develop`; that unrelated generated change is not part of this PR.

### DIFF
--- a/crates/common/src/msgbus/api.rs
+++ b/crates/common/src/msgbus/api.rs
@@ -1371,7 +1371,7 @@ pub fn send_response(correlation_id: &UUID4, message: &DataResponse) {
     let handler = {
         let bus = get_message_bus();
         let mut bus = bus.borrow_mut();
-        let handler = bus.get_response_handler(correlation_id).cloned();
+        let handler = bus.take_response_handler(correlation_id);
         bus.increment_res_count();
         handler
     };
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
Hi @xxxxxx-oss,

Thank you for the PR. I reviewed the request and response paths, and taking the handler before callback invocation is the right ordering. It matches the v1 and PyO3 message bus behavior, prevents completed handlers from accumulating, and preserves re-registering the same `correlation_id` from a reentrant callback. The added tests cover the important boundaries across response variants, taps and counters, joins, reset behavior, late delivery, and same-ID re-registration.

I also independently found the same defect while auditing this path and arrived at an almost identical `take_response_handler` fix. That provides further corroboration of the root cause and ordering here. This PR has broader coverage, so I’d rather merge it than duplicate the work.

There are two follow-ups that I don't think should block this change. Requests that fail, abort, or time out before producing a response still need an explicit cleanup path, and duplicate or late responses now reach the existing `ERROR` log while v1 logs that case at debug. I’ll track and address those separately rather than expand this focused fix.

This looks good to merge from my side. Thanks again for the careful work here 👍

Issue #4398, opened 2026-07-06
TimeBarAggregator anchors DAY/WEEK/MONTH bars to UTC midnight plus a fixed time_bars_origin_offset (pd.Timedelta/pd.DateOffset)


anteqkois:
# Feature Request

## Problem Statement

TimeBarAggregator anchors DAY/WEEK/MONTH bars to UTC midnight plus a fixed time_bars_origin_offset (pd.Timedelta/pd.DateOffset). find_closest_smaller_time computes day_start = now.floor("d") (UTC) + daily_time_origin, with no localization to a session timezone.
This cannot represent a trading session that begins/ends at a fixed local time in a DST-observing timezone. For example, most equity-index and FX CFDs use a "trading day" that closes at 17:00 America/New_York — which is 21:00 UTC during EDT but 22:00 UTC during EST. A single fixed offset is correct for only half the year; across a DST transition the daily boundary drifts by one hour, so internally-aggregated daily/weekly bars split or merge the session incorrectly (wrong O/H/L/C, wrong daily returns). The same applies to weekly/monthly bars whose start rides on that origin.

### Proposed Solution

Allow the time-bar origin to be timezone-aware: accept an IANA timezone (or a tz-aware anchor) so the DAY/WEEK/MONTH boundary is computed in that local calendar (DST-aware), rather than UTC floor + fixed offset. Concretely, find_closest_smaller_time (and the WEEK/MONTH/YEAR branches of get_start_time) would floor/step in the session-local calendar and convert back to UTC, so the boundary lands at the correct wall-clock instant on both sides of a DST change.

### Example Usage

```python
from nautilus_trader.data.aggregation import TimeBarAggregator

# Daily bars whose session closes at 17:00 America/New_York, DST-aware:
aggregator = TimeBarAggregator(
    instrument=instrument,
    bar_type=BarType.from_str("SPX.SIM-1-DAY-LAST-INTERNAL"),
    handler=handler,
    clock=clock,
    time_bars_origin_offset=pd.Timedelta(hours=17),  # 17:00 local
    time_bars_origin_tz="America/New_York",           # NEW: interpret the offset in this tz
    # boundary => 17:00 NY every day: 21:00 UTC (EDT) / 22:00 UTC (EST), no manual DST handling
)
```

### Alternatives Considered

[...]

PR #4584, opened 2026-07-27
Compute FX session boundaries with calendar-day arithmetic

folknor:
FX session boundaries in `crates/trading/src/sessions.rs` are stepped with `chrono::Duration`, which measures absolute time rather than calendar days, so any boundary whose computation spans a daylight-saving transition lands an hour away from its configured local wall clock.

## Session boundaries stepped in absolute time

`fx_next_start`, `fx_prev_start`, `fx_next_end` and `fx_prev_end` each built a zone-aware anchor for the session time on the current local date with `Tz::with_ymd_and_hms`, advanced or retreated it by `Duration::days(1)`, and then applied the Monday-to-Friday weekend skip with a second `Duration::days` step. `Duration::days(1)` is 86,400 absolute seconds, not one calendar day, so a step that spans a UTC offset transition shifts the local wall-clock time by the offset delta. Every zone here transitions on a Sunday by rule, which is precisely the day the Friday-to-Monday skip crosses, so the drift is reachable from an ordinary Friday or Monday query rather than only from a timestamp inside the transition weekend.

Concretely, `fx_next_start(ForexSession::London, ..)` for Friday 2024-03-29 18:00 GMT returned Monday 2024-04-01 09:00 BST instead of 08:00: the first step lands on Saturday 08:00 GMT, the weekend skip adds a further 48 absolute hours, and BST began at 01:00 UTC on the Sunday in between. In the other direction `fx_prev_start` for Monday 2024-04-01 06:00 BST returned Friday 07:00 GMT instead of 08:00. The end functions drift identically. Only the wall-clock hour is affected, not the selected weekday: a one-hour drift cannot cross local midnight for boundaries that sit between 07:00 and 18:00 local.

The four functions now delegate to two private helpers that perform all arithmetic on the local `NaiveDate` - one calendar-day step, then the weekend rule applied to that date - and resolve the final business date and session time in the session timezone exactly once. No offset-bearing intermediate value reaches the weekend rule, and a Sunday boundary that is about to be skipped is never resolved. The `LocalResult` contract is unchanged: the previous code panicked through `LocalResult::unwrap()` for a nonexistent or ambiguous local time, and `.single().expect(..)` preserves that behavior exactly, differing only in the panic message.

The tz-unaware daily bar origin reported in #4398 is a separate instance of this class in `TimeBarAggregator` and is not addressed here.

## Testing

Four parameterized tests - `test_fx_next_start_across_dst_weekend`, `test_fx_prev_start_across_dst_weekend`, `test_fx_next_end_across_dst_weekend` and `test_fx_prev_end_across_dst_weekend` - cover six real transition weekends: Sydney 2024-04-05 to 04-08 and 2024-10-04 to 10-07, London 2024-03-29 to 04-01 and 2024-10-25 to 10-28, New York 2024-03-08 to 03-11 and 2024-11-01 to 11-04. Each case builds both the input and the expected value as a known local zoned datetime and asserts a single exact UTC equality, which pins the date, the wall-clock time and the offset in one assertion rather than only the instant.

Two further cases cover the weekend arms the Friday and Monday fixtures never reach: a Saturday input that advances to Sunday before the weekend rule runs, and a Sunday input that retreats to Saturday. Neither arm was exercised by any test before this change, so a transposition between them would have passed the whole suite.

All 26 cases were verified to fail against the previous implementation, each by exactly 3,600 seconds. The tests are deterministic and free of wall-clock dependence: every input is a fixed historical date, with no sleeps and no reads of the current time. The existing epoch-based and 2020-dated cases are unchanged and continue to pass, including the Tokyo cases, whose zone observes no daylight saving and therefore acts as a control.

### DIFF
--- a/crates/trading/src/sessions.rs
+++ b/crates/trading/src/sessions.rs
@@ -24,7 +24,7 @@
 //! - London Session    0800-1600 (Europe / London)
 //! - New York Session  0800-1700 (America / New York)
 
-use chrono::{DateTime, Datelike, Duration, NaiveTime, TimeZone, Timelike, Utc};
+use chrono::{DateTime, Datelike, Days, NaiveTime, TimeZone, Utc, Weekday};
 use chrono_tz::{America::New_York, Asia::Tokyo, Australia::Sydney, Europe::London, Tz};
 use strum::{Display, EnumIter, EnumString, FromRepr};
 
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. This looks good to merge from my side 🙏. Just some small function ordering tidy up I'll do locally post-merge.

PR #4587, opened 2026-07-27
Fail fast when infrastructure tests have no Redis or PostgreSQL

folknor:
The `nautilus-infrastructure` tests that need a live Redis or PostgreSQL server give no indication that they do, and with no server listening they block rather than failing, so a developer running the suite without those services sees an apparent hang.

## Absent services are retried, not reported

87 tests across `crates/infrastructure` require a running service: 11 in the `serial_tests` module of `src/redis/msgbus.rs` and 76 across the four integration binaries under `tests/`. Nothing distinguishes them from ordinary tests - no marker, no feature gate, no probe - and connection refusal reaches a retry loop rather than the test.

The two paths block for different reasons. On the Redis side `create_redis_connection` builds a `redis::aio::ConnectionManagerConfig` carrying the configured retry count and exponential backoff, and a refused connection is a retryable condition for that manager, so an absent server is indistinguishable from a slow one. On the PostgreSQL side `connect_pg` is `PgPool::connect_with`, where SQLx applies a 30 second default pool acquire timeout and retries connection-refused with backoff. Either way the await does not return promptly: a single Redis test measured here was still blocked when killed at 20 seconds, and the full set at that rate is roughly half an hour of wall time before a developer learns anything.

CI never encounters this, because the Linux job provisions healthchecked Redis and PostgreSQL service containers before the suite runs. It is purely a local-developer failure mode, and an expensive one to diagnose from a stalled terminal.

The tests now use test-only connection settings that fail fast: the Redis configurations take zero retries and a one second connection timeout, and the PostgreSQL helpers are wrapped in a two second timeout, each surfacing an explicit message naming the service that is required. Production connection behavior is deliberately untouched - `create_redis_connection`'s manager settings and the SQLx pool defaults are what a real deployment wants, and nothing outside the test paths changes.

## Why not `#[ignore]`

The obvious alternative, marking these with `#[ignore = "requires a running Redis"]`, would silently remove them from CI. `make cargo-test` runs `cargo nextest run --workspace --lib --tests`, and neither it, the workflows, `.config/nextest.toml`, nor the CI scripts pass `--run-ignored`; nextest excludes ignored tests by default. The attribute would therefore trade roughly 85 tests of real coverage for local convenience. Skipping at runtime was rejected for a related reason: a Rust test that returns early records a pass, which manufactures green rather than reporting the truth. Failing quickly with a specific message keeps CI coverage identical and still tells a developer exactly what is missing.

## Testing

No tests are added or removed, and no assertion changes; this alters only how the existing tests obtain their connections. The two `#[ignore]`d tests awaiting PostgreSQL schema work keep their attributes and reason strings.

The differential was measured directly with no services running. `serial_tests::test_scan_keys_empty_database` previously had to be killed at 20 seconds, blocked in a futex wait; it now fails in 0.14 seconds with `A running Redis service is required for this test: Connection refused (os error 111)`. A representative PostgreSQL test fails at its two second boundary with the corresponding message. With both services running, the affected tests execute and pass exactly as before, verified locally against containerised Redis and PostgreSQL.

Because no test gains `#[ignore]`, the set nextest collects is unchanged, so a green CI run on this branch is itself the check that the services-present path is unaffected.

### DIFF
--- a/crates/infrastructure/src/redis/msgbus.rs
+++ b/crates/infrastructure/src/redis/msgbus.rs
@@ -1299,10 +1299,14 @@ mod serial_tests {
 
     #[fixture]
     async fn redis_connection() -> ConnectionManager {
-        let config = RedisMessageBusConfig::default();
+        let config = RedisMessageBusConfig {
+            connection_timeout: 1,
+            number_of_retries: 0,
+            ..Default::default()
+        };
         create_redis_connection(MSGBUS_STREAM, &config)
             .await
-            .unwrap()
+            .expect("A running Redis service is required for this test")
     }
 
     #[rstest]
--- a/crates/infrastructure/tests/test_cache_database_postgres.rs
+++ b/crates/infrastructure/tests/test_cache_database_postgres.rs
@@ -28,7 +28,7 @@ mod serial_tests {
     };
     use nautilus_core::{Params, UnixNanos};
     use nautilus_infrastructure::sql::{
-        cache::get_pg_cache_database,
+        cache::{PostgresCacheDatabase, get_pg_cache_database},
         pg::{connect_pg, get_postgres_connect_options, init_postgres},
         queries::DatabaseQueries,
     };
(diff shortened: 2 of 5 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR, this looks good to me. Keeping production connection behavior unchanged while making the tests fail quickly and clearly when Redis or PostgreSQL is missing is the right approach. I confirmed both missing‑service paths against the current `develop` merge, and this is safe to merge 👍

PR #4592, opened 2026-07-28
Reconstruct DataType cache fields on deserialization

folknor:
`DataType` derives its `Deserialize` over two fields that its own constructor computes, so a deserialized value can disagree with itself.

## The cached hash is treated as authoritative serialized state

`DataType` carries `type_name`, `metadata` and `identifier`, plus two caches derived from them by `DataType::new`: `topic`, built by `params_to_topic_suffix`, and `hash`, a `DefaultHasher` digest of that topic. Identity is split across the two - `PartialEq` and `Ord` compare `topic`, while `Hash` feeds only the cached `hash` field - and the derived `Deserialize` accepts whatever pair the payload carries without reconstructing either. Two values that are equal under `PartialEq` can therefore hash differently, which breaks the `Eq`/`Hash` contract every hash container depends on.

This does not require a forged payload. `DefaultHasher`'s algorithm is explicitly not guaranteed stable across Rust releases, so a version skew is enough: a value serialized by one build carries `hash` computed by one algorithm, a peer built against a different toolchain deserializes it verbatim and computes a different digest for the same topic locally, and the two compare equal while hashing apart. The consequences land in `DataClientAdapter::subscriptions_custom`, an `AHashSet<DataType>`, where an entry inserted from a deserialized command cannot be found by an equal locally constructed key, so the unsubscribe misses and the subscription leaks; and in the switchboard's `DataType`-keyed `custom_topics` map, where every lookup misses and re-enters topic construction.

The fix is a custom `Deserialize` that treats the cache fields as what they are. A payload's `topic` is authoritative and preserved when present, `hash` is always recomputed locally from that topic and the incoming value ignored, and a payload that omits the cache fields is constructed through `DataType::new` as before. `Serialize` is unchanged and still emits both fields, so existing readers parse the output as they always did; removing them from the wire would be a separate versioned change. The `PartialEq`, `Ord` and `Hash` field sets are untouched - this changes only whether the caches are trusted, not what identity means.

Because the cache fields are optional on the way in while `Serialize` emits them concretely, the implementation assumes a self-describing format. Every encoding of `DataType` in this repository is one: `serde_json`, and the `rmp_serde::to_vec_named` used on the event-store capture path.

The exposure is the derived impl, reached by the `DataType`-carrying command types such as `SubscribeCustomData`; the dedicated reconstruction paths, `from_persistence_json` and the custom-data registry parser, already rebuilt through `DataType::new` and are unchanged.

## Testing

`test_data_type_deserialization_recomputes_hash_from_topic` deserializes a payload whose `hash` has been inverted against its `topic` and asserts the result carries the digest `DataType::new` would compute. `test_data_type_serde_roundtrip_preserves_fields_and_repairs_hash` pins that `type_name`, `metadata`, `identifier` and `topic` survive a round trip while the poisoned digest is repaired, and `test_data_type_serialized_cache_fields_remain_wire_compatible` decodes the re-serialized value through a struct with the original five concrete fields, so a change that dropped `topic` or `hash` from the wire would fail.

`test_data_type_deserialization_without_cache_fields_uses_constructor` covers a payload omitting both caches, and `test_data_type_deserialization_preserves_topic_without_hash` and `test_data_type_deserialization_ignores_hash_without_topic` cover each present without the other - a regression that coupled topic preservation to the presence of a hash would otherwise pass. `test_data_type_persistence_result_hashes_like_equal_deserialized_value` and the registry test assert that all three construction routes - constructed, deserialized, persistence JSON, registry - agree on `type_name`, `metadata`, `identifier` and `topic` as well as on equality and hash, since equality alone observes only the topic and would not notice a route that dropped the identifier.

All of these were verified to fail against the unfixed implementation. Mutating the impl to trust the incoming `hash` fails four of them, including the cross-route consistency test; coupling topic preservation to the presence of a hash fails the independent-omission test and only that one.


cjdsellers:
Hey @folknor,

Thanks for the PR. The cache repair looks correct for JSON and MessagePack, but the new shadow struct breaks `DataType`'s existing Serde round trip for non-self-describing formats. `Serialize` still writes `topic` and `hash` as concrete `String` and `u64` fields, while `Deserialize` now reads those sequence positions as `Option<String>` and `Option<u64>`. With bincode, the first topic-length byte is read as an option tag and deserialization fails.

Could you preserve the original concrete field types when decoding sequence-based formats, and only make `topic` and `hash` optional for map-based inputs? Separate `visit_seq` and `visit_map` paths should retain the existing public round trip while still rebuilding the caches for JSON and named MessagePack.

folknor:
Hey @cjdsellers

You are right, and the asymmetry was mine: `Serialize` emits `topic` and `hash` concretely while the shadow struct read those positions as `Option`, so any format that keys by position and encodes `Option` with a discriminant mis-frames the payload. Fixed as you asked - `deserialize_struct` with a `Visitor` implementing both paths.

This split was considered before I filed and set aside deliberately: it trades a five-field shadow struct for a hand-written visitor, and I judged the extra surface disproportionate for a format no encoding in this repository uses. That was the wrong call - the round trip is public API regardless of what the tree itself encodes, and the cost turns out to be about a hundred lines that need no maintenance once written.

`visit_seq` reads the five fields in declaration order at their original concrete types (`String`, `Option<Params>`, `String`, `u64`, `Option<String>`), so it consumes exactly what `Serialize` writes; every position is required, and a short sequence is an `invalid_length` error as the derive made it. `visit_map` keeps `topic` and `hash` optional, treats missing and null alike as absent, ignores unknown keys, and rejects duplicate known keys. Both paths share one finalizer: `topic` is authoritative when supplied, the incoming `hash` is parsed and discarded, and the hash is always recomputed locally.

The regression test is `test_data_type_bincode_roundtrip_preserves_fields_and_repairs_hash`: it serializes a `DataType` whose `hash` field has been corrupted, decodes it, and asserts `type_name`, `metadata`, `topic`, `identifier` and the recomputed hash separately - `PartialEq` compares `topic` alone, so an equality assertion would prove almost nothing here. Against the previous shape that test fails with `InvalidTagEncoding(11)`, and the `11` is the mechanism in one number: `"ExampleType"` is eleven characters, so the decoder was consuming the first byte of the string's eight-byte length prefix as an option tag. `test_data_type_deserialization_rejects_duplicate_map_key_after_null` covers the map path, including a null first occurrence of each nullable field.

One intentional narrowing. Sequence inputs now require concrete values at the `topic` and `hash` positions, so a noncanonical array carrying `null` at either position is rejected where the previous shape accepted it. That is inherent to restoring the concrete contract. Nothing this repository emits is affected: `serde_json` and `rmp_serde::to_vec_named` both produce maps, and the map path is unchanged.

A limitation the sequence path does not remove and never could. `Params` is an `IndexMap<String, serde_json::Value>`, and `Value`'s `Deserialize` requires `deserialize_any`, which no non-self-describing format can serve - so a `DataType` carrying non-empty metadata still cannot be decoded from bincode. An empty `Params` round-trips, since no `Value` is visited. That is a pre-existing property of `Params` rather than anything this change introduces, which is why the test uses absent metadata to isolate the framing, and it means this restores the round trip the derive had rather than making `DataType` fully bincode-capable.

`bincode` is added as a `nautilus-model` dev-dependency for that test. It was already in `Cargo.lock` transitively, so the lockfile gains one line under `nautilus-model` and no new package or version.

I squashed this into the original commit and rebased onto current develop, which the dependency update at `2c03567711` made necessary for the lockfile.

folknor:
I'll redo this. Without `bincode` this time.

folknor:
Hey @cjdsellers

Correcting something I introduced in the last round. That revision added `bincode` as a `nautilus-model` dev-dependency to exercise the sequence path, and I justified it on the grounds that the crate was already in `Cargo.lock`. That was the wrong test to apply: `bincode` is unmaintained (RUSTSEC-2025-0141), and #4330 removed it from this workspace for exactly that reason, dropping the direct dependency and the `[workspace.dependencies]` entry and replacing it with a positional codec in `event_store`. Taking a direct dependency on it again one PR later, for a test, contradicts the argument that PR made. The lockfile point was true and beside the point.

The sequence-path regression therefore moves to `crates/event_store/src/codec.rs`, where it uses that crate's own codec through `encode_to_vec` / `decode_from_slice`. `nautilus-event-store` already depends on `nautilus-model`, so the test needs no new dependency edge and introduces no cycle, and `nautilus-model` gains nothing. `Cargo.toml` and `Cargo.lock` revert to their develop contents and leave this PR's diff entirely.

The differential survives intact, with the error renamed. The codec writes strings with a fixed eight-byte little-endian length prefix and encodes `Option` with a one-byte discriminant, so an `Option`-typed sequence position reads the first length byte as an option tag: against the pre-fix shape the test fails `InvalidOption(11)`, on the same eleven characters of `"ExampleType"` that produced `InvalidTagEncoding(11)` under bincode.

It is also a stronger test than the one it replaces. The bincode version corrupted the private `hash` field in memory before serializing, which was only possible from inside `nautilus-model`. `data_type_roundtrip_preserves_fields_and_repairs_hash` instead patches the poisoned hash into the encoded bytes at a computed offset, which is the case that actually matters here - a payload arriving from a build whose `DefaultHasher` disagrees. I verified it fails both ways it should: `InvalidOption(11)` if the sequence positions go back to `Option`, and on the hash assertion if the visitor is changed to carry the decoded hash through instead of recomputing it.

The `Params` / `deserialize_any` limitation is unchanged, and the comment recording it moved with the test: absent metadata isolates the framing, because `serde_json::Value` cannot be decoded by any non-self-describing format.

The dependency note in the description above is superseded by this comment; I have left the body as filed rather than rewriting it after the fact.

One thing I noticed and deliberately did not touch, since it predates this PR: the RUSTSEC-2025-0141 entry in `deny.toml` (and the matching line in `.cargo/audit.toml`) gives its reason as "dev-only via madsim", but `hypersync-client` also depends on `bincode` unconditionally, reachable through an optional adapter feature. `cargo deny` passes either way, since the ignore keys on the advisory id, but the stated provenance is not accurate. Happy to send a one-line fix for that separately if you want it.

cjdsellers:
Hey @folknor,

The separate sequence and map paths restore the existing positional Serde contract while preserving cache reconstruction for JSON and named MessagePack. The event-store codec regression test also covers the framing failure without reintroducing a direct `bincode` dependency. This looks good to me.

Thanks for the heads-up about the stale `RUSTSEC-2025-0141` provenance. I've corrected the corresponding supply-chain entries in `.cargo/audit.toml` and `deny.toml`, and I'll push that separately shortly.

cjdsellers (approved):
Hey @folknor,

I've completed a final pass against the resynced head, and all required CI is now green. The positional Serde compatibility issue is resolved, so this is good to merge from my side 👍.

PR #4595, opened 2026-07-28
Enforce venue order ID ownership in the cache

folknor:
A follow-up to the purge and index integrity work in #4569, covering the ownership half of the same index family.

## Venue order ID ownership was never enforced

`Cache::add_venue_order_id` maintains two maps, `index.client_order_ids` and `index.venue_order_ids`, but validated only one direction: whether the client order already had a different venue order ID. Nothing checked whether the incoming venue order ID was already owned by a *different* client order, so the reverse entry was rebound unconditionally. The first owner kept its forward mapping while the reverse entry named the claimant, and every lookup by venue order ID - order resolution on a venue report, reconciliation matching - resolved to the wrong order.

`refresh_order` had the complementary defect from the other side. Its guard tested `venue_order_ids.contains_key(..)`, that is existence rather than ownership, so when the key existed under another client order it did nothing at all: the order was left unindexed in both directions while the reverse entry continued to name the first owner.

The claim now validates both directions before mutating either map, and returns a typed `VenueOrderIdOwnershipError` naming the venue order ID, the existing owner and the claimant. On conflict neither map changes. `refresh_order` always performs the checked claim, which also repairs a missing forward entry when the reverse owner is already correct, and propagates an ownership conflict rather than logging it, so `replace_order` cannot commit a canonical snapshot that claims another order's venue order ID.

`Cache::update_order` preflights ownership only. A venue order ID owned by a different client order is rejected before the canonical order is committed, so a rejected event leaves the order in its prior state; a same-client forward mismatch remains the logged refresh inconsistency it was before this change. That distinction is load-bearing: enforcing the forward rule at the preflight too would reject a fill outright whenever its venue order ID disagreed with the acceptance, so an adapter emitting a stale or defaulted identifier on a fill would lose the fill entirely.

The two maps are deliberately not strict inverses and this change does not make them so. A cancel-replace legitimately retains the historical `V_old -> C` alias with no forward entry; the invariant enforced here is that a venue order ID has at most one owner, and that every current forward entry has a matching reverse entry. `overwrite` continues to permit the same client order changing its own venue order ID, and never permits taking one from another client.

## Generated IDs probe past a conflict rather than failing

Enforcing ownership makes the claim in `IdsGenerator::get_venue_order_id` fallible, and that path feeds the simulated matching engine, where a failure would surface in the middle of order lifecycles that assume ID allocation cannot fail. A freshly generated candidate has no external identity yet, so choosing a different one is the allocator's job rather than a failure the lifecycle must absorb. On an ownership conflict the generator now probes for the next free candidate, and only an ownership conflict is retried - every other error remains terminal. The first conflict is logged with the candidate, the existing owner and the claimant, and a count is logged when allocation succeeds, so a duplicate raw-ID namespace stays visible without producing a log entry per collision.

This changes collision policy, not the generated ID format: `{venue}-{raw_id}-{order_count}` is unchanged, and allocation on a clean cache produces exactly the same sequence as before.

Externally supplied and previously assigned venue order IDs are authoritative and are never replaced. They continue to fail fast at the cache boundary, and the execution engine no longer publishes or registers an externally materialized order whose claim was rejected.

## What this does not cover

Duplicate raw-ID namespaces become recoverable by probing rather than prevented; the error log is what makes the misconfiguration visible.

Counter exhaustion remains terminal, and the simulated matching engine unwraps it. Probing walks past a persisted prefix, so exhaustion means the counter itself is spent rather than a collision being unresolvable, but generated-ID allocation is not infallible.

`get_venue_order_id` returns an ID already present on the order or in the forward cache without revalidating its ownership; those IDs are trusted because they have already crossed the checked cache boundary.

A same-client forward mismatch stays logged rather than enforced. A fill carrying a stale venue order ID that no other order owns is applied while `index.client_order_ids` still holds the accepted one, which is the pre-existing logged inconsistency rather than cross-order corruption; if that identifier belongs to another client order, the ownership check still rejects it atomically.

`build_index` is unchanged. It rebuilds only reverse entries and overwrites duplicates in iteration order, so this change does not establish that a persisted cache is globally free of duplicate ownership, nor does `check_integrity` detect every such conflict.

Synthetic option-settlement IDs remain a best-effort legacy path outside this change: if one collides, cache ownership is preserved but that leg's accepted and filled events may be rejected, leaving settlement incomplete. That path derives both its client and venue identifiers from the same truncated UUID and registers its legs non-atomically, which needs its own change rather than being folded in here.

## Testing

The cache tests pin the invariant directly: a second client order claiming an owned venue order ID is rejected with both maps unchanged; an accepted or updated event carrying another client's venue order ID is rejected before the canonical order is modified; and a same-client cancel-replace still succeeds while retaining the historical reverse alias, which is the case that fails if an implementation "repairs" the maps into strict inverses.

`test_update_order_commits_canonical_state_when_database_update_fails` pins the preflight boundary specifically. Only a cross-order ownership conflict propagates; a backing-database failure is still logged after the canonical order is committed, so the ordering cannot regress into aborting an update with derived state already changed. A companion pair pins the narrowed semantics from both sides: a fill carrying a stale venue order ID owned by nobody is applied with the forward mismatch logged, while a fill carrying one owned by another client order is still rejected atomically.

On the allocator side, a matching engine whose next deterministic candidate is already owned probes past it and fills normally, parameterised over market and market-to-limit and asserted at the default acknowledgement configuration rather than only the non-default one.

All of these were verified to fail against the unfixed implementation. Several existing execution tests changed because they built fill events from local order objects that never received their acceptance, so `OrderFilledTestBuilder` fell back to a shared default venue order ID and multiple distinct client orders in one test claimed the same one; they now use the cached order, and two of them additionally needed distinct client order IDs that they had previously shared by default.

### DIFF
--- a/crates/backtest/tests/netting_fill_void.rs
+++ b/crates/backtest/tests/netting_fill_void.rs
@@ -237,9 +237,14 @@ fn process_filled_order(
         VenueOrderId::from(fill.venue_order_id),
     ));
 
+    let accepted_order = execution_engine
+        .cache()
+        .borrow()
+        .order_owned(&order.client_order_id())
+        .expect("accepted order");
     let instrument_any: InstrumentAny = instrument.clone().into();
     execution_engine.process(&TestOrderEventStubs::filled(
-        &order,
+        &accepted_order,
         &instrument_any,
         Some(TradeId::new(fill.trade_id)),
         Some(position_id),
--- a/crates/common/src/cache/error.rs
+++ b/crates/common/src/cache/error.rs
@@ -14,7 +14,7 @@
 // -------------------------------------------------------------------------------------------------
 
 use nautilus_model::identifiers::{
-    AccountId, ClientOrderId, InstrumentId, OrderListId, PositionId,
+    AccountId, ClientOrderId, InstrumentId, OrderListId, PositionId, VenueOrderId,
 };
 use thiserror::Error;
 use ustr::Ustr;
@@ -45,6 +45,21 @@ pub const ORDER_LIST_NOT_FOUND: &str = "order list not found in cache";
 
 /// Message used for a missing position lookup.
 pub const POSITION_NOT_FOUND: &str = "position not found in cache";
+
+/// Error returned when a venue order ID is already owned by another client order.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Error)]
+#[error(
+    "Venue order ID {venue_order_id} is already owned by client order \
+     {existing_client_order_id} and cannot be claimed by {claimant_client_order_id}"
+)]
+pub struct VenueOrderIdOwnershipError {
+    /// The venue order ID whose ownership conflicts.
+    pub venue_order_id: VenueOrderId,
+    /// The client order which already owns the venue order ID.
+    pub existing_client_order_id: ClientOrderId,
+    /// The client order attempting to claim the venue order ID.
+    pub claimant_client_order_id: ClientOrderId,
+}
 
 /// Error returned when an account cannot be resolved from a cache or store.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Error)]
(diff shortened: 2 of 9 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. The split between rejecting cross-order ownership and retaining the existing same-client forward mismatch behavior makes sense, and the cancel-replace alias handling stays intact.

I reviewed the cache, execution engine, and generated-ID paths against current `develop`, including the focused ownership and collision tests. Everything looks good to me, so I'm approving this 👍.

PR #4596, opened 2026-07-28
Register backtest venues atomically and validate instrument leverage

folknor:
A follow-up to the venue registration work in #4580, covering the setup half of the same area.

## `add_venue` publishes engine state before the step that can fail

`BacktestEngine::add_venue` performs its only propagating fallible call, `ExecutionEngine::register_client`, as the last statement before `Ok(())`. By that point it has already registered a global spread-quote endpoint for the venue, inserted the exchange into `self.venues`, registered the execution client into the exchange, and pushed that client into `self.exec_clients`. Both the endpoint registration and the venue map insert are silent replacements.

The observable case is a duplicate venue. `ExecutionEngine::register_client` rejects the already-registered client id, but by then the previous exchange has been displaced twice, and those were its only two ordinary strong owners: the spread-quote handler closure holds an `Rc` clone of the exchange, and `self.venues` holds the other. The exchange is therefore dropped, not merely shadowed. `BacktestExecutionClient` holds a `WeakCell`, so the execution engine keeps routing that venue to a client whose exchange reference no longer upgrades. `submit_order` then queues an `OrderSubmitted` event, fails the upgrade, logs, and returns `Ok(())` - the strategy observes its order as submitted while it reaches no exchange at all. A failed call destroys the working venue it collided with.

`add_venue` now rejects a duplicate venue up front and defers every publication until after `ExecutionEngine::register_client` succeeds. That call performs both of its conflict checks before mutating either of its maps, so its error path is non-mutating and reordering is sufficient; on any error the venue map, the quote endpoint, `exec_clients` and the execution engine are all left exactly as they were, and the candidate exchange is dropped. Rollback is not a viable alternative here: `register_quote_endpoint` returns `()`, the endpoint map is crate-private with no getter, and there is no quote-specific deregistration, so a displaced handler cannot be restored.

## `SimulatedVenueConfig` never validates its per-instrument leverages

`SimulatedVenueConfig::validate` checks `starting_balances`, `default_leverage` and `liquidation_trigger_ratio`, but never iterates its `leverages` map, while the sibling `BacktestVenueConfig::validate` checks every entry. The unvalidated value is not inert: `SimulatedExchange` copies the map, and account initialization calls `MarginAccount::set_leverage` for each entry, which is documented to panic on a non-positive leverage. A zero or negative entry therefore passes `build()` and panics later during initialization instead of being rejected at config time.

`validate` now applies the sibling's per-instrument rule, reporting the offending instrument under the same `leverages` field name.

## Testing

`test_add_duplicate_venue_preserves_original_exchange` holds a `Weak` reference to the first exchange rather than a strong one, so the test cannot itself keep a dropped exchange alive and mask the defect. After the failed duplicate it asserts the weak reference still upgrades, that `Rc::ptr_eq` holds between it and the live venue entry, that the venue list, `exec_clients` length and execution-engine client ids are unchanged, and that a quote sent to the venue endpoint moves the original exchange's best bid and ask. That last assertion is what proves the quote handler was preserved rather than merely that some endpoint is still registered.

`test_add_venue_execution_registration_failure_publishes_nothing` covers the other failure path, where no local venue entry exists but the execution engine already holds that client id. It registers a stub execution client for the venue, installs a sentinel quote handler, and asserts the failed `add_venue` published nothing: no venue entry, no execution client, unchanged execution-engine client ids, and a sentinel that still receives quotes. This is the test that pins the ordering itself - the duplicate test returns at the new pre-check and would stay green if publication were moved back ahead of registration.

`test_simulated_non_positive_instrument_leverage_rejected` is parameterised over zero and negative leverage, with `test_simulated_positive_instrument_leverage_accepted` guarding against blanket rejection.

All of these were verified to fail against the unfixed implementation. For the duplicate case the revert removed the new pre-check as well as the reordering, so the test could not pass merely by bailing out early; it fails there at the weak-reference upgrade, which is the dropped-exchange mechanism described above. The ordering test was verified against a variant with the pre-check left in place and publication moved back before registration.


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. `develop` moved and left a conflict in the backtest test imports, so I merged the current `develop` into your branch and retained both the state persistence and venue setup test imports.

PR #4597, opened 2026-07-28
Remove wall-clock dependency from own-book filtering

folknor:
A follow-up to the `DataType` identity work in #4592, covering another invariant in the same family: a query result that depends on when it was asked.

## Own-book filtering compared engine timestamps against the machine wall clock

`filter_orders` in `crates/model/src/orderbook/own.rs` substituted `nanos_since_unix_epoch` for a missing `ts_now`, then applied `ts_accepted + accepted_buffer_ns <= ts_now` unconditionally. The predicate was not gated on a buffer having been requested, so with `accepted_buffer_ns` absent it degraded to `ts_accepted <= wall_clock` and still consulted the system clock. Every caller that omitted `ts_now` was therefore wall-clock dependent, not only those asking for a grace period, and `bids_as_map`, `asks_as_map`, `bid_quantity` and `ask_quantity` all funnel through it.

Nothing else on this path is wall-clock driven. Backtests run on a simulated clock and `ts_accepted` values originate there, so the comparison mixed two unrelated time bases: an own-order query could include or exclude an order based on the machine clock at the moment it ran.

The tree already recorded the intended meaning in two places. The property-test reference implementation substituted `u64::MAX` for a missing `ts_now`, that is "no filtering", and the execution-engine tests' own comments describe `None, None` as "no filter". The oracle and the implementation had disagreed since the parameter was introduced.

`ts_now` is now the switch that controls acceptance-time filtering. With no time source the acceptance time is not inspected at all; with `Some(t)` the order must satisfy `ts_accepted + accepted_buffer_ns <= t`. A positive buffer with no `ts_now` is a caller error rather than a silently ignored request, matching the existing higher-level cache API which already rejects that combination. The addition is checked, so a buffer that would push eligibility past `u64::MAX` excludes the order instead of overflowing; saturating would wrongly admit it at `ts_now == u64::MAX` and wrapping would admit it near zero.

`nanos_since_unix_epoch` is no longer imported by the module.

## Grouped own-book rendering inherited the same dependency

`pprint_own_book` requests grouped quantities with no status, no buffer and no time. Under the previous semantics that still applied the wall-clock predicate, so grouped rendering could omit orders while the ungrouped branch, which walks the ladders directly, always showed them. The two branches disagreed about membership.

The new contract resolves this at the source: passing no time source performs no acceptance filtering, so grouped and ungrouped rendering now agree, and grouping changes bucketing and depth rather than which orders exist in the view. Rust `Display` never routed through this path and is unchanged.

## Testing

`test_own_order_book_missing_ts_now_skips_acceptance_filter` pins the core semantics with an order accepted at `UnixNanos::max()`: it is present with no time source and absent with an explicit `ts_now` of zero. The timestamp is deliberately `UnixNanos::max()` rather than a value derived from the current wall clock, so the assertion cannot pass by accident of when the suite runs.

`test_own_order_book_acceptance_buffer_overflow_excludes_order` covers the overflow boundary separately, and `test_own_order_book_positive_accepted_buffer_without_ts_now_panics` pins the caller error. `test_own_book_grouped_pprint_includes_future_accepted_order` asserts the future-dated order's price and quantity appear in grouped output, which protects the rendering path specifically; the map-level assertions alone would not prove rendering is membership-neutral.

Each was verified to fail against the unfixed implementation, and each fails for the mechanism it names: the first at its own membership assertion, the second on the overflow panic in the unchecked addition. These began as one test, which failed against the unfixed code only at the overflow call and so never exercised the wall-clock comparison; they were split so the differential evidence is unambiguous.

The 49 existing call sites in `crates/execution/tests/exec_engine.rs` that pass no buffer and no time are unmodified and their expected results are unchanged. Those tests use zero-based test clocks that are never advanced, so their acceptance timestamps sat far below the wall clock and the previous predicate already admitted the orders; the change removes a condition they were never asserting.

The property-test reference implementation was updated to match the new contract, replacing its `u64::MAX` approximation of "no filtering" with the same absent-time semantics.

### DIFF
--- a/crates/model/src/orderbook/book.rs
+++ b/crates/model/src/orderbook/book.rs
@@ -642,8 +642,13 @@ impl OrderBook {
     /// Maps bid prices to total public size per level, excluding own orders up to a depth limit.
     ///
     /// With `own_book`, subtracts own order sizes, filtered by `status` if provided.
-    /// Uses `accepted_buffer_ns` to include only orders accepted at least that many
-    /// nanoseconds before `now` (defaults to now).
+    /// When `now` is provided, only subtracts orders whose acceptance time plus
+    /// `accepted_buffer_ns` is at or before `now`. When `now` is `None`, acceptance-time
+    /// filtering is disabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `own_book` is `Some`, `accepted_buffer_ns` is positive, and `now` is `None`.
     #[must_use]
     pub fn bids_filtered_as_map(
         &self,
@@ -671,8 +676,13 @@ impl OrderBook {
     /// Maps ask prices to total public size per level, excluding own orders up to a depth limit.
     ///
     /// With `own_book`, subtracts own order sizes, filtered by `status` if provided.
-    /// Uses `accepted_buffer_ns` to include only orders accepted at least that many
-    /// nanoseconds before `now` (defaults to now).
+    /// When `now` is provided, only subtracts orders whose acceptance time plus
+    /// `accepted_buffer_ns` is at or before `now`. When `now` is `None`, acceptance-time
+    /// filtering is disabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `own_book` is `Some`, `accepted_buffer_ns` is positive, and `now` is `None`.
     #[must_use]
     pub fn asks_filtered_as_map(
         &self,
@@ -702,6 +712,7 @@ impl OrderBook {
     /// # Panics
     ///
     /// Panics if `self` and `own_book` have different instrument IDs.
+    /// Panics if `own_book` is `Some`, `accepted_buffer_ns` is positive, and `now` is `None`.
     ///
     /// [`Self::filtered_view_checked`] for fallible construction.
     #[must_use]
@@ -726,6 +737,7 @@ impl OrderBook {
     ///
     /// # Panics
     ///
+    /// Panics if `own_book` is `Some`, `accepted_buffer_ns` is positive, and `now` is `None`.
     /// Panics if `Price::from_decimal` or `Quantity::from_decimal` fails when
     /// reconstructing filtered levels.
     pub fn filtered_view_checked(
@@ -795,8 +807,13 @@ impl OrderBook {
     /// Groups bid quantities into price buckets, truncating to a maximum depth, excluding own orders.
     ///
     /// With `own_book`, subtracts own order sizes, filtered by `status` if provided.
-    /// Uses `accepted_buffer_ns` to include only orders accepted at least that many
-    /// nanoseconds before `now` (defaults to now).
+    /// When `now` is provided, only subtracts orders whose acceptance time plus
+    /// `accepted_buffer_ns` is at or before `now`. When `now` is `None`, acceptance-time
+    /// filtering is disabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `own_book` is `Some`, `accepted_buffer_ns` is positive, and `now` is `None`.
     #[must_use]
     pub fn group_bids_filtered(
         &self,
@@ -822,8 +839,13 @@ impl OrderBook {
     /// Groups ask quantities into price buckets, truncating to a maximum depth, excluding own orders.
     ///
     /// With `own_book`, subtracts own order sizes, filtered by `status` if provided.
-    /// Uses `accepted_buffer_ns` to include only orders accepted at least that many
-    /// nanoseconds before `now` (defaults to now).
+    /// When `now` is provided, only subtracts orders whose acceptance time plus
+    /// `accepted_buffer_ns` is at or before `now`. When `now` is `None`, acceptance-time
+    /// filtering is disabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `own_book` is `Some`, `accepted_buffer_ns` is positive, and `now` is `None`.
     #[must_use]
     pub fn group_asks_filtered(
         &self,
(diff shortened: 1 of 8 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. The explicit time-source semantics and checked overflow look good, but the new `assert!` in `filter_orders` is reachable from the public PyO3 methods when a caller supplies a positive `accepted_buffer_ns` without `ts_now`. For example, `OwnOrderBook.bids_to_dict(accepted_buffer_ns=500_000_000)` now reaches that assertion. The higher-level cache rejects the same combination with `ValueError`, but the direct v2 methods forward it unchanged.

This is more than a normal caller error in release wheels because the release profile uses `panic = "abort"`, so it can terminate the Python process. Could you make this combination fallible across the direct own-book and filtered-book Python entry points, or otherwise validate it before the assertion, and add a regression proving Python receives a normal exception?

Ideally the concept guide should also state that omitting `ts_now` disables acceptance-time filtering and that a positive buffer requires it.

folknor:
Thanks - confirmed, and it is worse. `[profile.release]` sets `panic = "abort"`, so in a release wheel this terminates the process rather than surfacing as `PanicException`. Nine methods could reach the assertion with caller-controlled arguments, not just the own-book ones: `bids_to_dict`, `asks_to_dict`, `bid_quantity` and `ask_quantity` on `OwnOrderBook`, plus `bids_filtered_to_dict`, `asks_filtered_to_dict`, `group_bids_filtered`, `group_asks_filtered` and `filtered_view` on `OrderBook`. `OwnOrderBook.pprint` also reaches `bid_quantity`/`ask_quantity` when grouping, but hardcodes `None, None` and so cannot violate the invariant.

The pushed change adds one validator next to `filter_orders` and uses it in both places. `filter_orders` consumes it and still panics, so the Rust signatures and their callers are untouched and the `# Panics` documentation stays true. Each of the nine PyO3 methods preflights it and maps failure with `to_pyvalue_err`, raising `ValueError` with the same message `Cache.own_bid_orders` and `own_ask_orders` already raise. Eight wrappers change return type from `T` to `PyResult<T>`; `filtered_view` was already fallible and only gained the preflight.

I kept the validation in the binding layer rather than making the core methods fallible. A core `Result` would propagate through the four `OwnOrderBook` methods and the `OrderBook` filtered family and force `.unwrap()` at roughly sixty call sites across `crates/execution`, `crates/model` and `crates/trading` that all pass `None, None` and can never hit the error. The hazard is specific to crossing the FFI boundary, so the check lives there.

Two consequences: Validation is unconditional, matching the cache, which checks the pair before it looks up the own book; for the filtered-book methods that makes the Python boundary stricter than Rust when `own_book` is `None`, because the assertion is unreachable in that case and the arguments are rejected as an invalid pair in their own right. And the panic text now comes from the validator, so it matches the cache wording and the existing `should_panic` test was updated - the old string appears nowhere else in the tree.

Testing: fourteen `rstest` cases driving the Python methods through `call_method`, asserting both `ValueError` and the exact message. Nine cover the entry points above; five more call the filtered-book methods with no `own_book` to pin the stricter-than-Rust behavior, which the first nine structurally cannot - they all supply an own book, so a preflight made conditional on `own_book.is_some()` would leave them green. Both sets were verified to fail against the unfixed code, and they fail differently: the nine fail with `PanicException` from the assertion, while the five fail because the call *succeeds*, which is the proof that only the boundary check produces the error on that path.

The concept guide now states that omitting `ts_now` disables acceptance-time filtering and that a positive `accepted_buffer_ns` requires it.

folknor:
@cjdsellers Worth noting that my local tooling failed on this now because of 3 pre-existing clippy errors in upstream `develop`. I might file a PR later or tomorrow to fix these if you don't:

1. `clippy::unused_async_trait_impl` - `crates/system/src/kernel.rs:747:5` - "unused async for async trait impl function with no .await statements".
2. `clippy::default_constructed_unit_structs` - `crates/adapters/derive/bin/flatten.rs:128:25` - "use of default to create a unit struct".
3. `clippy::default_constructed_unit_structs` - `crates/adapters/derive/bin/spot_research.rs:127:25` - same lint.

I have not dug into which of the commits on develop introduced these yet.

cjdsellers:
Hey @folknor,

The latest changes cleanly address the feedback. The Python boundary now returns `ValueError` across all nine affected methods, the tests cover both the direct paths and filtered-book calls without an own book, and the guide documents the updated contract. This looks good to me.

Separately, what Rust toolchain and Clippy command or config are you using for those three lint failures? I can reproduce the two `default_constructed_unit_structs` warnings with the pinned Rust 1.97.1 toolchain when explicitly checking those binaries, but `unused_async_trait_impl` only appears for me with nightly Rust 1.99.

I'm open to incrementally enabling useful additional lints or target scopes at the repository level. If you can share the exact command or config, we can compare it with the current gate and decide what makes sense to add, rather than chasing differences between local lint configurations.

cjdsellers (approved):

folknor:
@cjdsellers Here's my tooling for rust development, which I've adapted to work on nautilus_trader: https://gist.github.com/folknor/819d5204080a8995a4de083dd0ffaf6c?permalink_comment_id=6291052#gistcomment-6291052

I run the rust nightly toolchain. You might want to disable `disable_toolchain = true` at the top of the brokkr.toml.

EDIT: I want to make it clear, this doesn't replace all the CI gates you have, but it replaces what I consider important. Also, running bare `cargo` commands is blocked on my system - LLMs only have access to brokkr. LLMs are instructed to never `| tail` or otherwise mutate brokkr commands.

EDIT 2: My bad, there was a path dep in there you could not resolve. I've fixed that now.

faysou:
That's great. I was intrigued of how you find so many bugs. Thank you ! (Maybe brokkr is not the only method as well)

One comment, it would be useful to put an instruction for agents to generate one or several mermaid diagrams in PR summaries as this makes understanding what's going on much easier.

folknor:
I will give you guys direct and full commit access to brokkr if you want to use it. Perhaps try it for a bit.

The main benefit (and it's a big one) is that brokkrs standard output is extremely terse when it comes to lints and test failures and such. It saves on agent context.

@faysou Can you show me an example of a PR with a mermaid diagram?

PR #4603, opened 2026-07-29
Make Lighter reconnect auth-token rotation venue-visible

folknor:
A follow-up to the Lighter subscription-recovery rework in 1c555c143f73bd6e66c8960561b0c46493533660, whose duplicate-attempt coalescing can swallow the reconnect-time auth-token rotation.

## Reconnect auth rotation can be swallowed by attempt coalescing

On `Reconnected` the client replays the stored subscriptions, then the execution stream wakes the auth-rotation task, which mints a fresh token and re-issues `subscribe_account` for the five account channels. Both paths funnel into `queue_subscribe`, and the handler's `next()` loop selects unbiased between the command channel and inbound frames. When a fresh-token subscribe reaches `queue_subscribe` while that channel's replay attempt is still awaiting its ack, the request is coalesced into the live attempt - the stored auth is updated but no frame is sent - and the replay's ack then resolves the rotation's waiter. `rotate_auth_token_once` reports the rotation complete although the venue never saw the fresh token, and the next rotation is six hours out (`AUTH_TOKEN_REFRESH_INTERVAL`) against a seven-hour token TTL. The window is observable: `test_reconnect_replays_and_immediately_refreshes_authenticated_subscriptions` requires each account channel to see three subscribes (connect, replay, refresh) and fails intermittently on develop - roughly one run in six locally in isolation, deterministically once the mock's subscribe ack is delayed by 100ms - because the third frame is only sent when the replay's ack happens to win the race. nextest's process-per-test isolation does not help; the race is within a single test.

Coalescing now distinguishes queued from dispatched attempts. A changed token arriving for a queued attempt updates it in place, so the eventual frame carries the latest token. A changed token arriving for an in-flight attempt is recorded as a pending revision; when the in-flight generation completes, a successor generation carrying the fresh token is dispatched, and the rotation's waiter resolves only on the successor's own acknowledgement. Retries and reconnects fold the pending revision into the requeued attempt rather than sending an extra frame, a retry keeps its retry budget, and failure or cancellation drains both waiter sets. Duplicate subscribes carrying no token or the same token still coalesce.

## Completion frames are attributed to the request that owns them

The venue answers a subscribe with two completion-capable frames - the generic `subscribed` ack, then the typed `subscribed/<channel>` snapshot - and both resolved by topic alone, so the predecessor's trailing snapshot could complete a same-topic generation dispatched between the two: the serialized successor above, or a fresh attempt created after the predecessor completed. `complete_subscription` now takes the completion's kind, and a successful control-ack completion installs a one-shot per-topic tombstone that swallows the owed trailing typed frame before the inflight guard; typed and already-subscribed completions install none, and the map is cleared on reconnect. A swallowed frame is only excluded from completion attribution - it still flows through `handle_frame`, so the snapshot's account state is processed normally.

## Testing

- `changed_inflight_auth_waits_for_serialized_successor` pins the serialized successor: the old ack resolves only the old waiters, the trailing typed snapshot is swallowed with the successor's generation intact, and the fresh waiter resolves only on the successor's own ack.
- `changed_queued_auth_updates_existing_generation_before_dispatch` pins the queued case: one generation, dispatched with the latest token, both waiters resolved by its ack.
- `retry_folds_pending_auth_without_resetting_retry_budget` pins the fold: the retry carries the fresh token, no extra frame, and the retry count preserved.
- `stale_typed_frame_does_not_complete_fresh_attempt_after_predecessor_removed` pins the attribution interleaving where no successor existed when the tombstone was installed.
- The reconnect integration test now delays the mock's subscribe acks by 100ms after initial setup, turning the former race into a deterministic path, and asserts per channel that the replay carries the stored token and the third subscribe carries a different, freshly minted one.

All new assertions were verified to fail against the unfixed code: the integration test fails its 10-second wait with the handler changes reverted, and both attribution tests fail at their swallow assertions with the tombstone check disabled. The integration test passed fifteen consecutive isolated runs; no wall-clock waits were added beyond the bounded 100ms ack delay that makes the race deterministic.

### DIFF
--- a/crates/adapters/lighter/src/websocket/handler.rs
+++ b/crates/adapters/lighter/src/websocket/handler.rs
@@ -232,6 +232,7 @@ pub(super) struct FeedHandler {
     pending_subs: VecDeque<(Ustr, u64)>,
     inflight_subs: AHashMap<Ustr, u64>,
     subscription_attempts: AHashMap<Ustr, SubscriptionAttempt>,
+    ignored_completions: AHashMap<Ustr, CompletionKind>,
     next_subscription_generation: u64,
     instruments: AHashMap<i16, InstrumentAny>,
     book_delta_subs: AHashSet<i16>,
@@ -252,9 +253,36 @@ struct CachedOrderBook {
 struct SubscriptionAttempt {
     channel: LighterWsChannel,
     auth: Option<String>,
+    pending_auth: Option<String>,
     generation: u64,
     retries: u8,
     response_txs: Vec<tokio::sync::oneshot::Sender<Result<(), String>>>,
+    pending_response_txs: Vec<tokio::sync::oneshot::Sender<Result<(), String>>>,
+}
+
+impl SubscriptionAttempt {
+    fn fold_pending_auth(&mut self) {
+        if let Some(auth) = self.pending_auth.take() {
+            self.auth = Some(auth);
+            self.response_txs.append(&mut self.pending_response_txs);
+        } else {
+            debug_assert!(self.pending_response_txs.is_empty());
+        }
+    }
+}
+
+/// Which venue frame produced a subscription completion.
+///
+/// The venue answers a subscribe with a generic control ack followed by a
+/// typed `subscribed/<channel>` snapshot, and both complete by topic. The kind
+/// lets `complete_subscription` swallow the predecessor's trailing frame so a
+/// same-topic generation dispatched between the two cannot be completed by a
+/// response that belongs to the earlier request.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum CompletionKind {
+    ControlAck,
+    Typed,
+    AlreadySubscribed,
 }
 
 impl FeedHandler {
@@ -279,6 +307,7 @@ impl FeedHandler {
             pending_subs: VecDeque::new(),
             inflight_subs: AHashMap::new(),
             subscription_attempts: AHashMap::new(),
+            ignored_completions: AHashMap::new(),
             next_subscription_generation: 1,
             instruments: AHashMap::new(),
             book_delta_subs: AHashSet::new(),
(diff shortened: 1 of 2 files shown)


folknor:
@cjdsellers This fixes the "random" CI failures you're probably seeing on develop and in PRs.

folknor:
Hey @cjdsellers

Rebased onto 22e802bc23. That commit put `#[ignore = "Flaky reconnect timing; re-enable after synchronization is deterministic"]` on `test_reconnect_replays_and_immediately_refreshes_authenticated_subscriptions`, and this branch lifts it again - so I want to be explicit about that rather than have it show up as a silent revert.

The flake was the symptom this PR is about. The test requires three subscribes per account channel (connect, replay, refresh), and the third was only sent when the replay's ack happened to beat the rotation's `queue_subscribe`; when it lost, the fresh-token subscribe was coalesced into the still-unacked replay attempt, no frame went out, and the test sat out its ten-second wait. So the intermittency was not test-side timing noise - it was the venue-invisible rotation, sampled. With a changed token on an in-flight attempt now dispatched as a serialized successor generation, the third frame is sent unconditionally, and delaying the mock's subscribe acks by 100ms after initial setup makes the interleaving that used to be the unlucky one the only one. Fifteen consecutive isolated runs, no wall-clock waits beyond that bounded delay.

The order-identity work does not overlap with this. It changes `execution.rs` above the auth-rotation path and the `OrderIdentity` / `RetiredOrderCache` / `WsDispatchState` paths in `dispatch.rs`; `websocket/handler.rs` is untouched by it, and `rotate_auth_token_once` and both `subscribe_account` call sites are byte-identical. The rebase absorbed nothing - the diff is still +316/-25 across the same two files, the one extra deletion being the ignore attribute.

cjdsellers (approved):
Hey @folknor,

Thanks for the PR and for making the reconnect race deterministic. I reviewed the retry, reconnect, failure, and completion-attribution paths. The fresh-token waiter now resolves only after the successor carrying that token is acknowledged, and the predecessor’s typed snapshot cannot complete it.

The focused handler tests and reconnect integration test both pass locally. I found no blocking issues, so this looks good to merge 👌.

There is one non-blocking follow-up worth adding separately: a direct unit test for folding pending auth through a second reconnect (I'll action this).

PR #4608, opened 2026-07-30
Fix Clippy findings in kernel and Derive bins

folknor:
Three findings from a workspace-wide `cargo clippy --all-targets` sweep that no current CI lane reaches: one from clippy 1.98's new lints, following the nightly Clippy fixes in `88901da1a0`, and two stable-clippy findings introduced by `99f56028bc`.

## Suppress `unused_async_trait_impl` on `NautilusKernel::finalize_stop`

Clippy 1.98 adds `unused_async_trait_impl` as a sibling of `unused_async`, and its `check_impl_item` pass examines every `async fn` in an impl block without checking whether the impl implements a trait - so it also fires on the inherent `finalize_stop`, which already carries `#[expect(clippy::unused_async)]` for the same underlying condition. The fix applies the same idiom `88901da1a0` introduced for `reconcile_execution_mass_status` and the two websocket client sites: `#[allow(unknown_lints)]` (the lint id is unknown to 1.97) above an expect naming both lints. The `async` signature is unchanged.

## Construct `NonceManager` directly in the derive bins

`default_constructed_unit_structs` fires on `NonceManager::default()` in `bin/flatten.rs` and `bin/spot_research.rs`: `99f56028bc` made `NonceManager` a process-wide unit struct but left the two bins constructing it via `default()`. The lint is stable and warn-by-default, and went unseen because those files were not part of that change, so changed-files linting never revisited them. Both sites now use the unit-struct literal.

## Testing

No test changes - nothing behavioral changes. The differential is Clippy itself: a workspace `cargo clippy --all-targets` sweep reports exactly these three errors without the change and none with it. The two `default_constructed_unit_structs` findings are visible to the pinned 1.97.1 under the same sweep; `unused_async_trait_impl` is new in clippy 1.98, and on 1.97.1 the kernel attribute is inert (`unknown_lints` allowed, and the unit-struct literal is idiomatic under both).

### DIFF
--- a/crates/adapters/derive/bin/flatten.rs
+++ b/crates/adapters/derive/bin/flatten.rs
@@ -125,7 +125,7 @@ async fn main() -> anyhow::Result<()> {
     )?;
     let domain_separator = parse_b256(domain_separator_for(environment), "domain_separator")?;
     let action_typehash_bytes = parse_b256(ACTION_TYPEHASH, "ACTION_TYPEHASH")?;
-    let nonce_manager = NonceManager::default();
+    let nonce_manager = NonceManager;
 
     cancel_all_orders(&client, subaccount_id).await?;
     verify_no_open_orders(&client, subaccount_id).await?;
--- a/crates/adapters/derive/bin/spot_research.rs
+++ b/crates/adapters/derive/bin/spot_research.rs
@@ -124,7 +124,7 @@ async fn main() -> anyhow::Result<()> {
     let action_typehash: B256 = ACTION_TYPEHASH
         .parse()
         .map_err(|e| anyhow::anyhow!("failed to parse ACTION_TYPEHASH: {e}"))?;
-    let nonce_manager = NonceManager::default();
+    let nonce_manager = NonceManager;
 
     let instruments = client
         .get_instruments("ETH", DeriveInstrumentType::Erc20, false)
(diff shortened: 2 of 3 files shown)


cjdsellers:
Hey @folknor,

Thanks for putting this together. I've incorporated the two `NonceManager` fixes into a local change and expanded `scripts/clippy-changed.sh` to lint binary targets under the pinned Rust 1.97.1 toolchain. That will be included in an upcoming push.

As I mentioned on #4597, lint fixes need to be backed by a failure under the repository's current toolchain and lint configuration. I don't want to add speculative suppressions for future Clippy releases or chase differences from stricter local or nightly configurations. `unused_async_trait_impl` is a good example: it is new in 1.98, and this particular finding is a false positive because it fires on an inherent impl even though [the lint is intended for trait impl methods](https://github.com/rust-lang/rust-clippy/pull/16244).

I'm open to discussing scoped, gradual tightening of the repository's Clippy configuration and target coverage, as long as we adopt those changes at the repository level and validate them against the pinned toolchain.

I'll close this PR since the relevant stable fixes and missing binary coverage are now handled locally and will be pushed shortly :+1:

[edit]: Now pushed https://github.com/nautechsystems/nautilus_trader/commit/450d83da01c00aa4fca83426f58fae66bfc7b448.

PR #4610, opened 2026-07-30
Remove load-sensitive timing from the Hyperliquid stale-quote recovery test

folknor:
## Split the ladder claim from the wiring claim

`test_data_client_stale_quote_recovery_heals_without_reconnect` drove the stale-stream recovery ladder end to end against the mock server with every threshold at its 1-second floor, a finite resubscribe budget (`= 3`, with a comment admitting it papered over late heals), and a raw 3-second sleep as the window for the no-reconnect assertion. Under a loaded parallel `cargo test` run, scheduler latency can hold quote delivery across several 1-second health ticks, exhaust the budget, and escalate a healing stream to reconnect - failing the test for timing reasons nextest's process-per-test isolation happens to mask.

The escalation boundary is now pinned where time is synthetic: `test_stream_health_receive_resets_recovery_state` runs `MarketDataStreamHealthMonitor` with `max_targeted_resubscribes = 1` through warn, resubscribe, healing receive, then a fresh stale episode that warns and resubscribes *again* - possible only because healing reset the ladder - and reconnects only when the new episode exhausts its own allowance.

## Drive the integration test by causal completion markers

The integration test now owns only the WebSocket wiring, with every wait keyed on an observable event: the mock gates the initial and healing bbo messages behind explicit releases (`tokio::sync::Notify`, one permit per release, every post-initial subscribe gated so no uncontrolled send can heal the stream); the test awaits initial quote delivery, the logged resubscribe decision together with the server-observed bbo unsubscribe and re-subscribe, then releases and awaits the healing quote. `max_targeted_resubscribes` is `u32::MAX` here by design - this test no longer claims anything about budget exhaustion. The no-reconnect assertion moved after `disconnect()`, which cancels and awaits the health monitor, so the negative check has a real terminal fence instead of an arbitrary sleep.

## Testing

Removing the recovery-state reset fails the unit ladder at the new resubscribe-not-reconnect assertion, and omitting the resubscribe command times out the server-observation wait. The restructured integration test passed two consecutive full-workspace parallel `cargo test` sweeps - the load shape under which the previous version failed reproducibly - plus the adapter's full suite (952 tests).

### DIFF
--- a/crates/adapters/hyperliquid/src/data.rs
+++ b/crates/adapters/hyperliquid/src/data.rs
@@ -2355,7 +2355,7 @@ mod tests {
     fn test_stream_health_receive_resets_recovery_state() {
         let mut monitor =
             MarketDataStreamHealthMonitor::new(Duration::from_secs(5), Duration::from_secs(10))
-                .with_recovery(Duration::from_secs(10), 2);
+                .with_recovery(Duration::from_secs(10), 1);
         let instrument_id = btc_perp_id();
         let start = Instant::now();
 
@@ -2388,10 +2388,6 @@ mod tests {
         );
         assert_eq!(
             check_at(&mut monitor, start, 41)[0].action,
-            StaleStreamAction::Resubscribe,
-        );
-        assert_eq!(
-            check_at(&mut monitor, start, 51)[0].action,
             StaleStreamAction::Reconnect,
         );
     }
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I reviewed the exact head against current `develop`, including the recovery ladder and gated WebSocket flow, and the focused tests pass. I only noticed a minor comment-style issue around the post-initial BBO gate, which doesn't need to block this. This looks good to me 👍

PR #4611, opened 2026-07-30
Align Subscription equality and ordering and name the delivery comparator

folknor:
Extends the `Eq`/`Ord` agreement work merged in #4598 to the message bus subscription types, where the misaligned key is also the delivery-order mechanism.

## Align `Ord` with identity on both subscription types

`Subscription` and `TypedSubscription` define equality and hashing from `(pattern, handler_id)` but ordered on reversed `priority` first, so two subscriptions differing only in priority were `==` while `cmp` returned `Less`/`Greater` - violating the requirement that `a == b` exactly when `a.cmp(&b) == Ordering::Equal`. Both `Ord` implementations now compare `pattern` then `handler_id`, agreeing with `PartialEq`/`Eq`/`Hash`, which stay untouched: priority deliberately does not participate in identity, since duplicate registration ignores it and the first registration wins. The impls remain manual (`MStr` does not derive `Ord` and the handler is not orderable).

## Move delivery order to a named comparator

Delivery order lived implicitly in the old `Ord`, consumed by four bare `.sort()` calls (`MessageBus::inner_matching_subscriptions`, the uncached `publish_any` path, `subscribe_any`'s backfill into already-cached topics, and `TopicRouter::subscribe`). Each now uses an explicit `delivery_order` comparator carrying the exact previous key - priority descending, then pattern, then handler ID - so delivery semantics are byte-for-byte unchanged, including tie order among equal-priority handlers. Keeping the full key matters: a priority-only comparator under a stable sort would resolve equal-priority ties by input order, which differs per path (insertion order in the typed router, prior cache order in the backfill, hash-set iteration order in fresh matching) and would make delivery path-dependent.

## Testing

`test_subscription_ordering_laws` and `test_typed_subscription_ordering_laws` pin the contract over a variant matrix (same identity with different priorities is equal, hashes equal, and compares `Equal`; `partial_cmp` agrees with `cmp`; comparison is antisymmetric) - both fail against the previous implementations. Four delivery-path tests cover each former sort site with mixed and equal priorities inserted in reverse of the expected order, asserting the full delivery sequence: cache-miss `matching_subscriptions`, first uncached `publish_any`, a late subscription backfilled into an already-cached topic, and typed-router publish. The sequence assertions were verified to fail against a priority-only comparator variant, and the cached late-subscription case rejects that variant deterministically.

NB: The legacy Cython `Subscription` in `common/component.pyx` carries the analogous priority-in-comparison arrangement and is deliberately not touched here.

### DIFF
--- a/crates/common/src/msgbus/api.rs
+++ b/crates/common/src/msgbus/api.rs
@@ -258,7 +258,7 @@ pub fn subscribe_any(
     for (topic, subs) in &mut msgbus_ref_mut.topics {
         if is_matching_backtracking(*topic, sub.pattern) {
             subs.push(sub.clone());
-            subs.sort();
+            subs.sort_by(Subscription::delivery_order);
             log::debug!("Added subscription for '{topic}'");
         }
     }
--- a/crates/common/src/msgbus/core.rs
+++ b/crates/common/src/msgbus/core.rs
@@ -165,6 +165,14 @@ impl Subscription {
             priority: priority.unwrap_or(0),
         }
     }
+
+    pub(crate) fn delivery_order(&self, other: &Self) -> std::cmp::Ordering {
+        other
+            .priority
+            .cmp(&self.priority)
+            .then_with(|| self.pattern.cmp(&other.pattern))
+            .then_with(|| self.handler_id.cmp(&other.handler_id))
+    }
 }
 
 impl PartialEq<Self> for Subscription {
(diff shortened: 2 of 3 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. Separating subscription identity from delivery ordering fixes the `Eq`/`Ord` contract while preserving the existing priority and tie-break behavior across every delivery path. The focused tests pass against current `develop`, and I couldn't find any follow-ups. This looks good to merge.

PR #4612, opened 2026-07-30
Fix simulated venue liquidation, rollover, and reset state handling

folknor:
## Liquidate every breached currency in one pass

`SimulatedExchange::process_liquidations` evaluated all account currencies but stopped at the first breach: a `break` ended the loop after triggering one currency's liquidation, and `last_liquidation_ns` suppresses another pass at the same timestamp, so a simultaneous breach in a second currency stayed open until some later tick re-triggered. Liquidation is Rust-only (introduced in #4077, which scoped each breach to instruments settling in that currency - a scope `liquidate_open_positions` already enforces), so removing the `break` lets each independently breached currency trigger without widening any single breach's scope.

## Make rollover application atomic and mark only on success

`FXRolloverInterestModule::process` set `rollover_applied` before calculating, and the calculation's `Vec<Money>` return conflated success, no-work, partial application, and failure: instruments could be silently skipped for a missing matching engine, market price, or rate record, and a missing base-currency exchange rate substituted `0.0` - an economic no-op booked as done. Any of these permanently skipped that day's rollover (the legacy Cython module marks the day only after applying, and raises on missing prices). The calculation now returns an explicit `Completed`/`Retry` outcome: every open FX instrument's book, rate, and required exchange rate are preflighted, adjustments are generated all-or-nothing, and `rollover_applied` and `rollover_totals` update only on `Completed` - so a `Retry` day is re-attempted on subsequent process calls until the inputs resolve. Failure warnings are deduplicated per (instrument, failure kind) within the day, with repeats at debug, since retrying re-runs the calculation each call. `rollover_totals` now sums the booked `Money` adjustments rather than pre-rounding floats, so diagnostics match what was actually applied.

## Key the rollover day by date, not day-of-year

The day-change check compared `eastern_dt.ordinal()` alone, so a sparse data jump landing on the same ordinal of another year - day N of year Y to day N of year Y+1 - was treated as the same day, keeping the stale `rollover_applied` and skipping that day entirely. The key is now the Eastern `NaiveDate`.

## Clear the inflight counter on reset

`SimulatedExchange::reset` cleared the message and inflight queues but not `inflight_counter`, so counter entries survived into the next run and post-reset commands at a previously-seen timestamp received gapped sequence numbers. The counter is now cleared alongside the queues, matching the Cython engine's reset.

## Testing

`test_liquidation_closes_all_breached_currencies_in_one_pass` runs a margin account with two collateral currencies and a losing position settled in each, drives one liquidation pass through the real breach math and matching-engine fill path, and asserts both positions close - parameterized over both balance insertion orders so it cannot pass by traversal order. `test_fx_rollover_retries_after_quote_arrives` pins the retry cycle (no price at rollover, no adjustment and the day stays retryable; a quote later the same Eastern day yields exactly one adjustment, with no duplicate after). `test_fx_rollover_is_atomic_across_instruments` holds two priced-later instruments and asserts neither is applied until both can be valued. `test_fx_rollover_distinguishes_same_ordinal_across_years` asserts two rollovers across a year boundary on the same ordinal. `test_reset_clears_inflight_counter` pins the reset. Every assertion was verified to fail against the unfixed code: the liquidation cases in both orders against the restored `break`, the retry test against a mark-before-calculation variant, the cross-year test against an ordinal-only key, and the reset test with the clear removed.

### DIFF
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -1470,6 +1470,10 @@ impl SimulatedExchange {
     /// Must be called once per time step after all command queues have fully
     /// settled, not inside the settle loop.
     pub fn process_modules(&mut self, ts_now: UnixNanos) {
+        if self.frozen_account || self.exec_client.is_none() {
+            return;
+        }
+
         let results = {
             let cache = self.cache.borrow();
             let ctx = ExchangeContext {
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -60,7 +60,10 @@ use ustr::Ustr;
 
 use crate::{
     config::SimulatedVenueConfig,
-    modules::{ExchangeContext, SimulationModule},
+    modules::{
+        AccountAdjustmentError, AccountAdjustmentOutcome, ExchangeContext, SimulationModule,
+        SimulationModuleResult,
+    },
 };
 
 /// Represents commands with simulated network latency in a min-heap priority queue.
@@ -561,26 +564,44 @@ impl SimulatedExchange {
         }
 
         if let Some(exec_client) = &self.exec_client {
+            log::debug!("Adjusting account for venue {}", exec_client.venue());
+        }
+
+        match self.try_adjust_account(adjustment) {
+            Ok(()) => true,
+            Err(e) => {
+                log::error!("{e}");
+                false
+            }
+        }
+    }
+
+    /// Tries to adjust the account balance by the given amount without logging failures.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the account or currency balance is unavailable, the
+    /// resulting balance exceeds [`Money`] bounds, or account state generation fails.
+    pub fn try_adjust_account(&mut self, adjustment: Money) -> Result<(), AccountAdjustmentError> {
+        if self.frozen_account {
+            // Nothing to adjust
+            return Ok(());
+        }
+
+        if let Some(exec_client) = &self.exec_client {
             let venue = exec_client.venue();
-            log::debug!("Adjusting account for venue {venue}");
             let account_state = {
                 let cache = self.cache.borrow();
                 if let Some(account) = cache.account_for_venue(&venue) {
                     if let Some(balance) = account.balance(Some(adjustment.currency)) {
                         let mut current_balance = *balance;
                         let Some(total) = current_balance.total.checked_add(adjustment) else {
-                            log::error!(
-                                "Cannot adjust account: {} total exceeds Money bounds",
-                                adjustment.currency
-                            );
-                            return false;
+                            return Err(AccountAdjustmentError::TotalOverflow(adjustment.currency));
                         };
                         let Some(free) = current_balance.free.checked_add(adjustment) else {
-                            log::error!(
-                                "Cannot adjust account: {} free balance exceeds Money bounds",
-                                adjustment.currency
-                            );
-                            return false;
+                            return Err(AccountAdjustmentError::FreeBalanceOverflow(
+                                adjustment.currency,
+                            ));
                         };
                         current_balance.total = total;
                         current_balance.free = free;
@@ -596,30 +617,20 @@ impl SimulatedExchange {
                             self.clock.borrow().timestamp_ns(),
                         ))
                     } else {
-                        log::error!(
-                            "Cannot adjust account: no balance for currency {}",
-                            adjustment.currency
-                        );
-                        None
+                        return Err(AccountAdjustmentError::MissingBalance(adjustment.currency));
                     }
                 } else {
-                    log::error!("Cannot adjust account: no account for venue {venue}");
-                    None
+                    return Err(AccountAdjustmentError::MissingAccount(venue));
                 }
             };
 
             if let Some((balances, margins, ts_event)) = account_state {
-                if let Err(e) =
-                    exec_client.generate_account_state(balances, margins, true, ts_event)
-                {
-                    log::error!("Cannot adjust account: failed to generate account state: {e}");
-                    return false;
-                }
-            } else {
-                return false;
+                exec_client
+                    .generate_account_state(balances, margins, true, ts_event)
+                    .map_err(|e| AccountAdjustmentError::AccountStateGeneration(e.to_string()))?;
             }
         }
-        true
+        Ok(())
     }
 
     /// Returns whether there are pending commands at or before `ts_now`.
@@ -1459,7 +1470,7 @@ impl SimulatedExchange {
     /// Must be called once per time step after all command queues have fully
     /// settled, not inside the settle loop.
     pub fn process_modules(&mut self, ts_now: UnixNanos) {
-        let adjustments = {
+        let results = {
             let cache = self.cache.borrow();
             let ctx = ExchangeContext {
                 venue: self.id,
@@ -1470,12 +1481,22 @@ impl SimulatedExchange {
             };
             self.modules
                 .iter()
-                .flat_map(|m| m.process(ts_now, &ctx))
-                .collect::<Vec<Money>>()
+                .enumerate()
+                .map(|(module_index, module)| (module_index, module.process(ts_now, &ctx)))
+                .collect::<Vec<_>>()
         };
 
-        for adjustment in adjustments {
-            self.adjust_account(adjustment);
+        for (module_index, result) in results {
+            if let SimulationModuleResult::Completed(adjustments) = result {
+                let outcomes = adjustments
+                    .into_iter()
+                    .map(|adjustment| match self.try_adjust_account(adjustment) {
+                        Ok(()) => AccountAdjustmentOutcome::Applied,
+                        Err(e) => AccountAdjustmentOutcome::Failed(e),
+                    })
+                    .collect::<Vec<_>>();
+                self.modules[module_index].acknowledge(&outcomes);
+            }
         }
     }
 
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -1440,6 +1440,7 @@ impl SimulatedExchange {
         self.funding_settled_through.clear();
         self.message_queue.clear();
         self.inflight_queue.clear();
+        self.inflight_counter.clear();
 
         log::info!("Resetting exchange state");
     }
@@ -1561,8 +1562,6 @@ impl SimulatedExchange {
                     currency,
                 );
             }
-
-            break;
         }
     }
 
@@ -1773,13 +1772,17 @@ impl SimulatedExchange {
 
 #[cfg(test)]
 mod tests {
-    use nautilus_common::messages::execution::{QueryAccount, QueryOrder};
+    use nautilus_common::messages::execution::{QueryAccount, QueryOrder, SubmitOrder};
     use nautilus_execution::models::latency::StaticLatencyModel;
     use nautilus_model::{
-        enums::{AccountType, BookType},
+        accounts::MarginAccount,
+        enums::{AccountType, BookType, OrderSide, OrderType},
+        events::AccountState,
         identifiers::{ClientOrderId, StrategyId, TraderId},
         instruments::{CurrencyPair, InstrumentAny, stubs::audusd_sim},
+        orders::{OrderTestBuilder, stubs::TestOrderEventStubs},
         stubs::TestDefault,
+        types::AccountBalance,
     };
     use rstest::rstest;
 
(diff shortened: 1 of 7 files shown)


cjdsellers:
Hey @folknor,

Thank you for the PR. I resolved the current `develop` conflict and pushed signed merge commit `f177a5eaab` to `fix-backtest-venue-module-state`. The resolution preserves `develop`'s fallible rate and `Money` validation alongside this PR's retry state.

Local validation passed: nightly Rust formatting, 17 focused rollover, liquidation, and reset tests, the `nautilus-backtest` Python‑feature compile, and all repository commit hooks.

**I found three rollover issues that still need work before this can merge:**

- A calculation that returns `Retry` is abandoned when the Eastern date changes. If a missing quote or exchange rate arrives the next day, the prior‑day rollover is never booked.
- The module updates `rollover_totals` and sets `rollover_applied` before `adjust_account` runs. `process_modules` also ignores a failed adjustment, so booking can be partial or absent while rollover is treated as complete.
- `Cache::get_xrate` logs an error on every retry before the module's warning deduplication applies, which can flood logs at simulation cadence.

The branch is conflict‑free now. `pre-commit`, `pre-commit-v2`, and the Python v2 builds have passed; :ok_hand:

folknor:
Thanks for the merge and the review - all three are addressed on the branch, pushed on top of `f177a5eaab`.

On the abandoned retry: the module now keeps an explicit economic-day state - the date, its 17:00 Eastern boundary, the warning-dedup set, and any pending adjustment batch - and an incomplete day is no longer replaced when the Eastern date changes; it keeps processing priority across the calendar change until it books. Rate lookup and the Wednesday/Friday triple multiplier derive from the stored economic date rather than the retry timestamp, so a Wednesday rollover that resolves on Thursday still books with Wednesday's rates and multiplier. Once it completes, the observed day initializes and is immediately due if its boundary has passed. `AccountState` timestamps stay at actual application time rather than being backdated to the boundary: account events are appended in arrival order with `last_event()` meaning last arrival, event purging keys on `ts_event` age so a backdated state could qualify for immediate removal, and the portfolio analyzer assigns balances to calendar days by `ts_event`, so a backdated snapshot could overwrite a prior day's closing balance with a state that never existed at that time. The lateness is instead reported once, naming the economic date, the scheduled boundary, and the actual booking time.

On the premature commit: `SimulationModule::process` now returns `NotReady | Completed(Vec<Money>)` (distinguishing completed-with-nothing-to-book from not-ready) and the trait gains `acknowledge`, taking ordered per-adjustment `Applied | Failed(error)` outcomes. `process_modules` runs every module against the same read-only pre-adjustment snapshot first, then applies each completed batch and acknowledges the outcomes back to the originating module in the same call - so later modules no longer observe earlier modules' account changes, which the previous flat-map ordering had implicitly guaranteed. The application logic moved into a non-logging `try_adjust_account` returning a typed `AccountAdjustmentError`; `adjust_account` remains as the logging boolean wrapper with its existing behavior for the funding path and other callers. The module records a completed batch as pending before returning it, adds only acknowledged successes to `rollover_totals`, retains only failures for retry, and sets `rollover_applied` only when nothing remains pending - a partial application never replays succeeded entries, and application failures warn once per day per (currency, failure kind) before demoting to debug, so a persistently missing balance does not flood at simulation cadence either. The trait change is source-breaking for external `SimulationModule` implementors; there is no Python exposure of `process`, `acknowledge`, or `process_modules` (the v2 binding wraps the concrete module internally), so no stub changes are owed.

On the `get_xrate` flooding: `Cache` gains `try_get_xrate`, the same calculation returning `Result<Option<Decimal>>` with no logging, and `get_xrate` keeps its exact signature and logging as a wrapper over it - the portfolio, facade, and Python callers are untouched. Only the rollover module switched to `try_get_xrate`, folding both the error and the no-rate case into its existing deduplicated per-day warning with the error text preserved.

Testing: `test_fx_rollover_retry_keeps_prior_economic_day` pins the cross-day booking (Wednesday retries on Thursday, books `-8.22 USD` from January's rates with the triple multiplier, then Thursday initializes immediately due and books `10.96 USD` from February's); `test_partial_acknowledgement_retries_only_failed_adjustments` pins failed-only retry and the `rollover_applied` contract; `test_process_modules_forwards_real_outcomes_after_shared_snapshot` drives two modules through `process_modules` with a genuinely failing second adjustment, asserting the module receives `[Applied, Failed(MissingBalance)]` and that the account-state emission happens only after every module's process call (the shared-snapshot ordering); `test_missing_xrate_retries_emit_one_module_warning_and_no_cache_errors` captures logs while driving two real missing-xrate rollover attempts, asserting one warning and no `Failed to calculate xrate` lines; and `test_acknowledgement_count_panic_preserves_pending_batch` pins that a protocol-violating acknowledgement cannot corrupt the pending batch. Each was verified to fail against the previous branch head or a targeted regression: the cross-day test against the date-change abandonment, the seam test against both all-true outcome forwarding and an interleaved process-and-apply loop, the logging test against reverting the module to `get_xrate`, and the partial-ack test against discarding failed entries.

cjdsellers:
Hey @folknor,

Thank you, the current head addresses useful parts of the three earlier findings, but it still is not ready to merge:

- The rollover state preserves one pending economic day, but it skips intervening dates when a retry lasts more than one day. If Wednesday completes on Friday, the next call initializes Friday directly and Thursday is never booked.
- Missing rate data is not retryable because the calculator records are immutable. An unsupported currency or uncovered date therefore stalls that economic day, and every other instrument, for the rest of the run.
- `try_adjust_account` returns `Ok(())` for a frozen account and when no execution client is present. `process_modules` acknowledges that as `Applied`, so rollover totals and completion state report a booking even though the balance did not change.

The original `Cache::get_xrate` logging flood is fixed. There is also a smaller related issue: an invalid `Money` adjustment logs at error level before returning `Retry`, so that path still floods logs at simulation cadence.

folknor:
Thank you. No worries, we go as many rounds as you want :-)

folknor:
Thanks for the round - all four are addressed on the branch, pushed on top of `ac1aff9a7d`.

On the skipped intervening days: the economic-day cursor now advances sequentially. When a completed day's date is strictly earlier than the observed Eastern date, the next day initialized is `date.succ_opt()` - the completed day's successor - rather than the observed date. Each catch-up day has a boundary already in the past, so it is immediately due; a day with no eligible positions completes as an empty batch and the cursor moves on, so a Wednesday that books on Friday is followed by Thursday (single multiplier, Thursday's rates) and then Friday (triple), one economic day per `process_modules` call. The cursor never moves backward on a timestamp regression, and an incomplete day is preserved unchanged as before. There is deliberately no drain loop inside `process_modules`: a module that always returns an empty completed batch would loop forever, and later iterations would lose the shared pre-adjustment snapshot.

On the permanent rate stall: failure kinds now carry a disposition - `RetryDay` for the transient inputs (missing matching engine, price, exchange rate, which later market data can supply) and `SkipInstrument` for the permanent ones (missing rate data, an unrepresentable `Money` adjustment - the calculator's records are immutable for the run, so retrying can never succeed). A skipped instrument warns once through the existing per-(instrument, kind) day dedupe, naming the economic date, and the batch completes with the remaining instruments; if everything is skipped the day completes empty so later days still run. The skip is scoped to the economic day, not cached for the run, since a calculator may lack one month's data while holding the next. The rate lookup was also moved ahead of the engine/price/xrate checks, so a permanently unrated instrument that also happens to lack a quote is skipped rather than stalling the day behind the transient no-price retry. Transient failures still abort the whole batch, keeping the cross-instrument atomicity intact for inputs that can recover.

On the false `Applied`: `process_modules` now returns early when the account is frozen or no execution client is registered, before the read-only snapshot - modules are neither processed nor acknowledged, so `rollover_totals` and day completion can no longer record bookings that never touched a balance. Frozen is configured inapplicability rather than an application error, and mapping it to `Failed` would have recreated the permanent-retry stall; other exchange paths already gate on the frozen flag, and the engine registers the execution client before any module processing, so the no-client return is a setup boundary rather than a reachable simulation state. `adjust_account` keeps its existing contract for the funding path. The `rollover_applied` flag is renamed `rollover_completed`, since it marks day completion (it is set after an empty batch too) rather than an applied adjustment.

On the `Money` error flood: the conversion no longer logs directly. Failure is reported through the same deduplicated per-day warning as the other calculation failures (one warning per instrument per day, repeats at debug) and is treated as a permanent skip for that economic day - the next day recalculates from fresh inputs as usual.

Testing: `test_fx_rollover_catches_up_each_economic_day_in_order` extends the cross-day case to pin Wednesday `-8.22`, Thursday `10.96`, and Friday `32.88 USD` booking in order after a Friday resolution; `test_missing_rates_skip_instrument_and_do_not_stall_later_days` books the supported instrument while an unrated instrument - deliberately left without a quote, so the skip cannot hide behind the transient no-price retry - is skipped with exactly one warning, and later days complete empty; `test_process_modules_skips_when_account_adjustments_are_unavailable` drives frozen and no-client venues and asserts no module processing, no acknowledgement, and no account-state emission; `test_unrepresentable_money_warns_once_without_error_across_recalculation` pins one warning and no error-level lines across a recalculation forced by a separate transient failure. Each was verified to fail against the previous branch head, including the reordering: restoring the engine/price checks ahead of the rate lookup fails the missing-rates test.

cjdsellers:
Hey @folknor,

Thanks for sticking with the review and working through these rounds.

The current head addresses the earlier `get_xrate` logging, missing-rate, false-`Applied`, and `Money` logging issues, but two rollover paths still need work:

- `process` advances with `date.succ_opt()` over calendar days. After a normal Friday-to-Monday gap it books Saturday and Sunday even though Friday's triple already covers the weekend. Because the module advances only one economic day per call, sparse or daily data can also leave the cursor permanently behind the observed date.
- A failed account adjustment remains in `pending_adjustments` without any permanent-failure disposition. A missing quote-currency balance can therefore retry forever, preventing the economic day from completing and blocking every later rollover.

Could you add regressions for a Friday-to-Monday data gap and a permanently failed adjustment, then update the state transitions so weekends are not double-booked and an application failure cannot stall the module indefinitely?

folknor:
Thank you, @cjdsellers - both paths are fixed in the latest commit.

## Weekend double-booking and cursor lag

The cursor now enumerates rollover booking cutoffs rather than calendar days, so Saturday and Sunday are never cursor values: the sequence is Monday through Friday, then Monday. A multiplier weights one booking event and does not license skipping another weekday cutoff, so Wednesday is still followed by Thursday and Friday by Monday.

The one-day-per-call advance is replaced by a drain that walks every due weekday in order through the observed cutoff, calculating each with that date's own rate and multiplier and returning them as a single pending batch. On acknowledgement the cursor moves to the batch's stored end date rather than to the next day, which is what actually clears the lag - advancing by one would leave a drained three-week batch just as far behind as before.

Skip-ahead to the observed date was considered and rejected. It is cheap, but it permanently deletes every skipped rollover obligation, and it is worst across a rate-period boundary where the skipped dates carry different rates.

## Adjustment failures

`AccountAdjustmentFailureKind` now has a disposition split mirroring the one already on the calculation side. `TotalOverflow`, `FreeBalanceOverflow`, and `AccountStateGeneration` are transient and still retry: later debits or credits can restore representable headroom, free balance and locks can change, and a generation failure is a delivery problem rather than a rejection of the adjustment. `MissingBalance` and `MissingAccount` are permanent - retrying cannot create an absent configured currency balance, and once module processing is active the venue account should already exist.

A permanent failure warns once naming the date, currency, amount and kind, is added to a separate `unapplied_rollover_totals` ledger reported by its own diagnostics line, and is removed from pending so the batch can complete. It deliberately does not go into `rollover_totals`, which means successfully applied amounts. The whole day is not abandoned either: applications are performed sequentially before acknowledgement, so earlier adjustments in the same batch may already have succeeded and there is no rollback protocol.

The calculation-failure warn dedupe was keyed `(instrument, kind)`, which was correct when one state meant one day but silently swallowed information once a batch spans many dates - the first date warned and every later date dropped to debug, and a skipped `Rate` or `Money` calculation produces no unapplied total, so that loss was total. Both dedupe keys now carry the booking date.

## What this does not fix

Two limitations are documented beside the drain rather than hidden by it.

The catch-up is complete in booked-day count but not economically exact. The rate lookup is date-correct, but open positions, prices, and exchange rates are read as of the call, so a position opened near the end of a long gap can be charged for the whole gap while one closed before the next call receives no catch-up at all. Exact catch-up needs historical cutoff snapshots or reconstructable position and price history, which is a larger change than this round. Related, the work is proportional to gap length times instrument count, so a very long gap makes one expensive call. This is also a weekday calendar rather than a pair-specific business-day calendar.

Separately, and left alone deliberately: standard spot FX applies the weekend triple on Wednesday only, because T+2 moves settlement from Friday to Monday, and charges nothing directly for the weekend. The existing rule triples both Wednesday and Friday, so even with the weekend cursor entries removed it books more than the usual weekly accrual. I have treated that as deliberate parity and out of scope here, but it is worth a look if it was not intentional.

## Testing

`test_fx_rollover_friday_to_monday_gap_books_monday_once` completes a Friday and then ticks on Monday, asserting Friday books its triple and Monday books once with nothing between. Against the previous cursor it books three entries instead of one - Saturday, Sunday, and Monday - which is the double-booking made visible.

`test_permanent_adjustment_failure_completes_batch` drives two `MissingBalance` outcomes across two booking dates and asserts the batch completes, the amounts land in the unapplied ledger and not in the applied totals, both dated dedupe identities are recorded, and a later rollover still runs. Against the previous code it fails at the completion assertion, which is the stall. `test_transient_adjustment_failure_retries_only_failed_adjustments` keeps the transient path honest, so the fix does not turn every failure into a permanent skip, and `test_calculation_failure_dedupe_is_keyed_per_booking_date` pins that repeats still dedupe while a new date still warns.

The catch-up and cross-year tests were updated for the batching semantics: the former now receives the full due batch in one call, and the latter books the 23 intervening weekdays it previously skipped.

cjdsellers (approved):
Hey @folknor,

Thanks again, the latest head addresses the two remaining rollover findings. The weekday drain avoids weekend double-booking and cursor lag, while permanent adjustment failures no longer block later rollover dates. CI is green, and the branch merges cleanly with current `develop`, so this is safe to merge.

There is one small documentation mismatch in `process_liquidations`; I'll follow that up separately, and it doesn't need to hold this PR.

I think this is a good one to get in before the v2 transition.

PR #4613, opened 2026-07-30
Skip pong frames during reconnect and cap their payload

folknor:
## Skip pong frames raised against a connection that is gone

`WebSocketClient::send_pong` called `wait_for_active`, so a pong raised against one connection was held through the reconnection and enqueued on its replacement - a peer that never sent the matching ping. It now samples the connection mode and skips the frame when the client is not active. This narrows the window rather than closing it: `WriterCommand::Send` carries no epoch, so a pong already handed to the writer during reconnection is still replayed; strict epoch ownership is a separate follow-up.

## Reject oversized pong payloads before they reach the wire

`send_pong` accepted payloads beyond the 125-byte RFC 6455 control-frame limit. tungstenite 0.30 validates that limit only on inbound frames, so the frame was emitted and the peer closed the connection with a protocol error. The payload is now validated before enqueueing, returning a new `SendError::InvalidInput`, which `should_retry_lighter_ws_error` classifies as non-retryable. Validation precedes the inactive skip, so an oversized pong during reconnection is still a reported caller error.

## Testing

`test_send_pong_skips_gated_reconnect` holds the replacement handshake behind a gate: the pong call must return while the gate is shut, and exactly one fresh pong arrives on the replacement - deterministic, no timers. Two further tests pin the validation-before-skip ordering and that a rejected oversized pong leaves the connection undisturbed. Differentials verified: restoring `wait_for_active` fails the gated test; raising the byte cap fails all three rejection assertions.

### DIFF
--- a/crates/network/src/websocket/client.rs
+++ b/crates/network/src/websocket/client.rs
@@ -113,16 +113,6 @@ const WRITE_TIMEOUT_SECS: u64 = 5;
 /// The RFC 6455 control-frame payload limit.
 const MAX_CONTROL_FRAME_PAYLOAD_BYTES: usize = 125;
 
-fn validate_pong_payload(data: &[u8]) -> Result<(), SendError> {
-    if data.len() > MAX_CONTROL_FRAME_PAYLOAD_BYTES {
-        return Err(SendError::InvalidInput(format!(
-            "pong payload exceeds {MAX_CONTROL_FRAME_PAYLOAD_BYTES} bytes"
-        )));
-    }
-
-    Ok(())
-}
-
 /// Shared headers used by future automatic WebSocket reconnects.
 ///
 /// Updating these headers does not affect the active connection or trigger a reconnect.
@@ -2605,6 +2595,16 @@ fn fail_registered_auth(auth_tracker: &OnceLock<AuthTracker>, reason: &str) {
     if let Some(tracker) = auth_tracker.get() {
         tracker.fail(reason);
     }
+}
+
+fn validate_pong_payload(data: &[u8]) -> Result<(), SendError> {
+    if data.len() > MAX_CONTROL_FRAME_PAYLOAD_BYTES {
+        return Err(SendError::InvalidInput(format!(
+            "pong payload exceeds {MAX_CONTROL_FRAME_PAYLOAD_BYTES} bytes"
+        )));
+    }
+
+    Ok(())
 }
 
 impl Drop for WebSocketClient {
--- a/crates/network/src/websocket/client.rs
+++ b/crates/network/src/websocket/client.rs
@@ -110,6 +110,19 @@ use crate::{
 
 const WRITE_TIMEOUT_SECS: u64 = 5;
 
+/// The RFC 6455 control-frame payload limit.
+const MAX_CONTROL_FRAME_PAYLOAD_BYTES: usize = 125;
+
+fn validate_pong_payload(data: &[u8]) -> Result<(), SendError> {
+    if data.len() > MAX_CONTROL_FRAME_PAYLOAD_BYTES {
+        return Err(SendError::InvalidInput(format!(
+            "pong payload exceeds {MAX_CONTROL_FRAME_PAYLOAD_BYTES} bytes"
+        )));
+    }
+
+    Ok(())
+}
+
 /// Shared headers used by future automatic WebSocket reconnects.
 ///
 /// Updating these headers does not affect the active connection or trigger a reconnect.
@@ -2286,19 +2299,35 @@ impl WebSocketClient {
             .map_err(|e| SendError::BrokenPipe(e.to_string()))?
     }
 
-    /// Sends a pong frame back to the server.
+    /// Sends a pong frame back to the server when the connection is active.
+    ///
+    /// The pong is skipped silently if the connection is not active when called:
+    /// a pong belongs to the connection whose ping caused it, so this method does
+    /// not wait for reconnection before enqueueing it.
     ///
     /// # Errors
     ///
-    /// Returns a websocket error if unable to send.
+    /// Returns an error if:
+    /// - The payload exceeds 125 bytes, the RFC 6455 control-frame limit.
+    /// - The writer channel is broken.
+    #[allow(unknown_lints, reason = "Clippy lint is unavailable on Rust 1.97")]
+    #[expect(
+        clippy::unused_async,
+        clippy::unused_async_trait_impl,
+        reason = "skipping instead of waiting removes the only await; the signature is public API shared with the other send methods"
+    )]
     pub async fn send_pong(&self, data: Vec<u8>) -> Result<(), SendError> {
-        self.wait_for_active().await?;
+        validate_pong_payload(&data)?;
+
+        if !self.connection_mode().is_active() {
+            log::debug!("Skipping pong: connection not active");
+            return Ok(());
+        }
 
         log::trace!("Sending pong frame ({} bytes)", data.len());
 
-        let msg = Message::Pong(data.into());
         self.writer_tx
-            .send(WriterCommand::Send(msg))
+            .send(WriterCommand::Send(Message::Pong(data.into())))
             .map_err(|e| SendError::BrokenPipe(e.to_string()))
     }
 
@@ -2598,7 +2627,7 @@ mod tests {
     use std::{
         collections::HashMap,
         num::NonZeroU32,
-        sync::{Arc, Mutex},
+        sync::{Arc, Mutex, atomic::Ordering},
         time::Duration,
     };
 
@@ -2608,10 +2637,11 @@ mod tests {
     use rstest::rstest;
     use tokio::{
         net::TcpListener,
+        sync::{mpsc, oneshot},
         task::{self, JoinHandle},
     };
     use tokio_tungstenite::{
-        accept_hdr_async,
+        accept_async, accept_hdr_async,
         tungstenite::{
             Message as WsMessage,
             handshake::server::{self, Callback},
@@ -2620,6 +2650,7 @@ mod tests {
     };
 
     use crate::{
+        error::SendError,
         http::{HttpClient, Method},
         mode::ConnectionMode,
         ratelimiter::quota::Quota,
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. The reconnect and pong behavior looks good, and the focused tests pass, but there is one build issue to fix before merge.

The new `wait_for_terminal_state`, `send_message_after_reconnect`, and `send_pong_if_active` functions use `#[cfg(any(feature = "python", test))]`. Under `simulation + madsim`, `cfg(test)` includes those functions while both test modules that call them are excluded, so `nautilus-network` fails its warnings-as-errors compile gate with three `dead_code` errors. I reproduced this with:

`RUSTFLAGS='--cfg madsim' cargo check -p nautilus-network --tests --lib --features simulation`

Could you align those function gates with the test module's cfg while keeping the Python feature path enabled?

folknor:
Thanks - fixed and pushed, plus a merge of current `develop`.

The alignment mirrors every gate on the test module, not only the madsim arm: the module is also excluded under `feature = "turmoil"` and off Linux, and in either of those shapes `cfg(test)` alone would reproduce the same three `dead_code` errors. The helpers now carry `#[cfg(any(feature = "python", all(test, not(feature = "turmoil"), not(all(feature = "simulation", madsim)), target_os = "linux")))]`, with the `python` arm unconditional since `python/websocket.rs` calls `send_message_after_reconnect` and `send_pong_if_active` directly (and the former calls `wait_for_terminal_state`). Verified both directions against your repro shape: before the fix, `cargo clippy --all-targets -p nautilus-network --features simulation` with `RUSTFLAGS='--cfg madsim'` reports exactly your three `dead_code` errors (clippy additionally reports `unused_async` on `send_pong_if_active` there, since its awaits sit behind the excluded paths); after the fix the shape is clean, and the sim-compatible test subset per `make cargo-test-sim` passes across common, network, execution, core, and live.

While setting that verification up, linting the simulation+madsim shape surfaced four pre-existing findings on current `develop` that no CI lane currently reaches (`cargo-test-sim` runs tests without clippy, and the clippy jobs never pass `--cfg madsim`): `clippy::needless_return` and an unfulfilled `#[expect(clippy::needless_pass_by_value)]` in `crates/common/src/logging/logger.rs`, `clippy::manual_instant_elapsed` at `crates/live/tests/node.rs:1811` (`dst::time::Instant::now() - started_at`), and `clippy::unchecked_time_subtraction` in `crates/live/src/node/mod.rs`. All are stable-toolchain lints in the sim-only `cfg` paths, so they only appear under that shape. Happy to file a small follow-up PR making the simulation shape clippy-clean if you want it - or leave them with you, whichever you prefer.

cjdsellers:
Hey @folknor,

Thanks for the follow-up, the `simulation + madsim` cfg issue is fixed, and the original reproducer is clean.

I found one separate reconnect issue in the Rust `send_pong` path: after validating the payload, it still calls `wait_for_active()`. If reconnect starts between receiving the ping and this call, the pong waits and is enqueued on the replacement connection, even though the connection-ownership rationale says it belongs only to the connection that sent the ping. The Python path already skips while inactive.

Could you make Rust `send_pong` validate first, then return `Ok(())` unless the current connection is active, and add a reconnect regression test for the Rust entry point?

folknor:
@cjdsellers Thank you, I will follow up on all my 4 PRs you commented on within the next few hours.

folknor:

`WebSocketClient::send_pong` now delegates to the same `send_pong_if_active` helper the Python path already used, so there is one implementation of the sequence rather than two that can drift: validate the payload, then return `Ok(())` without enqueueing anything unless the connection is active when called. The helper's `cfg` gate is gone, since an unconditional caller makes it live in every build shape - that also removes one of the three gate mirrors from the previous round rather than adding to them. The doc comment picks up the helper's wording and no longer lists a wait timeout among the error cases, because it is no longer reachable.

Note that the old path reached `wait_for_active_state`, which returns `SendError::Closed` for `Disconnect` and `Closed`. Under delegation those modes are a silent `Ok(())` too. That matches the Python path and the connection-ownership rationale - a pong owed to a connection that is gone is not something the caller can act on - but it is a behaviour change on the terminal modes, not just on `Reconnect`. All twelve control-frame call sites already tolerate a non-error return.

## What this does not fix

The check narrows the window but does not make pong delivery strictly connection-owned. (1) Reconnect can begin after the mode is sampled active but before the writer consumes the command, and `WriterCommand::Send` is buffered during reconnect and replayed on the replacement connection. (2) A reconnect that completes between ping receipt and the call leaves the replacement looking active.

Strict ownership would mean carrying the inbound connection epoch with the ping and using an epoch-bound, no-replay writer command. The `SendOnConnection` machinery already has that shape and is rejected outright during reconnect, so the primitive exists - but binding pongs to it is an API change that reaches all twelve adapter ping handlers, and it would change the success path.

Happy to do it as a separate PR, or here as another follow-up if you want.

## Testing

`test_rust_send_pong_skips_gated_reconnect` drives the Rust entry point across a real reconnect. The server accepts the replacement TCP connection and signals the test, then holds the WebSocket upgrade; because the controller transitions to `Reconnect` before dialling, that signal is a causal barrier rather than a timing assumption, and the held upgrade keeps the client from reaching `Active` while the assertions run. Inside that window the test checks that a 126-byte payload is still rejected with `InvalidInput` - validation precedes the skip - and then polls a valid `send_pong` future exactly once, requiring `Poll::Ready`. The previous implementation returns `Pending` there, and the new test was verified to fail against it before the fix.

The gate is then released and the reconnection callback used as the completion barrier, confirming the epoch advanced exactly once. A fresh pong is sent and the replacement connection must observe exactly that payload. Because a replayed stale pong would necessarily arrive ahead of it in the writer FIFO, that assertion is what proves the skipped pong was never enqueued; the closing check that no second frame arrived is corroboration. There are no sleeps or polling loops - timeouts appear only as hang watchdogs around channel receives.

cjdsellers:
Hey @folknor,

Thanks, the latest changes address the specific cfg and Rust pong issues I raised.

Taking a step back and looking at this again in the context of the v2 cutover, I think we should pause this work for now. We do not intend to expose the raw `WebSocketClient` to Python after the cutover, and sends will not enter through this PyO3 path. **Removing those Python-specific paths should leave us with a clearer and simpler Rust architecture.**

I'm also concerned that this patch is growing the complexity of `nautilus-network` around state-wait helpers, feature and test cfg gates, parallel Rust and Python send behavior, and connection-ownership edge cases. I don't think it makes sense to keep refining this transitional path before its final boundary is clear.

Let's hold this until after the v2 cutover, remove the Python paths we no longer need, and then reassess the remaining Rust behavior and the smallest design from there.

I intend to circle back here post v2 cutover, otherwise ping me if it's feeling forgotten about. Thank you again for all of your work thus far :pray:

folknor:
@cjdsellers That's fair.

If convenient now or at some future point, please tell me what I can do to help reduce the time until v2 cutover.

cjdsellers:
@folknor, you won't be waiting long - preparations have already commenced and we can expect the last legacy v1 release imminently.

cjdsellers:
Hi @folknor,

Quick update on these network-related PRs: we're currently stripping out the unnecessary Python paths, this will greatly simplify the code and reduce the chance of bugs and edge cases lurking around. Let's reassess these networking PRs after today's pass - I expect some merge conflicts. Thanks for your patience.

folknor:
@cjdsellers Thank you. I will wait for your "go" and then I will rebase all my open PRs.

cjdsellers:
Hey @folknor,

The coast is clear on `nautilus-network` now, so you can rebase your PRs when you get a chance. I removed the redundant Python paths, expanded and strengthened test coverage, and took another refactoring pass to simplify the remaining code.

Some of the PRs may now be partly or fully redundant, but the rebase should make that clear. Let's see what remains.

folknor:
Rebased onto `develop`.

What remains is the pong semantics:

- `send_pong` waited for the connection to become active, so a pong raised against one connection was enqueued on its replacement. It now skips when the client is not active - the round-2 fix, re-expressed against the new code. As disclosed then, this narrows the window rather than closing it: `WriterCommand::Send` carries no epoch, so a pong that passes the mode check and is read by the writer during reconnection is still buffered and replayed. The strict epoch-ownership version remains offered as a separate PR whenever you want it; it reaches all twelve adapter ping handlers.
- The 125-byte control-frame cap, with `SendError::InvalidInput` and the Lighter retry classification.

Do you want me to rewrite the PR description?

cjdsellers:
Hey @folknor,

Thanks, the rebased Rust change looks sound, including the 125-byte boundary and reconnect-skip test.

Yes, please rewrite the PR title and description to the current Rust-only scope. The disclosed `WriterCommand::Send` replay window can remain a separate epoch-ownership follow-up.

One broader request: if possible, could you make future PR summaries about half this length while retaining the same high-signal detail? That would make them easier to review from a maintainer perspective.

cjdsellers (dismissed):

crates/network/src/websocket/client.rs
```diff
@@ -110,6 +110,19 @@ use crate::{
 
 const WRITE_TIMEOUT_SECS: u64 = 5;
 
+/// The RFC 6455 control-frame payload limit.
+const MAX_CONTROL_FRAME_PAYLOAD_BYTES: usize = 125;
+
+fn validate_pong_payload(data: &[u8]) -> Result<(), SendError> {
```
cjdsellers: nit: ideally we'd like to see these free functions below the main structs in their respective modules.

folknor:
Thank you, @cjdsellers - the title and description are rewritten to the Rust-only scope, and `validate_pong_payload` now sits below the structs beside `fail_registered_auth`. Noted on both counts for future PRs: helper functions below the primary types, and summaries at about half this length.

The epoch-ownership follow-up is on my list; I'll bring it as its own PR.

cjdsellers (approved):

folknor:
@cjdsellers Two clippy questions left over from this PR, if you want them.

Linting the `simulation + madsim` shape - which no CI lane does - reports four pre-existing findings on develop: `needless_return` and an unfulfilled `#[expect(needless_pass_by_value)]` in `crates/common/src/logging/logger.rs`, `manual_instant_elapsed` in `crates/live/tests/node.rs`, and `unchecked_time_subtraction` in `crates/live/src/node/mod.rs`. All stable-toolchain lints in sim-only cfg paths.

Separately, four clippy defaults absent from `[workspace.lints.clippy]` fire on toolchains newer than the 1.97.1 pin, so they arrive in your CI on the next bump: `clone_on_copy` (~95 files, most of `crates/model` and every adapter's `common/enums.rs`), `assert_is_empty` (~110 sites across 55 files), `unfulfilled_lint_expectations` (5 files), `double_must_use` (5 sites).

Do you want either? The second is mechanical but wide, so one PR per lint or per crate, or whichever is easier to review.

cjdsellers:
@folknor, Thanks for flagging these. Taking a closer look, you're right about the `simulation + madsim` set: this is a real gap in a supported stable configuration. I reproduced it with the repository's pinned Rust 1.97.1 toolchain, and it currently surfaces nine stable Clippy errors on `develop`, including the four you listed.

I'll handle this: fix the current errors and add a stable Clippy pass for `simulation + madsim` to the nightly DST checks. I won't add it to pre-commit because that cfg and feature set requires a separate Cargo build (and pre-commit run time is already too long as it is, pending refinement).

For the four defaults that only fire on toolchains newer than our pin, the position from [the earlier Clippy discussion](https://github.com/nautechsystems/nautilus_trader/pull/4597#issuecomment-5128902255) still stands. **We don't want to preemptively fix beta, nightly, or otherwise speculative lints before the repository's tracked configuration on latest stable reaches them. They can still produce false positives, crash, or change before stabilization, leading to avoidable churn.**

We're open to focused changes to the tracked stable Clippy configuration, including enabling a new lint or raising an existing lint's severity. Once a change is agreed and tracked, fixes reachable from that exact configuration are welcome :ok_hand:

[edit]: Nightly Clippy checks for DST / simulation landed here https://github.com/nautechsystems/nautilus_trader/commit/d6160e665b82d88cc3466282e4b0a935adff214c.

PR #4618, opened 2026-07-31
Make option expiry settlement transactional in the matching engine

folknor:
Option expiry settlement in `OrderMatchingEngine` dispatched events for legs it had not successfully registered, and committed expiration before settlement had run.

## All-or-neither registration and a single dispatch barrier

`option_register_settlement_order` ignored `add_order` failures as "already in cache", called `add_venue_order_id` with `.ok()`, and emitted `OrderAccepted` unconditionally; physical settlement registered its two legs separately and then dispatched both fills regardless, so one leg could apply while the other was rejected. Settlement is now planned as a unit across every open position of the expiring instrument: all legs register first, and only once every registration succeeds are the initialized, accepted, and filled events dispatched for all legs. A registration error dispatches nothing and marks settlement terminally failed for the run.

## Expiration commit deferred until settlement completes

`check_instrument_expiration` set `expiration_processed`, cleared `pending_resolution`, and consumed `instrument_close` before `process_option_expiry` ran - and that function returned silently on a missing underlying, an uncached underlying instrument, or a missing underlying price, permanently stranding every open position because the guard then blocked all retries. For `OptionContract` and `CryptoOption` the commit now happens only after settlement has dispatched; the missing-data cases defer with a deduplicated warning and retry with `instrument_close` preserved. Open orders are canceled at the first expiration trigger, since `iterate` matches resting orders ahead of the expiration check, and the cancellation is latched so a queueing handler does not receive a duplicate from the next retry's scan. The first trigger also enters pending resolution, so a deferred settlement leaves admission closed rather than accepting new resting orders the latch would prevent a retry from canceling.

## Independent settlement identifiers

The three identifiers for each leg derived from one truncated eight-character UUID, so the space was birthday-bounded and a collision hit all three at once. Each leg now takes independent full-width UUIDs: `EXPIRATION-{venue}-{uuid}` for the client and venue order IDs, matching the generic expiration path in the same function, and a bare UUID for the trade ID, which is a `StackStr` capped at exactly 36 characters. The leg role stays in the order tag.

The generic flat-close expiration path and the `BinaryOption` pending-resolution transition are unchanged.

## Testing

All timing is virtual (`TestClock`), with no wall-clock waits.

- `test_option_physical_settlement_second_registration_failure_dispatches_nothing`: a database failure on the second physical leg dispatches no acceptance or fill for either leg, does not commit expiration, blocks re-entry, and still cancels the resting order at the first trigger. It also settles a second, independent option afterwards, pinning that the residue of the failed registration does not obstruct later settlement.
- `test_process_option_expiry_missing_underlying_price_retries_with_close_preserved`: a missing price defers without committing or filling, and settles on a later check once a price arrives with the preserved `InstrumentClose` honored.
- `test_option_expiration_cancellation_latched_across_deferred_retries`: two deferred expiration checks dispatch exactly one `Canceled`.
- A deferred early `ContractExpired` close leaves admission closed: a probe order submitted afterwards is rejected.

The new assertions were verified to fail against the unfixed implementation: swallowing registration errors fails the second-leg test at its no-dispatch assertion, committing expiration on the deferred path fails the missing-price test, removing first-trigger cancellation fails both cancellation assertions, removing the latch dispatches two `Canceled` events, and the probe order is accepted without the pending-resolution entry.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6314,7 +6314,6 @@ dependencies = [
  "nautilus-common",
  "nautilus-core",
  "nautilus-model",
- "nautilus-testkit",
  "proptest",
  "pyo3",
  "pyo3-stub-gen",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6305,6 +6305,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "bon",
+ "bytes",
  "criterion",
  "futures",
  "indexmap 2.14.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6314,6 +6314,7 @@ dependencies = [
  "nautilus-common",
  "nautilus-core",
  "nautilus-model",
+ "nautilus-testkit",
  "proptest",
  "pyo3",
  "pyo3-stub-gen",
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -4489,6 +4489,18 @@ impl Cache {
 
         log::debug!("Adding {order:?}");
 
+        if let Some(database) = &mut self.database {
+            database.add_order(&order, client_id)?;
+            // TODO: Implement
+            // if self.config.snapshot_orders {
+            //     database.snapshot_order_state(order)?;
+            // }
+        }
+
+        if let Some(position_id) = position_id {
+            self.persist_order_position(&position_id, &client_order_id)?;
+        }
+
         self.index.orders.insert(client_order_id);
 
         if order.is_active_local() {
@@ -4558,26 +4570,18 @@ impl Cache {
 
         // Index position ID if provided
         if let Some(position_id) = position_id {
-            self.add_position_id(
+            self.commit_order_position(
                 &position_id,
                 &order.instrument_id().venue,
                 &client_order_id,
                 &strategy_id,
-            )?;
+            );
         }
 
         // Index client ID if provided
         if let Some(client_id) = client_id {
             self.index.order_client.insert(client_order_id, client_id);
             log::debug!("Indexed {client_id:?}");
-        }
-
-        if let Some(database) = &mut self.database {
-            database.add_order(&order, client_id)?;
-            // TODO: Implement
-            // if self.config.snapshot_orders {
-            //     database.snapshot_order_state(order)?;
-            // }
         }
 
         match self.orders.get(&client_order_id) {
@@ -4623,14 +4627,37 @@ impl Cache {
         client_order_id: &ClientOrderId,
         strategy_id: &StrategyId,
     ) -> anyhow::Result<()> {
+        // Legacy ordering preserved deliberately: existing callers that ignore or
+        // log this error rely on `order_position` surviving a database failure.
         self.index
             .order_position
             .insert(*client_order_id, *position_id);
+        self.persist_order_position(position_id, client_order_id)?;
+        self.commit_order_position(position_id, venue, client_order_id, strategy_id);
+        Ok(())
+    }
 
-        // Index: ClientOrderId -> PositionId
+    fn persist_order_position(
+        &mut self,
+        position_id: &PositionId,
+        client_order_id: &ClientOrderId,
+    ) -> anyhow::Result<()> {
         if let Some(database) = &mut self.database {
             database.index_order_position(*client_order_id, *position_id)?;
         }
+        Ok(())
+    }
+
+    fn commit_order_position(
+        &mut self,
+        position_id: &PositionId,
+        venue: &Venue,
+        client_order_id: &ClientOrderId,
+        strategy_id: &StrategyId,
+    ) {
+        self.index
+            .order_position
+            .insert(*client_order_id, *position_id);
 
         // Index: PositionId -> StrategyId
         self.index
@@ -4657,8 +4684,6 @@ impl Cache {
             .entry(*venue)
             .or_default()
             .insert(*position_id);
-
-        Ok(())
     }
 
     // Propagates parent OTO `position_id` to contingent children that are missing one.
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -4489,18 +4489,6 @@ impl Cache {
 
         log::debug!("Adding {order:?}");
 
-        if let Some(database) = &mut self.database {
-            database.add_order(&order, client_id)?;
-            // TODO: Implement
-            // if self.config.snapshot_orders {
-            //     database.snapshot_order_state(order)?;
-            // }
-        }
-
-        if let Some(position_id) = position_id {
-            self.persist_order_position(&position_id, &client_order_id)?;
-        }
-
         self.index.orders.insert(client_order_id);
 
         if order.is_active_local() {
@@ -4570,18 +4558,26 @@ impl Cache {
 
         // Index position ID if provided
         if let Some(position_id) = position_id {
-            self.commit_order_position(
+            self.add_position_id(
                 &position_id,
                 &order.instrument_id().venue,
                 &client_order_id,
                 &strategy_id,
-            );
+            )?;
         }
 
         // Index client ID if provided
         if let Some(client_id) = client_id {
             self.index.order_client.insert(client_order_id, client_id);
             log::debug!("Indexed {client_id:?}");
+        }
+
+        if let Some(database) = &mut self.database {
+            database.add_order(&order, client_id)?;
+            // TODO: Implement
+            // if self.config.snapshot_orders {
+            //     database.snapshot_order_state(order)?;
+            // }
         }
 
         match self.orders.get(&client_order_id) {
@@ -4627,37 +4623,14 @@ impl Cache {
         client_order_id: &ClientOrderId,
         strategy_id: &StrategyId,
     ) -> anyhow::Result<()> {
-        // Legacy ordering preserved deliberately: existing callers that ignore or
-        // log this error rely on `order_position` surviving a database failure.
         self.index
             .order_position
             .insert(*client_order_id, *position_id);
-        self.persist_order_position(position_id, client_order_id)?;
-        self.commit_order_position(position_id, venue, client_order_id, strategy_id);
-        Ok(())
-    }
 
-    fn persist_order_position(
-        &mut self,
-        position_id: &PositionId,
-        client_order_id: &ClientOrderId,
-    ) -> anyhow::Result<()> {
+        // Index: ClientOrderId -> PositionId
         if let Some(database) = &mut self.database {
             database.index_order_position(*client_order_id, *position_id)?;
         }
-        Ok(())
-    }
-
-    fn commit_order_position(
-        &mut self,
-        position_id: &PositionId,
-        venue: &Venue,
-        client_order_id: &ClientOrderId,
-        strategy_id: &StrategyId,
-    ) {
-        self.index
-            .order_position
-            .insert(*client_order_id, *position_id);
 
         // Index: PositionId -> StrategyId
         self.index
@@ -4684,6 +4657,8 @@ impl Cache {
             .entry(*venue)
             .or_default()
             .insert(*position_id);
+
+        Ok(())
     }
 
     // Propagates parent OTO `position_id` to contingent children that are missing one.
--- a/crates/execution/Cargo.toml
+++ b/crates/execution/Cargo.toml
@@ -71,8 +71,8 @@ pyo3 = { workspace = true, optional = true }
 pyo3-stub-gen = { workspace = true, optional = true }
 
 [dev-dependencies]
+bytes = { workspace = true }
 criterion = { workspace = true }
-nautilus-testkit = { workspace = true, features = ["cache"] }
 proptest = { workspace = true }
 serde_json = { workspace = true }
 
--- a/crates/execution/Cargo.toml
+++ b/crates/execution/Cargo.toml
@@ -72,6 +72,7 @@ pyo3-stub-gen = { workspace = true, optional = true }
 
 [dev-dependencies]
 criterion = { workspace = true }
+nautilus-testkit = { workspace = true, features = ["cache"] }
 proptest = { workspace = true }
 serde_json = { workspace = true }
 
--- a/crates/execution/src/matching_engine/engine.rs
+++ b/crates/execution/src/matching_engine/engine.rs
@@ -2389,13 +2389,12 @@ impl OrderMatchingEngine {
             self.instrument,
             InstrumentAny::OptionContract(_) | InstrumentAny::CryptoOption(_)
         ) {
-            // `iterate` matches resting orders ahead of this check, so cancel at
-            // the first trigger rather than after settlement. Latched because a
-            // queueing handler leaves the cached status behind the dispatch, and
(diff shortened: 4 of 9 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. The settlement sequencing and failure handling look good.

There’s one small change I’d like before merging: `option_cached_trade_ids` currently uses unfiltered `cache.orders(...)` and `cache.positions(...)`. Those APIs clone the complete ID sets and sort every order and position, so each settling option scans and sorts the full cache history. That cost repeats across every expiring option in an option‑chain backtest.

Let’s keep the global trade‑ID collision check, but build the set through `client_order_ids_view` and `position_ids_view` with point lookups, or another path that avoids cloning and sorting the full cache.

Once that’s changed, this looks ready from the code side :ok_hand:

folknor:

`option_cached_trade_ids` now walks the two unfiltered ID views and does point lookups instead of going through `cache.orders(...)` and `cache.positions(...)`. With all four filters `None`, `client_order_ids_view` and `position_ids_view` take the unfiltered arm and hand back a `Cow::Borrowed` over the index bucket, so nothing is cloned; `Cache::order` and `Cache::position` then return borrows. That removes, per settling option, two owned ID-set materializations, two full `Vec` builds over the entire cache history, and two sorts. The sort was the clearest waste: the result is only ever consumed as a membership test in `option_settlement_plan_collides`, so nothing iterates the set and nothing depends on its order.

The check stays global, and the collision and retry sequencing are untouched.

Note that the lookups are deliberately fail-fast rather than `if let Some(...)`. An ID present in the authoritative index but missing from object storage is an invariant violation that the old path already treated as a panic inside `get_orders_for_ids` and `get_positions_for_ids`. Swallowing it here would silently build an incomplete trade-ID set, which is the one failure mode that matters for this function - it could let settlement reuse a trade ID that is already in the cache. The panic messages match the old ones.

I did consider swapping the `IndexSet<TradeId>` for a plain hash set, since the iteration order is never consumed, but deliberately did not: `matching_engine/engine.rs` is on the deterministic-simulation list that requires `IndexMap`/`IndexSet` in favour of the `AHash` types. There is an exemption marker for that rule, and using it to save a hash lookup in a private helper did not seem like a trade worth making. The set stays an `IndexSet`.

One cost this does not remove: `Order::trade_ids()` allocates a `Vec<&TradeId>` per order, so there is still a per-order allocation on the scan. Getting rid of that needs either a model API change or a nine-variant match on `OrderAny`, which seemed out of proportion here - happy to take it as a follow-up if you think it is worth it.

Unrelated to the fix: the commit also adds a missing serial comma to a comment in `crates/execution/tests/matching_engine.rs`, a file this PR already touches. We are firm believers in the Oxford comma ourselves and applaud the cleanup in `3f71cbcf32`, so unless you would rather keep such edits out of otherwise-unrelated diffs, we intend to keep fixing stray sites in passing in files we are already opening.

## Testing

`test_option_cached_trade_ids_preserves_order_and_position_union` pins the union: it caches an order carrying a trade ID that appears in no position, and a position carrying a different trade ID whose opening order is never added to the cache, then asserts both are present. Because `add_position` indexes the opening-order relationship without adding that order to the order index, the second ID exercises the position loop alone, so the test fails if either half of the union is dropped - verified by removing each loop in turn.

This is a pure cost change with no behavioural delta, so it is a behaviour-preservation test rather than a regression that fails against a bug. There was no bug to reproduce.

cjdsellers:
Hey @folknor,

Thanks, the cache-scan change is addressed, but the new `nautilus-execution` dev dependency on `nautilus-testkit` still leaves a Cargo publication cycle: `nautilus-execution -> nautilus-testkit -> nautilus-trading -> nautilus-execution`.

On exact head `998e196353d2`, `scripts/ci/publish-cargo-crates.sh --check` exits `5` because it cannot sort the publishable crates. The same command lists all 41 crates and exits `0` at the merge base. The narrow `cache` feature avoids selecting `nautilus-trading` during this test build, but the publication planner intentionally includes every local dependency edge, including optional and dev dependencies.

Let's remove this cycle before merging, either by keeping the fault-injection database double somewhere that does not depend back through `nautilus-trading`, or by removing the `nautilus-execution -> nautilus-testkit` edge another way.

On the Oxford comma standardization(s), feel free to leave those within the same change set :ok_hand:

folknor:
Thank you, @cjdsellers - the cycle is gone in the latest commit.

`nautilus-execution` no longer depends on `nautilus-testkit` in any section. The fault-injection double now lives beside its only consumer as `FailNthAddOrderDatabase` in `crates/execution/tests/matching_engine/cache_database.rs`, which is deliberately narrow rather than a second general-purpose testkit: shared control state for the target call, a counting `add_order` that fails on it, and inert implementations elsewhere. `index_order_position` returns `Ok` because the first physical-settlement leg calls it, so a blanket-error stub would change which failure the test exercises. That leaves `bytes` as the one new dev-dependency, since `CacheDatabaseAdapter` signatures use it and execution did not depend on it before.

`crates/testkit` is reverted to develop byte for byte: the `cache` feature split and the `add_order` fault controls existed only to feed the edge you asked me to remove, and no other consumer used them. Its other users in system, backtest, and live are untouched.

## Testing

No settlement behaviour changed and nothing under `crates/execution/src` was touched.

The fault-injection regression matters most here, because swapping the double out is exactly the change that could quietly turn it into a test that passes for the wrong reason. Disarming the new stub's injection makes `test_option_physical_settlement_second_registration_failure_dispatches_nothing` fail at its no-dispatch assertion, so the failure it reports is still driven by an injected registration error rather than by an adapter that silently succeeds.

cjdsellers (dismissed):
Hey @folknor,

The cache scan now preserves the global trade-ID collision check without cloning and sorting full cache snapshots, and moving the fault-injection database beside the execution test removes the publication cycle. I reran `scripts/ci/publish-cargo-crates.sh --check` against the current `develop` merge, and it lists all 41 crates successfully. All required checks are green 👌.

There’s one structural change I’d still like. We generally avoid placing "auxiliary" structures above the primary structure in main modules. Let’s move `OptionSettlementLeg` and `OptionSettlementPlan` into a separate module, or place them below the main `OrderMatchingEngine` struct, or another location at your discretion.

Another nit is that there are a couple of fairly verbose inline comments. Preferably we should aim for these to be essential non-obvious information only (granted this is undocumented style preference, next time I'll comment inline as such).

After that, everything I raised is addressed. We’ll hold this PR until the v2 transition rather than merge it immediately.

folknor:
@cjdsellers Excellent. If you don't mind, I will wait until after the transition to update the PR because I have no real understanding of the ramifications yet. But I'm excited to find out!

folknor:
Thank you, @cjdsellers - both points are addressed in the latest commit, now rebased onto current `develop`.

`OptionSettlementLeg` and `OptionSettlementPlan` now sit below `OrderMatchingEngine` and its `Debug` impl rather than above the primary struct. I kept them in `engine.rs` rather than taking the separate-module option: `matching_engine/mod.rs` declares every sibling as `pub mod`, so a new module would either make two private helper types public or break that file's uniformity, and neither trade seemed worth it for eleven lines. Happy to move them out if you would rather have the module.

On the inline comments, three sites are trimmed. The cancellation block keeps the two facts that are not visible from the code - that `iterate` matches resting orders ahead of the expiration check, and that the latch exists because a queueing handler leaves the cached status behind the dispatch - and drops the restatement around them. The trade-ID note keeps the complexity argument and loses the prose. The third was a comment on `test_option_cached_trade_ids_preserves_order_and_position_union` describing the change as "not a regression for a behavioral bug", which was talking about the patch rather than the code, so it is gone entirely.

## Testing

No behaviour changed: the struct move is a relocation with no edit to the definitions, and the rest is comments. The full workspace test suite is green across the default, ffi, live and simulation configurations, and `scripts/ci/publish-cargo-crates.sh --check` still orders the workspace, so the round-2 cycle fix survived the rebase.

The rebase absorbed nothing: `git diff upstream/develop...` is unchanged except for one Oxford-comma hunk in `crates/execution/tests/matching_engine.rs` that you fixed identically on `develop`, so it now converges out of this branch.

## Two unrelated things noticed on current develop

Neither is from this branch, and both reproduce with an empty working tree - flagging them since they are cheap to fix and I would otherwise be sitting on them.

`crates/backtest/Cargo.toml` registers `examples/backtest_node_workload.rs` as both an `[[example]]` and a `[[test]]` with the same `name = "backtest-node-workload"`, added in `d84efc081d`. Cargo warns `file ... found to be present in multiple build targets` on every workspace build and test run, so it is in the output of any `cargo test --workspace`. Giving the two targets distinct names, or dropping one and reaching the fixture the other way, would clear it.

`clippy::large_enum_variant` fires on `PendingRunnerEvent` at `crates/live/src/runner.rs:161`, added in `ba24ad1305`. This one is shape-dependent and I suspect your gate does not see it: it appears under `cargo clippy --all-targets -p nautilus-common -p nautilus-live --features live`, but not in a full-workspace resolution, where the adapters' `high-precision` defaults change the variant sizes enough to clear the threshold. Since `scripts/clippy-changed.sh` carries `high-precision` in `DESIRED_FEATURES`, a standard-precision build with `node` enabled looks like the uncovered configuration. Boxing the largest variant would settle it either way.

cjdsellers:
Hey @folknor,

Thanks for the follow-ups, I found two remaining interactions in the new option-expiry state.

A `ContractExpired` close can trigger settlement while the engine clock is still before the instrument’s timestamp expiration. If settlement defers, the code latches cancellation but leaves `pending_resolution` false and the market open. A later `process_order` can therefore accept a new resting order, and the latch prevents a retry from canceling it. Let’s make the first option-expiry trigger enter pending resolution or otherwise close admission as part of the same transition.

The partial-registration residue also is not inert. `Cache::add_order` adds the client order ID to its indexes before the database call and inserts the order object only after that call succeeds. The injected failure can therefore leave an indexed ID with no order. The next option settlement walks the global ID view in `option_cached_trade_ids` and panics on that entry.

Let’s make failed registration roll back its cache mutations, add an atomic batch-registration path, or otherwise make the global scan safe for this exact residue. Please add regressions for a new order after an early deferred close and for a second option expiry using the cache after an injected registration failure.

folknor:
Thank you, @cjdsellers - both fixed in the latest commit.

The first option-expiry trigger now enters pending resolution rather than only latching cancellation, so a deferred settlement leaves the market closed and `process_order` rejects new orders; admission stays closed on terminal failure.

`Cache::add_order` is now failure-atomic: the database call runs before any in-memory mutation, so a failed registration leaves the cache untouched and the global scan can never meet an indexed ID without an order object. This is per-order atomicity - an earlier leg of a partially registered plan stays cached, but consistently: its object and position relationship resolve, and it never reaches the matching core. The public `add_position_id` keeps its previous ordering so existing callers are unchanged.

Both regressions are in and were verified to fail unfixed: the post-defer probe order was accepted, and the second expiry panicked in `option_cached_trade_ids`.

cjdsellers:
@folknor, Thanks for the follow-ups on both points.

**Normally I wouldn't re-review this deeply or reverse something I'd already asked for, but these changes sit under the whole system and there's something I missed in a previous round.**

I don't want the database call ahead of the in-memory mutations in `Cache::add_order`. It doesn't actually provide proper atomicity (as the database write isn't - _and shouldn't be_ - awaited), and then it makes "database first" the ordering convention for the cache, and every other write-through method would have to follow it to stay consistent.

It's also narrower than it reads. Both production adapters enqueue rather than write: `PostgresCacheDatabase::add_order` is a `self.tx.send(query)` and the Redis path bottoms out in the same channel send, so that `Err` means the receiver is gone or serialization failed, never that a write failed.

**Let's drop the global trade-ID scan and the 16-attempt loop:** Full-width `UUID4`s make the collision preflight unnecessary, and removing it takes the panic with it, so `Cache::add_order` can go back to how it was.

Two smaller things. `test_option_settlement_collision_dispatches_nothing` is in the description but I can't find it in the diff, and nothing else covers the collision path; it goes away with the preflight but worth reconciling either way.

Also, the identifiers lost their readable prefix. That's forced for `TradeId` at 36 characters, but not for `ClientOrderId` or `VenueOrderId`, and the generic expiration path in the same function already uses `EXPIRATION-{venue}-{uuid}` with a full-width UUID, so it'd be good to keep the two paths consistent.

The settlement sequencing itself still looks good though, and the second-leg registration test is a nice one. We just have to focus the scope of this change set though.

folknor:
@cjdsellers No worries friend, we do as many rounds as we need! However I'll be away today so expect a patch in 32+ hours :joy:

folknor:
Thank you, @cjdsellers - all four are addressed in the latest commit.

`option_apply_settlement_plan` now creates one plan, registers it once, and dispatches it once. Your two asks turned out to be one change: the preflight's real hazard was never a collision, it was that the scan panicked on an index entry whose object was missing - so removing the scan is what makes the reverted ordering safe again, rather than the two being traded off against each other.

This does drop trade-ID collision detection. Client order IDs colliding with a registered order are still rejected by `add_order` and venue order IDs by `add_venue_order_id`, both before anything dispatches, but nothing checks trade IDs now. At full-width `UUID4` I agree that is the right trade.

`test_option_settlement_collision_dispatches_nothing` was in `engine.rs`'s inline `mod tests`; it is removed with the preflight.

## Testing

`test_option_physical_settlement_second_registration_failure_dispatches_nothing` changed with the ordering. The failing second `add_order` now indexes its client order ID before the enqueue fails, so the test pins two indexed IDs with exactly one resolving to an order object, and only that leg holding a position relationship. Unchanged: no acceptance or fill for either leg, expiration not committed, re-entry blocked, and the resting order canceled at the first trigger. It then settles a second, independent option, which is what shows the stale index entry is inert.

Disarming the fault injection fails the test, because both legs then resolve to order objects.

cjdsellers:
Hey @folknor,

Thanks for working through the latest follow-ups. The cache scan, publication cycle, collision preflight, and identifier changes are addressed, but I found one reset interaction before this is ready.

The option path now calls `enter_pending_resolution()`, which sets `market_status` to `Closed`. Successful settlement clears `pending_resolution`, but `OrderMatchingEngine::reset()` does not restore `market_status`. `SimulatedExchange::reset()` reuses the same matching engines, so after a run in which an option expires, the next backtest run starts that engine closed and rejects orders even before expiry.

I confirmed this on 8be0ace7d970 with a focused regression: expire the option, call `reset()`, and assert the initial `Open` status. It fails with `Closed`.

Let’s restore the engine’s initial market status in `reset()` and add that regression. After that, everything previously raised looks addressed from the code side.

folknor:

`OrderMatchingEngine::reset()` now restores `market_status` to `MarketStatus::Open`, the constructor's initial value, alongside the expiry fields it already cleared.

That same line covers the venue-status case, which had the same problem before this PR: a `Halt`, `Pause`, `Suspend` or `Close` through `process_status` was also left in place by `reset()`. So there are two tests. `test_reset_restores_market_status_after_option_expiry` expires the option, asserts the engine reached `Closed`, resets, and asserts `Open`; `test_reset_restores_market_status_after_venue_status` does the same across the four actions. Both fail against the unfixed `reset()` - the first at `Closed`.

cjdsellers (approved):
Thanks @folknor, looks great now.

As a follow-up I'll add a `settlement.rs` module and centralize some of this machinery there.

PR #4620, opened 2026-07-31
Add LiveNode coordinated external-order-claim lifecycle API

folknor:
## Post-build claim registration reaching both tiers

External order claims live in two maps: the live reconciliation manager's (`ExecutionManager::external_order_claims`, consulted when reconciliation attributes venue-originated activity) and the execution engine's (consulted when materializing venue-originated status and fills). The only operation that registered claims with both was `LiveNode::add_strategy`, which requires `NodeState::Idle`; the coordinating `register_external_order_claims` was `pub(crate)`. A strategy added to a running node through `Controller`/`Trader::add_strategy` therefore could not populate the manager tier at all - the engine's `register_external_order_claims` is public, but `Controller` holds only the `Trader` and cannot reach the node-owned manager - so external orders on that strategy's instruments were attributed to `EXTERNAL` by reconciliation regardless of the engine-tier claim. `LiveNode::register_external_order_claims` is now public with no node-state restriction: it acquires the engine's `RefCell` with `try_borrow_mut` before any mutation, rejects in-request duplicates, preflights every instrument against both maps, and only then commits the whole batch to both tiers through infallible inserts. Every error path leaves both maps unchanged.

## Owner-scoped claim removal

Both maps were insert-only: the manager's `claim_external_orders` rejects an existing claim and nothing ever removed one, so a stopped-and-removed strategy's claims lingered on both tiers - a successor strategy on the same instrument was refused the claim and ran unclaimed, while stale claims kept routing external activity to the removed strategy ID. `LiveNode::deregister_external_order_claims` removes all claims owned by a `StrategyId` from both tiers: it collects the owner's instrument set from each map under the held engine borrow, errors without mutation if the sets diverge, and otherwise removes the full set from both. Removal is keyed by strategy rather than instrument so a caller cannot omit one configured instrument or remove another strategy's claim; removal with no claims present is an idempotent success. `Trader::remove_strategy` does not (and structurally cannot) touch the node-owned manager, so the documented sequence is: remove the strategy through the trader or controller, deregister its claims through the node, then register the successor's.

## No claims committed for a strategy the trader rejected

`LiveNode::add_strategy` registered claims before `Trader::add_strategy`, so a trader-side failure (for example a duplicate registration) left orphaned claims on both tiers with no strategy behind them. The path is now a preflight-and-commit split: claims are validated against both tiers first, the trader addition runs, and only on its success are the claims and the OMS override committed - through a new infallible `ExecutionEngine::commit_external_order_claims` whose contract requires the preflight, so no fallible operation follows the trader call. The engine borrow is taken only when the strategy actually configures claims or an OMS override, so adding a plain strategy no longer requires the engine `RefCell` to be available.

Both new methods are synchronous with no await points and are usable while the node is idle, between `poll` calls after a manual `start`, or after stopping; they cannot be called while `run(&mut self)` owns the node. The only Python-side change is a regenerated wrapper docstring; no signatures or stubs changed.

A deliberate follow-up, not part of this change: routing runtime strategy lifecycle through a node-owned command queue with responses. That would make claim cleanup automatic on strategy removal (covering the `Controller` path this PR cannot reach), and allow claim mutation while `run` owns the node.

## Testing

Eight new `rstest` cases in `crates/live/src/node/mod.rs` pin: registration after `build()` reaches both tiers; registration on a running node (after manual `start`) reaches both tiers; a batch containing one conflicting instrument leaves every new instrument absent from both tiers; a conflict present in only one tier changes neither; owner-scoped removal clears both tiers and a successor strategy can then claim the released instruments; divergent owner sets fail removal without mutation; removal with no claims is idempotent; and a forced `Trader::add_strategy` failure after successful claim preflight leaves neither tier populated. A ninth test holds an engine `borrow_mut` across a claimless `add_strategy` to pin that no borrow is required. The assertions were verified to fail against the unfixed implementation where expressible: the trader-failure case fails against the old commit-before-add ordering (the manager retained the claim), a preflight-bypass mutation fails the one-tier-conflict test, and a neutered divergence check fails the divergence test.

### DIFF
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -346,6 +346,18 @@ impl ExecutionEngine {
         self.external_order_claims.get(instrument_id).copied()
     }
 
+    /// Returns the instruments with external order claims owned by `strategy_id`.
+    #[must_use]
+    pub fn get_external_order_claims_for_strategy(
+        &self,
+        strategy_id: StrategyId,
+    ) -> HashSet<InstrumentId> {
+        self.external_order_claims
+            .iter()
+            .filter_map(|(instrument_id, owner)| (*owner == strategy_id).then_some(*instrument_id))
+            .collect()
+    }
+
     /// Registers a new execution client.
     ///
     /// # Errors
@@ -602,6 +614,41 @@ impl ExecutionEngine {
         }
 
         Ok(())
+    }
+
+    /// Commits external order claims for `strategy_id` without validation.
+    ///
+    /// The caller must have preflighted every instrument against
+    /// [`Self::get_external_order_claim`]: an existing claim is overwritten
+    /// without error. Coordinated live-node callers should use
+    /// `LiveNode::register_external_order_claims`, which preflights both the
+    /// execution engine and the reconciliation manager before committing;
+    /// ordinary callers should use
+    /// [`Self::register_external_order_claims`] instead.
+    pub fn commit_external_order_claims(
+        &mut self,
+        strategy_id: StrategyId,
+        instrument_ids: &HashSet<InstrumentId>,
+    ) {
+        self.external_order_claims.extend(
+            instrument_ids
+                .iter()
+                .map(|instrument_id| (*instrument_id, strategy_id)),
+        );
+
+        if !instrument_ids.is_empty() {
+            log::info!("Registered external order claims for {strategy_id}: {instrument_ids:?}");
+        }
+    }
+
+    /// Deregisters all external order claims owned by `strategy_id`.
+    ///
+    /// Coordinated live-node callers should use
+    /// `LiveNode::deregister_external_order_claims` so the execution engine and
+    /// reconciliation manager remain consistent.
+    pub fn deregister_external_order_claims(&mut self, strategy_id: StrategyId) {
+        self.external_order_claims
+            .retain(|_, owner| *owner != strategy_id);
     }
 
     /// # Errors
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -18,7 +18,10 @@
 //! This module provides the execution manager for reconciling execution state between
 //! the local cache and connected venues, as well as purging old state during live trading.
 
-use std::{cell::RefCell, fmt::Debug, rc::Rc, str::FromStr, sync::LazyLock, time::Duration};
+use std::{
+    cell::RefCell, collections::HashSet, fmt::Debug, rc::Rc, str::FromStr, sync::LazyLock,
+    time::Duration,
+};
 
 use indexmap::{IndexMap, IndexSet};
 use nautilus_common::{
@@ -2087,6 +2090,18 @@ impl ExecutionManager {
         self.external_order_claims.get(instrument_id).copied()
     }
 
+    /// Returns the instruments with external order claims owned by `strategy_id`.
+    #[must_use]
+    pub(crate) fn get_external_order_claims_for_strategy(
+        &self,
+        strategy_id: StrategyId,
+    ) -> HashSet<InstrumentId> {
+        self.external_order_claims
+            .iter()
+            .filter_map(|(instrument_id, owner)| (*owner == strategy_id).then_some(*instrument_id))
+            .collect()
+    }
+
     /// Claims external orders for a specific strategy and instrument.
     ///
     /// # Errors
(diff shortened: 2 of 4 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. The coordinated preflight and commit ordering and owner-scoped deregistration look sound. The current-base merge is clean, CI is green, and I found no hard blockers.

I only found two small follow-ups would be great to get in prior to merge: the engine-borrow error and docs should cover the OMS-only path, and the symmetric engine-only conflict branch would benefit from a regression test :ok_hand:.

Small additional note that "external order claims" is likely to receive a rename for clarity after the v2 cutover (there will be a large config option naming convention audit at that point). This is a minor detail though as renames are trivial and relates more to API versioning and breaking changes.

folknor:
Thank you, @cjdsellers - both follow-ups are in the latest commit.

The engine borrow in `add_strategy` is taken when a strategy configures claims *or* an OMS type override, but the `map_err` said only "Cannot register external order claims" and the `# Errors` list attributed the borrow failure to strategies configuring claims. The message is now "Cannot register external order claims or OMS type", and the docs split the two failure modes apart: one bullet for the claim conflicts, one for the borrow, with the borrow bullet naming both triggers and stating that a strategy configuring neither never takes the borrow and so cannot fail that way. The wording is "one or more external order claims" throughout, since a strategy returning `Some(vec![])` takes the empty path.

`preflight_external_order_claims` checks the manager tier and then the engine tier, and the existing one-tier test seeded the conflict only in the manager, so the engine branch had no coverage. `test_register_external_order_claims_engine_only_conflict_changes_neither_tier` is the mirror image: it seeds through `ExecutionEngine::register_external_order_claims` and asserts the same both-tiers-unchanged invariant. I also renamed `test_add_strategy_without_claims_does_not_require_engine_borrow` to `..._without_claims_or_oms_type_...`, since the old name carried the same claims-only framing the docs did.

## One divergence this leaves in place

The new `# Errors` text is accurate for the native `add_strategy`, but the Python wrapper does not behave that way on the OMS-only path. `python/node.rs` registers the OMS type with `borrow_mut()` *after* committing the strategy, so a borrow conflict there panics instead of returning this error, and its claims path reports the claims-only message. The wrapper docstring is generated from the native one, so it inherits the new text either way.

I have left that alone. The change would have to be made in the wrapper body itself rather than behind the binding, and it is a behavioural change to the Python path rather than the doc and error fix you asked for, so the Python half is yours to take or leave.

## Testing

The new test uses two distinct instruments and seeds only the engine tier, so neither the in-request duplicate check nor the manager branch can fire first; whichever order the preflight set iterates, it must reach the engine-conflicting instrument. Neutering the engine-tier conflict check fails that test alone, while the manager-only test and the rest of the `test_register_external_order_claims` family keep passing - which is what shows it covers a branch that previously had none rather than duplicating existing coverage.

cjdsellers:
Hey @folknor,

Thanks, the two follow-ups from my last comment are addressed.

I found one blocking feature-build issue on the updated head: the three new `ExecutionManager` methods are only called from the `node` module, but they compile when `node` is disabled. `cargo check -p nautilus-live --no-default-features --lib` fails under `-D warnings` because all three are dead code.

Add `#[cfg(feature = "node")]` to `get_external_order_claims_for_strategy`, `register_external_order_claims`, and `deregister_external_order_claims`, matching `remove_targeted_order_queries` above them.

The Python lifecycle divergence you called out predates this PR, so I don't think this fix needs to expand into that path :ok_hand:

folknor:

All three `ExecutionManager` methods now carry `#[cfg(feature = "node")]`, matching `remove_targeted_order_queries` above them. The neighbouring `get_external_order_claim` and `claim_external_orders` are unaffected because they are `pub` and so never counted as dead; the three new ones are `pub(crate)` with their only callers in `node/mod.rs`, which is exactly the combination that fails once `node` is off.

Gating them orphaned the `collections::HashSet` import, since the two claim-set signatures were its only remaining users in that module, so that import is now gated the same way.

Verified in the shape you used rather than only the default one: `nautilus-live` with default features disabled is clean and its 50 tests pass, and the normal `nautilus-live` plus `nautilus-execution` build with the `node` feature is unchanged at 1488.

Understood on the Python lifecycle divergence - leaving it alone.

folknor:
@cjdsellers The remaining CI gate failure here seems like a fluke.

cjdsellers (approved):
Hey @folknor,

Thanks, the earlier follow-ups and feature-gating issue are addressed, and I found no further issues 👌.

The remaining CI failure has also cleared and all required checks pass. I'll approve this, but we'll hold it back with the other PRs until we're past the v2 cutover.

cjdsellers (approved):
@folknor, I re-reviewed the rebased head against the current post-cutover `develop`. The patch is unchanged, the merge is clean, all required checks pass, and I found no new issues, so this is good to merge 👍.

The behavioral summary remains accurate for the Rust-native API. Minor nit: the current diff adds ten tests, not eight plus a ninth; that is a description-only correction and does not block the merge.

PR #4621, opened 2026-07-31
Log connection lifecycle transitions at Info in network clients

folknor:
## Connection lifecycle transitions logged at Info

Both `nautilus_network` controllers log their connection lifecycle at Debug, so an operator running at Info cannot observe a dead connection and its recovery at all - a venue connection can die and cleanly reconnect with no trace at the default operational level. The developer guide places connection lifecycle events and reconnection summaries at Info (`docs/developer_guide/rust.md`), and these six lines are the lifecycle transitions: the dead-connection detection in each controller loop (`websocket/client.rs` "Detected dead connection, transitioning to {target:?}", `socket/client.rs` "Detected dead read task, transitioning to RECONNECT"), the reconnect attempt start ("Reconnecting" in each client's `reconnect`), and the reconnect success ("Reconnect succeeded" in each, emitted after the `Reconnect -> Active` state transition and replacement reader installation). All six are raised from `log::debug!` to `log::info!` with wording unchanged.

The surrounding diagnostics deliberately stay at Debug: the numbered per-attempt detail line (which would duplicate the attempt start at Info), the websocket controller's post-callback completion line, the raw socket controller's secondary completion line (which can also fire after an aborted reconnect), and all abort, interruption, backoff, and writer-swap lines. The result at Info is exactly one line per transition: connection died, reconnect started, reconnect succeeded.

## Testing

No behavior changes; the existing reconnect tests exercise all six sites (`cargo test -p nautilus-network`, 488 tests). No new test asserts the levels themselves: pinning log levels reliably would require an isolated test binary with real TCP and WebSocket servers plus a process-global logger, since the existing parallel test binaries emit these same messages concurrently - disproportionate to a six-line level change.

### DIFF
--- a/crates/network/src/socket/client.rs
+++ b/crates/network/src/socket/client.rs
@@ -350,7 +350,7 @@ impl SocketClientInner {
     /// reader to a timeout; the writer task bounds both the old-writer
     /// shutdown and the buffer drain with its graceful-shutdown timeout.
     async fn reconnect(&mut self) -> Result<(), Error> {
-        log::debug!("Reconnecting");
+        log::info!("Reconnecting");
 
         if ConnectionMode::from_atomic(&self.connection_mode).is_disconnect() {
             log::debug!("Reconnect aborted due to disconnect state");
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. This is a nice refinement for v2, and the changes look good 👍.

I'll hold this until the v2 cutover so it can land without conflicting with the Python‑layer network lifecycle logging still used by v1 (likely within 48 hours now).

PR #4622, opened 2026-07-31
Add a runtime log-level ceiling to the logging subsystem

folknor:
## A runtime log-level ceiling beside the existing global controls

There is no way to change logging verbosity after initialization: `Logger.config` is an initialization snapshot, the logger installs via `set_boxed_logger` with no handle kept, the writers bake `stdout_level`/`fileout_level` at construction, and initialization unconditionally calls `set_max_level(Trace)`. A consumer wanting runtime verbosity control (for example gating a live run's log volume per session) can only call `log::set_max_level` directly - which works today because the `log` macros consult the global maximum before `Logger::enabled`, but rides on the unenforced assumption that Nautilus never re-sets the maximum mid-run; if that ever changed, the external caller's setting would be silently clobbered.

`logging_set_level(LevelFilter) -> anyhow::Result<()>` now sits beside `logging_set_bypass` and owns that fast path as a process-global runtime ceiling. It composes with, and never rewrites, the existing configuration: effective admission remains bounded by `log::STATIC_MAX_LEVEL`, the configured per-sink thresholds, and the component/module filters, so a runtime `Debug` cannot make a file configured at `Info` emit debug records, cannot create a file writer that initialization declined, and cannot bypass a component cap. No control event or per-record atomic is added - the `log` macros already read the global maximum per record; records admitted before a change may drain at the previous level.

Called before initialization, the validated level is stored and applied by initialization in place of the unconditional `Trace` (the store and the initialization read are both serialized under the logger lifecycle lock, so a concurrent setter cannot be clobbered by publication). Called while running, it applies immediately. After terminal shutdown, valid levels are accepted as no-ops and `log::max_level` stays `Off`. When never called, initialization behavior is unchanged.

## Two levels are refused rather than accepted-and-documented

`Off` returns an error directing the caller to `logging_shutdown`: the shutdown-on-error mechanism lives inside `Logger::log`, so suppressing all records at the macro fast path would also blind error detection - a setter that accepted `Off` would report success while disabling a safety mechanism.

`Trace` returns an error naming `release_max_level_debug` whenever `log::STATIC_MAX_LEVEL < Trace`. The workspace enables that `log` feature, which pins the static maximum to `Debug` in builds without debug assertions - and feature unification imposes it on every downstream consumer's `log` as well - so trace macro sites are compiled out and a runtime `Trace` could never surface a line. Accepting it silently would produce a setter that reports success and does nothing; where the static maximum permits `Trace` (debug-assertion builds), it is accepted normally. Validation precedes lifecycle dispatch, so both refusals hold before initialization and after shutdown too.

## Testing

Seven new tests. Under `serial_tests` (process-isolated by the nextest convention): a raise/lower sequence against file output with unique markers and `logging_sync_to_disk` (runtime `Info` admits Info and rejects Debug, a raise to `Debug` admits a new debug marker, a lower to `Warn` rejects a new info marker); a pre-initialization store honored by initialization (`Warn` set before init, info marker absent after); filter composition (runtime `Debug` does not override a component capped at `Info` while an unfiltered debug marker appears); the `Off` refusal with the documented error text and no state change; and a `Trace` test branching on the real `STATIC_MAX_LEVEL`. The pure validator is additionally tested with an injected static ceiling - `(Trace, Debug)` rejects naming `release_max_level_debug`, `(Trace, Trace)` accepts - because the test profile enables debug assertions, making the real static maximum `Trace` in-tree; the injected-ceiling test is what covers the release-build refusal branch. The post-shutdown no-op runs in the existing lifecycle child-process harness, since shutdown permanently terminates the process-global logger. The pre-initialization and raise/lower assertions were verified to fail against the unfixed behavior: with initialization's unconditional `set_max_level(Trace)` restored, the pre-init test fails at its suppressed-marker assertion.

### DIFF
--- a/crates/common/src/logging/logger.rs
+++ b/crates/common/src/logging/logger.rs
@@ -116,6 +116,9 @@ enum LoggerLifecycle {
 /// Process-global logger lifecycle serialization.
 static LOGGER_LIFECYCLE: Mutex<LoggerLifecycle> = Mutex::new(LoggerLifecycle::Uninitialized);
 
+/// Runtime ceiling stored before initialization or applied while the logger is running.
+static RUNTIME_LEVEL_OVERRIDE: Mutex<Option<LevelFilter>> = Mutex::new(None);
+
 #[cfg(all(test, not(all(feature = "simulation", madsim))))]
 struct InitPublishHook {
     reached: std::sync::mpsc::Sender<()>,
@@ -1051,7 +1054,10 @@ impl Logger {
             super::logging_set_bypass();
         }
 
-        let max_level = log::LevelFilter::Trace;
+        let max_level = RUNTIME_LEVEL_OVERRIDE
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .unwrap_or(LevelFilter::Trace);
         set_max_level(max_level);
 
         if print_config {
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. I reviewed the current head, and the implementation and tests look sound :+1:.

I'm going to park this until after the v2 transition. This adds useful Rust‑side control, but it is process‑global, has no in‑tree production caller, and is not exposed through PyO3. It is therefore not needed for the cutover, and the default behavior remains unchanged.

Let's revisit it once the transition has settled.

cjdsellers (approved):

cjdsellers:
@folknor, I've taken another look at this, and I'm going to hold off on expanding the runtime logging surface for now. Release builds already compile out Trace logging, while the existing component‑level filters and granular logger configuration cover the majority of use cases. The runtime ceiling is a useful nice‑to‑have, but it adds process‑global state and lifecycle complexity without enough practical gain at this stage.

This is something I may reconsider once v2 is more stable and a strong use case emerges.

PR #4623, opened 2026-07-31
Distinguish aborted reconnects from completed reconnections

folknor:
## Aborted reconnects are indistinguishable from completed ones

Both `nautilus_network` controllers drive reconnection through an `async fn reconnect()` returning `Result<(), _>`, and that type cannot express the third outcome the function actually has. Nine early returns unwind without reconnecting - the stream-mode refusal, three `is_disconnect()` checks in each client (before dialing, after `connect_with_server`, and after the graceful-shutdown delay), and the final `compare_exchange(Reconnect -> Active)` in each - and every one of them returns `Ok(())`, the same value a completed reconnection returns. The controller task consuming that result has no way to tell the two apart, so it runs its success bookkeeping either way and then re-derives the outcome by reading the connection-mode atomic a second time.

That re-read is not equivalent to knowing the outcome. It answers a different question - what the mode is now, not what the attempt did - and the two diverge in both directions: a reconnection that succeeded and whose replacement immediately died reads as not-active, while the mode being terminal does not distinguish an attempt that unwound from one that completed before the teardown landed.

This adds a `ReconnectOutcome` returned by both clients' `reconnect()`, with the final transition extracted to `ConnectionMode::complete_reconnect` in `mode.rs` alongside the existing `request_reconnect` and `request_disconnect`, which are its siblings - it performs the same compare-exchange both controllers previously open-coded and names the two results. `Err` keeps its present meaning of an attempt that failed and should warn and back off. The controllers now match on the outcome, and success bookkeeping runs only under `Reconnected`. `WebSocketClientInner::reconnect` is public and re-exported, so it keeps its `Result<(), _>` signature as a wrapper over the outcome-returning form rather than breaking that surface.

The `is_active()` check is deliberately retained around the recovery callbacks, because it was doing a second job unrelated to classification: suppressing venue-facing side effects once a teardown has been requested. Adapters register `post_reconnection` handlers that replay authentication and subscriptions, so firing them into a shutdown would put traffic on a connection being torn down. Classification now comes from the outcome; the guard now answers only whether the replacement is still usable.

## Raw socket controller logs success on an aborted reconnect

The raw socket controller emits `"Reconnected successfully"` as the first statement of its `Ok` arm, above and outside the `is_active()` guard, whereas the websocket controller emits its equivalent inside that guard. With every abort path returning the same `Ok(())`, the raw socket client therefore claims a successful reconnection every time one unwinds - including on the ordinary teardown path, where a disconnect races an in-flight reconnect. #4621 noted this line as one that "can also fire after an aborted reconnect" when it left the surrounding lifecycle lines at Debug; this is that case. The line is now reachable only under `ReconnectOutcome::Reconnected`.

Both controllers also record the reconnect timestamp that feeds the backoff stability ladder on every `Ok`, abort included. That bookkeeping has no observable consequence on the current code: every abort leaves the mode terminal, and both controller loops break on `is_disconnect()` and `is_closed()` before reaching the ladder, so the recorded instant is never consumed. It is corrected here because the contract makes it free to correct, not because a reachable defect was traced to it.

One ordering fix comes with it: the stream-mode refusal stored `Closed` and notified state waiters before failing pending authentication, so a waiter could observe the terminal mode before its bookkeeping completed. The notification now follows `fail_registered_auth`.

## Testing

`complete_reconnect_transitions` in `mode.rs` is a four-case table over the transition itself - `Reconnect` yields `Reconnected` and leaves the mode `Active`; `Disconnect`, `Closed` and `Active` each yield `Aborted` and leave the mode untouched. It needs no sockets and no clock, which is why the compare-exchange branch is pinned there rather than through a client: driving that branch through a real connection means racing a store against the reconnect path's internal delay, and a test that depends on winning a scheduling window is a flake waiting to happen on a loaded CI runner.

The per-client tests cover the remaining paths against real loopback servers, deterministically - each waits on a completion signal rather than a duration. `test_reconnect_outcome_is_aborted_before_dial` and `test_reconnect_outcome_is_reconnected` in the raw socket client; `test_reconnect_outcome_is_aborted_before_dial`, `test_stream_reconnect_outcome_is_aborted_and_notifies_closed` and `test_reconnect_outcome_is_reconnected_with_handlers` in the websocket client, the last asserting that a replacement connection's traffic reaches both the message and epoch handlers and that the mode is `Active`.

Two limitations are worth stating rather than leaving to be discovered. First, these tests pin the abort-versus-success contract at the `reconnect()` boundary; they are not a red-to-green differential, because the assertions name a type the previous code did not have, so they cannot be run against it. Second, the controller-loop consequences are not covered by an automated test - that an aborted attempt no longer records a reconnect timestamp, no longer notifies waiters, and no longer reaches the callback guard. Reaching those from a test needs either a controller injection seam or capture of a debug line through the process-global logger, and neither seemed proportionate to the defect; the existing turmoil reconnect tests continue to cover the success-path controller effects, including `RECONNECTED` delivery and `post_reconnection`.

`cargo test -p nautilus-network` passes, as does the crate under `--cfg madsim`.

### DIFF
--- a/crates/network/src/mode.rs
+++ b/crates/network/src/mode.rs
@@ -49,6 +49,20 @@ impl ReadSessionFence {
     }
 }
 
+/// Result of a reconnection attempt that did not fail outright.
+///
+/// A reconnect can finish without reconnecting: a teardown may be requested while
+/// it is in flight, at which point it unwinds and leaves the mode terminal. That is
+/// normal control flow rather than an error, so it needs to be distinguishable from
+/// a completed reconnection by the caller.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum ReconnectOutcome {
+    /// A replacement connection was established and the mode is now `Active`.
+    Reconnected,
+    /// The attempt unwound without reconnecting; the mode was left unchanged.
+    Aborted,
+}
+
 /// The lifecycle state of a socket client.
 ///
 /// Clients store the active, reconnecting, disconnecting, or closed state in an atomic flag so
@@ -127,6 +141,30 @@ impl ConnectionMode {
             .is_ok()
     }
 
+    /// Atomically completes a reconnection by transitioning `Reconnect` to `Active`.
+    ///
+    /// Returns [`ReconnectOutcome::Reconnected`] if this call performed the
+    /// transition, and [`ReconnectOutcome::Aborted`] if the mode had already moved
+    /// on - which a concurrent [`Self::request_disconnect`] or a terminal `Closed`
+    /// store can do while the reconnect is in flight. Aborting here is not an
+    /// error: it means a teardown won the race and the replacement connection must
+    /// not be adopted.
+    pub(crate) fn complete_reconnect(value: &AtomicU8) -> ReconnectOutcome {
+        if value
+            .compare_exchange(
+                Self::Reconnect.as_u8(),
+                Self::Active.as_u8(),
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            )
+            .is_ok()
+        {
+            ReconnectOutcome::Reconnected
+        } else {
+            ReconnectOutcome::Aborted
+        }
+    }
+
     /// Converts a [`ConnectionMode`] to its `u8` representation.
     #[inline]
     #[must_use]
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hey @folknor,

Thank you for the PR. I've reviewed the change and it looks good: extracting the final transition into `complete_reconnect` keeps the compare-exchange semantics intact, and every abort path leaves a terminal mode that's already notified at its source, so dropping the notify-on-abort can't strand waiters.

The branch now conflicts with `develop`, so it just needs a rebase before I can merge it. Are you able to rebase when you get a chance?

folknor:
@cjdsellers Done. The only conflicts were in socket/client.rs and websocket/client.rs, where develop had raised the "Reconnecting" / "Reconnect succeeded" log lines from debug to info.

cjdsellers (approved):

cjdsellers:
Thanks @folknor, it's nice when implicitness becomes useful explicitness 😄

PR #4638, opened 2026-08-03
Add opt-in Polymarket live book resync

### LINKED ISSUE #4635: Polymarket book snapshot request does not resync managed live book
## Problem

`request_book_snapshot` on the Polymarket data client returns historical book data, but it does not refresh the adapter's local order book or the managed Cache book.

This leaves no native recovery path when a managed L2 book is present but stale and a WebSocket unsubscribe/resubscribe does not produce a new snapshot.

## Reproduction evidence

An isolated public-data actor subscribed to managed L2 deltas and then called `request_book_snapshot` while recording historical callbacks, live callbacks, Cache update count, and book contents.

- `1.231.0a20260730`: no managed Cache refresh.
- `1.231.0a20260731` (after the Polymarket snapshot request fix): no managed Cache refresh.
- Normal live deltas continued, so method availability and a successful REST response were not treated as proof of recovery.

## Proposed behavior

Preserve the existing history-only behavior by default. When the request contains `params={"resync_live_book": true}`:

1. Coalesce concurrent requests per instrument.
2. Buffer that instrument's WebSocket deltas while the REST book is in flight.
3. Replace the adapter-local book and publish a full `OrderBookDeltas` snapshot so the managed Cache is refreshed.
4. Replay buffered deltas newer than the REST snapshot.
5. On HTTP failure, publish no snapshot and replay all buffered deltas against the old live book.
6. Let unsubscribe/reset and the existing tick-size epoch gate cancel the live resync.

The request should still return the normal historical response. No Python-side book store or direct Cache mutation is proposed.

6Kmfi6HP:
Closes #4635.

Adds an opt-in resync_live_book request parameter to Polymarket book snapshot requests. Standard requests retain historical-only behavior. Live resync requests buffer per-instrument WebSocket deltas, replace the adapter book from REST, publish an F_SNAPSHOT delta batch, and replay newer deltas in order.

Failure keeps the old book and replays all buffered deltas. Concurrent requests coalesce per instrument, with generation-safe response admission and buffer handoff. Tick-size epoch recovery remains authoritative.

Validation:
- cargo test -p nautilus-polymarket --lib --quiet: 816 passed
- cargo test -p nautilus-polymarket --test data_client --quiet: 16 passed
- cargo clippy -p nautilus-polymarket --all-targets -- -D warnings
- release wheel short public-data probe: historical response followed by matching F_SNAPSHOT and continuous ordinary deltas
- release wheel 60-minute public-data sandbox: 1,361 live callbacks, 0 failures, 0 Cache-count regressions, 0 ts_event regressions, normal exit 0

### DIFF
--- a/crates/adapters/polymarket/src/data/auto_load.rs
+++ b/crates/adapters/polymarket/src/data/auto_load.rs
@@ -110,6 +110,7 @@ impl PolymarketDataClient {
         let last_quotes = self.last_quotes.clone();
         let resolve_poll_watchlist = self.resolve_poll_watchlist.clone();
         let pending_snapshot_after_tick_change = self.pending_snapshot_after_tick_change.clone();
+        let live_book_resyncs = self.live_book_resyncs.clone();
 
         get_runtime().spawn(async move {
             // Coalesce concurrent misses into one Gamma call.
@@ -230,6 +231,7 @@ impl PolymarketDataClient {
                                 &active_trade_subs,
                                 &resolve_poll_watchlist,
                                 &pending_snapshot_after_tick_change,
+                                &live_book_resyncs,
                                 &pending,
                                 &ws_open_tokens,
                                 &ws_sub_mutex,
--- a/crates/adapters/polymarket/src/data/dispatch.rs
+++ b/crates/adapters/polymarket/src/data/dispatch.rs
@@ -42,7 +42,8 @@ use tokio_util::sync::CancellationToken;
 use ustr::Ustr;
 
 use super::{
-    NEW_MARKET_EMPTY_RECHECK_DELAY, NEW_MARKET_EMPTY_RECHECK_MAX_ATTEMPTS,
+    LiveBookResync, LiveBookResyncPhase, NEW_MARKET_EMPTY_RECHECK_DELAY,
+    NEW_MARKET_EMPTY_RECHECK_MAX_ATTEMPTS,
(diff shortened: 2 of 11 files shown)


6Kmfi6HP:
Final release-wheel public-data sandbox completed successfully (60 minutes, read-only).

- 1,366 recorded events; 1,361 live book callbacks; 0 request failures
- historical response arrived 38.52 ms after dispatch
- matching live F_SNAPSHOT arrived 39.01 ms after dispatch, 0.50 ms after the historical callback
- historical and live snapshot ts_event matched exactly: 1785738186504000000
- managed Cache update count moved atomically from 198 to 294, then ordinary deltas continued at 295, 296, 297, ... 3630
- Cache update-count regressions: 0
- live ts_event regressions: 0
- process restarts: 0; normal unsubscribe/disconnect and exit 0
- no out-of-order warnings observed

The tested artifact was a CPython 3.12 release wheel built from commit 078ff03.

cjdsellers:
Hi @6Kmfi6HP,

Following up on #4635, `request_book_snapshot` needs to remain a historical request without side effects on the adapter's live order book or managed `Cache` book. That means the `resync_live_book` parameter in this PR isn't the API direction we want to take.

I'm closing this in favor of #4643, which tracks explicit typed reconnect and order book resync commands through the engine. The snapshot and delta ordering work here remains useful context for that design.

PR #4639, opened 2026-08-03
Replace Chrono with Jiff

sunlei:
# Pull Request

## Summary

This replaces Chrono and Chrono-TZ throughout first-party Rust code and direct dependencies with [Jiff](https://github.com/BurntSushi/jiff), while preserving existing parsing, formatting, DST, wire-format, and Python datetime contracts. The migration follows Chrono's planned maintenance wind-down and removes the direct business risk of relying on a bundled time zone database that will no longer receive updates. First-party lookups use Jiff's bundled TZDB through a shared policy point for deterministic behavior across hosts and low-latency execution.

## Why migrate

In [chronotope/chrono#1768](https://github.com/chronotope/chrono/issues/1768), the Chrono maintainer explains that maintenance of `chrono` and `chrono-tz` is expected to wind down. The stated reason is not a newly disclosed or known unpatched security vulnerability: the APIs are dated, revising them would require substantial effort, and Jiff is the alternative the maintainer is comfortable recommending.

The most direct risk for NautilusTrader is time zone data freshness. On 2026-08-02, the maintainer [explicitly confirmed](https://github.com/chronotope/chrono/issues/1768#issuecomment-5156560960) that `chrono-tz` will no longer be updated for IANA TZDB changes.

| Dependency | Version | Bundled IANA TZDB | Update status |
| ---------- | ------- | ----------------- | ------------- |
| Existing `chrono-tz` | 0.10.4 | 2025b | No further TZDB updates planned |
| Replacement Jiff | 0.2.35 | 2026c | Updated in [Jiff 0.2.32](https://github.com/BurntSushi/jiff/blob/master/CHANGELOG.md#0232-2026-07-08) |

Stale rules can affect session boundaries, FX rollover dates, instrument expirations, and other civil-time calculations. As of this PR, RustSec has not classified Chrono as unmaintained; [the question raised in the upstream issue](https://github.com/chronotope/chrono/issues/1768#issuecomment-5157820360) has not received a maintainer response. This PR therefore treats the migration as proactive maintenance and data-correctness work, not a security advisory response.

## Related issues/PRs

- Upstream context: [chronotope/chrono#1768](https://github.com/chronotope/chrono/issues/1768)
- Replacement library: [BurntSushi/jiff](https://github.com/BurntSushi/jiff)

## Type of change

- [ ] Bug fix (non-breaking)
- [x] Documentation update
- [x] Maintenance / chore

## Implementation

- Replace direct `chrono` and `chrono-tz` dependencies and migrate timestamp parsing, formatting, arithmetic, time zone conversion, and test fixtures across the Rust workspace and adapters.
- Preserve established RFC 3339 input and output contracts, including Chrono-compatible fractional-second widths and accepted date forms.
- Preserve existing DST fold/gap handling and add Python boundary coverage for UTC, fixed-offset, `ZoneInfo`, and naive datetimes.
- Centralize first-party time zone lookup on the bundled Jiff database and cache static zones used on hot paths.
- Update PyO3 and `pyo3-stub-gen` type mappings while retaining `datetime.datetime` Python signatures.
- Add pre-commit guards for Jiff feature resolution, dependency declarations, and direct wall-clock access.

## Performance

The final implementation caches the Eastern `TimeZone` handle used by the FX rollover event path instead of performing a database lookup for every event.

| Path | `develop` Chrono-TZ | This PR Jiff | Latency reduction | Speedup |
| ---- | -------------------: | -----------: | ----------------: | ------: |
| FX rollover Eastern civil-date conversion | 17.5 ns/op | 12.1 ns/op | 5.4 ns/op (30.9%) | 1.45x |

Benchmark: release build on macOS, warm caches, 200k-500k iterations. Absolute timings vary by machine; the relative comparison is the useful signal.

## Documentation

## Testing

Validation completed:

- `cargo check --locked --workspace --all-targets`
- `cargo check --locked -p pyo3-stub-gen`
- Targeted tests for `nautilus-core`, `nautilus-model`, `nautilus-trading`, `nautilus-common`, `nautilus-persistence`, `nautilus-backtest`, and `nautilus-databento`: 5,643 passed
- `cargo test -p nautilus-core --features python python::datetime`: 6 passed
- Workspace Clippy with warnings denied
- Cargo, DST, Jiff feature-policy, and documentation pre-commit hooks
- Residual search confirms no Chrono or Chrono-TZ references in first-party Rust code or direct Cargo dependencies

<details>
<summary>Time zone strategy, experiments, and rejected alternatives</summary>

The existing `chrono-tz` behavior was already hermetic: its IANA data was compiled into the binary. This PR keeps that property with `TimeZoneDatabase::bundled()` because backtests must be reproducible across hosts and latency-sensitive paths should not acquire file-system tail latency. It also keeps deployments independent of host tzdata availability, including minimal container images.

Jiff generally recommends the system zoneinfo database for applications that should receive OS tzdata updates without rebuilding. That option was not selected here because it makes results depend on host state and introduces five-minute TTL revalidation with `stat` calls and potential directory rescans. PyO3 currently enables Jiff's default/system features transitively; first-party lookups still explicitly use the bundled policy point, and the new feature guard makes upstream feature changes visible in CI.

| Experiment | Result | Comparison |
| ---------- | -----: | ---------- |
| Fresh bundled database handle and lookup | 307 ns/op | Baseline |
| Fresh environment/system database and lookup | 1,062,388.5 ns/op | ~3,460x slower cold path |
| Warm bundled lookup | 27.2 ns/op | Baseline |
| Warm system zoneinfo lookup | 33.1 ns/op | 1.22x slower |
| Per-event bundled lookup plus conversion | 42.3 ns/op | Intermediate implementation |
| Cached Jiff zone plus conversion | 12.1 ns/op | 3.50x faster than intermediate implementation |

</details>


cjdsellers (approved):
Hi @sunlei,

Thank you for the PR. This is a substantial contribution and the care shows, the write-up alone made this much easier to review 🙏.

We agree this is the right call. With chrono maintenance winding down and chrono-tz no longer receiving IANA updates, tz data staleness would force this eventually anyway. Jiff is also an easy successor to trust given BurntSushi's track record.

On the implementation, this covers everything we'd want from a first pass. The parsing and formatting contracts are pinned to the legacy behavior rather than naively swapped, the DST gap/fold disambiguation maps chrono's semantics exactly, and the Python boundary keeps `datetime.datetime` with the aware/naive handling preserved. I particularly like that the bundled tzdb policy point and the new pre-commit guards turn the migration decisions into CI-enforced invariants rather than conventions, and the cached zone lookup coming out ahead of chrono-tz on the FX rollover path is a nice result.

The diff is enormous at 219 changed files, but the bulk is mechanical substitution, and the current high-churn v2 transition is the best window we'll get to land something like this. I'm happy to merge this.

I'll pick up a few follow-ups post-merge, none blocking:

- `chrono-tz` remains in the tree transitively via Arrow/DataFusion, so catalog SQL temporal functions still use the frozen 2025b tz data until upstream migrates. We'll document this and track it.
- Add a test driving both the Rust and Python feather writers across a DST transition, since the scheduled rotation logic has to agree across the two implementations.
- A rolling-upgrade check that state persisted by the previous build still reads cleanly.
- An ingest-path benchmark for timestamp parsing on a high-rate feed, since the PR benchmarks the rollover path only.

Thanks again for taking this on.

PR #4640, opened 2026-08-03
Fix FeeModel subclassing and inheritance

dfjmax:

## Summary

The base `FeeModel.__new__` accepted no arguments, so Python subclasses defining their own `__init__` raised `TypeError`, and the concrete fee models did not inherit `FeeModel`, breaking `isinstance` parity with the v1 Cython models. This widens the base constructor to forward `*args/**kwargs`, makes each concrete `extends FeeModel`, and hardens the boundary by validating `order`/`instrument` and using checked `Decimal` arithmetic so extreme inputs raise instead of panic.

## Related issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)

## Documentation

## Testing

Added Python tests in `tests/unit/execution/test_models.py` (subclassing with `__init__` args, `isinstance` inheritance, base `NotImplementedError`, boundary rejection) and Rust overflow tests in `crates/execution/src/models/fee.rs`.


cjdsellers:
Hi @dfjmax,

Thanks for the PR. The constructor forwarding and `isinstance` parity both look good, but there is one behavior gap before this can merge.

Now that the concrete models extend `FeeModel`, they inherit the base Python `get_commission`, which always raises `NotImplementedError`. None of the concrete `#[pymethods]` blocks exposes its Rust `FeeModel::get_commission` implementation, so calls such as `FixedFeeModel(...).get_commission(...)` fail even though the generated stub now advertises them. The v1 concrete models implemented this method directly, so the inheritance change should preserve that behavior too.

Could you bind `get_commission` for the concrete models and add representative Python coverage for direct calls?

One smaller test issue: `test_per_contract_fee_model_extreme_inputs_return_error` does not currently reach `mul_checked`. The multiplication remains within the `Decimal` range and the test passes later because `Money` rejects the amount. Please make it force a decimal overflow and assert `commission calculation overflow` so it proves the new path.

dfjmax:
Thanks for the review @cjdsellers.

I have addressed both points:

1. get_commission binding: bound it on all six concrete models (FixedFeeModel, MakerTakerFeeModel, PerContractFeeModel, ProbabilityPriceFeeModel, CappedOptionFeeModel, TieredNotionalOptionFeeModel) so they no 
longer fall through to the base NotImplementedError. Added Python direct-call coverage in test_models.py (success paths for Fixed/PerContract and representative error paths for the rest).

2. Reworked test_per_contract_fee_model_decimal_overflow_returns_error to force a real Decimal overflow (large per-contract commission × fill qty × a high spread leg-count) so it now reaches mul_checked and asserts "commission calculation overflow" rather than passing via the later Money rejection.

Also added a small test for the ProbabilityPriceFeeModel fill-price [0, 1] validation branch.

cjdsellers (approved):
Thank you for the follow-ups @dfjmax. This looks good to merge now 👌

Issue #4633, opened 2026-08-02
SIGSEGV on macOS arm64 when pyarrow/pandas is imported before nautilus_trader


ZhongxuanWang:
## Summary

On macOS arm64, `2.0.0rc2` segfaults when `pyarrow` (or `pandas`, which imports
it) is loaded **before** `nautilus_trader`. `2.0.0rc1` is clean under the same
conditions, so this looks like an rc1 → rc2 regression.

There are two symptoms. The second is the concerning one: the crash happens
*during a normal call*, not at interpreter teardown.

## Environment

| | |
|---|---|
| nautilus_trader | `2.0.0rc2` (`cp312-cp312-macosx_11_0_arm64` wheel from PyPI) |
| OS | macOS 26.5.2, arm64 (Apple silicon) |
| Python | 3.12.10 |
| pyarrow | 24.0.0 |
| pandas | 2.3.3 |

## Repro A — SIGSEGV at interpreter exit

```python
import pyarrow
import nautilus_trader
print("reached end")
```

```
reached end
[1]    ... segmentation fault
```

Exit code 139. The script completes, then the process dies during shutdown.

## Repro B — SIGSEGV during the call

```python
print("1 start", flush=True)
import nautilus_trader
print("2 nautilus top imported", flush=True)
import pandas
print("3 pandas imported", flush=True)
from nautilus_trader.model import Currency
print("4 model imported", flush=True)
c = Currency.from_str("USDC")
print("5 currency constructed:", c, flush=True)   # never reached
```

Output stops after `4 model imported`; `Currency.from_str` never returns.
Deterministic across runs (3/3).

## Works fine — reverse order

```python
import nautilus_trader
import pyarrow          # fine

from nautilus_trader.model import Currency
print(Currency.from_str("USDC"))   # fine
import pandas                       # fine
```

Importing `nautilus_trader.model` before pandas is **not** sufficient on its
own — what matters is that the currency registry is initialised before
pyarrow's Arrow runtime is loaded.

## Regression check against rc1

Same machine, same test, isolated environments differing only in the pinned
version:

| | `2.0.0rc1` | `2.0.0rc2` |
|---|---|---|
| Repro A | exit 0 | **SIGSEGV** |
| Repro B | exit 0 | **SIGSEGV** |
| Our package test suite (776 tests) | exit 0 | all tests pass, then **exit 139** |

## Platform scope

macOS arm64 reproduces consistently. **Linux x86_64 does not** — the same
[...]

PR #4641, opened 2026-08-04
fix: hide cdylib symbols on macOS to prevent pyarrow SIGSEGV (#4633)

alex09x:
## Problem
Issue #4633 reports a SIGSEGV when pyarrow or pandas is imported before nautilus_trader on macOS arm64.
This is caused by symbol collisions between statically linked Arrow in `_libnautilus` and pyarrow's Arrow runtime. By default on macOS, rustc exports all symbols in a `cdylib` if not explicitly hidden, meaning `libarrow` symbols leak into the flat namespace and interpose with `pyarrow`.

## Solution
Added a build script `crates/pyo3/build.rs` to restrict the export table of the `cdylib` to only the module initializer (`_PyInit__libnautilus` or `_PyInit_nautilus_pyo3` depending on the cython-compat feature). This is done using `cargo::rustc-link-arg-cdylib=-Wl,-exported_symbol,...` on macOS.
This ensures statically linked Arrow symbols become non-external and can no longer collide with pyarrow/pandas' Arrow runtime.

## Verification
- Statically linked Arrow symbols are now properly hidden in the shared object.
- `import pyarrow; import pandas; import nautilus_trader` executes cleanly on macOS arm64 without segfaulting.

### DIFF
--- a/crates/pyo3/Cargo.toml
+++ b/crates/pyo3/Cargo.toml
@@ -13,6 +13,7 @@ documentation.workspace = true
 repository.workspace = true
 homepage.workspace = true
 publish = false # Never publish to crates.io
+build = "build.rs" # Hides the cdylib export table on macOS (see issue #4633)
 
 [lints]
 workspace = true
--- a/crates/pyo3/build.rs
+++ b/crates/pyo3/build.rs
@@ -1,0 +1,49 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2026 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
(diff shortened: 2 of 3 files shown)


cjdsellers:
Hi @alex09x,

Thanks for tracking down the root cause and submitting the PR. I’ve implemented the same core linker fix locally, with coverage for the reported PyArrow and pandas import order.

I left out the `cython-compat` handling because it reintroduces a redundant compatibility concept into the v2 PyO3 module. Since the CLA also remains unsigned, it was cleaner to take the narrower fix on locally rather than merge this PR as-is.

I’ve credited you in the release notes, along with @ZhongxuanWang for the original report.

Separately, no final decision has been made on pandas versus Polars for v2. Both will remain possible, although we’re currently leaning toward first-class Polars support.

PR #4642, opened 2026-08-04
Fix cdylib symbols on macOS to prevent pyarrow SIGSEGV

alex09x:
## Problem
Issue #4633 reports a SIGSEGV when pyarrow or pandas is imported before nautilus_trader on macOS arm64.
This is caused by symbol collisions between statically linked Arrow in `_libnautilus` and pyarrow's Arrow runtime. By default on macOS, rustc exports all symbols in a `cdylib` if not explicitly hidden, meaning `libarrow` symbols leak into the flat namespace and interpose with `pyarrow`.

## Solution
Added a build script `crates/pyo3/build.rs` to restrict the export table of the `cdylib` to only the module initializer (`_PyInit__libnautilus` or `_PyInit_nautilus_pyo3` depending on the cython-compat feature). This is done using `cargo::rustc-link-arg-cdylib=-Wl,-exported_symbol,...` on macOS.
This ensures statically linked Arrow symbols become non-external and can no longer collide with pyarrow/pandas' Arrow runtime.

## Verification
- Statically linked Arrow symbols are now properly hidden in the shared object.
- `import pyarrow; import pandas; import nautilus_trader` executes cleanly on macOS arm64 without segfaulting.

### DIFF
--- a/crates/pyo3/Cargo.toml
+++ b/crates/pyo3/Cargo.toml
@@ -13,6 +13,7 @@ documentation.workspace = true
 repository.workspace = true
 homepage.workspace = true
 publish = false # Never publish to crates.io
+build = "build.rs" # Hides the cdylib export table on macOS (see issue #4633)
 
 [lints]
 workspace = true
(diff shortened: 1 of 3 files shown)


cjdsellers:
Thanks again @alex09x,

Closing as fixed from https://github.com/nautechsystems/nautilus_trader/commit/457bd66b2b8645f08b24e9155ac576c9499aa8fe.

Issue #4086, opened 2026-05-17
Polymarket bulk Gamma API instrument loading is broken


haimgel:
Gamma silently caps responses at 100 regardless of the requested limit. Polymarket's own docs confirm 100 [as the per-page maximum](https://docs.polymarket.com/api-reference/markets/list-markets-keyset-pagination) -- it's for the new keyset pagination endpoint, but the old one now behaves the same.

NautilusTrader requests pages of 500 and stops when less than that items are returned. So it always stops after the first page, does not load all markets. Affects both the Python adapter (`nautilus_trader/adapters/polymarket/`) and the new Rust crate (`crates/adapters/polymarket/`).

### Reproduction

  Configure `PolymarketInstrumentProviderConfig(load_ids=frozenset(<201 instrument IDs across 186 condition_ids>), use_gamma_markets=True)` on a `PolymarketLiveDataClient` and start the
  node:

```
  [INFO] PolymarketInstrumentProvider: Loading 201 instruments from 186 markets, using bulk load of all markets
  [INFO] PolymarketInstrumentProvider: Loaded 100 markets using Gamma API
  [WARN] PolymarketInstrumentProvider: No instruments were loaded, verify config if this is unexpected
```

### Fundamental problem with Gamma /markets API bulk-loading

Gamma API not only enforces maximum 100 items limit, it also refuses offsets over 10,000. I'm getting this:
```
  GET /markets?closed=false&limit=100&offset=10100
  422 {"type":"validation error","error":"offset exceeds maximum allowed for markets list queries"}
```

So an instrument provider that relies on bulk-walking /markets cannot in principle see Polymarket's full market set (which is well over 10k). The "fall back to bulk load" branch is not going to work in principle.

cjdsellers:
Hi @haimgel,

Thanks for the report, and for tracing both the 100-cap and the 10k-offset issues.

Just landed the fix on `develop`. The Gamma `/markets` paginator now uses 100-per-page in both Python and Rust, matching the server's silent cap. The >100 condition_ids path in `load_ids` no longer bulk-walks; it chunks into batches of 100 `condition_ids=` queries and unions the results, with any caller filters (e.g. `active`, `closed`) preserved across each chunk. Regression tests cover pagination, chunking, and filter preservation on both code paths.

[...]

PR #4650, opened 2026-08-04
Add Polymarket instrument provider series_ids scoping

### LINKED ISSUE #4651: [Polymarket] [V2]  Instrument provider cannot scope universes by Gamma series
# Feature Request

## Problem statement

`PolymarketInstrumentProviderConfig` cannot express a universe that combines recurring
market families with tag- or id-based selection, even though the Gamma API and the
adapter's own HTTP layer both support it.

Three limitations interact:

1. `event_slug_builder` (added in #3501) accepts a single `PolymarketUpDownEventSlugConfig`
   with a scalar `interval_mins`. Subscribing btc and bnb up/down markets at both 5m and
   15m (four recurring families) is inexpressible in one provider.
2. Scopes are silently mutually exclusive. When any slug scope is set (builder,
   `event_slugs`, `market_slugs`), `initialize()` returns before the `filters` branch runs
   and `should_load_all()` makes the `load_ids` branch unreachable (`providers.rs`). A
   config with `event_slug_builder` + `filters={"tag_id": "84"}` + `load_ids=[...]`
   validates, then loads only the slug component. Nothing warns that the other scopes
   were dropped.
3. The slug builder reconstructs event slugs from local clock arithmetic
   (`{asset}-updown-{interval_mins}m-{period_start}`). This duplicates, on the client,
   grouping that Gamma already maintains server-side: recurring markets belong to a
   series (event `btc-updown-5m-1785856500` → series `btc-up-or-down-5m`, id 10684,
   recurrence `5m`; `highest-temperature-in-hong-kong-on-august-5-2026` → series
   `hong-kong-daily-weather`, id 11312, recurrence `daily`). Any change to Polymarket's
   slug convention or period alignment silently breaks the builder, while series
   membership remains authoritative.

The plumbing already exists: `GetGammaEventsParams` has `series_id: Option<Vec<u64>>`
(`crates/adapters/polymarket/src/http/query.rs`), and
`GET /events?series_id=...&active=true&closed=false` returns the live events of a series
(verified against the live API). The instrument provider simply has no path that uses it.

Note this is unrelated to the bulk Gamma loading limits that led to #4086: a
series-scoped events query returns a bounded set (the active periods of a family ;
tens to a few hundred events), never a 150k-market walk, so the 100-per-page /
10k-offset ceilings are not in play.

## Proposed solution

1. Add `series_ids: Option<Vec<u64>>` to `PolymarketInstrumentProviderConfig` (Rust +
   Python binding). During `initialize()`, fetch events via the existing Gamma events
   query with `series_id`, `active=true`, `closed=false` (optionally `end_date_min=now`),
   and load all markets from the returned events — the same hydration path as the
   existing event-slug branch. In the Rust filter design from #3708 this is naturally a
   `SeriesFilter` alongside `EventSlugFilter`/`TagFilter`.
2. Make provider scopes compose as a union instead of early-returning: `series_ids`,
   slug scopes, `filters`, and `load_ids` each contribute markets, deduplicated by
   condition id. If union semantics for the legacy scopes are considered a behavior
   change, at minimum `series_ids` should compose with `filters` and `load_ids`, and the
   provider should log a warning when configured scopes are ignored rather than dropping
   them silently.

Because `update_instruments_interval_mins` re-runs the provider, series scoping makes
rolling families self-maintaining: each refresh picks up newly created periods with no
client-side slug math. It could also serve the pending Polymarket phase of the rolling
binary-option scope work (#4224 / #4227) as the discovery mechanism for scope rotation —
a series is exactly the venue-side identity of a rolling scope.

## Example usage

```python
from nautilus_trader.adapters.polymarket import (
    PolymarketDataClientConfig,
    PolymarketInstrumentProviderConfig,
)

instrument_config = PolymarketInstrumentProviderConfig(
    series_ids=[
        10684,  # btc-up-or-down-5m
        10192,  # btc-up-or-down-15m
        11326,  # bnb-up-or-down-5m
        11330,  # bnb-up-or-down-15m
        11312,  # hong-kong-daily-weather
    ],
    filters={"tag_id": "84", "active": "true", "closed": "false"},  # one-off weather markets
)

config = PolymarketDataClientConfig(
    instrument_config=instrument_config,
    update_instruments_interval_mins=5,  # refresh brings in new periods automatically
    subscribe_new_markets=True,
)
```

## Alternatives considered

- **Multiple data clients on one node**, one per scope: works today, but multiplies WS
  connections and client ids for what is logically one venue subscription, and still
  cannot express two up/down intervals per client.
- **Generating `event_slugs` externally and reloading**: reimplements Gamma's series
  grouping on the client with clock arithmetic, and still silently disables `filters` /
  `load_ids` due to the early-return.
- **Extending `event_slug_builder` to a list of builders**: fixes only limitation 1;
  keeps the fragile slug reconstruction and the silent scope exclusivity.
- **Waiting for rolling binary-option scopes (#4224 phase 2)**: solves strategy-side
  rotation rebinding, not provider-side universe selection (e.g. for recording), and its
  Polymarket scope manager will itself need a discovery mechanism — which series ids
  provide.

Series scoping is preferred because it delegates market-family membership to the venue's
own authoritative grouping, composes with existing scopes, and reuses an HTTP-layer
capability the adapter already ships.

## Additional context

- Adapter version inspected: 2.0.0rc3 (`crates/adapters/polymarket`).
- Relevant code: `providers.rs` (`initialize`, `should_load_all`), `http/query.rs`
  (`GetGammaEventsParams.series_id`), `python/mod.rs` (callable builder rejection).
- Related work: #3501 (event_slug_builder), #3708 (Rust `InstrumentFilter` design),
  #4086 (bulk Gamma limits — not applicable to scoped series queries), #4224/#4227
  (rolling binary-option scope RFC + generic plumbing).
- Live API verification: `GET /events?slug=...` responses embed the owning series;
  `GET /events?series_id=10684&active=true&closed=false` returns the series' events.
  Series docs: https://docs.polymarket.com/api-reference/series/list-series

mystic-io:
Closes #4651

## Summary

Adds `series_ids: Option<Vec<u64>>` to `PolymarketInstrumentProviderConfig` (Rust + Python binding + stub). The provider resolves each series to its active, unresolved events via the existing Gamma events plumbing (`GetGammaEventsParams.series_id`, `request_instruments_by_event_params`) and loads their markets additively.

Recurring market families (e.g. `btc-up-or-down-5m` id 10684, `hong-kong-daily-weather` id 11312) can now be scoped by the venue's own authoritative series grouping instead of client-side slug arithmetic, and roll forward naturally with `update_instruments_interval_mins` — the existing `event_slug_builder` supports only a single builder with a scalar `interval_mins`, so multi-interval up/down universes were previously inexpressible in one provider.

Series-scoped events queries return a bounded set (the active periods of a family), so the bulk-Gamma pagination limits from #4086 are not in play.

## Behavior change: bootstrap scopes now compose

Previously any slug scope silently disabled the `filters` map (the `initialize()` early-return) and made `load_ids` unreachable — a config combining them validated but only loaded the slug component, with no warning. Now:

- A non-empty `filters` map fetches alongside explicit slug/series scopes; only the *unfiltered* full-universe fetch remains suppressed under explicit scoping (preserving the "explicit scoping never broadens into a full-universe fetch" invariant).
- `load_ids` loads additively with `load_all`/scoped bootstrap instead of being unreachable.

Scope predicates are hoisted onto the config (`has_explicit_scope`, `has_series_ids`, `has_nonempty_filters`) and shared by both bootstrap paths (`load_scoped_all`, `fetch_configured_instruments`), which previously each maintained their own copy.

## Example

```python
instrument_config = PolymarketInstrumentProviderConfig(
    series_ids=[10684, 10192, 11326, 11330],   # btc/bnb up-or-down 5m/15m
    filters={"tag_id": "84", "active": "true", "closed": "false"},
)
```

## Testing

- `cargo test -p nautilus-polymarket`: 1,230 tests pass, including new integration tests asserting the series query shape (`series_id`/`active`/`closed` params, single batched request) and the filters+series composition, plus config unit tests for `should_load_all`/`has_series_ids`.
- `cargo clippy -p nautilus-polymarket --all-targets`: clean.
- `cargo +nightly fmt`: applied.
- Verified against the live Gamma API that `GET /events?series_id=...&active=true&closed=false` returns the expected events and that up/down and daily-weather events embed these series ids.

### DIFF
--- a/crates/adapters/polymarket/src/config.rs
+++ b/crates/adapters/polymarket/src/config.rs
@@ -190,6 +190,8 @@ pub struct PolymarketInstrumentProviderConfig {
     pub market_slugs: Option<Vec<String>>,
     /// Optional Rust-backed Up/Down event slug builder.
     pub event_slug_builder: Option<PolymarketUpDownEventSlugConfig>,
+    /// Optional Gamma series IDs whose active events resolve to markets during bootstrap.
+    pub series_ids: Option<Vec<u64>>,
     /// Whether provider warnings should be logged.
     #[builder(default = true)]
     pub log_warnings: bool,
@@ -208,6 +210,7 @@ nautilus_core::impl_pyo3_config_getters!(PolymarketInstrumentProviderConfig {
     event_slugs: Option<Vec<String>>,
     market_slugs: Option<Vec<String>>,
     event_slug_builder: Option<PolymarketUpDownEventSlugConfig>,
+    series_ids: Option<Vec<u64>>,
     log_warnings: bool,
     use_gamma_markets: bool,
 });
@@ -226,10 +229,26 @@ impl PolymarketInstrumentProviderConfig {
 
     #[must_use]
     pub fn should_load_all(&self) -> bool {
-        self.load_all
-            || self.event_slug_builder.is_some()
+        self.load_all || self.has_explicit_scope()
+    }
+
+    /// Returns whether any explicit bootstrap scope (slug, builder, or series) is configured.
+    #[must_use]
+    pub fn has_explicit_scope(&self) -> bool {
+        self.event_slug_builder.is_some()
             || self.event_slugs.as_ref().is_some_and(|s| !s.is_empty())
             || self.market_slugs.as_ref().is_some_and(|s| !s.is_empty())
+            || self.has_series_ids()
+    }
+
+    #[must_use]
+    pub fn has_series_ids(&self) -> bool {
+        self.series_ids.as_ref().is_some_and(|ids| !ids.is_empty())
+    }
+
+    #[must_use]
+    pub fn has_nonempty_filters(&self) -> bool {
+        self.filters.as_ref().is_some_and(|map| !map.is_empty())
     }
 
     #[must_use]
@@ -676,6 +695,28 @@ mod tests {
             err.to_string()
                 .contains("event_slug_builder.interval_mins must be positive")
         );
+    }
+
+    #[rstest]
+    fn provider_config_series_ids_trigger_load_all() {
+        let config = PolymarketInstrumentProviderConfig {
+            series_ids: Some(vec![10684]),
+            ..PolymarketInstrumentProviderConfig::default()
+        };
+
+        assert!(config.has_series_ids());
+        assert!(config.should_load_all());
+    }
+
+    #[rstest]
+    fn provider_config_empty_series_ids_do_not_trigger_load_all() {
+        let config = PolymarketInstrumentProviderConfig {
+            series_ids: Some(Vec::new()),
+            ..PolymarketInstrumentProviderConfig::default()
+        };
+
+        assert!(!config.has_series_ids());
+        assert!(!config.should_load_all());
     }
 
     #[rstest]
--- a/crates/adapters/polymarket/src/providers.rs
+++ b/crates/adapters/polymarket/src/providers.rs
@@ -215,46 +215,67 @@ impl PolymarketInstrumentProvider {
         Ok(())
     }
 
+    /// Loads instruments for the given Gamma series IDs additively into the store.
+    ///
+    /// Resolves each series to its active, unresolved events and loads their
+    /// markets. Unlike [`Self::load_all`], this does **not** clear existing
+    /// instruments or mark the store as initialized.
+    pub async fn load_by_series_ids(&mut self, series_ids: Vec<u64>) -> anyhow::Result<()> {
+        let instruments = self
+            .http_client
+            .request_instruments_by_event_params(series_events_params(series_ids))
+            .await?;
+        self.add_instruments(instruments);
+        Ok(())
+    }
+
     /// Initializes the provider using its configured bootstrap scope.
     pub async fn initialize(&mut self, reload: bool) -> anyhow::Result<()> {
         if self.store.is_initialized() && !reload {
             return Ok(());
         }
 
-        if self.config.should_load_all() {
-            self.load_scoped_all().await?;
-            self.store.set_initialized();
+        let should_load_all = self.config.should_load_all();
+        let has_load_ids = self.config.has_load_ids();
+
+        if !should_load_all && !has_load_ids {
+            if self.config.log_warnings {
+                log::warn!(
+                    "No Polymarket instrument bootstrap configured: set instrument_config.load_all, instrument_config.load_ids, instrument_config.event_slugs, instrument_config.market_slugs, instrument_config.event_slug_builder, or instrument_config.series_ids"
+                );
+            }
             return Ok(());
         }
 
-        if self.config.has_load_ids() {
-            let load_ids = self.config.load_ids.clone().unwrap_or_default();
-            let filters = self.config.filters.clone();
-            self.load_ids(&load_ids, filters.as_ref()).await?;
-            self.store.set_initialized();
-            return Ok(());
-        }
-
-        if self.config.log_warnings {
+        // Registered `InstrumentFilter`s take precedence over the Gamma filter map,
+        // so say when the map is being dropped rather than ignoring it silently.
+        if self.config.log_warnings
+            && !self.filters.is_empty()
+            && self.config.has_nonempty_filters()
+        {
             log::warn!(
-                "No Polymarket instrument bootstrap configured: set instrument_config.load_all, instrument_config.load_ids, instrument_config.event_slugs, instrument_config.market_slugs, or instrument_config.event_slug_builder"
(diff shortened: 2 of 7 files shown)


cjdsellers:
Hi @mystic-io,

Thanks for the PR. The series query itself looks sound, but I found a few issues that need addressing before merge:

- Append `series_ids` to the existing PyO3 constructor instead of inserting it before `log_warnings`; existing positional callers currently break.
- In `fetch_configured_instruments`, run configured `InstrumentFilter` queries alongside explicit series and slug scopes. Refreshes and `request_instruments` currently drop those filter-selected markets.
- Add focused coverage for `fetch_configured_instruments` with `series_ids`, non-empty filters, and `load_ids`, since this is the separate path used by interval refreshes.
- Avoid marking the store initialized until every additive bootstrap scope succeeds. A later series, slug, or ID request failure currently makes `initialize(false)` skip the missing scopes on retry.
- Document `series_ids` in the Polymarket instrument provider options and runtime preload guidance.

Once these are addressed and the build workflow has run, this should be ready for another pass.

mystic-io:
Hey @cjdsellers ,

Thanks for the review! all five are done and pushed.

- `series_ids` is appended at the end of the constructor now, so existing positional callers are fine. `fetch_configured_instruments` runs the configured filter queries alongside the explicit series and slug scopes, with focused tests covering series_ids, non-empty filters and load_ids. The initialized flag was the interesting one  `load_all` sets it internally, so the scoped bootstrap no longer routes through it and the store is only marked once every additive scope has succeeded. Docs are updated in the provider options and the preload guidance.
- Two extras while I was in there. `load_scoped_all` had the same filter-dropping problem you spotted in the fetch path. it was only checking the config filter map, not registered `InstrumentFilter`, so bootstrap and the interval refresh now agree. And the `load` fallback no longer broadens a scoped provider into a full-universe fetch, which would have quietly undone the initialize fix.

Thanks again!

cjdsellers:
Hi @mystic-io,

Thanks for the follow-ups. Most of the original concerns are addressed, but two scope composition issues remain:

- `load_ids` is still intersected with `config.filters` during bootstrap and refresh. Explicit IDs outside those filters will not load, despite the documented additive behavior. Please query `load_ids` using only the condition IDs and add coverage for `filters + load_ids`.
- Registered `InstrumentFilter`s still behave differently between bootstrap and refresh. `load_scoped_all` adds explicit scopes without applying `accept`, while `fetch_configured_instruments` applies the final filter to every collected instrument. A refresh can therefore remove series, slug, or ID instruments loaded during bootstrap. Please keep filter acceptance limited to results from queries driven by filters and add a test where an explicit series instrument does not match the registered filter.

### Workflow suggestions

The `pre-commit` job also currently fails on the em dash in `docs/integrations/polymarket.md` and the fully qualified `nautilus_polymarket::filters::InstrumentFilter` in `crates/adapters/polymarket/tests/http.rs`.

For the next update, follow the [`CONTRIBUTING.md` setup](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md#set-up-your-environment): run `prek install` once to install the Git hooks, then make sure `make pre-commit` (`prek run --all-files`) passes locally before submitting or updating a PR.

mystic-io:
Hey @cjdsellers ,
Both scope issues are sorted and pushed: `load_ids` now queries by condition ID alone in both paths, and `accept` only touches filter-driven results. Tests for both. Em dash and the qualified path are fixed too; turns out I just never had prek installed locally, which explains how those slipped by. All green now except `test wheel publishing`, which fails on a clean develop here as well. Thanks for the sharp eyes.

cjdsellers (approved):
Hey @mystic-io,

Thanks for turning both of those around so quickly. Querying `load_ids` by condition ID alone and keeping `accept` on filter-driven results only is exactly right, and the two extras you found on your own pass, `load_scoped_all` dropping registered filters and the `load` fallback broadening a scoped provider, are the ones that would have shown up later as a refresh quietly disagreeing with the bootstrap.

I went back over the whole change and it looks good. Deferring `set_initialized` until every additive scope succeeds closes the retry hole, `series_ids` lands last in the PyO3 signature so existing positional callers are unaffected, and the stub matches. Good call covering both `initialize` and `fetch_configured_instruments` for each scope combination rather than just the bootstrap path.

One thing that came out of the review and isn't for this PR: `should_load_all` still doesn't count `filters`, so a config with only a `filters` map and nothing else bootstraps nothing and logs the "No bootstrap configured" warning. That predates your change, and I'll raise it separately now that the docs describe scopes as composing.

Merging this, thanks again for both the issue and the fix 🙏

PR #4656, opened 2026-08-05
fix(polymarket): retain discovery replay on reconnect

6Kmfi6HP:
## Summary

- retain empty-asset new-market discovery as explicit desired subscription state
- rebuild reconnect replay as separate discovery and asset frames
- bind replay sends to a connection epoch and add shard/epoch replay observability
- emit feed_resumed only after the first inbound market frame for the replay epoch

## Tests

- `cargo test -p nautilus-polymarket websocket --lib` (89 passed)

This patch is based on exact nightly wheel source SHA `a930c8afe380025fc0a10c6b2cd6907d6b983e86` used by `1.231.0a20260730`.


cjdsellers:
Hi @6Kmfi6HP,

Thanks for the PR and for identifying this reconnect gap. Since the CLA is still unsigned, I can't merge the contribution directly, but I've independently reimplemented the fix locally. Polymarket now retains discovery and asset subscription intent across reconnects and rejects replay work from stale connection epochs.

The reimplementation has passed the focused tests and a live Polymarket smoke test. I'll close this PR, but I appreciate the work that helped point us at the issue.

Issue #4607, opened 2026-07-30
[Persistence] [v2] Custom Feather conversion silently writes no Parquet data


mystic-io:
# Bug Report

## Confirmation

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code and originating pull requests to confirm the behavior is not by design.
- [ ] I've tested this issue using a recent pre-release or development wheel (`2.0.0rcN`,
      `dev` develop, or `a` nightly) and can still reproduce it.

I reproduced this from source against current `develop` commit
`45903fc8b925adae6323035fb0b4fb5b49b4f89b`. I have not separately repeated the
reproduction against a packaged development wheel.

I searched issues and pull requests, including closed and merged results, for
`custom data parquet`, `convert_stream_to_data`, `Feather custom data`,
`to_snake_case custom`, and `custom parquet conversion`. I found related work,
including #3542, #3590, #1796, #3954, and #4297, but no existing report for this
specific marker and stream-layout defect.

This appears to be a bug rather than an enhancement request. Merged PR #3542
introduced the custom-data persistence architecture and explicitly included
conversion of custom data written through the Feather writer into Parquet.
`docs/concepts/custom_data.md` likewise states that the custom-data conversion
branch writes Feather stream data to Parquet. The implementation also contains
a `data_cls.starts_with("custom/")` conversion branch. That branch is currently
unreachable for Feather files written by `FeatherWriter`.

## Expected behavior

When registered custom data is written through `FeatherWriter`, calling:

```rust
catalog.convert_stream_to_data(
    instance_id,
    "custom/RustTestHashMapCustomData",
    Some("live"),
    None,
    false,
)?;
```

should discover the generated Feather files, convert their Arrow batches into
Parquet, and preserve the custom type and identifier layout.

A subsequent query for `RustTestHashMapCustomData` should return every record
written to the Feather stream. For example, writing three records for each of
two identifiers should result in six queryable Parquet records.

The expected flow is:

1. `FeatherWriter` writes the stream under
   `live/{instance_id}/data/custom/{TypeName}/{identifier}/...`.
2. `convert_stream_to_data` recognizes the logical name
   `custom/{TypeName}`.
3. Stream discovery maps that logical name to the physical
   `data/custom/{TypeName}` directory.
4. The Arrow batches are written under the catalog’s normal
[...]

PR #4658, opened 2026-08-05
fix(persistence): correctly handle custom feather paths for parquet conversion (#4607)

alex09x:
## Problem
Issue #4607 reports that `ParquetDataCatalog::convert_stream_to_data` silently finishes while failing to convert custom data into parquet files.
This is because:
1. `to_snake_case` incorrectly mutated the `custom/` prefix, breaking discovery for custom datatypes.
2. `list_feather_files` was searching beneath `{root}/live/{instance_id}/custom/{TypeName}/...` instead of `{root}/live/{instance_id}/data/custom/{TypeName}/...` where `FeatherWriter` actually places them.

## Solution
- Prevent `to_snake_case` from altering data class strings that begin with `custom/` (preserving exact casing for user types).
- Adjust the expected search path prefix within `list_feather_files` to look beneath `data/custom/...` for custom data.
- Added a full round-trip test `test_convert_stream_to_data_custom_data_daily_roundtrip` replicating the exact issue setup.

## Verification
- Added test successfully simulates writes via `FeatherWriter` followed by stream conversion and dynamically queries the expected data.

### DIFF
--- a/crates/persistence/src/backend/catalog.rs
+++ b/crates/persistence/src/backend/catalog.rs
@@ -3755,8 +3755,14 @@ impl ParquetDataCatalog {
                     continue;
                 };
 
+                let expected_prefix = if data_name.starts_with("custom/") {
+                    format!("data/{data_name}/")
+                } else {
+                    format!("{data_name}/")
+                };
+
                 if let Some(data_relative_path) =
-                    relative_path.strip_prefix(&format!("{data_name}/"))
+                    relative_path.strip_prefix(&expected_prefix)
                 {
                     if let Some(identifiers) = identifiers {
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @alex09x,

Thanks for the PR. As noted in #4607, we're deferring this fix to the new persistence path rather than putting more work into the current catalog (even if a PR has been generated for it), so I'll close this. The CLA also remains unsigned, and the PR title incorrectly uses Conventional Commits syntax, which [CONTRIBUTING.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md#before-you-open-a-pr) asks contributors not to use.

PR #4659, opened 2026-08-05
Fix CMO indicator value exceeding range bounds

### LINKED ISSUE #4654: ChandeMomentumOscillator can return a value outside [-100, 100], and VIDYA turns it into a weight above 1
# Bug Report

## Confirmation

- [x] I've reviewed or skimmed the source code and originating pull requests to confirm the behavior is not by design.
- [x] I've reproduced this from source against current `develop` (`91d057f`).

## Expected Behavior

`ChandeMomentumOscillator::value` stays within `[-100, 100]`, and the `cmo_pct` that `VariableIndexDynamicAverage` derives from it stays within `[0, 1]`. The second is asserted by the crate itself at `crates/indicators/src/average/vidya.rs:346`.

## Actual Behavior

When the gain average decays to exactly zero, the oscillator lands one ULP outside its range, and VIDYA reads that as a smoothing weight above one:

```
CMO     = -100.00000000000001
cmo_pct =  1.0000000000000002
```

## Steps to Reproduce

```rust
use nautilus_indicators::average::vidya::VariableIndexDynamicAverage;
use nautilus_indicators::indicator::MovingAverage;
use nautilus_indicators::momentum::cmo::ChandeMomentumOscillator;

let mut cmo = ChandeMomentumOscillator::new(14, None);
let mut vidya = VariableIndexDynamicAverage::new(14, None, None);
for price in std::iter::repeat_n(100.0, 21).chain(std::iter::repeat_n(50.0, 20)) {
    cmo.update_raw(price);
    vidya.update_raw(price);
}
assert!((-100.0..=100.0).contains(&cmo.value));      // fails
assert!((0.0..=1.0).contains(&vidya.cmo_pct));       // fails
```

Twenty-one flat prices then a drop and more flat prices: after the drop every gain is zero, so the gain average reaches exactly zero while the loss average stays positive.

## Cause

`crates/indicators/src/momentum/cmo.rs`:

```rust
self.value = 100.0 * (self.average_gain.value() - self.average_loss.value()) / divisor;
```

With `average_gain == 0.0` this is `(100 * -L) / L`, which is not exactly `-100` for every `L`, because the multiplication rounds first. Dividing first gives `(-L / L) * 100`, and `-1.0 * 100.0` is exact.

`stochastics.rs:331` already writes it that way, so the codebase has the shape in it:

```rust
let raw_k = 100.0 * ((close - k_min_low) / (k_max_high - k_min_low));
```

## Impact

Small in magnitude and not in kind: `cmo_pct` is used as a weight in `vidya.rs:139`, so a value above one inverts the sign of the complement `1 - alpha * cmo_pct` for `alpha` close to one. Any downstream comparison against the documented bounds also fails.

## Fix

One line, plus a regression test. I have it locally: all 537 tests in the crate pass, and the test fails on `develop` with `value -100.00000000000001 outside [-100, 100]`.

One thing worth flagging before I open the PR. The reorder moves `vidya::tests::test_value_with_ten_inputs` by a single ULP, from `0.046_813_474_863_949_87` to `0.046_813_474_863_949_864`, because that test pins a float with `assert_eq!` at full precision. I updated the constant. There are 27 such pinned constants in the crate against 10 uses of `approx_eq`, so any future change to the order of operations will look like a regression in the same way; happy to leave that alone or raise it separately, whichever you prefer.

Shall I open the PR against `develop`?

alex09x:
Fixes #4654.

## Problem
When `average_gain` decays to exactly zero, `ChandeMomentumOscillator` lands one ULP outside of the `[-100, 100]` bound (`-100.00000000000001`). `VIDYA` uses this value as a smoothing weight and expects it to be at most 1, resulting in weights above 1 which inverted the sign of the complement `1 - alpha * cmo_pct`.
This was caused by multiplying by `100.0` before dividing by `divisor`.

## Solution
Changed the CMO calculation order to `100.0 * ((average_gain - average_loss) / divisor)`. This ensures that when `average_gain` is 0, `(0 - L) / L` is exactly `-1.0` before being scaled to `-100.0`.
Updated one regression test constant in `vidya` to reflect the 1-ULP change and added the reproducer to `cmo.rs` tests.

### DIFF
--- a/crates/indicators/src/average/vidya.rs
+++ b/crates/indicators/src/average/vidya.rs
@@ -211,7 +211,7 @@ mod tests {
         indicator_vidya_10.update_raw(1.00020);
         indicator_vidya_10.update_raw(1.00010);
         indicator_vidya_10.update_raw(1.00000);
-        assert_eq!(indicator_vidya_10.value, 0.046_813_474_863_949_87);
+        assert_eq!(indicator_vidya_10.value, 0.046_813_474_863_949_864);
     }
 
     #[rstest]
--- a/crates/indicators/src/momentum/cmo.rs
+++ b/crates/indicators/src/momentum/cmo.rs
@@ -300,13 +300,13 @@ mod tests {
         assert!(cmo.value <= 100.0 && cmo.value >= -100.0);
     }
 
-    #[test]
+    #[rstest]
     fn test_cmo_value_stays_within_bounds() {
-        use crate::average::vidya::VariableIndexDynamicAverage;
+        use crate::{average::vidya::VariableIndexDynamicAverage, MovingAverage};
 
         let mut cmo = ChandeMomentumOscillator::new(14, None);
         let mut vidya = VariableIndexDynamicAverage::new(14, None, None);
-        for price in std::iter::repeat(100.0).take(21).chain(std::iter::repeat(50.0).take(20)) {
+        for price in std::iter::repeat_n(100.0, 21).chain(std::iter::repeat_n(50.0, 20)) {
             cmo.update_raw(price);
             vidya.update_raw(price);
         }
--- a/crates/indicators/src/momentum/cmo.rs
+++ b/crates/indicators/src/momentum/cmo.rs
@@ -137,7 +137,7 @@ impl ChandeMomentumOscillator {
                 self.value = 0.0;
             } else {
                 self.value =
-                    100.0 * (self.average_gain.value() - self.average_loss.value()) / divisor;
+                    100.0 * ((self.average_gain.value() - self.average_loss.value()) / divisor);
             }
         }
         self.previous_close = close;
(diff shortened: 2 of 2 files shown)


cjdsellers:
Hey @alex09x,

Thanks for the PR. The calculation change itself looks correct, but the new regression test does not compile because `VariableIndexDynamicAverage::update_raw` comes from the `MovingAverage` trait, which is not in scope in the `cmo.rs` test module. A focused check against f40c97af59 fails with `E0599` at line 311.

There are several things to address before this can move forward:

- The CLA has not been signed ([CONTRIBUTING.md](CONTRIBUTING.md#start-with-an-issue)).
- The PR title uses Conventional Commits syntax, which we ask contributors not to use ([CONTRIBUTING.md](CONTRIBUTING.md#before-you-open-a-pr)).
- The new regression test does not compile because `MovingAverage` is not in scope.
- The test uses `#[test]` instead of the required `#[rstest]`.
- The iterator triggers `clippy::manual_repeat_n`.
- The added code is not rustfmt-clean.
- `make pre-commit` will therefore fail as soon as it runs.

Please run `make pre-commit` and the relevant tests before updating the PR. We hold contributor PRs to a higher standard and consider these checks the minimum gates to clear before we can accept and merge a PR. Thanks again.

cjdsellers:
I'm closing this in favour of #4667 from @mkzung, who reported the bug in #4654 and had the same fix ready with a compiling test.

PR #4662, opened 2026-08-05
Fix Polymarket maker fill ownership and count dropped trades

seungpyoson:
## Summary

Maker-order ownership was decided by case-sensitive string equality on the maker address. Venue payloads carry EIP-55 checksummed (mixed-case) addresses while configured funder addresses are commonly lowercase, so an account's own confirmed maker fills failed the ownership test and the whole trade vanished from the generated mass status with no fill report, no counter, and no log line, silently understating executed quantity.

- Add `PolymarketMakerOrder::is_owned_by`: the address compare ignores ASCII case (hex letter case carries no account identity) while the API-key compare stays exact; both prior inline spellings (REST reconciliation and WS dispatch) now delegate to it
- Count confirmed maker trades dropped for holding no owned maker order in a typed `FillBuildDiscards`, and raise one error per mass-status pass when the count is nonzero; polled report paths are unaffected
- Reproduction tests fail at the parent commit and pass here; a config-side-normalization mutant is killed by the direction-pinned tests

## Known open, deliberately not bundled

The Python adapter has the identical defect: `nautilus_trader/adapters/polymarket/schemas/trade.py:79` and `schemas/user.py:226` compare `order.maker_address == maker_address or order.owner == api_key`, and callers `execution.py:851` (REST trade reports) and `execution.py:2335` (WS user trade) iterate the returned list, so an empty result is a silent no-op. This PR fixes the Rust adapter only to keep the change small and reviewable; the Python instance needs its own reproduction and fix.

## Testing

- `cargo nextest run -p nautilus-polymarket`: 1232 passed, 0 skipped
- `cargo clippy -p nautilus-polymarket --all-targets -- -D warnings`: clean
- `make pre-commit` (all files) run locally before the single push

### DIFF
--- a/crates/adapters/polymarket/src/common/models.rs
+++ b/crates/adapters/polymarket/src/common/models.rs
@@ -63,6 +63,22 @@ pub struct PolymarketMakerOrder {
     pub side: Option<PolymarketOrderSide>,
 }
 
+impl PolymarketMakerOrder {
+    /// Returns whether this maker order belongs to the account identified by
+    /// `user_address` and `api_key`.
+    ///
+    /// The address comparison ignores ASCII case: hex letter case carries no
+    /// account identity (EIP-55 checksumming encodes only a display checksum),
+    /// and the two sides routinely disagree: recorded venue payloads carry
+    /// checksummed maker addresses while configured funder addresses are
+    /// commonly lowercase, or vice versa. The API-key comparison stays exact
+    /// because keys are opaque credentials.
+    #[must_use]
+    pub fn is_owned_by(&self, user_address: &str, api_key: &str) -> bool {
+        self.maker_address.eq_ignore_ascii_case(user_address) || self.owner == api_key
+    }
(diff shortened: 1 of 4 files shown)


cjdsellers (approved):
Hi @seungpyoson,

Thank you for the PR, and for tracing this back to a real checksummed address mismatch. The local case‑insensitive comparison and typed discard counts are the right design here, and the tests cover the core ownership and reconciliation behavior well. This is good to merge as it stands, with no further changes needed on this PR 👌.

I'll carry the remaining refinements as follow‑ups: narrow `PolymarketMakerOrder::is_owned_by` to `pub(crate)`, add a focused WebSocket case‑variant maker‑fill regression, strengthen the REST test with the exact `venue_order_id`, and add maker/taker missing‑instrument counter coverage. I'll also clean up the "Known open" Python note, since those v1 paths are no longer present on `develop`.

PR #4664, opened 2026-08-05
Fix Interactive Brokers continuous futures historical bar requests

dfjmax:
# Pull Request

> workflow changes are maintainer‑only.

## Summary

Continuous futures (CONTFUT) historical bar downloads failed with IB error `10339` ("Setting end date/time for continuous future security type is not allowed") because the bars request always sent an end date/time. The request now omits the end date for continuous futures (matching the legacy v1 client, which blanked `endDateTime` for CONTFUT) and uses only the first duration segment, since without an end date the request cannot page backward. Behavior for every other security type is unchanged.

## Related issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

N/A — non-breaking. Continuous futures previously errored and now return bars; all other security types follow the original segmented request path unchanged.

## Documentation

No documentation changes.

## Testing

**Ensure new or changed logic is covered by tests.** Check at least one:

Added unit tests in `crates/adapters/interactive_brokers/src/historical/client.rs` (`historical::client::tests`) for the extracted `bar_request_segments` helper: non-continuous segments retain their end dates in order; continuous futures drop the end date and keep only the first (largest) duration segment; empty input yields nothing.

### DIFF
--- a/crates/adapters/interactive_brokers/src/data/convert.rs
+++ b/crates/adapters/interactive_brokers/src/data/convert.rs
@@ -363,6 +363,44 @@ pub fn calculate_duration_segments(
     results
 }
 
+/// Adapt duration segments for an IB historical bars request.
+///
+/// For continuous futures the end date is dropped and only the first segment is
+/// kept (IB rejects an explicit end date with error 10339), logging a warning
+/// when the requested range cannot be honored.
+pub fn bar_request_segments(
+    segments: Vec<(Timestamp, IBDuration)>,
+    is_continuous_future: bool,
+) -> Vec<(Option<Timestamp>, IBDuration)> {
+    if is_continuous_future {
+        // Treat end dates within the last second as "now" so requests whose
+        // end defaults to the current time do not trigger a spurious warning.
+        let now = Timestamp::now();
+        let end_in_past = segments
+            .first()
+            .is_some_and(|(end, _)| *end < now - jiff::SignedDuration::from_secs(1));
+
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hey @dfjmax,

Thanks for the PR. Omitting `endDateTime` is the right approach, and `take(1)` avoids repeating the same now-anchored request.

**I think two issues still block merging this:**

- The request silently ignores a past `end_date_time`, and ranges requiring multiple duration segments are truncated to the first segment. This can return bars outside the requested range without warning. Could we document this on the Rust and PyO3 APIs, and warn or return an error when the requested range cannot be honored?
- `InteractiveBrokersDataClient::request_bars` still calls `.ending(...)`, so `CONTFUT` requests through a `TradingNode` continue to fail with IB 10339. V1 covered both callers through the shared request code, so I think that path should be covered here too.

One small non-blocking test point: `historical::Duration` supports equality, so the retained duration can use an exact `assert_eq!`.

dfjmax:
Hi @cjdsellers, thanks for the review. 

I've addressed the points. Let me know if anything else needs to be changed.

cjdsellers (approved):
Hey @dfjmax,

Thanks for the follow-up. I rechecked the latest commit, and all three points are addressed. The range warning and docs are in place, the `TradingNode` request path now handles `CONTFUT`, and the duration check is exact. CI is green, so this is ready to merge.

PR #4666, opened 2026-08-05
Fix Indicator::reset leaving state behind in AMA and FuzzyCandlesticks

### LINKED ISSUE #4665: Indicator::reset leaves state behind in AdaptiveMovingAverage and FuzzyCandlesticks
Two indicators write `reset` twice, once as an inherent method and once for the `Indicator` trait, and in both cases the two copies have drifted apart. The inherent one is complete; the trait one is missing fields.

`AdaptiveMovingAverage`: the trait `reset` clears `value`, `count`, `has_inputs` and `initialized`, but not `prior_value`. The next `update_raw` after a reset then smooths against a price from before it, through `prior_value.unwrap()` in `ama.rs`.

`FuzzyCandlesticks`: the trait `reset` clears the four percent vectors and the last OHLC, but not `value` or `vector`. The previous candle's classification stays readable, and the next `vector` keeps the old one's length.

Calling `reset()` on a concrete instance picks the inherent method, so most code is fine. Anything reached through `dyn Indicator` or a generic `T: Indicator` bound gets the trait method and the stale state.

Reproduction, no data needed:

```rust
let mut ama = AdaptiveMovingAverage::new(10, 2, 30, None);
for i in 1..=12 { ama.update_raw(f64::from(i)); }
Indicator::reset(&mut ama);
assert_eq!(ama.prior_value, None); // fails: Some(9.581497043272726)
```

I checked all 37 indicators by comparing the fields mutated in `update_raw` against the fields cleared in `reset`. These two are the only ones that define `reset` twice, and neither pair matches.

Happy to open a PR: both trait impls delegate to the inherent method so they cannot drift again, with a test each that resets through the trait.

mkzung:
Closes #4665.

Exactly two indicators write reset twice, once as an inherent method and
once for the Indicator trait, and both had drifted apart. The inherent copy
is the complete one in each case.

AdaptiveMovingAverage's trait reset never cleared prior_value, so the first
update_raw after a reset smoothed against a price from before it.
FuzzyCandlesticks' trait reset never cleared value or vector, so the last
candle's classification stayed readable and the next vector kept the old
one's length.

Calling reset() on a concrete instance picks the inherent method and is
fine. Anything going through dyn Indicator or a generic Indicator bound
gets the trait method and the stale state, which is the path taken by
anything holding a boxed indicator.

Both trait impls now delegate to the inherent one. Rust prefers inherent
methods, so this resolves the way it reads; if the inherent method were
ever deleted it would become infinite recursion instead, and the two tests
added here would catch that.

Found by comparing, for all 37 indicators, the fields mutated in update_raw
against the fields cleared in reset. Only these two define reset twice, and
neither pair matched.

### DIFF
--- a/crates/indicators/src/average/ama.rs
+++ b/crates/indicators/src/average/ama.rs
@@ -101,10 +101,12 @@ impl Indicator for AdaptiveMovingAverage {
     }
 
     fn reset(&mut self) {
-        self.value = 0.0;
-        self.count = 0;
-        self.has_inputs = false;
-        self.initialized = false;
+        // Delegate rather than repeat the field list: the copy that used to
+        // live here had already fallen behind and left `prior_value` set.
+        // This resolves to the inherent `reset` because Rust prefers inherent
+        // methods; deleting that one would turn this into infinite recursion,
+        // which `trait_reset_clears_prior_value` below would catch.
+        Self::reset(self);
     }
 }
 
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @mkzung,

Thanks for the PR. Everything looks correct here except for the pre-commit.

It's easy to fix. Just run the following and push any changes to the same PR:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

Once that's done we should be able to merge after CI is green!

cjdsellers:
@mkzung, Your diagnosis of the duplicated reset paths was correct. I landed the fix directly in 8003bed6ef, with one extension: `AdaptiveMovingAverage` now also clears its nested `EfficiencyRatio` history so inputs and deltas cannot carry across a reset. Both `AdaptiveMovingAverage` and `FuzzyCandlesticks` now use one canonical reset implementation, with the complete reset contract tested through both dispatch paths.

I've credited you in the release notes for finding this. I'll close this PR in favor of the version now on `develop`.

PR #4669, opened 2026-08-06
Compose the report generators in the default generate_mass_status

folknor:
The `ExecutionClient` trait default for `generate_mass_status` logged not-implemented and returned `Ok(None)`, which the live node's reconciliation treats as "no mass status available" - a warn line and zero startup reconciliation. An adapter implementing only the three granular generators (`generate_order_status_reports`, `generate_fill_reports`, `generate_position_status_reports`) therefore got no reconciliation with no error. The legacy Python `LiveExecutionClient` composed the three generators in exactly this situation, and that fallback disappeared with the v2 cutover, so the trait default is now the only place this composition can live.

The default now snapshots the realtime atomic clock once, derives `start` from `lookback_mins` (`checked_mins_to_nanos`, error on overflow, saturating at the epoch), runs the three generators concurrently, and returns a populated `ExecutionMassStatus`. Granular errors propagate with context naming the failing generator rather than degrading to `Ok(None)` - `Err` already means "reconciliation must not proceed" at the node, and returning `None` on failure would recreate the silent skip. Three empty report sets return `Some(empty)`, since an inactive account is a valid snapshot. The doc comment states the trade: the default is clock-correct for live/realtime clients; a client on a mocked or backtest clock must override. Every in-tree adapter overrides `generate_mass_status`, so no existing adapter changes behavior.

## Testing

A fake client implements only the three granular generators: the composition test asserts the result carries all three fixtures under the expected keys, each generator is called exactly once, every command shares the mass status's `ts_init` and derived `start`, and `open_only` is false with no filters set - the report fixtures' own timestamps differ from the mass timestamp, proving the result is not timestamped from report contents. A second test asserts a sentinel generator error propagates with its context, never `None` or partial data. Both tests were verified to fail against the unfixed default, which returns `None` without invoking any generator.

### DIFF
--- a/crates/common/src/clients/execution.rs
+++ b/crates/common/src/clients/execution.rs
@@ -15,8 +15,9 @@
 
 //! Execution client trait definition.
 
+use anyhow::Context;
 use async_trait::async_trait;
-use nautilus_core::UnixNanos;
+use nautilus_core::{UnixNanos, datetime::checked_mins_to_nanos, time::get_atomic_clock_realtime};
 use nautilus_model::{
     accounts::AccountAny,
     enums::{LiquiditySide, OmsType},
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I checked the shared timestamp and lookback window, report aggregation, and error path.

Looks good to merge from my side 👌

PR #4672, opened 2026-08-06
Add post-start data-client registration to LiveNode

folknor:
Data clients could only be registered on `LiveNodeBuilder` before `build()`; a running node had no path to bring up a new venue's data client and had to be rebuilt.

## Add post-start data-client registration to LiveNode

`LiveNode::add_data_client` / `add_data_client_with_routing` create the client from its factory, start and connect it under the node's `timeout_connection` with no engine borrow held across the awaits, and only then commit registration - a client that fails to start or connect leaves no engine registration, route, or name behind (a drop guard stops and disposes it, with best-effort disconnect under `timeout_disconnection`). Holding no borrow across `connect()` matters because a `DataClient::connect` may synchronously reach a message-bus endpoint whose `DataEngine` handler borrows the engine. Committed clients participate in normal shutdown, and name collisions are refused with the builder's error.

## Checked atomic registration in DataEngine

`register_client` panics on duplicate IDs, and venue routing could fail after insertion. `register_client_checked` validates the client ID, default selection, and every explicit route first, then commits infallibly; the builder now registers through it so build-time and post-start semantics cannot diverge. A primary venue keeps `register_client`'s overwrite semantics and explicit routes conflict only when routed to a different client, so existing builder configurations are unaffected.

## Testing

Current-thread lifecycle tests cover the happy path (factory/start/connect ordering, venue routing reachable via `get_client`, a connect-emitted instrument cached before the call returns, disconnect and stop on node shutdown), a name collision with a build-time client that never invokes the second factory, a pending connect that times out leaving nothing registered, a route conflict that fails before start or connect, and a hostile client whose `connect()` synchronously sends into a `DataEngine` endpoint - pinning that no engine borrow spans the connect. An engine regression pins the preserved primary-venue overwrite. Differentials verified: neutering the name preflight fails the collision test, and a borrow-across-await variant panics the hostile test with `RefCell already borrowed`.

### DIFF
--- a/crates/data/src/engine/mod.rs
+++ b/crates/data/src/engine/mod.rs
@@ -460,6 +460,87 @@ impl DataEngine {
         log::debug!("Registered client {client_id}");
     }
 
+    /// Validates that a client and all of its routing can be registered atomically.
+    ///
+    /// A primary venue keeps [`Self::register_client`]'s overwrite semantics and never
+    /// conflicts; explicit venue routes conflict when routed to a different client.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the client ID, default routing, or any explicit venue routing
+    /// conflicts with an existing registration.
+    pub fn validate_client_registration(
+        &self,
+        client_id: ClientId,
+        primary_venue: Option<Venue>,
+        default: bool,
+        venue_routes: &[Venue],
+    ) -> anyhow::Result<()> {
+        if self.clients.contains_key(&client_id) {
+            anyhow::bail!("Data client {client_id} is already registered");
+        }
+
+        if default && self.default_client_id.is_some_and(|id| id != client_id) {
+            anyhow::bail!("default client already registered");
+        }
+
+        for venue in venue_routes {
+            if primary_venue == Some(*venue) {
+                continue;
+            }
+
+            if let Some(existing_client_id) = self.routing_map.get(venue)
+                && *existing_client_id != client_id
+            {
+                anyhow::bail!(
+                    "Venue {venue} already routed to {existing_client_id}, cannot re-route to {client_id}"
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Registers a client and all routing after validating that the complete operation can
+    /// succeed without mutating the engine.
+    ///
+    /// A primary venue keeps [`Self::register_client`]'s overwrite semantics; explicit venue
+    /// routes conflict when routed to a different client.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the client ID, default routing, or any explicit venue routing
+    /// conflicts with an existing registration.
+    pub fn register_client_checked(
+        &mut self,
+        client: DataClientAdapter,
+        primary_venue: Option<Venue>,
+        default: bool,
+        venue_routes: &[Venue],
+    ) -> anyhow::Result<()> {
+        let client_id = client.client_id();
+        self.validate_client_registration(client_id, primary_venue, default, venue_routes)?;
+
+        if let Some(venue) = primary_venue {
+            self.routing_map.insert(venue, client_id);
+            log::debug!("Set client {client_id} routing for {venue}");
+        }
+
+        for venue in venue_routes {
+            self.routing_map.insert(*venue, client_id);
+            log::debug!("Set client {client_id} routing for {venue}");
+        }
+
+        if default || (client.venue.is_none() && self.default_client_id.is_none()) {
+            self.default_client_id = Some(client_id);
+            log::debug!("Registered client {client_id} for default routing");
+        }
+
+        self.clients.insert(client_id, client);
+        log::debug!("Registered client {client_id}");
+        Ok(())
+    }
+
     /// Deregisters the client for the `client_id`.
     ///
     /// # Panics
--- a/crates/data/tests/engine.rs
+++ b/crates/data/tests/engine.rs
@@ -634,6 +634,74 @@ fn test_register_and_deregister_client(
 }
 
 #[rstest]
+fn test_register_client_checked_preserves_primary_venue_overwrite(
+    data_engine: Rc<RefCell<DataEngine>>,
+    clock: Rc<RefCell<TestClock>>,
+    cache: Rc<RefCell<Cache>>,
+) {
+    let mut data_engine = data_engine.borrow_mut();
+
+    let venue = Venue::test_default();
+    let client_id1 = ClientId::new("C1");
+    let data_client1 = DataClientAdapter::new(
+        client_id1,
+        Some(venue),
+        true,
+        true,
+        Box::new(MockDataClient::new(
+            clock.clone(),
+            cache.clone(),
+            client_id1,
+            Some(venue),
+        )),
+    );
+    data_engine
+        .register_client_checked(data_client1, Some(venue), false, &[])
+        .unwrap();
+
+    // A second primary registration for the same venue keeps register_client's
+    // overwrite semantics: the later client takes the route.
+    let client_id2 = ClientId::new("C2");
+    let data_client2 = DataClientAdapter::new(
+        client_id2,
+        Some(venue),
+        true,
+        true,
+        Box::new(MockDataClient::new(
+            clock.clone(),
+            cache.clone(),
+            client_id2,
+            Some(venue),
+        )),
+    );
+    data_engine
+        .register_client_checked(data_client2, Some(venue), false, &[])
+        .unwrap();
+    assert_eq!(
+        data_engine
+            .get_client(None, Some(&venue))
+            .unwrap()
+            .client_id(),
+        client_id2
+    );
+
+    // An explicit route to a venue owned by a different client still conflicts.
+    let client_id3 = ClientId::new("C3");
+    let data_client3 = DataClientAdapter::new(
+        client_id3,
+        None,
+        true,
+        true,
+        Box::new(MockDataClient::new(clock, cache, client_id3, Some(venue))),
+    );
+    let error = data_engine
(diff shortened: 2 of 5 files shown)


cjdsellers:
Hi @folknor,

Thank you for the PR. The atomic engine registration and borrow handling look solid.

**I found three issues that need changes before this is safe to merge:**

- In the documented `start()`/`poll()` mode, `add_data_client_with_routing` holds `&mut LiveNode` while awaiting connection and readiness, so `poll()` cannot service existing time, execution, or data traffic. Only data is flushed after success. A normal connection delay can therefore hold fills, order commands, and timers for up to `timeout_connection`. Can you keep the existing runner serviced during registration and add a regression for existing execution and time traffic?
- The factory-created client is not placed under `LateDataClientGuard` until after route and engine validation. Duplicate IDs, default or route conflicts, and a failed engine borrow return after creation without `stop()` or `dispose()`. Can you establish cleanup ownership before those fallible steps and assert disposal in the route-conflict test?
- `Venue::new` panics on invalid `RoutingConfig.venues`; release builds use `panic = "abort"`, so malformed runtime input terminates the live node after the factory has already created the client. This should use checked parsing and return an error through the documented API.

**I also found three non-blocking follow-ups:**

- Have the readiness wait react to `LiveNodeHandle` and kernel shutdown requests.
- Preserve the original error chain when cleanup also fails.
- `LifecycleClientState` has grown to seven independent atomic flags. I think an ordered `LifecycleEvent` trace would reduce the state surface and let these tests assert lifecycle sequences directly. This can happen separately, but it would be worth addressing before the fixture grows further.

  A simpler state enum could be:

```rust
  #[derive(Clone, Copy, Debug, PartialEq, Eq)]
  enum LifecycleState {
      Created,
      Started,
      Connecting,
      Connected,
      Disconnecting,
      Disconnected,
      Stopped,
      Disposed,
  }
```

folknor:
Thank you, @cjdsellers - all three blocking issues and the first two follow-ups are addressed in the latest commit.

## Registration no longer starves the runner

Connect and readiness are now one staged future driven by a single loop that services `runner.recv()` between polls, so time events, execution events and execution commands are processed while a registration is in flight. The loop also re-checks `LiveNodeHandle` and kernel shutdown on every pass, which covers your first follow-up for a blocked `connect()` as well as the readiness wait.

Three properties you can't see in the diff: (1) no `DataEngine` borrow enters the loop, so the synchronous msgbus re-entry case still holds; (2) a zero `timeout_connection` still gets exactly one biased poll before timing out, so an immediately ready client continues to register; (3) an absent runner is not an error - `run()` owns the receivers, so that mode falls back to the previous unserviced wait rather than acquiring a new precondition. Closed runner channels abort through the guarded cleanup path.

## Cleanup ownership precedes every fallible step

Routing is parsed before `factory.create`, so a malformed route returns without a client existing at all. `DataClientAdapter::new` is infallible, which leaves nothing fallible between the factory returning and `LateDataClientGuard`, so duplicate IDs, default and route conflicts, and a failed engine borrow all dispose.

While adding that coverage I found one more hole in the same area: `LateDataClientGuard::cleanup` moved the adapter out of the guard before awaiting `disconnect()`, so cancelling the registration mid-disconnect skipped `stop()` and `dispose()`. The adapter now stays in the guard across that await and is taken only once the synchronous half runs, which is also what keeps it from running twice.

## Testing

`test_add_data_client_services_runner_while_connect_is_blocked` and `test_add_data_client_services_runner_during_readiness_wait` gate the client inside `connect()`, enqueue a time event, an execution event and an execution command, and require all three acknowledgements before releasing the gate - barriers throughout, no timing windows. Disabling the servicing arm while the loop still ticks fails both at the acknowledgement wait.

`test_add_data_client_cancelled_during_cleanup_still_disposes` drops the registration future while `disconnect()` is pending; restoring the old ownership fails it at the `stopped` assertion. Duplicate ID, default conflict, held engine borrow and route conflict each assert disposal, `test_add_data_client_rejects_malformed_route_before_factory` asserts the factory never ran, the cleanup-chain test walks `error.chain()`, and two shutdown tests cover a blocked connect and a readiness wait. Each new assertion fails against the previous implementation.

## On `LifecycleClientState`

The connect gates added two `Notify` handles to that struct, so it is a little larger than when you raised it, not smaller. I'll follow up with the `LifecycleEvent` trace PR after this lands.

cjdsellers (dismissed):
Hey @folknor,

Thanks for the detailed follow-up. I re-reviewed f37747168e60a07d4f63e55a6fce99de903d6452. The runner servicing, cleanup ownership, checked route parsing, shutdown handling, and preserved error chain address the blocking points and follow-ups I raised. The focused tests also cover the relevant failure, cancellation, and in-flight runner paths.

This looks good to me now. The `LifecycleEvent` trace can follow separately and doesn't need to hold this PR 👍

cjdsellers:
@folknor, looks like some merge conflicts crept in when you get a chance.

folknor:
@cjdsellers One thing this raised that I just noticed.

`LateDataClientGuard` makes registration cancellation-safe - drop it mid-flight and the client is stopped and disposed. `LiveNode` has no equivalent: there is no `Drop` impl in `crates/system/src` or `crates/live/src`, and `stop()` / `dispose()` are the only paths reaching `NautilusKernel::cancel_timers`. So dropping the run future - a `select!` losing branch, a cancelled task, an aborted signal handler - leaves timers registered.

Is that something you want closed? `Drop` on `LiveNode` overlaps `dispose()`, and this is not my decision to make.

I can pick it up after the `LifecycleEvent` trace if you want it, and let me know how you prefer it.

cjdsellers:
Hey @folknor,

Thanks for raising the cancellation question. I re-reviewed f3df11fd646c8d90f3b7f5a8da8ef66ca089f018. The earlier connection/readiness starvation, cleanup ownership, route parsing, shutdown handling, and error-chain points are addressed.

**I found one remaining starvation path:**

Every failed late registration awaits `cleanup_late_data_client()`, and `LateDataClientGuard::cleanup()` can remain blocked in `disconnect()` for up to `timeout_disconnection` without servicing the runner. `test_add_data_client_cancelled_during_cleanup_still_disposes` already demonstrates that this await can remain pending. While the node is still running, time events, execution events, and order commands remain queued for the whole cleanup wait.

Can you drive cleanup through the same runner-servicing path and add a regression that acknowledges existing time and execution traffic while disconnect cleanup is blocked?

The broader `LiveNode::run()` cancellation contract can stay separate. A `Drop for LiveNode` alone would not cover a losing `select!` branch because cancelling that future leaves the node owned by the caller. Happy to review a separate PR for this :+1:

folknor:
Thank you, @cjdsellers - the remaining cleanup starvation is addressed in the latest commit.

## Cleanup no longer starves the runner

`cleanup_late_data_client` is now a `LiveNode` method that pins the existing `LateDataClientGuard::cleanup` future and drives it in a `select!` against `runner.recv()`, so time events, execution events and execution commands are processed while `disconnect()` is pending. All four post-construction failure exits route through it.

The servicing layer wraps the existing `dst::time::timeout(timeout_disconnection, disconnect())` rather than replacing it, which keeps the single-poll behaviour a zero disconnect timeout relies on, and keeps the adapter inside the guard across the await so a cancelled registration still runs `stop()` and `dispose()`.

Cleanup has no shutdown check, unlike the registration loop: a shutdown arriving mid-cleanup would otherwise skip `stop()` and `dispose()`. A closed runner also ends the selection and awaits cleanup directly, where registration treats one as a failure.

## Testing

`test_add_data_client_services_runner_while_cleanup_is_blocked` gates the client inside `disconnect()`, enqueues a time event, an execution event and an execution command once cleanup is in flight, and requires all three acknowledgements before releasing the gate - barriers throughout, with the timeouts acting only as deadlock guards. Discarding the serviced event while the loop still runs fails the test at the acknowledgement wait. The existing cancellation and disconnect-timeout tests keep a permanently pending client, so their behaviour is unchanged.

cjdsellers:
Hey @folknor,

Thank you for the care you put into each review follow-up. I’m sorry for taking this through several rounds of review and only reaching this decision now. The project is moving quickly, and I focused too narrowly on making the runtime registration path correct rather than stepping back earlier to reassess whether we should add the capability at all.

At this final review point, we’ve decided that the added lifecycle, runner servicing, cleanup, routing, and state management complexity is not worth supporting the addition of entire data clients after startup. We have concrete use cases for adding actors and strategies at runtime, since they operate within a node’s configured connectivity, but we don’t see the same need for changing the node’s external client topology while it is running.

We prefer the simplicity and reliability guarantees of keeping that topology immutable after construction. It is reasonable for a node to be configured for its intended task from the outset. If it needs to connect to a new provider or venue, bringing it down, reconfiguring it, and starting it again is an acceptable operational boundary.

Your follow-ups handled the review points carefully, and this decision is about the capability and its long-term maintenance cost, not the quality of your implementation. Apologies for the late reversal and the time it cost you. I’m going to close this PR.

folknor:
Hey @cjdsellers,

Thank you for the decision and for the review rounds that got here.

The argument I forgot to make in the description is: the node whose data dependencies are not knowable at construction. Workers accept strategies at runtime - the case you name as concrete - and each strategy declares the instruments it reads. If a strategy needs data from a venue the node was not built to connect to, there is no path to it today, and the strategy is refused before it starts.

Restart is a different operational boundary there than it is at deployment. The node holds open positions on the venues it already has, so bringing it down to add a read-only lane means reconciling live exposure through a restart to acquire data it does not trade on.

That is the coupling: runtime actors and strategies are bounded by build-time connectivity, and the bound is invisible at config time - it depends on what the strategy asks for.

Happy to leave this closed. If the shape is ever wanted, the constrained version is narrower than what this PR built: data clients only, no execution clients, no removal.

PR #4680, opened 2026-08-07
Emit quote ticks from delayed IB market data

### LINKED ISSUE #4679: Interactive Brokers: no quote ticks emitted on a delayed market data feed
Interactive Brokers: no quote ticks emitted on a delayed market data feed

### Summary

`subscribe_quote_ticks` yields no `QuoteTick` at all when the feed is delayed, even though TWS is sending complete bid/ask data. This affects any account without a real-time market data subscription for the instrument, and any configuration using `MarketDataType::Delayed` / `DelayedFrozen`.

### Cause

Quote assembly in `crates/adapters/interactive_brokers/src/data/core_streams.rs` matches only the real-time tick types:

```rust
match price.tick_type {
    TickType::Bid => cache.update_bid_price(...),
    TickType::Ask => cache.update_ask_price(...),
    TickType::Last => None,
    _ => None,          // DelayedBid / DelayedAsk land here
}
```

TWS reports a delayed feed as `DelayedBid` (66) / `DelayedAsk` (67) and `DelayedBidSize` (69) / `DelayedAskSize` (70). Those fall through to `_ => None`, so the `QuoteCache` is never updated and nothing is emitted. The same pattern is in `update_quote_from_size_tick` and `update_quote_from_price_size_tick`.

There is no warning — the subscription succeeds, ticks arrive, and the strategy simply never receives a quote, which makes it look like a connectivity or entitlement problem rather than a mapping gap.

### Reproduce

Against a paper account with no market data subscription:

1. `reqMarketDataType(4)`  (DelayedFrozen)
2. Subscribe to quotes for e.g. `AAPL.NASDAQ`
3. TWS returns error 10167 ("Requested market data is not subscribed. Displaying delayed market data") and then streams delayed ticks
4. No `QuoteTick` is ever produced

Raw `ibapi` on the same connection confirms the data is present:

```
66:DELAYED_BID=312.44   67:DELAYED_ASK=312.45
69:DELAYED_BID_SIZE     70:DELAYED_ASK_SIZE
```

### Verification that mapping is the only blocker

Applying the equivalent mapping to the released v1 Python adapter (1.231.0) and re-running an otherwise unchanged connectivity script against the same paper account produces quotes immediately:

```
quote received AAPL.NASDAQ bid=311.75 ask=311.76
quote received SPY.ARCA    bid=771.48 ask=771.50
```

### Environment

- Bug located on `develop` (Rust adapter, `nautilus-interactive-brokers` v0.62.0)
- Also present in the released v1 Python adapter 1.231.0 (`client/market_data.py`, `_try_create_quote_tick_from_market_data`)
- TWS API via `ghcr.io/gnzsnz/ib-gateway:stable`, IBKR paper account

Happy to submit a PR — I have one ready with regression tests.

G14rb:
Fixes #4679.

### Problem

`subscribe_quote_ticks` emits no `QuoteTick` at all when the Interactive Brokers feed is delayed, even though TWS is streaming complete bid/ask data. This hits any account without a real-time subscription for the instrument (TWS error 10167) and any configuration using `MarketDataType::Delayed` / `DelayedFrozen`.

Quote assembly matched only the real-time tick types, so the delayed variants fell through `_ => None`:

```rust
TickType::Bid => cache.update_bid_price(...),
TickType::Ask => cache.update_ask_price(...),
TickType::Last => None,
_ => None,          // DelayedBid (66) / DelayedAsk (67) land here
```

The failure is silent — the subscription succeeds and ticks arrive, but no quote is ever produced, which reads as a connectivity or entitlement problem rather than a mapping gap.

### Change

Add the delayed variants to the matching arms in the three quote-assembly functions in `data/core_streams.rs`:

| Function | Added |
|---|---|
| `update_quote_from_price_tick` | `DelayedBid`, `DelayedAsk`, `DelayedLast` |
| `update_quote_from_size_tick` | `DelayedBidSize`, `DelayedAskSize` |
| `update_quote_from_price_size_tick` | `DelayedBid`, `DelayedAsk`, `DelayedLast` |

Delayed ticks carry identical semantics to their real-time counterparts, so they route to the same `QuoteCache` updates. `DelayedLast` is mapped alongside `Last` (both `None`) so the arms stay symmetrical and a future change to `Last` handling does not silently skip the delayed case.

No behaviour changes for real-time feeds — this only adds previously unmatched variants.

### Tests

Two regression tests in the existing `core_streams` test module:

- `test_process_quote_tick_result_emits_quote_from_delayed_price_ticks` — `DelayedBid` + `DelayedAsk` produce a quote
- `test_process_quote_tick_result_emits_quote_from_delayed_size_ticks` — delayed prices and sizes produce a fully populated quote

```
cargo test -p nautilus-interactive-brokers --lib
test result: ok. 380 passed; 0 failed
```

### Live verification

Verified end to end against an IBKR paper account with no market data subscription, US session open. Before the fix the account produced no quotes; applying the equivalent mapping to the released v1 Python adapter (1.231.0) and re-running an otherwise unchanged connectivity script gave quotes immediately:

```
quote received AAPL.NASDAQ bid=311.75 ask=311.76
quote received SPY.ARCA    bid=771.48 ask=771.50
```

Raw `ibapi` on the same connection confirmed TWS was always sending the data:

```
66:DELAYED_BID=312.44   67:DELAYED_ASK=312.45
69:DELAYED_BID_SIZE     70:DELAYED_ASK_SIZE
```

### Note

The same bug exists in the v1 Python adapter (`client/market_data.py`, `_try_create_quote_tick_from_market_data`, which reads only tick types 0/1/2/3). Happy to open a companion PR against `develop_v1` if that branch is still taking fixes.

### DIFF
--- a/crates/adapters/interactive_brokers/src/data/core_streams.rs
+++ b/crates/adapters/interactive_brokers/src/data/core_streams.rs
@@ -1845,7 +1845,11 @@ fn update_quote_from_price_tick(
     ts_init: UnixNanos,
 ) -> Option<QuoteTick> {
     match price.tick_type {
-        TickType::Bid => cache.update_bid_price(
+        // Delayed variants carry identical semantics to their real-time counterparts.
+        // TWS emits them whenever the account has no real-time subscription for the
+        // instrument (error 10167) or `MarketDataType::Delayed`/`DelayedFrozen` is
+        // configured, so ignoring them drops every quote on a delayed feed.
+        TickType::Bid | TickType::DelayedBid => cache.update_bid_price(
             instrument_id,
             price.price,
             price_precision,
@@ -1853,7 +1857,7 @@ fn update_quote_from_price_tick(
             ts_event,
             ts_init,
         ),
-        TickType::Ask => cache.update_ask_price(
+        TickType::Ask | TickType::DelayedAsk => cache.update_ask_price(
             instrument_id,
             price.price,
             price_precision,
@@ -1861,7 +1865,7 @@ fn update_quote_from_price_tick(
             ts_event,
             ts_init,
         ),
-        TickType::Last => None,
+        TickType::Last | TickType::DelayedLast => None,
         _ => None,
     }
 }
@@ -1878,7 +1882,7 @@ fn update_quote_from_size_tick(
     ignore_size_updates: bool,
 ) -> Option<QuoteTick> {
     match size.tick_type {
-        TickType::BidSize => cache.update_bid_size_with_filter(
+        TickType::BidSize | TickType::DelayedBidSize => cache.update_bid_size_with_filter(
             instrument_id,
             size.size,
             price_precision,
@@ -1887,7 +1891,7 @@ fn update_quote_from_size_tick(
             ts_init,
             ignore_size_updates,
         ),
-        TickType::AskSize => cache.update_ask_size_with_filter(
+        TickType::AskSize | TickType::DelayedAskSize => cache.update_ask_size_with_filter(
             instrument_id,
             size.size,
             price_precision,
@@ -1911,7 +1915,7 @@ fn update_quote_from_price_size_tick(
     ts_init: UnixNanos,
 ) -> Option<QuoteTick> {
     let quote = match price_size.price_tick_type {
-        TickType::Bid => cache.update_bid_price(
+        TickType::Bid | TickType::DelayedBid => cache.update_bid_price(
             instrument_id,
             price_size.price,
             price_precision,
@@ -1919,7 +1923,7 @@ fn update_quote_from_price_size_tick(
             ts_event,
             ts_init,
         ),
-        TickType::Ask => cache.update_ask_price(
+        TickType::Ask | TickType::DelayedAsk => cache.update_ask_price(
             instrument_id,
             price_size.price,
             price_precision,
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @G14rb,

Thanks for the PR and for tracking this down with clear live verification. The mapping changes look right, but there’s one small coverage gap.

`ibapi` decodes a delayed price update that includes size as `TickTypes::PriceSize`, while the new tests only exercise separate `Price` and `Size` ticks. Would you be able to add a regression with delayed bid and ask `PriceSize` ticks when you get a chance, asserting both prices and sizes? That would cover the delayed arms in `update_quote_from_price_size_tick`.

cjdsellers:
@G14rb, Looks like the pre-commit is also failing, but this is easy to fix. Just run the following and push any changes to the same PR:
```
make pre-commit
```
or (if you're on Windows):
```
prek run --all-files
```

```
prek install
```

faysou:
@G14rb do you want to finish the PR ? If you don't have time I can finish it in another PR and you will be credited in the release for the contribution,

G14rb:
@faysou thanks for the help. If you have time, feel free to finish it, I have few issue to solve and it's taking more than expected.

faysou:
Finishing in another PR.

PR #4683, opened 2026-08-08
Bind pong frames to the connection that received the ping

folknor:
A follow-up to the pong handling in #4613, which narrowed the cross-connection window without closing it and disclosed the remainder as separate work.

## Pongs can outlive the connection that asked for them

`send_pong` samples the connection mode and then enqueues `WriterCommand::Send`. That command carries no connection identity and is buffered for replay during reconnection, so a pong can be delivered on the replacement connection: either because reconnection begins between the mode sample and the writer dequeuing the command, or because the buffered frame is drained onto the new sink. A pong belongs to the connection whose ping caused it; the new peer never asked for it.

The reader already knows which connection a ping arrived on, but discards that epoch at the handler callback boundary. Sampling the epoch inside `send_pong` instead does not fix it - a delayed handler task would read the epoch after the replacement is installed and label a stale pong as current.

## Carrying connection identity from the ping to the pong

`EpochPingHandler` receives the originating connection epoch alongside the payload, dispatched through an internal `IncomingPingHandler` enum in the same shape as the existing `IncomingHandler`. `send_pong_on_connection` tags the pong with that epoch and enqueues `WriterCommand::SendPongOnConnection`, which the writer discards during reconnection and drops on an epoch mismatch, and which never enters the replay buffer. Sink replacement and the epoch increment are both serialized in the writer task, so no replacement can interleave between the comparison and the send.

This is opt-in. `PingHandler`, `send_pong`, and every existing constructor are unchanged, so `send_pong` keeps its replayable behaviour for stream-mode and manual callers; ownership is available to callers that ask for it via the new handler and the new epoch-aware constructor. No adapter is affected.

## Testing

`test_pong_is_bound_to_connection_epoch` drives the writer task directly with a recording transport, so there are no sockets, no sleeps, and no scheduling assumptions - command-channel FIFO order and two acknowledgements are the barriers. It enqueues an `Update` installing epoch 1 followed by a pong tagged epoch 0, awaits the update acknowledgement, then uses an acknowledged `SendOnConnection` sentinel as a processing barrier before asserting the stale pong never reached the replacement sink. It then sends a pong tagged epoch 1 with a second sentinel and asserts that one was written, so an implementation that dropped every pong could not pass.

Removing the writer's epoch comparison fails the test with the stale pong recorded ahead of the sentinel. The oversized-payload test now covers both entry points, pinning that validation still precedes the ownership check so an over-long pong during reconnection remains a reported caller error.

### DIFF
--- a/crates/network/src/websocket/client.rs
+++ b/crates/network/src/websocket/client.rs
@@ -85,8 +85,8 @@ use super::{
         GRACEFUL_SHUTDOWN_TIMEOUT_SECS,
     },
     types::{
-        EpochMessageHandler, MessageHandler, MessageReader, MessageWriter, PingHandler,
-        WriterCommand,
+        EpochMessageHandler, EpochPingHandler, MessageHandler, MessageReader, MessageWriter,
+        PingHandler, WriterCommand,
     },
 };
 #[cfg(feature = "turmoil")]
@@ -203,7 +203,7 @@ pub struct WebSocketClientInner {
     config: WebSocketConfig,
     reconnect_headers: ReconnectHeaders,
     handler: Option<IncomingHandler>,
-    ping_handler: Option<PingHandler>,
+    ping_handler: Option<IncomingPingHandler>,
     read_task: Option<tokio::task::JoinHandle<()>>,
     read_fence: Option<ReadSessionFence>,
     write_task: tokio::task::JoinHandle<()>,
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR, and for following this through. Carrying the reader epoch across the ping handler boundary and checking it in the writer task closes the remaining window without changing the existing `PingHandler` or `send_pong` behavior. The regression test covers both stale and current-epoch pongs, and the required checks are green.

This looks good to me.

PR #4684, opened 2026-08-08
Average bet position price by stake on increase

folknor:
`BetPosition` reports a cost basis it never paid once a position is increased at more than one price.

## The position keeps its first price

`BetPosition::position_increase` assigns `self.price` only when the position is empty, so a same-side increase at a different price accumulates exposure while the price stays at the first bet's odds. Every reader of that field then works from the wrong basis: `as_bet` reconstructs the aggregate stake as `exposure / price`, and `flattening_bet`, `unrealized_pnl`, `total_pnl` and `position_decrease` all derive from it.

The result is that profit and loss stops conserving. Backing 100 at 2.0 and then 50 at 4.0 gives an aggregate that should pay 250 on a win, matching the two bets taken separately. Holding the price at 2.0 makes `as_bet` report a stake of 200 and a payoff of 200, losing 50 from the accounting.

## Averaging the odds by stake

The conserving aggregate is the stake-weighted arithmetic mean of the decimal odds. Writing `A` for the absolute exposure and `S` for the total stake, the constituents of a BACK position pay `A - S` on a win while the aggregate pays `A - A/P`; those agree only when `P = A/S`. LAY gives the same result with both signs flipped, and the losing outcome agrees as well. Applied incrementally against the existing state, `P_new = (A + a) / (A/P + s)` for an incoming bet of stake `s` and exposure `a`.

Averaging applies to a same-side bet with a positive current price, incoming price, and stake. `add_bet` routes an opposite side to `position_decrease` and real odds are positive, but `position_increase` is public and `Bet` validates neither field, so anything outside that keeps the price it had rather than mixing signed and absolute quantities or dividing by a cancelled stake. Those conditions also make the denominator positive. An empty position still adopts the incoming price directly, and `position_decrease` is unchanged - its three branches consume the corrected price as they always did.

## Testing

`test_position_increase_back_averages_price_and_conserves_pnl` and its LAY counterpart build the 100 at 2.0 plus 50 at 4.0 position, assert the averaged price, and assert `total_pnl` at a settlement price equals `calc_bets_pnl` over each constituent bet paired with its own hedge - stating the conservation property rather than a precomputed number.

Four more cover the guard, one per condition: an opposite-side bet through the public method, a cancelling stake, a negative stake that still leaves a positive aggregate stake, and a zero or negative incoming price. Each asserts the price is left alone.

The new assertions were checked against the unfixed implementation. Removing the averaging fails the price assertion at 2.0 against 400/150, and fails the conservation assertion at -66.67 against -16.67. Removing the stake condition misprices the position to 7.0 against 2.0, and removing the cancellation condition panics in `rust_decimal` with a division by zero.

### DIFF
--- a/crates/model/src/data/bet.rs
+++ b/crates/model/src/data/bet.rs
@@ -275,9 +275,27 @@ impl BetPosition {
     }
 
     /// Increases the position with the provided bet.
+    ///
+    /// A same-side increase sets the price to the stake-weighted mean of the decimal odds.
     pub fn position_increase(&mut self, bet: &Bet) {
         if self.side().is_none() {
             self.price = bet.price;
+        } else {
+            let abs_self_exposure = self.exposure.abs();
+            let abs_bet_exposure = bet.exposure().abs();
+            // The mean is only meaningful for a same-side bet with well-formed odds:
+            // `add_bet` routes the opposite side to `position_decrease`, but this
+            // method is public, and `Bet` enforces neither a positive price nor a
+            // positive stake. These conditions also guarantee a positive denominator
+            // below. Anything else keeps the price it had.
+            if self.side() == Some(bet.side)
+                && self.price > Decimal::ZERO
+                && bet.price > Decimal::ZERO
+                && bet.stake > Decimal::ZERO
+            {
+                let total_stake = abs_self_exposure / self.price + bet.stake;
+                self.price = (abs_self_exposure + abs_bet_exposure) / total_stake;
+            }
         }
         self.exposure += bet.exposure();
     }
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. The stake‑weighted price is correct for both BACK and LAY, including Betfair’s convention that LAY `size` remains the backer’s stake.

One small test gap: the new `self.price > Decimal::ZERO` guard isn’t covered. Because `Bet::new` accepts unvalidated values, a negative initial price can leave a nonempty position with a negative current price. On the next same‑side increase, removing that guard makes `total_stake` zero and the division panics. Could you add a test for the non‑positive current‑price case when you get a chance?

Otherwise, this looks good to me.

folknor:
Thank you, @cjdsellers - covered in the latest commit.

`test_position_increase_non_positive_current_price_preserves_price` reaches that arm the only way it is reachable: `Bet::new(-3.0, 100.0, Lay)` opens the position at a price of `-3.0` with an exposure of `+300`, so `side()` is BACK. The next bet is same-side with a positive price and stake, which leaves the current price as the only condition rejecting the averaging, and its stake is chosen so the aggregate stake would be `300 / -3 + 100`. Removing `self.price > Decimal::ZERO` fails the test with `Division by zero`.

The existing incoming-price test is renamed to `test_position_increase_non_positive_incoming_price_preserves_price`, so the two read as the two halves of the same guard.

cjdsellers (approved):
Thanks @folknor, looks great 👌

PR #4685, opened 2026-08-08
Retain backtest data past the window end and bound the no-data horizon

folknor:
Two boundary defects in `BacktestEngine::run_impl`: one drops a data point at every streaming window, the other gives a data-less run a horizon in the year 2100.

## The first item past `end_ns` is consumed and discarded

The run loop pulled the next item at the bottom of the body and tested it against `end_ns` at the top of the following pass. By then the item had already left `BacktestDataIterator`, so breaking out of the loop discarded it. Streaming a backtest as consecutive windows therefore lost exactly one data point per boundary, silently.

`BacktestDataIterator` gains a non-consuming `peek`, covering both the single-stream index path and the multi-stream heap path, and the loop now inspects the next item before deciding to take it: an item beyond `end_ns` ends the run while staying in the iterator, so the next window delivers it. The lookahead that drives the timestamp-boundary flush is a peek for the same reason. Ordering is otherwise unchanged - the item is processed, the next timestamp inspected, then `flush_accumulator_events` and `finalize_timestamp` run against the current timestamp.

## A run with no data adopts a year-2100 horizon

With no explicit `end` and no data loaded, `end_ns` fell back to a hardcoded `4_102_444_800_000_000_000`, or 2100-01-01, so any registered timer advanced simulated time across seventy-odd years of nothing. The horizon now falls back to `start_ns`: without data or an explicit end there is no evidence simulated time should move past the requested start. An explicit `end` is still honoured with no data, since that is a deliberate timer-only run, and a run with data still uses the last data timestamp.

## Testing

`test_streaming_windows_retain_first_item_past_end` runs two consecutive windows over one iterator without clearing it between them, asserting all three timestamps arrive. Restoring the old consume-then-test order yields `[1, 2]` against `[1, 2, 3]`.

`test_run_with_empty_data_does_not_advance_recurring_timer` registers a recurring timer on an engine with no data and asserts it never fires and the clock does not advance. Under the old horizon the timer fires and requests shutdown, giving one event instead of none.

`test_peek_does_not_consume_single_stream_item` and `test_peek_does_not_consume_multi_stream_heap_item` cover both cursor implementations, since the fast path advances an index while the merge path pops a heap.

### DIFF
--- a/crates/backtest/src/data_iterator.rs
+++ b/crates/backtest/src/data_iterator.rs
@@ -183,6 +183,18 @@ impl BacktestDataIterator {
         self.rebuild_heap();
     }
 
+    /// Returns the next backtest data element without advancing the stream cursor.
+    pub(crate) fn peek(&self) -> Option<&Data> {
+        if let Some(p) = self.single_priority {
+            let data = self.streams.get(&p)?;
+            let idx = *self.indices.get(&p)?;
+            return data.get(idx);
+        }
+
+        let entry = self.heap.peek()?;
+        self.streams.get(&entry.priority)?.get(entry.index)
+    }
+
     /// Returns the next backtest data element across all streams in replay order.
     pub(crate) fn next_item(&mut self) -> Option<Data> {
         // Fast path for single stream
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
@folknor, Thanks for the PR. I checked the `peek`/consume ordering across both cursor paths, the timer and timestamp-finalization paths, and the current-base interaction. This looks good to me 👍

PR #4686, opened 2026-08-08
Fix Databento MBO snapshot sequence handling

faysou:
### Summary

- Set decoded Databento MBO snapshot sequences to zero so replay snapshots do not advance the order book incremental sequence high-water mark.
- Preserve source packet sequences for incremental MBO records.
- Cover both snapshot and incremental decoding with exact sequence assertions.

### Testing

- `cargo fmt --all --check`
- `CARGO_BUILD_JOBS=16 cargo test -p nautilus-databento test_decode_mbo_msg_snapshot_does_not_advance_sequence -- --nocapture`
- `CARGO_BUILD_JOBS=16 cargo test -p nautilus-databento --lib` (301 passed)
- Commit-time pre-commit hooks passed, including Cargo fmt, Clippy, docs, and repository convention checks.

### DIFF
--- a/crates/adapters/databento/src/decode/market_data.rs
+++ b/crates/adapters/databento/src/decode/market_data.rs
@@ -194,12 +194,20 @@ pub fn decode_mbo_msg(
     let ts_event = msg.ts_recv.into();
     let ts_init = ts_init.unwrap_or(ts_event);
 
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @faysou,

Thanks for the PR. Good catch, this is the right way to handle `sequence` on `F_SNAPSHOT` deltas 👍

PR #4687, opened 2026-08-08
Return no orders for an unknown instrument in the simulated exchange

folknor:
## An unknown instrument returns every instrument's orders

`SimulatedExchange::get_open_orders`, `get_open_bid_orders` and `get_open_ask_orders` each pass the caller's `Option<InstrumentId>` through `and_then` over the `matching_engines` lookup, then fall back to aggregating every engine. `and_then` collapses a missing engine into `None`, which is the same value the caller supplies to mean "no filter" - so `Some(unknown_id)` takes the unfiltered branch and returns every instrument's open orders instead of none. `None` and `Some(known_id)` are both correct; only the unmatched filter is wrong, and it is silent, since the caller cannot tell the superset from a real answer.

Each accessor now matches on the caller's `Option` so the two absences stay distinct: a present ID resolves through `map_or_else(Vec::new, ...)`, and only `None` aggregates. Three parallel `match` bodies rather than a shared helper, since abstracting one short branch behind a function parameter would bury the distinction that caused the defect.

## Testing

`test_open_order_accessors_filter_by_instrument_id` registers two instruments on one venue and rests a bid on the first and an ask on the second - a buy limit at 90.00 against a 100/101 market and a sell limit at 210.00 against 200/201, so neither crosses - then queries a third, unregistered instrument ID. It asserts the unfiltered counts, both known IDs including the side-specific empties, and empty for the unknown ID from all three accessors.

The resting orders on two different engines are what make the unknown-ID assertions non-vacuous: against the previous implementation they return two, one and one orders respectively, and the assertions fail. The same fixture pins the multi-engine aggregation and known-ID paths that stay unchanged.

## PS: an unrelated failure on current develop

Not from this branch, but I hit it validating this one and it is cheap to fix. `cargo test --workspace` fails on `node::tests::test_republish_external_msgbus_message_logs_topic_and_error_chain` (`crates/live/src/node/mod.rs`, added in `6624f9bf07`) with `test logger already installed: SetLoggerError(())`.

The test calls `log::set_logger` unconditionally, and that succeeds at most once per process. `NautilusKernel::new` installs the global logger through `init_logging` (`crates/system/src/kernel.rs:361`), so any earlier test in the same binary that calls `LiveNode::build` has already claimed it. Under nextest each test gets its own process and the call always wins, which is why CI is green; a plain in-process `cargo test --workspace` fails deterministically. Running the test alone also passes.

Either the `serial_tests::` convention or tolerating an already-installed logger would settle it.

### DIFF
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -490,54 +490,57 @@ impl SimulatedExchange {
     }
 
     /// Returns all open orders, optionally filtered by instrument ID.
+    ///
+    /// An instrument ID with no matching engine returns no orders.
     #[must_use]
     pub fn get_open_orders(&self, instrument_id: Option<InstrumentId>) -> Vec<RestingOrder> {
-        instrument_id
-            .and_then(|id| {
-                self.matching_engines
-                    .get(&id)
-                    .map(OrderMatchingEngine::get_open_orders)
-            })
-            .unwrap_or_else(|| {
-                self.matching_engines
-                    .values()
-                    .flat_map(OrderMatchingEngine::get_open_orders)
-                    .collect()
-            })
+        match instrument_id {
+            Some(id) => self
+                .matching_engines
+                .get(&id)
+                .map_or_else(Vec::new, OrderMatchingEngine::get_open_orders),
+            None => self
+                .matching_engines
+                .values()
+                .flat_map(OrderMatchingEngine::get_open_orders)
+                .collect(),
+        }
     }
 
     /// Returns all open bid orders, optionally filtered by instrument ID.
+    ///
+    /// An instrument ID with no matching engine returns no orders.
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. The explicit `match` keeps the caller's `None` distinct from a failed matching engine lookup, and the regression with two engines covers the previous fallback behavior cleanly. The code and current CI look good.

Thanks also for flagging the logger behavior. We support the normal Rust unit and integration suite through `cargo nextest`, which provides the process isolation these tests rely on; plain `cargo test --workspace` isn't a supported gate. The test passes under the supported runner and in CI, so no change is needed for it.

(A follow‑up for me is to document this explicitly as policy rather than leave it as a tip in `README.md`.)

PR #4688, opened 2026-08-08
Fix orderless position cache indexing

### LINKED ISSUE #4628: [Interactive Brokers] Synthetic spread leg fills create invalid order-position cache index
# Bug Report

## Confirmation

- [x] I've re-read the relevant cache and live-execution documentation.
- [x] I've searched existing issues and discussions for `order_position`, synthetic leg fills, generic spreads, BAG legs, and cache integrity.
- [x] I've reviewed the adapter, execution-engine, and cache source paths involved.
- [x] I've reproduced this on the current development wheel `1.231.0a20260731` as well as `1.228.0`.

## Expected behavior

Interactive Brokers generic-spread leg fills should create/update the native per-leg positions needed by the portfolio without making `Cache.check_integrity()` fail.

If a synthetic leg fill intentionally has no corresponding leg `Order`, its synthetic client order ID should not be persisted in an index whose integrity contract requires every key to exist in `Cache._orders`.

## Actual behavior

The IB adapter emits per-leg `OrderFilled` events with client order IDs shaped as:

```text
{parent_client_order_id}-LEG-{leg_symbol}
```

No leg `Order` is created, by design. `ExecutionEngine._handle_leg_fill_without_order()` correctly creates or updates a native per-leg `Position`. When the position is first added, however, `Cache.add_position()` unconditionally calls:

```python
self.add_position_id(
    position.id,
    position.instrument_id.venue,
    position.opening_order_id,
    position.strategy_id,
)
```

That writes the synthetic leg ID into `_index_order_position` (and a configured database adapter persists it), even though `cache.order(synthetic_leg_id)` is `None`.

`Cache.check_integrity()` then returns `False` because every `_index_order_position` key is required to exist in `_orders`.

The resulting native state is internally contradictory:

```text
cache.order(synthetic_leg_id) is None
cache.position_id(synthetic_leg_id) is not None
cache.position(cache.position_id(synthetic_leg_id)) is not None
cache.check_integrity() is False
```

Repeated partial fills correctly update the same NETTING positions, but the invalid order-to-position keys remain.

## Steps to reproduce

This requires no IB connection and no Redis database:

1. Create two `OptionContract` legs and one `OptionSpread` using `new_generic_spread_id`.
2. Add the instruments, a margin account, and one parent spread `Order` to `Cache(database=None)`.
3. Create `ExecutionEngine` with that cache.
4. For each leg, pass an `OrderFilled` through public `ExecutionEngine.process()` using:
   - the leg instrument ID;
   - client order ID `{parent}-LEG-{leg_symbol}`;
   - `position_id=None`;
   - the parent strategy/account IDs.
5. Repeat both leg fills to exercise partial-fill/netting updates.
6. Observe that both per-leg positions are correct, each synthetic leg ID has no cached Order but does resolve to a Position, and `cache.check_integrity()` returns `False`.

The causal assertions in the reproducer are:

```python
for leg in legs:
    synthetic_id = ClientOrderId(f"{parent.client_order_id}-LEG-{leg.id.symbol.value}")
    assert cache.order(synthetic_id) is None
    position_id = cache.position_id(synthetic_id)
    assert position_id is not None
    assert cache.position(position_id) is not None

assert cache.check_integrity() is True  # fails: actual value is False
```

## Code pointers

- IB synthetic leg ID emission: `nautilus_trader/adapters/interactive_brokers/execution.py::_generate_leg_fill`
- Intentional no-order handling: `nautilus_trader/execution/engine.pyx::_handle_leg_fill_without_order`
- Unconditional order index write: `nautilus_trader/cache/cache.pyx::add_position` → `add_position_id`
- Integrity invariant: `nautilus_trader/cache/cache.pyx::check_integrity`, `_index_order_position`

A likely repair boundary is the no-order leg-position creation path: preserve the per-leg Position and its instrument/account/strategy indexes, but do not create an order-to-position entry for a client order ID that intentionally has no Order. I have not fabricated leg Order objects or removed the integrity check because either would hide the invariant mismatch.

## Specifications

- OS platform: macOS arm64
- Python version: 3.13.13
- `nautilus_trader` versions:
  - `1.228.0` from PyPI
  - `1.231.0a20260731` from `https://packages.nautechsystems.io/simple`
- Adapter/venue: Interactive Brokers generic spread / BAG event shape; reproduction itself is no-broker and in-memory

pengpengyi92:

## Summary

Synthetic spread leg fills intentionally create positions without backing orders, but the cache previously persisted and rebuilt order associations for the absent synthetic order.

This change adds an explicit orderless-position path which:

- preserves position, strategy, venue, instrument, and account indexes without creating an invalid `order_position` mapping;
- keeps an empty `position_orders` bucket so orderless position queries and index rebuilds remain valid;
- filters legacy persisted mappings whose client order no longer exists during cache load; and
- replays orderless flips by splitting the source fill consistently with live execution, closing the predecessor, and binding the opening fragment to the following `PositionOpened` replacement.

## Related issues/PRs

Closes #4628

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not applicable.

## Documentation

No documentation or PyO3 binding changes.

## Testing

- `cargo test -p nautilus-model events::order::filled::tests` (10 passed)
- `cargo test -p nautilus-common cache::tests` (282 passed; 2 ignored)
- `cargo test -p nautilus-event-store replay::tests` (44 passed)
- `cargo test -p nautilus-event-store` (all test targets passed)
- `cargo test -p nautilus-execution --test exec_engine` (222 passed)
- `cargo clippy -p nautilus-model -p nautilus-common -p nautilus-execution -p nautilus-event-store --lib --tests -- -D warnings`

Regression coverage includes live and replayed HEDGING flips, subsequent-fill correlation, legacy cache upgrade loading, empty orderless reverse-index buckets before and after `clear_index() -> build_index()`, and cache integrity.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -2656,6 +2656,10 @@ impl Cache {
                 .entry(instrument_id)
                 .or_default()
                 .insert(*position_id);
+            self.index
+                .instrument_orders
+                .entry(instrument_id)
+                .or_default();
 
             // 5: Build index.strategy_positions -> {StrategyId, {PositionId}}
             self.index
@@ -4632,6 +4636,18 @@ impl Cache {
             database.index_order_position(*client_order_id, *position_id)?;
         }
 
+        self.index_position_id(position_id, venue, client_order_id, strategy_id);
+
+        Ok(())
+    }
+
+    fn index_position_id(
+        &mut self,
+        position_id: &PositionId,
+        venue: &Venue,
+        client_order_id: &ClientOrderId,
+        strategy_id: &StrategyId,
+    ) {
         // Index: PositionId -> StrategyId
         self.index
             .position_strategy
@@ -4657,8 +4673,6 @@ impl Cache {
             .entry(*venue)
             .or_default()
             .insert(*position_id);
-
-        Ok(())
     }
 
     // Propagates parent OTO `position_id` to contingent children that are missing one.
@@ -4722,6 +4736,28 @@ impl Cache {
     ///
     /// Returns an error if persisting the position to the backing database fails.
     pub fn add_position(&mut self, position: &Position, oms_type: OmsType) -> anyhow::Result<()> {
+        self.add_position_inner(position, oms_type, true)
+    }
+
+    /// Adds a position whose opening fill intentionally has no backing order.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if persisting the position to the backing database fails.
+    pub fn add_position_without_order(
+        &mut self,
+        position: &Position,
+        oms_type: OmsType,
+    ) -> anyhow::Result<()> {
+        self.add_position_inner(position, oms_type, false)
+    }
+
+    fn add_position_inner(
+        &mut self,
+        position: &Position,
+        oms_type: OmsType,
+        index_order: bool,
+    ) -> anyhow::Result<()> {
         self.positions
             .insert(position.id, SharedCell::new(position.clone()));
         self.index.position_oms.insert(position.id, oms_type);
@@ -4731,12 +4767,21 @@ impl Cache {
 
         log::debug!("Adding {position}");
 
-        self.add_position_id(
-            &position.id,
-            &position.instrument_id.venue,
-            &position.opening_order_id,
-            &position.strategy_id,
-        )?;
+        if index_order {
+            self.add_position_id(
+                &position.id,
+                &position.instrument_id.venue,
+                &position.opening_order_id,
+                &position.strategy_id,
+            )?;
+        } else {
+            self.index_position_id(
+                &position.id,
+                &position.instrument_id.venue,
+                &position.opening_order_id,
+                &position.strategy_id,
+            );
+        }
 
         let venue = position.instrument_id.venue;
         let venue_positions = self.index.venue_positions.entry(venue).or_default();
@@ -4750,6 +4795,10 @@ impl Cache {
             .entry(instrument_id)
             .or_default();
         instrument_positions.insert(position.id);
+        self.index
+            .instrument_orders
+            .entry(instrument_id)
+            .or_default();
 
         // Index: AccountId -> AHashSet<PositionId>
         self.index
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -2347,7 +2347,8 @@ impl Cache {
             .collect();
 
         if let Some(db) = &self.database {
-            self.index.order_position = db.load_index_order_position()?;
+            let order_position = db.load_index_order_position()?;
+            self.index.order_position = self.sanitize_order_position_index(order_position);
             self.index.order_client = db.load_index_order_client()?;
         }
 
@@ -2444,7 +2445,8 @@ impl Cache {
         };
 
         if let Some(db) = &self.database {
-            self.index.order_position = db.load_index_order_position()?;
+            let order_position = db.load_index_order_position()?;
+            self.index.order_position = self.sanitize_order_position_index(order_position);
             self.index.order_client = db.load_index_order_client()?;
         }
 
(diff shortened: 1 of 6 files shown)


cjdsellers:
Hi @pengpengyi92,

Thanks for the PR. The NETTING change correctly removes the invalid `order_position` entry.

**I found three issues to address before merge:**

- With a strategy-level `oms_type=HEDGING` override, removing the mapping also removes the correlation used by `determine_leg_fill_position_id`. In a focused test, two fills for the same synthetic leg ID produced two positions instead of updating the first.
- `add_position_without_order` is not covered by the event-store recovery contract, which causes the current Rust CI failure. Replay also assumes every `OrderFilled` has a cached order, so this path needs explicit recovery handling, not just a classification entry.
- The rebuilt-index behavior needs a regression test covering `clear_index()`, `build_index()`, the absent order mapping, and `check_integrity()`.

Note: the HEDGING correlation contract is not clearly documented, so I'll follow up on that policy and documentation work. Until then, I think this PR should preserve the existing behavior.

When you're able, could you address the recovery and index-rebuild gaps, and preserve HEDGING correlation for now.

cjdsellers:
Hi @pengpengyi92,

Thanks for addressing the earlier feedback.

**I found one remaining HEDGING edge case:**

When an orderless leg position flips, the closed position and the new open position both retain the same `opening_order_id`. `orderless_hedging_leg_position_id` searches all matching positions, so a later fill sees two matches and generates another position instead of updating the open flipped position.

When you're able, could you restrict that lookup to the open position, or otherwise select the current flipped position, and add a regression covering an initial fill, an opposite-side overfill that flips the position, and a subsequent fill for the same synthetic leg ID?

pengpengyi92:
Thanks for catching this, Chris. I restricted the orderless HEDGING correlation lookup to open positions, so after a virtual position flips, the closed predecessor is ignored and the current open flipped position is reused.

I added a regression covering an initial buy fill, an opposite-side sell overfill that flips it, and a subsequent sell fill with the same synthetic leg ID; it asserts that only the closed original and open flipped position remain, and that the later fill updates the flipped position. Verification: the focused regression passes, and cargo test -p nautilus-execution completed with 1031 passed, 0 failed, and 1 ignored.

This change is intentionally limited to this orderless HEDGING flip edge case and does not try to define the broader undocumented correlation policy. I would be very happy to keep discussing or refine the contract if you think the selection semantics should be expressed differently.

cjdsellers:
Hi @pengpengyi92,

Thanks for addressing the HEDGING flip case. Restricting the lookup to open positions fixes the live execution path.

**I found three remaining issues, two blockers and one smaller cache issue:**

- Event-store replay does not recreate the new virtual position opened by an orderless HEDGING flip. It applies the unsplit fill to the original position, then ignores the `PositionOpened` event because the replacement position does not exist in the recovered cache.
- Existing invalid `order_position` entries survive an upgrade. Cache startup reloads the persisted mapping after clearing the in-memory indexes, and index rebuilding does not remove entries whose orders do not exist.
- The orderless path still puts the synthetic ID in `position_orders`, so `orders_for_position` panics for an otherwise valid orderless position.

When you are able, could you make replay reproduce the fill -> flip state, filter or migrate stale persisted mappings during load, and keep an empty `position_orders` bucket for orderless positions?

pengpengyi92:
Thanks Chris — I addressed all three points in `e85c56f`.

1. **Replay:** snapshot-tail replay now detects an orderless overfill, uses the same fill-splitting helper as live execution, applies only the closing fragment to the predecessor, and defers the opening fragment until the matching `PositionOpened` supplies the replacement ID. The recovered fragment preserves proportional commission and causation to the original venue fill; a distinct replacement ID also restores the HEDGING OMS metadata. Ambiguous or unresolved matches fail explicitly, and the single-entry API rejects this cross-entry case before mutating the position.

2. **Upgrade loading:** both `cache_all()` and `cache_orders()` filter persisted `order_position` entries whose client order is absent. This intentionally uses the filter option rather than expanding this PR into a cross-adapter physical migration; valid mappings backed by real orders remain intact.

3. **Reverse index:** orderless positions now receive an empty `position_orders` bucket. `build_index()` only adds IDs backed by cached orders, so `orders_for_position()` returns an empty result before and after rebuild without weakening its strict behavior for genuine corruption.

Regression coverage includes the complete `BUY 1 -> SELL 2 flip -> SELL 1` replay sequence, split quantity/commission/provenance, legacy load filtering through both entry points, and orderless queries across `clear_index() -> build_index()`.

Verification:
- model order-filled tests: 10 passed
- cache module: 282 passed, 2 existing ignored
- event-store replay module: 44 passed; full event-store package passed
- execution engine integration: 222 passed
- repository formatting and strict Clippy for the four affected crates passed

I kept the change scoped to the orderless position lifecycle and did not try to define the broader HEDGING correlation policy or add an adapter schema migration.

cjdsellers:
Hi @pengpengyi92,

Thanks for the thorough follow-up. I rechecked and the earlier replay and cache-index issues are addressed. This now looks ready from my side.

I found one small event-store replay edge case where an opposite-side fill reopening a closed NETTING position is routed through flip handling and fails with `closing quantity was zero`. The fix is a narrow closed-position guard with a regression test, so I don't think it needs to hold this PR. I'll follow that up separately.

Thanks again for working through all the feedback :pray:

cjdsellers (approved):

PR #4690, opened 2026-08-08
Retire Polymarket instruments on venue liveness not endDate

### LINKED ISSUE #4689: Polymarket data client drops markets the venue reports as live
## Expected behavior

The data client keeps an instrument while the venue reports the market as open. The subscription
stays open, and market data continues to flow. The client removes the instrument when the venue
reports the market as closed.

## Actual behavior

The client uses the Gamma `endDate` field to decide when to remove an instrument. It does not read
the venue liveness fields at this point.

There are two results. The client does not load a market that is past its `endDate`, even when the
market is open and trades continue. The client also removes a loaded market at the first expiry
sweep after `endDate`.

The removal takes the WebSocket subscription, the order book, the quote data, and the routing
entry. The refresh does not put the instrument back, and a new subscription fails.

Three exceptions keep a small amount of state, but none lets data flow again: a watchlisted
instrument keeps metadata for settlement, the execution lookup entry stays while open orders or
positions exist, and a subscription to an uncached market starts an auto-load that discards the
result.

## Steps to reproduce

**1. Find a market that the venue reports as open, and whose `endDate` is in the past.**

```bash
curl -s 'https://gamma-api.polymarket.com/markets?closed=false&active=true&limit=100&order=volume24hr&ascending=false' \
  | jq -r '.[] | select(.acceptingOrders and (.endDate < (now|todate))) | "\(.endDate)  \(.slug)"'
```

The result changes continuously as markets close. A sample at 2026-08-08T10:00Z included:

```
2026-06-01T00:00:00Z  will-berhanu-nega-be-the-next-prime-minister-of-ethiopia
```

This market is 68 days past its `endDate`, and its 24-hour volume is more than one million dollars.
The CLOB also reports it as open: `active`, `accepting_orders`, and a book with quotes.

**2. Configure a data client for that market.** Set `instrument_config.market_slugs` to any slug
from step 1.

**3. Look at the log at `DEBUG` level.** The cache stays empty.

```
Parsed 2 instruments from slug queries
Skipping expired instrument ... during live cache publish
Published 0 Polymarket instruments to data engine
```

The client gets the market from Gamma correctly, and then discards it.

The auto-load path has the same fault. Configure a scope that does not contain the market, then
subscribe to its quotes, trades, or book with the `instrument_id`. A `subscribe_instrument` call
alone does not start an auto-load. The log then shows
`Skipping expired auto-loaded instrument ...`.

## Why `endDate` is not the end of trading

This table shows `closedTime` minus `endDate` for 300 markets that closed recently, sampled at
2026-08-08T09:00Z.

| Family | n | median | max | within 1 h |
| --- | --- | --- | --- | --- |
| crypto Up/Down | 110 | 0.02 h | 0.04 h | **110/110** |
| sports | 80 | 4.14 h | 32.27 h | 0/80 |
| weather | 6 | 19.69 h | 19.73 h | 0/6 |
| other | 104 | −165.58 h | 5360.77 h | 0/104 |

To repeat: `GET /markets?closed=true&limit=100&offset=N&order=closedTime&ascending=false` with `N`
= `0`, `100`, `200` (`limit` maximum is 100 for each page). Sort by slug: `-updown-\d+m-` is crypto
Up/Down, `^highest-temperature` is weather, a league code at the start is sports, the rest is other.
The counts change as markets close, but the shape does not.

`endDate` is frequently a value of the parent event: in the top 100 markets, about four of five had
the same `endDate` as their parent. This is usual, but not a rule.

The weather sample above is small, so measure that family separately with
`GET /events?slug=highest-temperature-in-{city}-on-july-{day}-2026`, `day` 20 to 31, `city` =
atlanta, nyc, los-angeles, seattle, chicago, denver, miami, houston, phoenix, boston, philadelphia,
dallas. 108 of the 144 city-days exist, and give 1,188 markets. The minimum gap is 16.55 h, the
median 18.25 h, the maximum 25.22 h. All 1,188 gaps are more than 6 h, and none is less than 1 h.

One more test: for a single day, `endDate` has the same value for 7 cities in 4 US time zones. A
field that does not change with the time zone of the market is not an end time for that market.

## Mechanism

The client reads `endDate` (`http/models.rs:171`), puts it in `PolymarketInstrumentDef`
(`http/parse.rs:167`), changes it to nanoseconds as `expiration_ns` (`http/parse.rs:197-201`), and
gives it to `BinaryOption::new_checked` (`http/parse.rs:216-222`). If the field is absent, or the
client cannot read it, the value is `0`.

The function `is_expired` at `filters.rs:262-263` then controls every path that can keep the market:

- `data/runtime.rs:184` : the sweep that removes the instrument, which operates also when
  `resolve_poll_enabled = false`
- `data/runtime.rs:123` : removes the subscriptions, books, quotes, and `token_meta`
- `data/instruments.rs:64` : the refresh does not put the instrument back
- `data/mod.rs:299` : a new subscription is not possible
- `data/runtime.rs:45` : the token metadata seed skips expired instruments, so message routing does
  not start again when the handler starts again
- `execution/lifecycle.rs:705` : the execution client removes its lookup entry

The venue supplies the necessary data. `parse.rs:139-141` calculates
`active && !closed && accepting_orders` and keeps the result, but no other code reads it. The same
applies to `enable_order_book`.

## Suggested direction

Remove an instrument only when the time is after `expiration_ns` **and** the venue reports the
market as closed. Use the `closed` field. Do not use `acceptingOrders`, because the venue sets that
field to false for short periods while the market stays open.

Keep the `endDate` value in `expiration_ns`. It is correct for Up/Down markets, and settlement uses
it (`resolve/watchlist.rs:94,228,240`). If you remove it, the client removes no instrument at all,
which is a regression against #4272.

One constraint applies to any correction. Gamma does not supply closed markets unless you ask:

```
condition_ids=<closed market>              -> 0 results
condition_ids=<closed market>&closed=true  -> 1 result
condition_ids=<live market>                -> 1 result   (control)
```

A liveness query must therefore ask for closed markets. If it does not, the client sees only that
the market is absent, and absence is a weak indication.

I have an implementation with tests, and a draft PR is ready with the full design. The change controls
liveness on a path that the data client and the execution client both use, so I prefer to agree the
direction first.

## Relationship to existing work

- #4278 : a different fault, on the Python path, with the date-only `endDateIso` field. This report
  shows that `endDate` itself is not an end of trading.
- #4272 : added the expiry sweep. The sweep operates as designed; this report is about its input.
- #4066 : designed the expiration control that uses time. #4253 : gave the reason for the cleanup
  of retained data. I found no other issue for this fault.

## Specifications

- OS macOS 26.5 (arm64) · Python 3.13.14 · `nautilus_trader` 2.0.0rc3.dev20260807 and current
  `develop` · installed from package index wheel and from source · Polymarket Rust v2 data client

mystic-io:
Closes #4689

## Summary

The data client used the Gamma `endDate` field as the only test of liveness. That field is a
scheduled end, and the venue frequently keeps a market open after it. The client thus removed
markets that still traded, and it also refused to load them again.

This change keeps an instrument until two conditions are true: the time is after `expiration_ns`,
and the venue reports the market as closed. The `closed` field supplies the venue state, and the
client reads it at each fetch.

## Details

- The `closed` field goes into the instrument `info` map. Instruments in the cache from before this
  change have no venue state, and they retire on expiration as they did before.
- The expiry sweep re-observes liveness for each carried instrument by `condition_id`. This covers
  markets that auto-load or discovery brought in, which the scope refresh does not touch.
- Gamma does not supply closed markets unless you ask. A condition that is absent from the plain
  lookup thus gets a second lookup with `closed=true`. A market that is confirmed closed retires
  immediately. A condition that is absent from both lookups is a weaker signal, and it must repeat
  three times.
- The probe takes only the venue state from the response. A definition that a `tick_size_change`
  rebuilt thus keeps its precision.
- The probe runs on the instrument refresh interval, not on each sweep tick.
- The execution client uses the same test, so the two clients agree about liveness.
- Retirement now writes an INFO log. The previous path wrote no log at any level.

## Documentation

No PyO3 binding changed, thus the stubs do not change.

## Testing

New tests drive the liveness probe against a mock Gamma server: a market that is confirmed closed,
a condition that is absent from both lookups, a market that stays open, a request that fails, and a
tick size that the venue lowered. Other tests show that settlement and message routing do not
change. `make pre-commit` is green, apart from `test wheel publishing`, which fails in the same way
on a clean `develop`.

I also ran the change against the live venue. The observations are in #4689.

## Limits of the evidence

- Retirement latency includes the propagation delay of Gamma. My measurements are an upper limit,
  not a measurement of this code alone.
- Only unit tests cover the three-lookup limit and the precision behavior. Neither condition
  occurred in the live runs.
- One risk stays open. A condition with empty `clobTokenIds` is held as transient, and thus it does
  not retire. I found no such payload in 500 closed markets, but I also produced no positive
  example, so the sample does not close the question.

### DIFF
--- a/crates/adapters/polymarket/src/data/auto_load.rs
+++ b/crates/adapters/polymarket/src/data/auto_load.rs
@@ -25,7 +25,7 @@ use nautilus_model::{
 use super::{
     PolymarketDataClient,
     instruments::cache_instrument,
-    runtime::{is_instrument_expired, retire_local_instrument_state},
+    runtime::{is_instrument_retirable, retire_local_instrument_state},
     subscriptions::{resolve_token_id_from, sync_ws_subscription_async},
 };
 use crate::{
@@ -216,7 +216,7 @@ impl PolymarketDataClient {
                             continue;
                         }
 
-                        if is_instrument_expired(&inst, clock.get_time_ns()) {
+                        if is_instrument_retirable(&inst, clock.get_time_ns()) {
                             log::debug!("Skipping expired auto-loaded instrument {}", inst.id());
                             retired_expired_ids.insert(inst.id());
                             retire_local_instrument_state(
--- a/crates/adapters/polymarket/src/data/dispatch.rs
+++ b/crates/adapters/polymarket/src/data/dispatch.rs
@@ -1054,6 +1054,7 @@ mod tests {
         market_id: Option<&'a str>,
         condition_id: Option<&'a str>,
         expiration_ns: Option<UnixNanos>,
+        venue_closed: Option<bool>,
     }
 
     fn seed_instrument_with_context(
@@ -1093,6 +1094,13 @@ mod tests {
                 info.insert(
                     "condition_id".to_string(),
                     serde_json::Value::String(condition_id.to_string()),
+                );
+            }
+
+            if let Some(venue_closed) = seed_ctx.venue_closed {
+                info.insert(
+                    "venue_closed".to_string(),
+                    serde_json::Value::Bool(venue_closed),
                 );
             }
 
@@ -1205,6 +1213,22 @@ mod tests {
     fn gamma_market_expired_fixture_value() -> Value {
         serde_json::from_str(include_str!("../../test_data/gamma_market.json"))
             .expect("gamma market fixture json")
+    }
+
+    /// Returns the expired fixture with the venue also reporting the market closed.
+    ///
+    /// Gamma keeps trading many markets past `endDate`, so expiration alone no longer retires an
+    /// instrument. Paths that must reject a dead market need the venue to agree it is dead.
+    fn gamma_market_expired_closed_fixture_value() -> Value {
+        let mut value = gamma_market_expired_fixture_value();
+
+        if let Some(root) = value.as_object_mut() {
+            root.insert("active".to_string(), Value::Bool(false));
+            root.insert("closed".to_string(), Value::Bool(true));
+            root.insert("acceptingOrders".to_string(), Value::Bool(false));
+        }
+
+        value
     }
 
     fn gamma_market_recheck_fixture_value() -> Value {
@@ -2182,6 +2206,7 @@ mod tests {
                 market_id: Some("1778973900"),
                 condition_id: Some("0xCOND-BTC"),
                 expiration_ns: Some(expiration_ns),
+                venue_closed: None,
             },
         );
         let no = seed_instrument_with_context(
@@ -2194,6 +2219,7 @@ mod tests {
                 market_id: Some("1778973900"),
                 condition_id: Some("0xCOND-BTC"),
                 expiration_ns: Some(expiration_ns),
+                venue_closed: None,
             },
         );
 
@@ -2294,6 +2320,7 @@ mod tests {
                 market_id: Some("1778973900"),
                 condition_id: Some("0xCOND-BTC"),
                 expiration_ns: Some(expiration_ns),
+                venue_closed: None,
             },
         );
         let no = seed_instrument_with_context(
@@ -2306,6 +2333,7 @@ mod tests {
                 market_id: Some("1778973900"),
                 condition_id: Some("0xCOND-BTC"),
                 expiration_ns: Some(expiration_ns),
+                venue_closed: None,
             },
         );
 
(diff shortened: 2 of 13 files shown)


cjdsellers:
Hi @mystic-io,

Thank you for the PR. This is quite a large change. Would you be open to splitting it into a few smaller, incremental PRs? I’ll wait until you mark this ready for review before taking a deeper look.

mystic-io:
@cjdsellers sure, I am happy to split this. I will replace this with three smaller PRs.

PR #4691, opened 2026-08-08
Count and surface Polymarket mass-status omissions

seungpyoson:
## Summary

The Polymarket adapter omitted venue orders and positions it could not represent - and holdings it filters as dust - from the mass status it returned, still reported success, and let no count of either omission reach the report. A mass status could be silently partial: an order the venue holds could be absent from the report with the pass still succeeding.

This PR makes omission visible and rejects what cannot be represented:

- Every dropped or degenerate input row is counted exactly once and surfaced in one aggregated per-pass message. Anomalies (unmapped instruments, unrepresentable orders or positions, invalid fill timestamps or values, unowned maker trades, invalid average prices) emit one `error!` per pass; policy dust, withheld Flat reports, and lookback removals `warn!`. The polled order, fill, and position report routes log the same omission counts (warn on any anomaly).
- Unrepresentable order and fill quantities, prices outside the binary unit interval, terminal filled orders with no matched quantity, negative position sizes, and non-zero positions with a missing-interval average price are rejected and counted instead of being coerced into present-but-wrong reports (a zeroed report asserts a state the venue never reported; a position report without an open average price is silently dropped by the live reconciliation engine).
- The mass-status lookback applies to fetched orders and trades before building reports, with saturating cutoff arithmetic; rows with unparsable timestamps are retained so their parsers count them. Removed rows are counted in an aggregated warning, since the cutoff follows the local clock and clock skew otherwise silently filters venue-current rows.
- Explicitly zero-size venue positions now produce `Flat` position reports, so a startup mass status can close a stale cached position (previously the closed position row was dropped and the stale cache entry survived startup indefinitely). Positive dust remains omitted and counted.

## Behavior changes

- Flat position reports are now emitted for explicit zero-size rows (previously dropped). The netting consumer treats them as authoritative zero and closes any stale cached position.
- A Flat report is withheld and counted when the same mass-status pass contains an order or fill report for that instrument, because the Data API positions indexer is a separate service that can lag current CLOB activity; without this guard a stale zero row could close a live position.
- Non-zero positions with an average price outside `(0, 1)` are rejected and counted (previously the average price was cleared and the position retained, which the engine then dropped silently).
- Degenerate values that were previously coerced (zeroed) into reports are now rejected and counted.

## Test plan

- `cargo nextest run -p nautilus-polymarket`: 1,364 passed, 0 skipped (new: per-anomaly counting, flat-withhold guard, invalid-value rejection for taker and maker fills, invalid-average-price rejection, lookback-removal warning, per-pass log severity, polled-route omission summaries).
- `cargo clippy -p nautilus-polymarket --all-targets -- -D warnings`: clean. `prek run --all-files`: all hooks pass.
- Negative controls: each new test was confirmed to fail with its production change reverted, then restored.

## Notes

- This PR and the settlement-pending-fills PR (#4692) both edit `execution/reconciliation.rs` (different concerns: omission counting vs fill building). They are independent and either order merges; the second may need a small rebase, which we will handle promptly.

### DIFF
--- a/crates/adapters/polymarket/benches/data.rs
+++ b/crates/adapters/polymarket/benches/data.rs
@@ -255,7 +255,7 @@ fn bench_order_fill_maker(c: &mut Criterion) {
             let reports: Vec<_> = trade
                 .maker_orders
                 .iter()
-                .map(|order| {
+                .filter_map(|order| {
                     build_maker_fill_report(
                         order,
                         &trade.id,
--- a/crates/adapters/polymarket/src/execution/parse.rs
+++ b/crates/adapters/polymarket/src/execution/parse.rs
@@ -125,7 +125,7 @@ pub fn parse_order_status_report(
     price_precision: u8,
     size_precision: u8,
     ts_init: UnixNanos,
-) -> OrderStatusReport {
+) -> Option<OrderStatusReport> {
     let venue_order_id = VenueOrderId::from(order.id.as_str());
     let order_side = OrderSide::from(order.side);
     let time_in_force = TimeInForce::from(order.order_type);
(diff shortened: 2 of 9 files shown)


cjdsellers (dismissed):
Hey @seungpyoson,

Thank you for the PR. I reviewed the exact current head and found no blocking issues. The omission accounting, lookback filtering, fail-closed parsing, and `Flat` position report path look coherent, and the required checks are green. These are solid changes.

I appreciate you already accounting for the rebase with #4692, and I've reviewed this on that understanding 👌

cjdsellers:
@seungpyoson the PR is currently a draft. Mark it ready for review when you're able and happy for it to be merged. I can also review additional changes.

seungpyoson:
Rebased to a single commit at 59d3950866 after an adversarial review round. Changes since first filing: Flat reports are now withheld (and counted) when the same pass has order/fill activity for the instrument; invalid average prices on non-zero positions are rejected and counted instead of cleared-and-retained; unrepresentable taker/maker fill values are rejected and counted instead of zero-coerced; lookback removals are counted in an aggregated warning; the polled report routes log their omission counts. Battery re-run at this exact head: nextest 1,364 passed / 0 skipped, clippy -D warnings clean, prek all hooks pass. One disclosed residual: a non-zero position with avgPrice absent (rather than present-but-invalid) is still emitted without an open average price; the Data API always returns the field, so this is theoretical.

seungpyoson:

PR #4692, opened 2026-08-08
Reconcile Polymarket settlement-pending fills and their corrections

seungpyoson:
## Summary

The Polymarket adapter discarded non-confirmed trades with no fill report, no counter and no log, and capped order filled quantity against confirmed fills with a zero local-filled floor. A matched-but-unconfirmed trade understated filled quantity; restoring the quantity without the fill would have converted the understatement into a double count when the venue's later confirmation applied on top of an inferred fill.

This PR reports settlement-pending fills and completes their correction path:

- **Settlement-pending fills flow with identity.** MATCHED trades produce fill reports carrying the venue trade id, price, fee and liquidity side, paired with their order reports, so the quantity is restored and the venue's later CONFIRMED update reconciles by trade id instead of stacking. Runtime routes stay confirmed-only; the mass status emits pending fills only paired with an order report, counting and warning on unpaired ones (routine during settlement-window restarts).
- **Withhold-before-cap.** Lookback filtering applies to fetched inputs before building reports, and pending fills are recorded and withheld before the order cap, so a FAILED-noted fill neither ships nor props up filled quantity.
- **FAILED trades void completely and atomically.** The void covers the per-venue-order union of WebSocket-local fills and REST-applied evidence. Evidence is identity-complete: it records the applied fill's own trade id (composite for maker legs), client order id, strategy and order type at apply time, so the correction matches what the engine indexed. When identity or evidence may be incomplete (identity-registry, matched-fill, raw-fill or evidence-cache eviction), the adapter defers loudly and keeps the trade retry-eligible instead of emitting a correction built on defaults; **deferrals never expire on mass-status passes** and are bounded, with one `error!` if the bound evicts. A FAILED trade with no applied fill and no eviction signal is ignored with a warning (previously debug). Correction state keys on the venue trade id alone in dispatch and at restart restore; consumption is tracked per maker leg so an invalid leg — or a maker payload holding no owned leg — leaves its trade retry-eligible without re-emitting applied legs; confirmation side effects defer until a trade is fully processed; corrupt order updates carrying a terminal status still close the order, and a corrupt partial-FAK MATCHED is salvaged as Canceled when `size_matched` parses below `original_size`. Every bounded correction, dedup and identity insert warns on eviction.
- **Replay after dedup eviction cannot double count.** After a WebSocket dedup-cache eviction, a redelivered fill for a tracker-registered order is still emitted for engine-side trade-id dedup but no longer re-accumulates adapter fill quantity or triggers a BUY overfill bump, so the replay is rejected by the engine without having enlarged the order first.
- **WS order prevalidation.** Order updates with unrepresentable sizes or prices outside the binary unit interval are rejected (previously zero-coerced); market-order updates that omit the price or the matched size remain accepted, matching current venue behavior; one shared helper owns the partial-FAK-to-Canceled mapping across REST and WS.
- **Bulk order reports resolve `client_order_id`** from the identity registry with a cache fallback. Previously every bulk report carried `None`, so periodic reconciliation ignored venue-returned orders entirely (and with `open_check_open_only: false` could synthesize `NOT_FOUND_AT_VENUE` rejections for orders the venue did return).
- **Order polling emits exact fills for the queried order only.** The singular `query_order` route emits the fetched confirmed fills belonging to the queried order before the capped status report (FIFO channel, ordering tested), so the engine applies real fills instead of inferring one synthetic fill with a fabricated trade id and no commission. Fills belonging to other orders on the same instrument are filtered out.
- **Collision-resistant maker composite trade ids.** Readable trade prefix plus 16 hex chars of `keccak256(trade_id | venue_order_id)` within the 36-char `TradeId` limit, replacing a truncation under which two maker legs sharing an order-id suffix collided and one leg was silently dropped.

## Behavior changes

- WS order updates with unrepresentable quantity/price are now dropped with a warning instead of reported with zero-coerced values.
- Bulk open-order polling intentionally does **not** emit fills (see notes).
- Maker composite trade id format changes for new fills (see upgrade note).
- Deferral semantics change: safety deferrals for FAILED corrections no longer expire when a mass-status pass completes without re-recording the trade; they remain retry-eligible (bounded, loud on eviction).

## Known limitations

Introduced by this PR:

- REST pending-fill evidence is in-memory; a restart before settlement loses the correction identity for REST-applied pending fills, and the later FAILED is ignored with a warning instead of voiding (documented in the integration guide).
- The replay guard above is deliberately conservative: once any WS dedup eviction has occurred, subsequent fills for tracker-registered orders update the engine but not the adapter's cumulative tracker or overfill bumps until the next mass-status pass re-syncs. Mass-status REST passes rebuild tracker truth, so the divergence is bounded and self-healing.

Pre-existing, changed only in reach:

- An evidence-derived void for a fill the engine never applied (possible only when the manager is configured with `reconciliation_instrument_ids` or `filtered_client_order_ids` exclusions) still relies on the model's unknown-fill void acceptance; this PR adds a second void source reaching that path. Tightening the acceptance is core-crate work we intend to propose separately. Default configurations are unaffected.
- The bulk open-order route still reduces fetched fills to a cumulative quantity (the engine may infer a synthetic fill on that route). The engine-side inference was unreachable at base because bulk reports carried no `client_order_id`; resolving it here makes the route live, so this limitation is newly activated, not merely inherited. Fixing it requires a consumer-side ordering change, intended as a separate proposal.
- A void for an order purged from the cache after closing cannot be addressed (engine-side retention policy, unchanged).

## Upgrade note

Cached orders with maker fills persisted under the old composite format would not dedup against re-emitted fills after upgrade. The REST paths therefore check both the new and the legacy derivation against the cached order's applied fills before emitting; WS redeliveries are suppressed by restored correction state, which keys on the raw venue trade id and is format-independent.

## Test plan

- `cargo nextest run -p nautilus-polymarket`: 1,389 passed, 0 skipped, including real-client WebSocket integration tests driving the full mass-status lifecycle (withhold-before-cap, snap-to-submitted, snapped-evidence-void) and the new correction-semantics tests (deferral survival across passes, identity-eviction deferral, no-owned-leg retry, replay-after-eviction suppression, partial-FAK salvage).
- `cargo clippy -p nautilus-polymarket --all-targets -- -D warnings`: clean. `prek run --all-files`: all hooks pass.
- Each fix carries a regression test confirmed to fail with its production change reverted, then restored.

## Notes

- This PR and the mass-status omission-counting PR (#4691) both edit `execution/reconciliation.rs` (different concerns). They are independent and either order merges; the second may need a small rebase, which we will handle promptly.

### DIFF
--- a/crates/adapters/polymarket/src/common/fifo_ext.rs
+++ b/crates/adapters/polymarket/src/common/fifo_ext.rs
@@ -1,0 +1,122 @@
+// -------------------------------------------------------------------------------------------------
+//  Copyright (C) 2015-2026 Nautech Systems Pty Ltd. All rights reserved.
+//  https://nautechsystems.io
+//
+//  Licensed under the GNU Lesser General Public License Version 3.0 (the "License");
+//  You may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at https://www.gnu.org/licenses/lgpl-3.0.en.html
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------------------------------
+
+//! Loud insertion helpers for bounded correction and identity caches.
+
+use std::{fmt::Debug, hash::Hash};
+
+use nautilus_common::cache::fifo::{FifoCache, FifoCacheMap};
+
+/// Adds `id` to the cache, warning when the insertion evicts the oldest entry.
+pub(crate) fn add_to_fifo_with_eviction_warn<T, const N: usize>(
+    cache: &mut FifoCache<T, N>,
+    id: T,
+    what: &str,
+) -> bool
+where
+    T: Clone + Debug + Eq + Hash,
+{
+    let will_evict = cache.len() == N && !cache.contains(&id);
+
+    if will_evict {
+        log::warn!("{what} cache is at capacity ({N}); evicting its oldest entry");
+    }
+
+    cache.add(id);
+    will_evict
+}
+
+/// Inserts `key` into the cache, warning when the insertion evicts the oldest entry.
+pub(crate) fn add_to_fifo_map_with_eviction_warn<K, V, const N: usize>(
+    cache: &mut FifoCacheMap<K, V, N>,
+    key: K,
+    value: V,
+    what: &str,
+) -> bool
+where
+    K: Clone + Debug + Eq + Hash,
+{
+    let will_evict = cache.len() == N && !cache.contains_key(&key);
+
+    if will_evict {
+        log::warn!("{what} cache is at capacity ({N}); evicting its oldest entry");
+    }
+
+    cache.insert(key, value);
+    will_evict
+}
+
+#[cfg(test)]
+mod tests {
+    use nautilus_common::cache::fifo::{FifoCache, FifoCacheMap};
+    use rstest::rstest;
+
+    use super::{add_to_fifo_map_with_eviction_warn, add_to_fifo_with_eviction_warn};
+    use crate::common::test_logger::{capture_start, records_since};
+
+    #[rstest]
+    fn fifo_insertions_warn_only_when_they_evict() {
+        let log_start = capture_start();
+        let mut cache = FifoCache::<u8, 1>::default();
+
+        add_to_fifo_with_eviction_warn(&mut cache, 1, "test FIFO");
+        add_to_fifo_with_eviction_warn(&mut cache, 1, "test FIFO");
+
+        assert!(
+            records_since(log_start)
+                .iter()
+                .all(|(_, message)| !message.contains("test FIFO cache"))
+        );
+
+        add_to_fifo_with_eviction_warn(&mut cache, 2, "test FIFO");
+
+        let records = records_since(log_start)
+            .into_iter()
+            .filter(|(_, message)| message.contains("test FIFO cache"))
+            .collect::<Vec<_>>();
+        assert_eq!(records.len(), 1);
+        assert_eq!(records[0].0, log::Level::Warn);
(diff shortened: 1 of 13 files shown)


cjdsellers:
Hey @seungpyoson,

Thank you for the PR. I appreciate you accounting for the rebase with #4691. I reviewed this on that understanding.

**I found three blocking correction paths at the current head:**

- `query_order` fetches confirmed fills for the whole instrument and emits fills for unrelated venue orders. This needs to be scoped to the queried `venue_order_id`, with a two-order regression test.
- Once a failed trade has incomplete REST evidence, a later mass-status pass cannot replace it with newly resolved identity data, so the correction remains permanently deferred.
- When a live fill falls back to `FillReport` because order identity is unavailable, the emitted fill is not retained as correction evidence. A later `FAILED` update can complete without emitting `OrderFillVoided`.

I reproduced each path with a focused test against d06a09993ba8. The required checks are green, but these need fixing before merge.

seungpyoson:
Rebased to a single commit at 6bd9a4e8a0 after an adversarial review round. Changes since first filing: query_order now emits fills for the queried order only (previously instrument-wide, which could materialize external orders for uncached orders); safety deferrals for FAILED corrections no longer expire on mass-status passes (the absence-based promotion was unsound) and are bounded with an error on eviction; identity-registry eviction now feeds the deferral gate and raw-applied fills are retained in bounded evidence; a maker payload with no owned leg stays retry-eligible instead of consuming the trade id; after a WS dedup-cache eviction, redelivered fills update neither the adapter tracker nor the order quantity, so a replay is rejected by the engine without side effects; corrupt partial-FAK MATCHED updates are salvaged as Canceled when size_matched parses below original_size; the no-evidence ignore branch is now warn-level. The body re-labels two limitations honestly: in-memory REST evidence is introduced here (not pre-existing), and the bulk-route cumulative inference is newly activated by resolving client_order_id. Battery re-run at this exact head: nextest 1,389 passed / 0 skipped, clippy -D warnings clean, prek all hooks pass (except the wheel-publishing hook, which fails identically on untouched trees in this environment).

seungpyoson:

PR #4693, opened 2026-08-08
Count and surface dropped venue position reports in live reconciliation

seungpyoson:
## Summary

The live execution manager dropped a venue position report whose instrument was absent from its cache with no else branch, no counter and no log, so a mass status the adapter built completely could still lose positions inside the engine. The same silence covered sub-lot quantities, cross-zero legs, unrepresentable values, and periodic discrepancy handling.

This PR makes every such drop visible and closes the silent-acceptance paths:

- Every drop or skip of a venue position report is counted exactly once per reconciliation pass and surfaced as one aggregated warning naming the count; per-site detail logs at debug (no per-report error spam).
- Position reconciliation is scoped to each report's `account_id`, with a fail-closed cache-account gate for received reports (previously reports were reconciled under the mass-status envelope account regardless of their own account). The gate runs before any early return: mass-status position reports are gated up front, and the periodic path validates every venue report's account before quantity matching.
- A hedge position report whose batch fills establish a different quantity than reported is counted as a drop (no corrective order is generated; reconciliation outcomes are unchanged).
- Flat reports remain exempt only when the instrument is uncached and no same-account open position is cached; a flat report for a cached instrument from an unknown account is dropped and counted.
- A cached position that cannot be reconciled because no venue report arrived is counted separately (`unreconcilable_cached_positions`), so the pass message never claims a report was dropped when nothing was received; cross-zero leg failures count their report once.
- The reconciliation completion line now carries both the applied-event count and the dropped-report count on every path (previously the direct mass-reconciliation line omitted drops).
- `LiveNode::start()` now drains pending runner events (including account state) before startup reconciliation, matching the `run()` path, so the account gate's precondition holds on both startup paths.

## Behavior changes (deliberate, fail-closed)

- Venue reports whose price or quantity cannot be represented at instrument precision are now dropped and counted, instead of generating a synthetic reconciliation order with a missing or downstream-clamped price (four sites: periodic difference, cross-zero open/close legs, position creation).
- Position reports for accounts absent from the cache are dropped and counted instead of being reconciled under the envelope account. The gating happens before mass-status fill adjustment, so an unknown-account report can no longer steer synthetic fills on the envelope account.

## Known limitations (pre-existing)

- After `position_check_retries` is exhausted, later periodic passes discard recurring reports without counting (the retry ladder's existing terminal behavior).
- A cached open position whose account was removed from the cache, with no venue report arriving, follows the pre-existing periodic path semantics.

## Test plan

- `cargo nextest run -p nautilus-live`: 555 passed, 2 skipped (new: per-site drop counting, flat exemption both directions, account gating before fill adjustment and before periodic early returns, hedge batch-fill quantity mismatch, startup event draining, completion-line counts).
- `cargo clippy -p nautilus-live --all-targets -- -D warnings`: clean. `prek run --all-files`: all hooks pass.
- Negative controls: each new test was confirmed to fail with its production change reverted, then restored.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -20,7 +20,14 @@
 
 #[cfg(feature = "node")]
 use std::collections::HashSet;
-use std::{cell::RefCell, fmt::Debug, rc::Rc, str::FromStr, sync::LazyLock, time::Duration};
+use std::{
+    cell::{Cell, RefCell},
+    fmt::Debug,
+    rc::Rc,
+    str::FromStr,
+    sync::LazyLock,
+    time::Duration,
+};
 
 use indexmap::{IndexMap, IndexSet};
 use nautilus_common::{
@@ -89,6 +96,17 @@ static TAG_RECONCILIATION: LazyLock<Ustr> = LazyLock::new(|| Ustr::from("RECONCI
 pub type InstrumentAccountKey = (InstrumentId, AccountId);
 type FillKey = (AccountId, InstrumentId, TradeId);
 
+enum CrossZeroLegReportError {
+    Quantity,
+    Price,
+}
+
+fn position_reconciliation_applied(events: &[OrderEventAny]) -> bool {
+    events
+        .iter()
+        .any(|event| matches!(event, OrderEventAny::Filled(_)))
+}
+
 #[expect(clippy::too_many_arguments)]
 fn build_cross_zero_leg_report(
     instrument: &InstrumentAny,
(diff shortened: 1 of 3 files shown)


seungpyoson:
Heads-up on scope before this leaves draft: although framed as drop-counting, this PR deliberately changes shared live-node behavior, and I would rather name that upfront than have it surface in review:

1. A fail-closed account gate in the live execution manager: position reports whose account_id is absent from the cache are dropped and counted instead of reconciling under the mass-status envelope account. The gate runs ahead of mass-status fill adjustment and ahead of the periodic path's early returns. I audited all 13 live adapters: every PositionStatusReport is built from the registered client account_id, so no currently-reconciling deployment shape changes behavior.
2. LiveNode::start() now drains pending runner events (including account state) before startup reconciliation, matching the run() path; without it the gate's account-in-cache precondition held on only one of the two startup paths.
3. A hedge position report whose batch fills establish a different quantity than reported is now counted as dropped (no corrective order either way; outcomes unchanged).

If you would rather review the account gate as its own PR, or settle the shape in an issue first, say so and I will split it. Battery re-run at this exact head (5623f19472): nextest 555 passed / 2 skipped, clippy -D warnings clean, prek all hooks pass. Leaving this as draft until one of #4691/#4692 lands, to keep at most two active PRs from us.

seungpyoson:

PR #4698, opened 2026-08-09
Isolate indicator failures from sibling indicators and actor handlers

folknor:
A single registered indicator returning `Err` currently suppresses both its sibling indicators and the actor's own data handler for that data point.

## Registry fan-out aborts at the first failure

`Indicators::handle_quote` drove the per-instrument indicator list with `?` inside the loop, so the first `Err` returned and every indicator registered after the failing one was skipped for that quote; `handle_trade` and `handle_bar` have the same shape. The bulk forms propagate per element, so one failing element also abandoned the remainder of a historical batch. The registry now completes the whole fan-out - across the remaining indicators for a failed item, and across the remaining items of a batch - and returns one aggregate error carrying the first three distinct messages plus the total failure count.

## Actor handlers treated the failure as fatal for the data point

`DataActor::handle_quote` logged the registry error and returned before `on_quote`, so the subscribed handler never ran. Same at `handle_trade`, `handle_bar`, and the three `handle_*_response` paths, each returning before its `on_historical_*` call. The `return` is dropped at all six sites: the error is logged and delivery continues.

## Behaviour change

Handlers now run after an indicator update fails, so a handler reading the failed indicator can observe stale or partially updated state where previously it did not run at all. The failure is logged before the callback. With shutdown-on-error armed, the trigger is recorded synchronously but drained asynchronously, so a handler can still act on the datum that produced the error.

## Testing

Registry: `test_handle_{quote,trade,bar}_continues_after_indicator_error` register tracking/failing/tracking in that order and assert both trackers received the item while the call still reports an error. The bulk equivalents use two items with the failure on the first, asserting both trackers received both. `test_handle_quote_error_retains_distinct_details_up_to_the_bound` asserts the first three distinct messages appear and the fourth is accounted for by the count; `test_handle_quote_error_collapses_identical_details_and_counts_all` asserts four identical failures collapse to one detail with all four counted.

Actor: `test_{quote,trade,bar}_handler_runs_after_indicator_error` register one failing indicator on a running actor and assert the handler ran once with that item. `test_historical_{quotes,trades,bars}_handler_receives_complete_response_after_indicator_error` assert the historical handler ran and received the complete slice rather than a prefix.

Both halves were checked against the unfixed code independently: restoring the actor-side returns fails all six actor tests, and making the registry stop at the first failure fails all six continuation tests on the count of the indicator registered after the failing one.

PS: this stops at not losing data. A repeatedly failing indicator keeps being called and keeps failing, and the actor carries on. Automatic deregistration, or degrading the actor after a number of failures, is a policy decision rather than part of this defect, so I left it alone - if you have a view on what the behaviour should be there, I am happy to do it as a follow-up.

### DIFF
--- a/crates/common/src/actor/data_actor.rs
+++ b/crates/common/src/actor/data_actor.rs
@@ -900,7 +900,6 @@ pub trait DataActor: Component {
 
         if let Err(e) = self.core().handle_indicators_for_quote(quote) {
             log_error(&e);
-            return;
         }
 
         if self.not_running() {
@@ -922,7 +921,6 @@ pub trait DataActor: Component {
 
         if let Err(e) = self.core().handle_indicators_for_trade(trade) {
             log_error(&e);
-            return;
         }
 
         if self.not_running() {
@@ -944,7 +942,6 @@ pub trait DataActor: Component {
 
         if let Err(e) = self.core().handle_indicators_for_bar(bar) {
             log_error(&e);
-            return;
         }
 
         if self.not_running() {
@@ -1228,7 +1225,6 @@ pub trait DataActor: Component {
 
         if let Err(e) = self.core().handle_indicators_for_quotes(&resp.data) {
             log_error(&e);
-            return;
         }
 
         if let Err(e) = self.on_historical_quotes(&resp.data) {
@@ -1246,7 +1242,6 @@ pub trait DataActor: Component {
 
         if let Err(e) = self.core().handle_indicators_for_trades(&resp.data) {
             log_error(&e);
-            return;
         }
 
         if let Err(e) = self.on_historical_trades(&resp.data) {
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @folknor,

Thank you for the PR, and for the detailed explanation and tests. This is a valid failure-isolation approach, but it doesn't match the failure model we want for NautilusTrader.

Our [data integrity and fail-fast policy](https://github.com/nautechsystems/nautilus_trader/blob/develop/docs/concepts/architecture.md#data-integrity-and-fail-fast-policy) prioritizes data integrity over availability for trading operations and avoids allowing invalid state to propagate. We apply that principle here: an indicator returning an error indicates broken indicator code, invalid configuration, or an upstream data issue that should be repaired. Our intended response is to fail closed and bring the node down, then fix the indicator or upstream adapter. We don't want to automatically deregister the indicator or degrade the actor while allowing it to continue.

Supporting partial continuation also adds an aggregate-error contract and changes how sibling indicators, historical batches, actor callbacks, and shutdown behave. We don't think the value of avoiding cascading indicator failures justifies that additional behavioral and maintenance complexity, so we won't merge this change.

PR #4700, opened 2026-08-09
Restrict portfolio greeks to open positions

folknor:
`GreeksCalculator::portfolio_greeks` documents itself as aggregating open positions but queries the full position index.

## A closed position can abort the whole call

`portfolio_greeks` binds `cache.positions(...)` to a local named `open_positions`. `Cache::positions` returns every position, open and closed, while `Cache::positions_open` sits directly beside it. A closed position has `signed_qty == 0.0`, so `quantity * &instrument_greeks` contributes exactly nothing to the total - but `instrument_greeks` is still called for it and its error propagates, so a position the portfolio no longer holds can fail an entire portfolio snapshot once its price data is gone or its contract has expired. The query now uses `positions_open` with the same filter arguments, which also stops greeks being computed for closed positions and multiplied by zero.

Error propagation for positions actually held is unchanged.

## Behaviour change with side = Flat

`portfolio_greeks(side = PositionSide::Flat)` previously selected closed positions, computed their greeks and returned zero or an error. It now selects nothing and returns zero. An open position cannot be flat under the `Position` invariants, so this removes behaviour that contradicted the method's own contract. `NoPositionSide`, `Long` and `Short` are unaffected.

## Testing

`test_portfolio_greeks_ignores_closed_position_with_missing_price` holds one open option position and one closed futures position whose instrument is registered without a price. The closed position is built the way the engine builds one - opened, offsetting fill applied, then `cache.update_position` - and the test asserts it satisfies `is_closed()` and has left the open index before calling. Against the unfixed query it fails with `No price available for CLOSED.GLBX`; with the fix it returns the open position's aggregate.

`test_portfolio_greeks_preserves_open_position_aggregate_and_side_filters` characterizes the success path over a long and a short position with non-unit quantities, comparing every `PortfolioGreeks` field against an independently computed `signed_qty * greeks` sum, asserting those expectations are non-zero first, then repeating for the `Long` and `Short` filters. It passes against the unfixed code by design - a closed position already contributed zero, so the open-book aggregate is unchanged - and the missing-price test is the one that fails without the fix.

### DIFF
--- a/crates/common/src/greeks.rs
+++ b/crates/common/src/greeks.rs
@@ -974,7 +974,7 @@ impl GreeksCalculator {
         let side = side.unwrap_or(PositionSide::NoPositionSide);
 
         let cache = self.cache.borrow();
-        let open_positions = cache.positions(
+        let open_positions = cache.positions_open(
             venue.as_ref(),
             instrument_id.as_ref(),
             strategy_id.as_ref(),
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. The semantic change is coherent: `portfolio_greeks` operates on open positions, and an open position cannot have `PositionSide::Flat`, so returning zero for that filter is the natural empty aggregation. Using the existing open-position query also fixes the stale closed-position failure without adding any new production complexity.

One non-blocking nice-to-have: could you add a direct assertion that `PositionSide::Flat` returns zero if you get a chance? The PR explicitly describes that behavior change, so it would be useful to pin it in the regression coverage.

This looks good to me, and I see nothing blocking.

PR #4701, opened 2026-08-09
Clear option greeks on cache reset

folknor:
`Cache::reset` clears the other market-derived state maps but leaves `option_greeks` populated.

## Option greeks survive a reset

`reset` clears roughly twenty maps, `greeks` and `yield_curves` among them, but never `option_greeks` - a separate map written by `add_option_greeks` and read by `option_greeks`. A second backtest run in the same process therefore starts with the previous run's option greeks readable. The per-instrument purge path already removes an entry from that map, so it is maintained everywhere except the wholesale reset. One `self.option_greeks.clear()` beside the existing `greeks` clear.

## Testing

`test_reset_clears_option_greeks_even_when_instruments_retained` mirrors the existing `test_reset_clears_mark_xrate_even_when_instruments_retained`: it builds the cache with `drop_instruments_on_reset(false)`, adds an instrument and a non-default `OptionGreeks`, asserts the stored value reads back field for field, resets, then asserts the instrument survives while `option_greeks` returns `None`. Retaining the instrument is what pins the lifecycle distinction - reusable definitions outlive a reset, run-specific market state does not. The final assertion fails against the unfixed `reset`.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -3876,6 +3876,7 @@ impl Cache {
         self.position_snapshots.clear();
         self.position_snapshot_revisions.clear();
         self.greeks.clear();
+        self.option_greeks.clear();
         self.yield_curves.clear();
 
         if self.config.drop_instruments_on_reset {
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I checked the cache reset path, the regression, and the merge with current `develop`. Clearing `option_greeks` here is correct, the test covers the retained-instrument lifecycle, and all required checks pass. This looks good to merge.

PR #4704, opened 2026-08-09
Make Polymarket reconciliation authoritative

seungpyoson:
## Summary

Polymarket REST reconciliation currently builds a mass status from independently indexed order, trade, and position rows even when some venue evidence cannot be represented. This change makes that construction an atomic authority boundary:

- A typed omission ledger accompanies every report collection. Material omissions reject the complete mass status instead of emitting partial state; policy omissions remain visible in one collection summary.
- One fallible parser contract rejects unrepresentable or degenerate order and fill values. The same contract is used by mass, polling, FOK recovery, and maker WebSocket paths.
- Settlement-pending trades make the REST snapshot unavailable. The existing user WebSocket remains the sole provisional fill and FAILED-correction lifecycle.
- Explicit zero Data API rows remain visible as omitted evidence but cannot synthesize Flat closes, because the independently indexed row has no freshness/version proof.
- Maker WebSocket trade validation is atomic: every owned leg is valid before any leg mutates fill state, and rejected trades remain retryable.

## Supersession

This replaces the per-instrument Flat guard proposed in #4691 with a snapshot-level authority contract. It intentionally does not adopt #4692’s second pending-fill lifecycle, replay heuristics, compatibility keys, or additional eviction state; those mechanisms are unnecessary when provisional ownership remains singular.

## Verification

Exact-head evidence is posted separately so this body remains stable as the branch evolves.

### DIFF
--- a/crates/adapters/polymarket/benches/data.rs
+++ b/crates/adapters/polymarket/benches/data.rs
@@ -197,7 +197,8 @@ fn bench_order_event(c: &mut Criterion) {
                 px_prec,
                 sz_prec,
                 ts_init,
-            );
+            )
+            .expect("benchmark fixture order should be valid");
             black_box(report);
         });
     });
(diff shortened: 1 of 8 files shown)


seungpyoson:
Closed because this replacement was published before the full local adjudication set was complete. Work will remain private to the fork until the complete design and verification gate pass.

PR #4706, opened 2026-08-09
Retain open Polymarket markets past endDate

mystic-io:
## Summary

Observed: Polymarket can keep a market open after its Gamma `endDate`. One captured market was 68 days past that date while Gamma still reported `closed=false` and `acceptingOrders=true`. The existing expiry sweep removed cached live state based on time alone.

From code inspection: `expiration_ns` was the only cleanup gate. This PR records Gamma’s `closed` state, rechecks expired cached markets through the open and closed lookups, and removes their live state only after Gamma positively reports `closed=true`. Loading and execution paths remain in separate follow-up changes.

## Related issue

Related to #4689

## Verification

I verified this with:

- Live Gamma check: 12 of 12 closed markets were absent from the open lookup and present with `closed=true`
- Live lifecycle run: six instruments were removed after positive closure responses, while four past-end open controls remained
- `cargo test -p nautilus-polymarket --lib --tests`: passed
- `cargo nextest run -p nautilus-polymarket --features python`: 1,349 passed, 2 skipped
- `make pytest`: passed
- `cargo clippy -p nautilus-polymarket --all-targets -- -D warnings`: passed through pre-commit
- `cargo +nightly fmt -p nautilus-polymarket -- --check`: passed through pre-commit
- `make pre-commit`: all hooks passed except `test-wheel-publishing`, which failed with the known stale index link fixture

### DIFF
--- a/crates/adapters/polymarket/src/data/dispatch.rs
+++ b/crates/adapters/polymarket/src/data/dispatch.rs
@@ -486,6 +486,8 @@ fn handle_market_message(message: MarketWsMessage, ctx: &WsMessageContext) {
                 return;
             }
 
+            drop(instruments);
+
             log::debug!(
                 "Tick size changed for {}: {} -> {}",
                 change.asset_id,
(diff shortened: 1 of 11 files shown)


cjdsellers:
Hi @mystic-io,

Thank you for the PR.

**I found two data lifecycle gaps that need addressing before this merges:**

- The new `is_expired_and_not_reported_open` predicate is only used during runtime seeding and retirement. `ensure_live_subscription_allowed` and `ensure_market_data_request_allowed` still use the time-only expiry check, so a retained market with `closed=false` rejects later quote, trade, and book subscriptions, plus book and trade requests. This is one of the failure paths described in #4689. Could these guards use the same closure-aware rule, with regression coverage for an expired cached instrument carrying `info.closed=false`?
- `refresh_expired_market_closure` is throttled by `update_instruments_interval_mins`, which defaults to 60 minutes, while retirement still ticks every 30 seconds. A production instrument can therefore retain subscriptions, books, quotes, routing entries, and WebSocket state for almost an hour after Gamma closes it. The cleanup and churn tests currently omit `closed`, so they exercise `None` and still take the immediate-retirement path. Could closure detection keep a suitably short cadence, with a regression covering the real `closed=false` to `closed=true` transition?

The positive closure lookup itself looks sound, and CI is green. Once these two lifecycle paths are covered, I think the change should be in good shape.

mystic-io:
@cjdsellers Thanks Chris, both points are addressed. I've applied both fixes and added regression coverage for each case. Local tests are passing and it’s ready for another look

cjdsellers:
@mystic-io, Thanks for addressing both lifecycle gaps. The guard change and move off the hourly refresh cadence are now in place.

**I found two further blockers:**

- Adding `closed` in `build_info_json` also changes `PolymarketDataLoader`, so historical instruments now expose their current terminal state through `instrument.info`. This conflicts with the loader contract that keeps resolution-bearing fields in `resolution_metadata`, and can leak future state into a backtest. Could the live client retain the closure state without adding it to historical loader instruments?
- `refresh_expired_market_closure` and the WebSocket tick-size path both rebuild and replace the whole cached instrument from an earlier snapshot. If they overlap, the last write can either revert `closed=true` and delay retirement, or discard the new tick size and publish stale instrument data. Could these updates compose atomically against the latest cached value?

There are also two smaller points worth fixing in the same pass: the extra elapsed-time gate makes closure refresh run on every second resolve tick, and an error in a later condition-ID chunk discards positive closure results from earlier chunks.

I expect that after the next round, I can address any remaining P3s locally, but let's see where it lands.

mystic-io:
@cjdsellers Changes implemented! Ready for another look whenever you have time!

cjdsellers (approved):
Hi @mystic-io,

Thanks for the follow-up and for addressing the remaining points. I've reviewed 5de44fe, the previous blockers are resolved, and CI is green.

There is one remaining P3 around out-of-order `DataEvent::Instrument` delivery if the closure refresh and a tick-size update overlap. It doesn't block this change, and we'll follow it up separately on our side.

This is good to merge.

PR #4709, opened 2026-08-10
Correct the f32 exp approximation outside its normal exponent range

folknor:
`<f32 as BlackScholesReal>::exp` in `crates/model/src/data/black_scholes.rs` reconstructs its `2^k` scale factor from raw IEEE-754 bits without checking that `k` fits an exponent field.

## Out-of-range exponent reconstruction

The approximation ends with `r.mul_add(res, 1.0) * f32::from_bits(((k + 127) as u32) << 23)`, where `k` is the range-reduction exponent. Once `k` leaves `[-126, 127]` the biased value is no longer a valid exponent field: it goes negative, the `as u32` wraps, and the shift writes those bits into the sign and exponent. The corruption then depends on `k` rather than taking one form. Against the correctly rounded `f32::exp`, `k = -127` gives exactly zero where `7.54e-39` is right, `k = -128` gives `-inf` where `3.88e-39` is right, `k = -129` gives `-2.31e+38` where `1.99e-39` is right, and on the positive side `k = 128` gives `inf` where the finite `2.72e+38` is right. A large enough argument additionally saturates the `f32 -> i32` conversion of `k`, which leaves `r` unreduced and the polynomial itself invalid, so repairing only the scale factor would not be sufficient.

`exp` now classifies `k` before evaluating the polynomial: `k <= -151` returns zero, since every exact result there is below half the smallest subnormal, and `k >= 129` returns infinity. The ordinary `-126..=127` range keeps the existing single-factor expression, so arithmetic on the common path is unchanged. The two remaining intervals, `-150..=-127` and `128`, straddle the subnormal and overflow boundaries and are computed by splitting `2^k` across two valid biased exponents, applied to the polynomial one after the other. Pre-multiplying the two scale factors instead underflows to zero at `k = -150` before the polynomial can round the result up into the subnormal range, and a comment records that so the multiply is not folded later.

The trait method is reached from the generic kernels through the discount and carry factors, `-r*t` and `(b-r)*t`, in `compute_greeks`, `compute_iv_and_greeks` and `pricing_kernel`. Reaching the negative tail needs `r*t >= 87.7`, which realistic rates and tenors do not produce, and the `cdf_with_pdf` calls resolve to the inherent `f32::exp` rather than to this method. So this corrects an exported numerical primitive; it does not change option greeks under ordinary inputs.

## Testing

`test_exp_exponent_boundaries` pins both sides of the `k = -126/-127` and `k = -127/-128` transitions, a point inside `k = -150` that must round to the smallest subnormal, and the finite and overflowing sides of the positive cutoff near `ln(f32::MAX)`. Each case is compared against `f32::exp` on classification, sign and a three-ULP bound; bit distance is used rather than relative error because the subnormal results carry progressively fewer significand bits. The `k = -150` case is what catches a later rewrite that pre-multiplies the two scale factors.

`test_exp_tail_sweep` walks the argument range that spans every affected exponent and asserts the result is non-negative, monotonically non-decreasing, never NaN, and equal to `f32::exp` within the same bound. `test_exp_extreme_inputs` covers the saturating conversions and the infinities.

All three fail against the unfixed implementation, reporting a premature zero at the first boundary, `-2.9e31` for `exp(-105)`, and `inf` for `exp(-1e10)`. The existing accuracy tests pass with their tolerances unchanged, which is the check that the common path is untouched.

## Question on the `cdf_with_pdf` binding

`cdf_with_pdf` calls the inherent `f32::exp` twice per calculation, while the discount and carry expressions call this approximation three times. Was that distinction intentional? I have left the behaviour as it is, since changing it would move numerical output on the pricing path.

### DIFF
--- a/crates/model/src/data/black_scholes.rs
+++ b/crates/model/src/data/black_scholes.rs
@@ -128,6 +128,15 @@ impl BlackScholesReal for f32 {
             std::f32::consts::LOG2_E,
             if self > 0.0 { 0.5 } else { -0.5 },
         )) as i32;
+
+        if k <= -151 {
+            return 0.0;
+        }
+
+        if k >= 129 {
+            return Self::INFINITY;
+        }
+
         let r = self - (k as Self * 0.693_145_75) - (k as Self * 1.428_606_8e-6);
         let mut res = 0.001_388_89_f32;
         res = r.mul_add(res, 0.008_333_33);
@@ -135,7 +144,19 @@ impl BlackScholesReal for f32 {
         res = r.mul_add(res, 0.166_666_67);
         res = r.mul_add(res, 0.5);
         res = r.mul_add(res, 1.0);
-        r.mul_add(res, 1.0) * Self::from_bits(((k + 127) as u32) << 23)
+        if (-126..=127).contains(&k) {
+            r.mul_add(res, 1.0) * Self::from_bits(((k + 127) as u32) << 23)
+        } else {
+            // Split 2^k across two valid biased exponents; a single one would be
+            // out of range here. Apply them in sequence: pre-multiplying the two
+            // factors underflows to zero at k = -150 before the polynomial can
+            // round the result up into the subnormal range.
+            let ka = k >> 1;
+            let kb = k - ka;
+            r.mul_add(res, 1.0)
+                * Self::from_bits(((ka + 127) as u32) << 23)
+                * Self::from_bits(((kb + 127) as u32) << 23)
+        }
     }
 
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


faysou:
Regarding Question on the cdf_with_pdf binding if you find some optimisation that makes sense like what I see you have been doing recently, you can do it, thanks.

faysou:
I think using the full exp was not intentional, the whole purpose of these approximations is to be fast and replace full approximations, so using the full exp in another approximation of cdf doesn't make sense. If I were you I would fix this in this PR as well. Sorry for not looking at it more yesterday, I was travelling during the whole day.

The exp in cdf case can also be solved in another PR, it's up to you.

folknor:
@faysou Thank you - agreed the inherent `f32::exp` there is not deliberate. I will take it as a separate PR rather than folding it in, since it moves numerical output on the pricing path and this one deliberately does not: the ordinary `-126..=127` range keeps the original single-factor expression, and the description claims greeks are unchanged under realistic inputs.

The two are ordered, though. `cdf_with_pdf` computes `exp(-0.5 * x * x)`, so `|x| >= 13.2` drives `k` below `-127`, and routing that call through the approximation before this lands would feed wrapped-exponent results into the pdf. `|d1|` reaches that magnitude when `vol * sqrt(t)` is small.

On other optimisations in the same file: `ln` has the mirror-image issue. It does the same bit manipulation as `exp` and assumes a positive, finite, normal `f32` without checking any of it. The sign bit is ignored, so a negative input returns `ln(|x|)` rather than NaN; subnormals are read as normal mantissas and come back roughly 15 natural-log units off; and `ln(0.0)` gives about `-88.03` instead of `-inf`. Reachability is weaker than the exp case - normal positive `s` and `k` protect the ordinary path, and the argument is `(s / k).ln()` - so I would file it separately after this one, unless you would rather it rode along.

faysou:
You can work on these findings thank you, I'm still working a lot on a data side refactor and new features of nautilus. This should get merged or at least PRed in the next weeks under the supervision of @cjdsellers (I mean I did the work, but I'll wait for his approval on which PRs to do).

faysou:
The pricing code is designed to work with simd when simd will land in rust stable. Who knows it could be useful for market making or risk management like computing some portfolio margins like what CME does for example which could be useful in backtesting.

cjdsellers:
Hey @folknor and @faysou,

I've reviewed this and it looks good to merge as scoped. I agree with the split: let's keep this PR limited to correcting `BlackScholesReal::exp`, then handle the `cdf_with_pdf` binding in a separate PR after this lands, with accuracy comparisons and before-and-after benchmarks since it changes the pricing path.

The `ln` special-value and subnormal handling should also stay in its own focused PR. Let's retain the `BlackScholesReal` abstraction for future SIMD implementations.

cjdsellers (approved):

folknor:
@faysou @cjdsellers Thank you both - I implemented the `cdf_with_pdf` dispatch change and ran the accuracy and benchmark comparisons you asked for. The results argue against making it, so they are here rather than in a pull request.

The binding is as described: inside `impl BlackScholesReal for f32` the receiver of `(-0.5 * self * self).exp()` is a concrete `f32`, so method resolution takes the inherent `f32::exp` and the trait approximation defined 47 lines above is never called. The change is one line, `<Self as BlackScholesReal>::exp(-0.5 * self * self)`.

## Accuracy

Maximum PDF error against an f64 reference evaluated from the promoted input, sampled over `[-16, 16]` at a step of 1/1024:

| \|x\| | inherent `f32::exp` | trait `exp` |
| --- | --- | --- |
| [0, 3) | 2 ULP | 2 ULP |
| [3, 6) | 9 ULP | 9 ULP |
| [6, 9) | 33 ULP | 33 ULP |
| [9, 12) | 55 ULP | 55 ULP |
| [12, 16) | 56 ULP | 56 ULP |

Identical maxima in every band. The error is dominated by the f32 formation of `-0.5 * x * x`, not by either exponential: a relative perturbation of `2^-24` in the exponent `y` reaches the result as roughly `|y| * m / 2` ULP, which is 40-80 ULP by `y = -85`. The two exponentials differ on 120 of 32,769 samples by at most 2 ULP, and the composed CDF moves by at most 16 ULP.

## Benchmarks

`black_scholes_criterion`, 300 samples, 10s measurement time, each side measured twice with under 0.3% spread between repeats:

| | change |
| --- | --- |
| `compute_iv_and_greeks_f32` | +9.3% |
| `compute_greeks_f32` | +12.2% |

The approximation is slower than the library call it would replace on this path, which inverts the premise for the change: same accuracy, roughly 10% of the two central pricing operations.

On the SIMD case, `cdf_with_pdf` is a required trait method with no default body, so an `f32x8` implementation supplies its own and never inherits this call site. Correcting the scalar binding does not change what a future vector type does.

My reading is that the inherent call should stay. If you would like the site to stop reading as an oversight, I can send a one-line comment recording that the inherent `f32::exp` is deliberate there, with these numbers as the reason. I am also happy to open the dispatch change as a pull request if you would rather have the consistency at that cost.

faysou:
Thank you, it would actually be interesting to check if changing exp to the inherent exp everywhere is faster.

PR #4710, opened 2026-08-10
Validate child instrument IDs in OrderBookDeltas

folknor:
`OrderBookDeltas` pairs a wrapper `instrument_id` with children that each carry their own, and nothing requires the two to agree.

## Children are never checked against the wrapper

`OrderBookDeltas::new_checked` validates only that the vector is non-empty, then takes `flags`, `sequence`, `ts_event` and `ts_init` from the last child. `OrderBook::apply_deltas` compares the wrapper against the book and delegates to `apply_deltas_unchecked`, which applies each child through `apply_delta_unchecked`, so no child's instrument ID is compared to anything. The singular `apply_delta` does check its delta's ID before delegating; the batch path has no equivalent.

`new_checked` now rejects a batch whose children disagree with the wrapper. The comparison is pointer-first with a string-value fallback, because `InstrumentId` equality is `Ustr` pointer equality and that is the reason `apply_delta_unchecked` exists at all: the fallback compares `symbol` and `venue` as strings, so identifiers interned in different pools compare equal when their values match. Ordinary in-process data takes the pointer path and never evaluates the fallback.

No in-tree caller is affected, since every built-in assembler resolves one instrument and uses it for the wrapper and each child.

The check is not repeated in `apply_deltas`. A batch is constructed once and may be applied by several consumers, and application is the hot path. Construction is also not a complete boundary: the struct's fields are public, the Cap'n Proto and SBE decoders build it literally, and derived `Deserialize` bypasses it. This closes the ordinary construction route rather than making the type invariant-preserving.

## Testing

`test_order_book_deltas_new_checked_rejects_mismatched_instrument` covers a mismatched first child and a mismatched later child, so the whole vector is examined rather than its head; a homogeneous multi-child case is the control. Both rejection cases fail against the unfixed constructor.

`test_apply_deltas_does_not_validate_children_constructed_literally` builds the struct literally to bypass the constructor and pins what remains: `apply_deltas` returns `Ok` with the foreign child's level resting in the book, and `apply_deltas_unchecked` still applies one. It records the boundary rather than asserting a rollback that does not occur.

Two existing tests met the new contract in opposite ways. `test_order_book_deltas_single_delta` paired a `BTCUSD.CRYPTO` wrapper with a `EURUSD.SIM` fixture child while asserting only metadata propagation, and now takes its wrapper from the child. The SBE round-trip fixture `test_order_book_deltas_preserve_delta_instrument_ids` is heterogeneous on purpose, since it exists to prove each child's ID is encoded independently, so it constructs its value literally and preserves that coverage; the encoded input is unchanged.

### DIFF
--- a/crates/data/src/engine/mod.rs
+++ b/crates/data/src/engine/mod.rs
@@ -2048,43 +2048,13 @@ impl DataEngine {
         Some(rebuilt)
     }
 
-    // Replays a day-start snapshot forward to the request's original start: when the first delta
-    // is an F_SNAPSHOT on a UTC day boundary, rebuilds the book from the pre-start deltas and
-    // replaces them with one snapshot keyed at the original start, then forwards the rest.
+    // Replays each instrument's day-start snapshot forward to the request's original start when
+    // that partition's first delta is an F_SNAPSHOT on a UTC day boundary, replacing its
+    // pre-start deltas with one snapshot keyed at the original start before forwarding the rest.
     fn book_deltas_snapshot_replay(&self, resp: &mut BookDeltasResponse) {
         let Some(original_start_ns) = resp.start else {
             return;
         };
-
-        let Some(first) = resp.data.first().copied() else {
-            return;
-        };
-
-        if !RecordFlag::F_SNAPSHOT.matches(first.flags) {
-            return;
-        }
-
-        if first.ts_init.as_u64() % NANOSECONDS_IN_DAY != 0 {
-            return;
-        }
-
-        // Nothing to fast-forward when the request starts at or before the day-start snapshot
-        if original_start_ns <= first.ts_init {
-            return;
-        }
-
-        if self
-            .cache
-            .borrow()
-            .instrument(&resp.instrument_id)
-            .is_none()
-        {
-            log::warn!(
-                "Instrument {} not found in cache, skipping snapshot replay",
-                resp.instrument_id,
-            );
-            return;
-        }
 
         let book_type = resp
             .params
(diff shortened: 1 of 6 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR. The constructor check itself works for ordinary homogeneous batches, but it currently breaks heterogeneous `RequestJoin` responses for book deltas.

`rebuild_pipeline_response` retains the first leg's `BookDeltasResponse.instrument_id` while concatenating deltas from every leg. `finalize_request_join` routes the result through `handle_book_deltas_response`, which calls `OrderBookDeltas::new(resp.instrument_id, frame)`. When a later leg has another instrument, the new predicate returns an error and `new()` panics. Release builds use `panic = "abort"`, so this can terminate the process.

Could you preserve the supported heterogeneous join path without a panic and add a regression covering `BookDeltas` from two instruments? The `OrderBookDeltas` concept page should also list the new matching-ID construction rule.

folknor:
Thank you - confirmed, and there is a third construction site on the same response: `book_deltas_snapshot_replay` builds both an `OrderBook` and a batch from the wrapper ID, and runs before `handle_book_deltas_response`.

The wrapper `instrument_id` is now used only when the response is homogeneous. Both functions partition `resp.data` by each child's own `instrument_id` in first-seen order and key every book-scoped operation off the partition: cache ownership and book lookup, `F_LAST` framing, batch construction, and the publish topic. Partitioning precedes framing rather than following it, because `rebuild_pipeline_response` sorts the legs together by `ts_init`, so `A(non-last), B(non-last), A(last)` would otherwise frame `[A, B, A]`. A single-instrument response takes a borrowed fast path that allocates no partitions.

This also corrects cache routing on the same path. Ownership was tested only for the wrapper, so a live-owned wrapper suppressed cache writes for every other instrument, and foreign children were pushed at the wrapper's book where `apply_delta` rejected them.

One behaviour change: snapshot-replay eligibility was evaluated once against whichever leg sorted first, and is now evaluated per instrument against that partition's own first delta. Skipping replay for mixed responses instead would let the parent-window trim discard the pre-start deltas that replay exists to fold into the snapshot.

`new_checked` is unchanged, and `rebuild_pipeline_response` still accepts the heterogeneous join. The `OrderBookDeltas` concept page now lists the matching-ID construction rule.

## Testing

`test_request_join_publishes_book_deltas_on_each_instrument_topic` drives a two-instrument join end to end and asserts each child arrives on its own topic.

`test_request_join_replays_later_eligible_partition_when_first_is_ineligible` and `test_request_join_replays_first_eligible_partition_when_later_is_ineligible` cover both replay directions through the pipeline. The parent window is `1_000..3_000` and both legs carry data from zero, so the asserted `CLEAR` at `1_500` can only appear if replay ran before the parent trim.

`test_book_deltas_response_routes_cache_by_partition_ownership` is parameterized over which instrument holds the live subscription and asserts both update-count outcomes, including that wrapper ownership no longer suppresses the other instrument.

`test_book_deltas_response_partitions_interleaved_instruments_before_framing` pins the straddle case above, and `test_book_deltas_response_publishes_trailing_partitions_in_first_seen_order` pins publication order using one recorder subscribed to both topics.

All five fail against the previous commit.

cjdsellers:
Hey @folknor,

Thanks for the thorough follow-up. The additional constructor validation is a great improvement, and we definitely want to keep it: requiring every child `instrument_id` to match the wrapper closes a real gap in the normal construction path.

I need to correct my earlier request, though. I should have challenged whether heterogeneous `RequestJoin` responses are a supported contract before asking you to preserve them. That's on me, and I'm sorry I sent you down the larger implementation path.

After tracing the callers, I don't think we should add that complexity here. The production pipelines split one logical request across sources or time windows, so the legs retain the same instrument, and the pre-existing `RequestJoin` coverage is also same-instrument.

Would you be open to paring this back to the constructor validation, documentation, focused tests, and a small guard in `rebuild_pipeline_response` that rejects `BookDeltas` legs whose wrapper instrument differs? That would keep the valuable validation and fail the unsupported join upstream without adding the per-instrument replay, cache, framing, and publication machinery.

folknor:
Thank you - pared back as you describe, and no apology needed.

`rebuild_pipeline_response` now refuses a `BookDeltas` leg whose wrapper instrument differs from the accumulator's, logging the pipeline and both identifiers and returning `None`, which is the existing rebuild-failure path and clears the pipeline and pending-join staging. Legs match by value as well as identity, mirroring `OrderBookDeltas::new_checked`, since legs crossing the FFI boundary do not share an intern pool.

`BookDeltasResponse` is not a validated type - `new` stores the wrapper and the vector as given, the fields are public, and `Deserialize` is derived - so a producer whose wrapper disagrees with its own children still reaches `OrderBookDeltas::new` in `handle_book_deltas_response`. Time-range child responses reach that handler directly, without passing through `rebuild_pipeline_response`.

## Testing

`test_request_join_rebuilds_same_instrument_book_deltas_legs` drives a two-leg same-instrument join and asserts the rebuilt response keeps its wrapper and concatenates in `ts_init` order.

`test_request_join_mixed_instrument_book_deltas_cleans_up_join_staging` asserts a two-instrument join emits no parent response and leaves the pipeline and pending-join counts at zero. Both tests advance the test clock past the leg timestamps, otherwise the parent-window trim empties the response before it reaches the handler. Without the guard the second test aborts in `OrderBookDeltas::new`, which is the panic you reported.

cjdsellers (dismissed):
Thanks for paring this back. I don't see a code blocker in the current head, and all required checks pass.

I found two small follow-ups that would make the new failure paths easier to maintain and diagnose. `rebuild_pipeline_response` now also returns `None` for mixed-instrument `BookDeltas` legs, but its rustdoc and cleanup comment don't include that failure mode. `OrderBookDeltas::new` can also abort the process in release builds on a child mismatch without reporting the wrapper ID, offending child ID, or child index.

If you're able, could you tighten those in this PR? I don't think either needs to hold it up, so they could also follow separately if you'd rather keep this patch focused.

folknor:
@cjdsellers Thank you - both tightened.

`OrderBookDeltas::new_checked` now reports the wrapper ID, the child index, the batch length and the offending child ID. Getting the position meant moving from an `all` fold to a `find`, so the predicate is now the negation of the acceptance condition rather than the condition itself; it is the exact negation, and identifiers from different intern pools whose symbol and venue strings match are still accepted. `test_order_book_deltas_new_checked_rejects_mismatched_instrument` is cased over a first-delta and a later-delta mismatch and now asserts the whole message, which pins that the reported index follows the offending delta.

`rebuild_pipeline_response`'s rustdoc and the cleanup comment at its call site both name the mixed-instrument `BookDeltas` path now.

cjdsellers (approved):
@folknor, both tightened changes look good. We're down to just one remaining edge case, and it occurs in just one supported Databento response shape: a historical parent-symbol MBO response can carry deltas for multiple child contracts.

I'll address that locally after this merges with a Databento adapter-level follow-up, including the response handling needed to partition the data into per-instrument `OrderBookDeltas` before cache application, snapshot replay, and publication. It doesn't need to hold up this PR, and no further changes are needed from you here.

Thanks again for sticking with the review cycles on this.

PR #4718, opened 2026-08-11
Scale indicator test tolerance with magnitude

### LINKED ISSUE #4678: Indicator tests compare f64 with assert_eq!
Split out of #4654 at @cjdsellers' suggestion.

Several indicator tests pin expected values with `assert_eq!` on `f64`. That holds only
while the arithmetic stays bit-reproducible, so a legitimate refactor that moves a result
by one ULP looks like a regression. #4667 is the case in point: dividing before comparing
is the right order, the ULP move was harmless, and the test would still have flagged it.

An approximate comparison with an explicit tolerance would say the intent out loud, and the
tolerance becomes a documented property of the indicator rather than an accident of
evaluation order. Happy to do the sweep.

mkzung:
Closes #4678.

`f64::EPSILON` is the step at 1.0, about 2.2e-16. Near 1e6 the spacing of `f64` is already
~1.2e-10, so `(a - b).abs() < f64::EPSILON` was exact equality for anything of indicator
magnitude. The docstring example hid that: `0.1 + 0.2` vs `0.3` passes only because the
values are below 1.

`approx_equal` now takes a relative bound of 1e-9 with an absolute floor of 1e-15 for
values near zero, `approx_equal_with` takes both explicitly, and `assert_approx_equal`
reports the two values and the absolute and relative errors on failure. NaN never compares
equal; infinities do only with matching signs.

Per your note on the issue I have left invariants exact. Only the five existing
`approx_equal` call sites move over, and the two ATR expectations that were pinned to
`0.000_099_999_999_999_988_99` now read `0.0001`.

## Testing

`cargo test -p nautilus-indicators`: 543 lib tests, 2 doctests. `cargo clippy
--all-targets` and `cargo fmt --check` clean, and `prek` passes on the changed files.

Three of the four new tests fail against the old implementation, so they distinguish it;
the fourth asserts that real differences still separate and passes either way.

Two things I would rather flag than hide. The ATR cleanup is not evidence for this change:
the old helper accepts `0.0001` too, because at 1e-4 the noise fits under `f64::EPSILON`.
And `assert_approx_equal` is a `#[track_caller]` function rather than a macro, since a
`macro_rules!` body is only type-checked where it is expanded and this crate exports no
macros today.

Claude Code was used for implementation assistance, covering the helper and this
description. All changes have been reviewed, understood and manually tested by me.

### DIFF
--- a/crates/indicators/src/average/ama.rs
+++ b/crates/indicators/src/average/ama.rs
@@ -223,6 +223,7 @@ mod tests {
         average::ama::AdaptiveMovingAverage,
         indicator::{Indicator, MovingAverage},
         stubs::*,
+        testing::assert_approx_equal,
     };
 
     #[rstest]
@@ -244,7 +245,7 @@ mod tests {
     fn test_value_with_two_inputs(mut indicator_ama_10: AdaptiveMovingAverage) {
         indicator_ama_10.update_raw(1.0);
         indicator_ama_10.update_raw(2.0);
-        assert_eq!(indicator_ama_10.value, 1.444_444_444_444_444_2);
+        assert_approx_equal(indicator_ama_10.value, 1.44444444444);
     }
 
     #[rstest]
@@ -252,7 +253,7 @@ mod tests {
         indicator_ama_10.update_raw(1.0);
         indicator_ama_10.update_raw(2.0);
         indicator_ama_10.update_raw(3.0);
-        assert_eq!(indicator_ama_10.value, 2.135_802_469_135_802);
+        assert_approx_equal(indicator_ama_10.value, 2.13580246914);
     }
 
     #[rstest]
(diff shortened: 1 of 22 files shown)


cjdsellers:
Hi @mkzung,

Thanks for the PR. The `pre-commit` check needs to pass before I can look any further. The current failure is the missing blank line above the `if` in `crates/indicators/src/testing.rs:66`.

This should be easy to fix. Run the following from the repository root, address anything it reports, and push the changes to the same PR:

```bash
make pre-commit
```

On Windows, or if `make` isn't available, run:

```bash
prek run --all-files
```

You can also install the hooks so they run before future commits:

```bash
prek install
```

cjdsellers (approved):
Hey @mkzung,

Thanks for fixing the `pre-commit` issue. I've reviewed the updated PR, and everything looks good from my side now.

PR #4723, opened 2026-08-12
Bound nesting depth in the synthetic formula parser

folknor:
The synthetic-instrument formula language enforces two compile-time resource limits, an evaluation-stack depth of 32 and 16 local variables, but nothing bounds how deeply a formula may nest.

## Unbounded recursion in the parser

`parse_prefix` in `crates/model/src/expressions/parser.rs` recurses into `parse_expression` for parentheses, both prefix operators and call arguments, and `parse_expression` recurses for right-associative `^`. None of those sites tracked depth, so a sufficiently nested formula consumes more call-stack budget than the thread has and ends the process rather than returning the typed error every other malformed formula produces. The existing `MAX_STACK` check cannot cover this: it bounds the compiled VM's operand stack and is evaluated in `compile` after the recursive walk has already run.

Parser recursion depth and tree depth are separate quantities and both need bounding. `((((1))))` recurses deeply but yields a single leaf, while `1 + 1 + 1 + ...` is assembled by the loop in `parse_expression` with O(1) recursion into an O(n)-deep left-associated tree, which is then walked recursively by `compile_expr` and by drop glue, and which passes `MAX_STACK` because a left spine needs about two operand slots.

The parser now carries each subtree's depth in its parse result and threads a syntax-nesting depth through the recursive calls, checking both against `MAX_EXPRESSION_DEPTH = 128` before constructing each parent node, so no over-limit tree is ever built and an error path never owns one. Exceeding it returns `ExpressionDepthExceeded { depth, max }`; the top-level expression counts as one level, so 127 nested parentheses are accepted. The limit is empirical: on a 2 MiB thread in debug builds the walks fail at roughly 300 for parentheses and prefix chains, 400 for compiling and dropping a left-associated chain, and 1200 for a right-associative chain, with release clearing 1024 on every walk, so 128 holds better than a 2x margin below the lowest. A weighted sum of N components builds a tree of depth N + 1, which caps such a formula at 127 components.

The `### Limits` table in `docs/concepts/synthetics.md` gains a row. The sentence claiming the limits are "generous for any realistic pricing formula" is removed, since an eight-component example does not establish what realistic formulas require and the table now states a component ceiling.

## Testing

`test_expression_depth_limit_on_constrained_stack` parses, compiles and drops each shape at exactly the limit and asserts rejection one level past it, covering nested parentheses, a prefix chain, a left-associated spine and a right-associative `^` chain. `test_deep_expression_returns_error_without_crashing_process` submits 10,000-deep prefix and `^` inputs. `test_new_checked_rejects_excessive_expression_depth` covers the public `SyntheticInstrument` path. Both parser tests run on an explicitly sized 2 MiB thread so the bound is exercised against a known stack budget rather than the harness default.

The new assertions fail against the unmodified parser: the boundary cases parse successfully and `unwrap_err` panics, and the 10,000-deep inputs exhaust the child thread instead of returning the error.

### DIFF
--- a/crates/model/src/expressions/error.rs
+++ b/crates/model/src/expressions/error.rs
@@ -71,6 +71,10 @@ pub(crate) enum ExpressionError {
     InputCountMismatch { expected: usize, actual: usize },
     #[error("Expression requires {depth} stack slots, maximum is {max}")]
     StackOverflow { depth: usize, max: usize },
+    #[error(
+        "Expression nesting depth {depth} exceeds maximum {max} (the top-level expression counts as one level)"
+    )]
+    ExpressionDepthExceeded { depth: usize, max: usize },
     #[error("Expression defines {count} local variables, maximum is {max}")]
     TooManyLocals { count: usize, max: usize },
     #[error("Expression result is empty")]
--- a/crates/model/src/expressions/parser.rs
+++ b/crates/model/src/expressions/parser.rs
@@ -19,6 +19,21 @@ use super::{
     lexer::{Token, TokenKind},
 };
 
+/// Maximum syntax nesting and constructed AST depth accepted by the parser,
+/// counting the top-level expression as one level.
+///
+/// Bounds both the parser's own recursion (parentheses, prefix operators,
+/// right-associative `^`, call arguments) and the depth of the tree it
+/// constructs (a left-associated operator chain builds an O(n)-deep tree with
+/// O(1) parser recursion), so parsing, `compile_expr`, and drop glue all stay
+/// within a fixed stack budget. On a 2 MiB thread in debug builds the walks
+/// overflow at depths of roughly 300 (nested parentheses and prefix chains),
+/// 400 (compiling and dropping a left-associated chain), and 1200 (a
+/// right-associative chain); release builds clear depth 1024 on every walk.
+/// 128 keeps better than a 2x margin below the lowest of those cliffs. The
+/// documented eight-component weighted sum has depth 9.
+const MAX_EXPRESSION_DEPTH: usize = 128;
+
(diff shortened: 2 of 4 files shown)


cjdsellers (dismissed):
Hey @folknor,

Thanks for the PR. The implementation looks good overall, and I found no blocking issues. The depth accounting correctly bounds both parser recursion and the constructed tree.

I found one non-blocking test gap: nested function calls are the only recursive shape without boundary coverage. I confirmed locally that nested `abs(...)` calls at depth 128 parse, compile, and drop on the 2 MiB stack, while depth 129 returns `ExpressionDepthExceeded`. Would you be able to add those cases to `test_expression_depth_limit_on_constrained_stack` when you get a chance? That would lock in the documented 128/129 boundary for the remaining recursive shape and prevent future parser changes from reopening the stack exhaustion path.

folknor:
@cjdsellers Thank you - nested calls now have both boundary cases in `test_expression_depth_limit_on_constrained_stack`, the accepted one parsing and compiling as you described.

One note on the numbers, since the test reads one lower than your report. A call advances the syntax nesting depth and takes the parent depth over its arguments, so `n` nested calls produce depth `n + 1`: your accepted depth of 128 is `MAX_EXPRESSION_DEPTH - 1` nested `abs(`, and 129 is `MAX_EXPRESSION_DEPTH` of them, which reports `ExpressionDepthExceeded { depth: 129, max: 128 }`. Both cases are written against the constant rather than the literals, so they follow the limit if it ever moves - the same form as the paren, prefix and spine cases already there.

cjdsellers (approved):
@folknor, thanks for adding the nested-call boundary coverage and for clarifying the depth accounting. I rechecked the latest head and found no further issues 👌

PR #4724, opened 2026-08-12
Expose the risk engine's price sanity check

folknor:
`RiskEngine::check_price` is a side-effect-free instrument-level check that callers outside the crate cannot reach.

## Making the price check callable

`check_price` is an associated function taking only an `InstrumentAny` and an `Option<Price>`, reading `price_precision` and `allows_negative_price` and returning a reason or `None`. It touches no config, cache, clock, portfolio or account state, and emits nothing. This PR makes it `pub`.

This exposes the existing rule so Rust callers that intentionally route orders directly to execution can apply exactly the same pre-trade check without duplicating policy or emitting denial events. There is no in-tree caller for it.

The rustdoc states what a `None` result does not mean: only decimal precision and the non-positive case are checked, and venue increment rules, notional bounds, and account state are not. One consequence to note is that the human-readable reason becomes public surface, so its wording is harder to change afterwards.

## Correcting a comment in `check_quantity`

Separately, its comment said the skipped `min_quantity`/`max_quantity` bounds "will be checked in `check_orders_risk` using `effective_quantity`". They are not - `check_orders_risk_for_account` guards the same comparisons with `if !order.is_quote_quantity()` and skips them for the same reason. The behaviour is deliberate and unchanged; only the comment described a second check that does not exist. I bundled that here rather than as a separate PR. Let me know if that was a mistake.

## Testing

Four tests in `crates/risk/tests/risk_engine.rs`, which call the function from outside the crate - the surface being exposed. They cover an absent price, a valid price, an excess-precision rejection, and a non-positive rejection where the instrument disallows one, matching the function's four control-flow outcomes. No behaviour changed, so there is no differential claim to make: they pin the newly public contract rather than a fix.

### DIFF
--- a/crates/risk/src/engine/mod.rs
+++ b/crates/risk/src/engine/mod.rs
@@ -1914,7 +1914,15 @@ impl RiskEngine {
         );
     }
 
-    fn check_price(instrument: &InstrumentAny, price: Option<Price>) -> Option<String> {
+    /// Checks `price` against `instrument`, returning a reason when it is invalid.
+    ///
+    /// Returns `None` when `price` is absent or passes these checks, or a human-readable
+    /// reason when it does not. A `None` result is not a statement that the order is
+    /// valid: only decimal precision against the instrument's `price_precision`, and a
+    /// non-positive price where the instrument disallows one, are checked here. Venue
+    /// increment rules, notional bounds, and account state are not.
+    #[must_use]
+    pub fn check_price(instrument: &InstrumentAny, price: Option<Price>) -> Option<String> {
         let price_val = price?;
 
         if price_val.precision > instrument.price_precision() {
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. I don't think we should expose this as `RiskEngine::check_price`. It would establish a public API shape where callers bypassing the risk engine still name `RiskEngine` and depend on the entire risk crate for a stateless validation function. That suggests they are applying the risk engine's validation when they are only applying two narrow checks, and it also makes the human-readable reason part of the public contract.

A standalone price validation function may make sense in a lower-level crate, but there is no defined caller or demonstrated demand for that API yet, so I don't think we should design it speculatively. If and when we need it, opening a fresh PR around the concrete caller and intended contract will be cleaner than repurposing this one. The `check_quantity` comment correction is accurate and wasn't a problem to include, but it doesn't change the API decision.

PR #4728, opened 2026-08-12
Retain open Polymarket markets during auto-load

mystic-io:
# Pull Request

- [x] A maintainer agreed on the problem and approach in an issue
- [x] I have read and followed CONTRIBUTING.md and AI_POLICY.md
- [x] I understand and can explain every submitted change
- [x] This change is complete and locally validated
- [x] I ran `make format`, then `make pre-commit` locally
- [x] I ran the relevant tests

## Summary

Polymarket’s normal Gamma lookup can omit a closed market or return it without enough usable token information to classify it. When no classifiable instrument is returned for a condition, the auto-load path performs a positive Gamma `closed=true` lookup before deciding whether the condition is `Closed` or remains `Unknown`.

This change gives each auto-loaded condition one of three lifecycle outcomes:

- `Open`: apply the normal filters, then cache, publish, and subscribe;
- `Closed`: retire live subscriptions and runtime state, preserve any watchlisted metadata needed for pending settlement, and stop retrying;
- `Unknown`: preserve subscription intent and retry without overriding a known state.

A positive Gamma `closed=true` result is the terminal signal for live data. Closed conditions are classified separately and never pass through the normal filtering or live publication path. Once closure is confirmed, stale concurrent `Open` results and later subscription attempts cannot reopen the condition.

Gamma `endDate` remains instrument metadata and is not proof of closure. `market_resolved` remains separate settlement evidence.

Successful conditions are applied independently, so an absent condition or failed closure probe does not discard successful results from the same batch.

## Related issues/PRs

Related to #4689  
Follow-up to #4706
Lifecycle contract discussed in #4722

## Type of change

- [x] Bug fix (non-breaking)

## Breaking change details

None.

## Documentation

No documentation changes.

## Testing

Added regression coverage for:

- open conditions being cached, published, and subscribed with one Gamma request;
- absent conditions receiving the additional `closed=true` lookup;
- positive closure evidence retiring condition-wide live state without retrying;
- closed lookup results bypassing filters and live publication;
- unknown and failed probes preserving subscription intent;
- successful conditions surviving failures elsewhere in the same request;
- terminal closure winning over stale `Open` results in either completion order;
- sibling outcomes for a closed condition being retired together;
- watchlisted metadata being retained while live state is removed;
- later subscriptions being unable to reopen a confirmed-closed condition;
- auto-load and periodic closure refresh sharing the same terminal state.

Validation completed:

- `cargo test -p nautilus-polymarket --lib --tests`
- `make format`
- `make pre-commit`
- `git diff --check`

### DIFF
--- a/crates/adapters/polymarket/src/data/auto_load.rs
+++ b/crates/adapters/polymarket/src/data/auto_load.rs
@@ -25,10 +25,12 @@ use nautilus_model::{
 use super::{
     PolymarketDataClient,
     instruments::{
-        cache_instrument, publish_cached_condition_closed, query_positive_closed_condition_ids,
+        apply_live_instrument, publish_cached_condition_closed, query_positive_closed_condition_ids,
     },
-    runtime::{is_condition_closed, retire_closed_condition_state},
-    subscriptions::{resolve_token_id_from, sync_ws_subscription_async},
+    runtime::{
+        is_condition_closed, register_closed_condition_for_live_data, retire_closed_condition_state,
+    },
+    subscriptions::{resolve_token_id_from, sync_ws_subscription_with_terminal_async},
 };
 use crate::{
     common::consts::GAMMA_CONDITION_IDS_BATCH_SIZE, filters::market_closed,
@@ -304,11 +306,7 @@ impl PolymarketDataClient {
                 for (condition_id, outcome) in &outcomes {
                     match outcome {
                         AutoLoadOutcome::Open(loaded) => {
-                            let closed = closed_condition_ids
-                                .lock()
-                                .expect("closed_condition_ids mutex poisoned");
-
-                            if cancellation.is_cancelled() || closed.contains(condition_id) {
+                            if cancellation.is_cancelled() {
                                 continue;
                             }
 
@@ -321,20 +319,37 @@ impl PolymarketDataClient {
                                     continue;
                                 }
 
-                                cache_instrument(&instruments, &token_meta, instrument);
                                 let instrument_id = instrument.id();
-                                if let Err(e) =
-                                    data_sender.send(DataEvent::Instrument(instrument.clone()))
-                                {
-                                    log::error!(
-                                        "Failed to emit auto-loaded instrument {instrument_id}: {e}"
-                                    );
-                                }
+                                apply_live_instrument(
+                                    &closed_condition_ids,
+                                    &instruments,
+                                    &token_meta,
+                                    instrument,
+                                    |instrument| {
+                                        if let Err(e) = data_sender
+                                            .send(DataEvent::Instrument(instrument.clone()))
+                                        {
+                                            log::error!(
+                                                "Failed to emit auto-loaded instrument {instrument_id}: {e}"
+                                            );
+                                        }
+                                    },
+                                );
                             }
                         }
                         AutoLoadOutcome::Closed => {
+                            if !register_closed_condition_for_live_data(
+                                &closed_condition_ids,
+                                &ws_sub_mutex,
+                                condition_id,
+                                Some(&cancellation),
+                            )
+                            .await
                             {
-                                let mut closed = closed_condition_ids
+                                return;
+                            }
+                            {
+                                let closed = closed_condition_ids
                                     .lock()
                                     .expect("closed_condition_ids mutex poisoned");
 
@@ -342,7 +357,7 @@ impl PolymarketDataClient {
                                     return;
                                 }
 
-                                closed.insert(condition_id.clone());
+                                debug_assert!(closed.contains(condition_id));
                                 publish_cached_condition_closed(
                                     condition_id,
                                     &instruments,
@@ -406,17 +421,20 @@ impl PolymarketDataClient {
                             if loaded_ids.contains(id)
                                 && let Ok(token_id) = resolve_token_id_from(&instruments, *id)
                             {
-                                sync_ws_subscription_async(
+                                sync_ws_subscription_with_terminal_async(
                                     *id,
                                     token_id,
                                     active_quote_subs.clone(),
                                     active_delta_subs.clone(),
                                     active_trade_subs.clone(),
+                                    closed_condition_ids.clone(),
                                     ws_open_tokens.clone(),
                                     ws_sub_mutex.clone(),
                                     ws_client.clone(),
                                 )
                                 .await;
+                            } else {
+                                next_batch.insert(*id);
                             }
                         }
                         Some(AutoLoadOutcome::Closed) => {
--- a/crates/adapters/polymarket/src/data/auto_load.rs
+++ b/crates/adapters/polymarket/src/data/auto_load.rs
@@ -15,7 +15,7 @@
 
 use std::time::Duration;
 
-use ahash::AHashSet;
+use ahash::{AHashMap, AHashSet};
 use nautilus_common::{live::get_runtime, messages::DataEvent};
 use nautilus_model::{
     identifiers::InstrumentId,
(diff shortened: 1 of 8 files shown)


cjdsellers:
Hi @mystic-io,

Thanks for the PR.

I found one auto-load lifecycle gap that still needs addressing. `ensure_auto_load_task` queries Gamma with only `condition_ids`, but Gamma omits closed markets unless the request includes `closed=true`. If a market closes before auto-load completes, it never reaches the new closure-aware predicate. The ID takes the full retry path, and its subscription intent remains active instead of being retired.

The new closed-market test masks this because its mock returns a `closed=true` market to the unfiltered request. Could the absent-ID path probe `closed=true` before classifying the market as transient, and could the test assert the exact open and closed query sequence plus the terminal state? The open path should still need one request; a closed path needs the positive closure lookup.

Current `develop` also adds `"initial_dump": true` to market subscription payloads. The new open-market test still expects the old payload, so the synthetic merge is clean but that assertion will fail. Could you rebase, update the expectation, and rerun the focused Polymarket tests when you get a chance? The core change looks good; these are the two remaining edges I found.

mystic-io:
Thank you @cjdsellers , I’ve addressed both points, rebased, rerun the checks, and pushed the update. Appreciate the  review!

cjdsellers:
Thanks for addressing both earlier points. After following the new fallback through `ensure_auto_load_task`, I think my previous request was too narrow. The extra `closed=true` lookup is required, but its result needs to classify lifecycle state rather than enter the normal `loaded` path as another instrument batch. The current shape leaves that contract implicit, which is why filters, partial request failures, retirement, and overlapping loads now produce different outcomes depending on order.

I also think #4689 combines two decisions that need separate rules: ending live market data and retaining instrument metadata for settlement. The direction proposed in #4722 concerns execution reconciliation rather than instrument loading, but its typed evidence, fail‑closed handling, and ordered application look directly relevant here. These appear to be separate but coupled state machines:

- Market evidence: `Unknown -> Open -> Closed`.
- Trade settlement: `Pending -> Confirmed | Failed`.

A market becoming closed should end live subscriptions without implying that every fill is final or removing metadata still needed for settlement.

Before another implementation revision, I'd like to agree these contracts:

- A positive Gamma `closed=true` result is terminal for live market‑data state; `endDate` remains scheduling and settlement metadata.
- Absence or a failed probe is `Unknown`; it must not override proven open or closed results or discard successful conditions from the same batch.
- The closed lookup classifies absent conditions as `Closed`; its instruments do not pass through provider filters or live cache publication.
- A newer `Closed` observation wins over a stale concurrent `Open` response, and later subscriptions cannot forget that closure merely because retirement removed the cached instrument.
- Live cache and WebSocket state are managed separately from metadata retained for pending settlement.

Could we pause the code changes here and work through this contract alongside #4722? Once agreed, I think the implementation can stay contained around a shared closure probe, per‑condition outcomes, and monotonic application rather than adding more predicate and ordering fixes.

mystic-io:
Hey @cjdsellers 

I ran a 10min probe. 2 markets well past Gamma `endDate` stayed open and orderbook was enabled throughout. To me `endDate` isn’t a reliable closure signal. I also caught 2 live closures. Gamma reported `closed=true` and the books were already unavailable, but CLOB metadata still reported `closed=false` and `accepting_orders=true`. The CLOB fields caught up bout 7-8mins later. Gamma `closed=true` and CLOB `closed=true` are related, but not synchronised

My current read is:
```
Gamam endDate:  metadata only; never proof of closure
Gamma closed: terminal signal for Gamma discovery/auto-load
CLOB closed: CLOB-side lifecycle signal, but may lag Gamma
market_resolved: final resolution evidence
accepting_orders: operational hint; not terminal and may also lag
```
I’d prefer Gamma `active=true&closed=false` for normal discovery, with the positive closed=true lookup when an ID is absent.

For this auto-load path, the positive Gamma 'closed=true' lookup seems correct. it doesn’t look safe to make CLOB closed the sole liveness authority. Does that match the lifecycle split you had in mind?

cjdsellers:
@mystic-io, thanks for running the probe. Yes, that matches the source of truth split I had in mind: `endDate` is not proof of closure, Gamma `closed=true` is terminal for live discovery and auto-load, and lagging CLOB fields must not reopen that state. `market_resolved` remains separate resolution evidence.

The remaining concern is how that evidence is applied. The `closed=true` lookup should produce a per-condition lifecycle outcome rather than feed instruments into the normal `loaded` path. An absent or failed lookup remains `Unknown`, while `Closed` must win over stale concurrent `Open` results and retire live runtime state without removing metadata needed for settlement. I’d keep the code changes paused until that application contract is agreed. Does that narrower split also match your read?

mystic-io:
@cjdsellers Yeah, agreed. This model makes sense to me. I'll hold off on the changes until the contract is settled.

cjdsellers:
@mystic-io, thanks for your patience!

The source-authority question in [#4722](https://github.com/nautechsystems/nautilus_trader/discussions/4722#discussioncomment-18023007) is now settled, and it confirms the same rule here: ambiguous evidence must not mutate runtime state before the designated authority resolves it.

**So we can now resume with these per-condition outcomes:**

- `Open`: apply the normal filters, then cache, publish, and subscribe.
- `Closed`: retire live subscriptions and runtime state, preserve any watchlisted metadata, and stop retrying. This is terminal for live data, so a stale concurrent `Open` result cannot reopen it.
- `Unknown`: preserve subscription intent and retry without overriding a known state.

The positive Gamma `closed=true` lookup provides the `Closed` outcome. `endDate` remains metadata, and `market_resolved` remains separate resolution evidence. The closed lookup should classify its result rather than add those instruments to the normal `loaded` path, while successful conditions from the same batch should continue normally.

I think that gives us a clear boundary for the next update. Once everything is in place I'll update the `polymarket.md` docs as a follow-up.

mystic-io:
Hey @cjdsellers Just updated the PR body and pushed the update. about 2k added lines are tests in dispatch.rs. The production change is about 600 lines across 5 modules. I looked into splitting it, but that would create PRs with incomplete lifecycle handling and race gaps. So I kept it as one coherent change, but I’m happy to revisit if you prefer a split. Ready for review whenever you have time.

cjdsellers:
@mystic-io, thanks for the substantial update and for keeping the lifecycle handling together. I've had another look at this, and the overall `Open` / `Closed` / `Unknown` direction looks right.

**I found two correctness gaps that I think we should address before merging:**

- `closed_condition_ids` protects new subscription intent, but it is not checked by every path that can restore or publish live state. `NewMarket`, instrument requests, and refreshes can still cache a terminally closed condition, while reconnect can rebuild its token routing from retained settlement metadata. WebSocket dispatch and an already queued subscription reconciliation can also run while retirement is in progress. A stale response or message can therefore publish the market as live, emit data after closure, or briefly reopen its subscription. Could the terminal state be enforced at the shared cache, publication, routing, dispatch, and reconciliation boundaries?
- When a condition is `Open` but the requested token is absent from `loaded_ids`, the branch neither subscribes nor returns the ID to the retry batch. This leaves the subscription intent active without another attempt or an error. Could this case remain in the retry or error path?

You mentioned the 10‑minute mainnet probe earlier. Was that run against this implementation? If so, could you share the setup and results, including which auto‑load, closure, reconnect, and subscription paths were exercised? If you have run any further Polymarket mainnet validation since then, those results would also be useful.

I'll also run independent Polymarket mainnet validation before merging once these two points are addressed.

Thanks for sticking with the review process.

mystic-io:
Thanks @cjdsellers . Just to clarify, the earlier 10min probe was a direct Gamma/CLOB observation, not a run against this implementation. I haven’t run any further mainnet validation since then. fixed both two correctness gaps and ready for your review

cjdsellers:
@mystic-io, thanks for the clarification on the probe and for closing both gaps. I've verified them: the terminal state now holds at the cache, publication, routing, dispatch, and reconciliation boundaries, and an open condition whose token is missing from the cache returns to the retry path.

I ran independent Polymarket mainnet validation against your branch merged with current `develop`, so no rebase is needed. Three markets from a match that resolved mid-run went terminal: Gamma omitted all three from the unfiltered lookup, the `closed=true` probe confirmed each, and the auto-load classified them closed without publishing them or consuming any retry budget. The reconnect a few minutes later restored only the open assets. Open markets well past `endDate` loaded in a single batch and streamed throughout.

This looks good to merge :ok_hand:

A few follow-ups I'll pick up on our side rather than hold this:

- Log the terminal transition. Right now a market going terminal produces no output at any level, so the only visible symptom is a subscription that stops delivering.
- Make the periodic retirement sweep converge instead of walking the whole accumulated set every cycle.
- Hold the terminal check without keeping the guard across downstream dispatch. The data client rules in `adapters.md` have covered that since late July, and my earlier request to enforce the terminal state at the dispatch boundary pointed you straight at it without mentioning the constraint. An `AtomicSet` suits the read-heavy access here better than a mutex anyway.
- Move the two `#[cfg(test)]` functions into their test modules, and drop the loop whose body is only a `debug_assert!` since it compiles to nothing in release.

Thanks again for sticking with all of the review rounds!

cjdsellers (approved):

Issue #4716, opened 2026-08-11
[OKX] Algo terminal states are not emitted or reconciled


silarin:
# Bug Report

## Confirmation

- [x] I've re-read the relevant NautilusTrader and OKX documentation.
- [x] I've searched existing issues and pull requests to avoid duplicates.
- [x] I've reviewed the current OKX adapter source to confirm the behavior is not intentional.
- [x] I've reproduced this using recent `develop` source built from commit `61dbddf7bca4e3e7080934e59f9168f815f91343` (2026-08-09). The relevant paths remain unchanged on current `develop` at `f3a0bed303bc8a6d9f83138742d085966ffd47d0`.

## Expected behavior

For an OKX algo order submitted by a tracked Nautilus strategy:

- a triggered stop-limit order should progress from accepted to triggered, then filled when its child order fills;
- a trailing-stop order should reach the filled terminal state when execution completes; and
- restart reconciliation should recover the terminal algo order and its fill within the configured lookback.

OKX exposes completed algo records through the [algo-order history endpoint](https://my.okx.com/docs-v5/en/#order-book-trading-algo-trading-get-algo-order-history), and publishes updates on the [algo orders](https://my.okx.com/docs-v5/en/#order-book-trading-algo-trading-ws-algo-orders-channel) and [advance algo orders](https://my.okx.com/docs-v5/en/#order-book-trading-algo-trading-ws-advance-algo-orders-channel) channels.

## Actual behavior

Two bounded OKX EEA Demo tests executed successfully at the venue, but the tracked Nautilus order lifecycle stopped after acceptance.

1. A stop-limit buy reached OKX algo state `effective`, reported a filled quantity and price, and exposed a child order ID. Nautilus emitted `OrderSubmitted` and `OrderAccepted`, but no `OrderTriggered` or `OrderFilled`.
2. A trailing-stop sell reached OKX algo state `effective` and reported a filled quantity and price. Its algo detail had no child order ID. Nautilus emitted `OrderSubmitted` and `OrderAccepted`, but no terminal fill event.
3. Restart mass-status reconciliation returned no matching terminal algo order or fill for either execution.

Venue balances independently confirmed both executions. All test-created exposure was restored and no open test orders or liabilities remained.

## Steps to reproduce

1. Build the Rust OKX adapter from `61dbddf7bca4e3e7080934e59f9168f815f91343`.
2. Connect a `LiveNode` to OKX EEA Demo Trading Services.
[...]

PR #4731, opened 2026-08-12
Fix OKX query order routing

### LINKED ISSUE #4717: [OKX] query_order probes algo details for known regular orders
# Bug Report

## Confirmation

- [x] I've re-read the relevant NautilusTrader and OKX documentation.
- [x] I've searched existing issues, discussions, and pull requests to avoid duplicates.
- [x] I've reviewed the current OKX adapter source and confirmed the behavior is not required by the OKX API.
- [x] I've reproduced the error path using current `develop` source at commit `718f0b704aa4144c3237eb32fb88be7e15cef3b7` (2026-08-11) with a focused mock HTTP test.

## Expected behavior

`QueryOrder` should use the OKX detail endpoint that corresponds to the tracked order:

- known regular orders should use `GET /api/v5/trade/order` only;
- known conditional, OCO, TP/SL, and trailing-stop orders should use `GET /api/v5/trade/order-algo` only; and
- when the order type is not cached, the client may query the regular detail endpoint first and fall back to the algo detail endpoint only after a no-match response.

OKX documents `instId` plus `ordId` or `clOrdId` for regular order details. It documents `algoId` or `algoClOrdId` for algo order details. When both identifiers are supplied, the venue-assigned ID takes precedence.

- [Regular order details](https://my.okx.com/docs-v5/en/#order-book-trading-trade-get-order-details)
- [Algo order details](https://my.okx.com/docs-v5/en/#order-book-trading-algo-trading-get-algo-order-details)
- [Error code 51603](https://my.okx.com/docs-v5/en/#error-code)

## Actual behavior

The current Rust OKX `query_order` implementation performs a broad regular-order status lookup and then unconditionally queries algo details for every supported non-spread instrument, without considering the cached order type.

For a known regular order, OKX correctly returns `51603` (`Order does not exist`) from the algo endpoint. The specific algo lookup propagates that expected no-match as an error, so `query_order` emits a warning even when the regular order was found:

```text
OKX query_order algo lookup failed for ETH-USDT.OKX: OKX error 51603: Order does not exist
```

Current source paths:

- [`query_order` queries regular status reports and then algo details for every supported non-spread instrument](https://github.com/nautechsystems/nautilus_trader/blob/718f0b704aa4144c3237eb32fb88be7e15cef3b7/crates/adapters/okx/src/execution.rs#L1248-L1301).
- [The exact algo-detail path propagates `51603` as an error](https://github.com/nautechsystems/nautilus_trader/blob/718f0b704aa4144c3237eb32fb88be7e15cef3b7/crates/adapters/okx/src/http/client.rs#L6177-L6195).
- [`GetOrderParams` includes undocumented `instType` and `posSide` request fields](https://github.com/nautechsystems/nautilus_trader/blob/718f0b704aa4144c3237eb32fb88be7e15cef3b7/crates/adapters/okx/src/http/query.rs#L699-L719).
- [`OKXOrderAlgo` models only the soon-to-be-deprecated `ordId`, not `ordIdList` or `subAlgoIdList`](https://github.com/nautechsystems/nautilus_trader/blob/718f0b704aa4144c3237eb32fb88be7e15cef3b7/crates/adapters/okx/src/http/models.rs#L1124-L1142).

## Steps to reproduce

1. Build the Rust OKX adapter from current `develop` commit `718f0b704aa4144c3237eb32fb88be7e15cef3b7`.
2. Configure a mock OKX HTTP server to return code `51603` for `GET /api/v5/trade/order-algo?algoClOrdId=<regular-client-order-id>`.
3. Call `request_algo_order_status_report` with that client order ID. The focused test confirms that current `develop` propagates `OKXHttpError::OkxError { error_code: "51603", ... }`.
4. Inspect or invoke `query_order` for a cached regular order on a supported non-spread instrument. The current routing always calls that exact algo-detail path after the regular lookup and logs the propagated error.

Focused reproduction result:

```text
running 1 test
test test_http_request_algo_order_status_report_propagates_missing_order ... ok

test result: ok. 1 passed; 0 failed
```

## Proposed approach

1. Route known regular and algo orders to their corresponding detail endpoints using the cached `OrderType`.
2. For an uncached order, query regular details first and fall back to algo details only after a no-match response.
3. Treat OKX `51603` as `None` only in specific order-detail lookup helpers; do not suppress it globally.
4. Restrict `GetOrderParams` to the documented request fields.
5. Prefer `algoId` only when it is known to be the stable pre-trigger venue ID; otherwise use `algoClOrdId`.
6. Deserialize `ordIdList` and `subAlgoIdList` while leaving multi-child lifecycle reconciliation outside this focused fix.

This is related to, but separate from, #4716. That issue covers terminal algo events and restart reconciliation; this report is limited to `QueryOrder` endpoint selection, request schema, and no-match handling.

## Code snippets or logs

No credentials, account identifiers, order identifiers, or production data are required to reproduce the issue. The focused reproduction uses the existing Axum mock HTTP test pattern.

## Specifications

- OS platform: macOS 26.6.1, arm64
- Python version: N/A (Rust adapter)
- Rust version: 1.97.1
- `nautilus_trader` version: current `develop` at `718f0b704aa4144c3237eb32fb88be7e15cef3b7`
- Adapter/venue: Rust `nautilus-okx` / OKX

silarin:
# Pull Request

  and, because AI assistance was used,

## Summary

Route `QueryOrder` through the regular or algo detail endpoint according to the cached order type, with regular-to-algo fallback only for uncached orders after an order-not-found response. The exact lookup helpers now treat OKX `51603` as a scoped no-match, use documented request fields and identifier precedence, and deserialize current algo child identifier lists without adding multi-child lifecycle behavior.

The regression coverage includes triggered conditionals whose cached venue ID has become the child `ordId`, ensuring the parent lookup uses `algoClOrdId`, and regular orders where a stale `ordId` must not override a valid `clOrdId`.

## Related issues/PRs

Resolves #4717

Related to #4716

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

None.

## Documentation

- [x] Documentation changes follow the style guide (not applicable; no documentation changed)
- [x] PyO3 stubs are current (not applicable; no PyO3 bindings or wrapped Rust docs changed)

## Testing

- `make format`
- `make pre-commit`
- `cargo test -p nautilus-okx --lib --test http --test exec_client`
  - 648 library tests passed
  - 58 execution-client tests passed
  - 100 HTTP tests passed

Issue #4717 also records a bounded OKX EEA Demo proof of the regular, cached-algo, and uncached fallback routes. The test-created orders were canceled and reconciled with no fills or residual exposure.

### DIFF
--- a/crates/adapters/okx/src/execution.rs
+++ b/crates/adapters/okx/src/execution.rs
@@ -1257,47 +1257,85 @@ impl ExecutionClient for OKXExecutionClient {
         let instrument_id = cmd.instrument_id;
         let client_order_id = cmd.client_order_id;
         let venue_order_id = cmd.venue_order_id;
-        let should_query_algo = !is_spread_instrument(instrument_id)
-            && supports_algo_orders(okx_instrument_type_from_symbol(
-                instrument_id.symbol.as_str(),
-            ));
+        let order_state = {
+            let cache = self.core.cache();
+            cache
+                .order(&client_order_id)
+                .map(|order| (order.order_type(), order.is_triggered()))
+        };
+        let route = query_order_route(instrument_id, order_state.map(|state| state.0));
+        let algo_id = order_state
+            .filter(|(order_type, is_triggered)| {
+                OKX_CONDITIONAL_ORDER_TYPES.contains(order_type) && *is_triggered == Some(false)
+            })
+            .and_then(|_| {
+                venue_order_id
+                    .as_ref()
+                    .map(|venue_order_id| venue_order_id.as_str().to_string())
+            });
 
         self.spawn_task("query_order", async move {
-            let mut reports = match http_client
-                .request_order_status_reports(
-                    account_id,
-                    None,
-                    Some(instrument_id),
-                    None,
-                    None,
-                    false,
-                    None,
-                )
-                .await
-            {
-                Ok(r) => r,
-                Err(e) => {
-                    log::error!("OKX query_order failed to fetch orders: {e}");
-                    Vec::new()
-                }
-            };
+            let mut reports = Vec::with_capacity(1);
+            let mut query_algo = route == QueryOrderRoute::Algo;
 
-            // Merge algo orders (stop, OCO, TP/SL, trailing) so query_order can
-            // resolve conditional orders as well.
-            if should_query_algo {
+            match route {
+                QueryOrderRoute::Spread => {
+                    match http_client
+                        .request_order_status_reports(
+                            account_id,
+                            None,
+                            Some(instrument_id),
+                            None,
+                            None,
+                            false,
+                            None,
+                        )
+                        .await
+                    {
+                        Ok(spread_reports) => reports.extend(spread_reports),
+                        Err(e) => {
+                            log::error!("OKX query_order failed to fetch spread order: {e}");
+                        }
+                    }
+                }
+                QueryOrderRoute::Regular | QueryOrderRoute::RegularThenAlgo => {
+                    match http_client
+                        .request_order_status_report(
+                            account_id,
+                            instrument_id,
+                            client_order_id,
+                        )
+                        .await
+                    {
+                        Ok(Some(report)) => reports.push(report),
+                        Ok(None) => {
+                            query_algo = route == QueryOrderRoute::RegularThenAlgo;
+                        }
+                        Err(e) => {
+                            log::error!("OKX query_order failed to fetch regular order: {e}");
+                        }
+                    }
+                }
+                QueryOrderRoute::Algo => {}
+            }
+
+            // Known conditional orders route directly to the algo endpoint. For
+            // an uncached order, only fall back when the regular detail lookup
+            // has no match (including OKX 51603).
+            if query_algo {
                 match http_client
                     .request_algo_order_status_reports(
                         account_id,
                         None,
                         Some(instrument_id),
-                        None,
+                        algo_id,
                         Some(client_order_id),
                         None,
-                        None,
+                        Some(1),
                     )
                     .await
                 {
-                    Ok(algo) => merge_order_status_reports(&mut reports, algo),
+                    Ok(algo_reports) => merge_order_status_reports(&mut reports, algo_reports),
                     Err(e) => {
                         log::warn!("OKX query_order algo lookup failed for {instrument_id}: {e}");
                     }
@@ -2364,6 +2402,14 @@ enum OrderCommandRoute {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum QueryOrderRoute {
+    Regular,
+    Algo,
+    RegularThenAlgo,
+    Spread,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum CancelAllOrdersRoute {
     BatchWs,
     MassCancelHttp,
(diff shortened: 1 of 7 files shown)


cjdsellers:
Hi @silarin,

Thanks for the PR. It currently has merge conflicts with `develop`, so I'll take a proper look once those are resolved.

silarin:
Hi @cjdsellers, the conflicts with `develop` have been resolved and the branch has been rebased. All required checks are passing, and the PR is ready for review.

cjdsellers:
Hi @silarin,

Thanks for the PR. The endpoint split is the right direction, but I found three reachable query paths that still need adjustment:

- Market-style conditionals can retain `is_triggered() == false` after the cached venue ID changes to the generated child `ordId`. That value is then sent as `algoId`, which takes precedence over the valid `algoClOrdId` and causes the lookup to miss.
- Once any cached conditional triggers, routing only to the algo endpoint loses the regular child's fill state. An `effective` parent report remains `Triggered` with zero filled quantity, so `QueryOrder` can no longer recover a missed child fill.
- Adopted external regular orders may have no venue `clOrdId`; their local client ID is synthesized from `ordId`. The new regular lookup sends that value as `clOrdId` and ignores the available venue order ID, so those orders return no report.

Could you adjust the identifier selection and preserve the regular-child lookup after trigger, then add regressions for a market-style child ID, a missed child fill, and an adopted external order without `clOrdId`?

silarin:
Thanks for identifying these cases. I have pushed an update that addresses all three:

- A cached conditional with a generated child ordId now queries the regular child by ordId and the algo parent by algoClOrdId. This includes market-style conditionals where is_triggered remains false.
- Triggered conditionals retain both regular-child and algo-parent lookup. Report selection prefers the more advanced state, and the regression recovers a missed filled child over an effective parent.
- An adopted external regular order without clOrdId now queries the regular endpoint using its cached ordId.

I kept the follow-up focused: it does not add recursive ordIdList or subAlgoIdList traversal or broader multi-child lifecycle reconciliation.

The branch is rebased onto current develop. Local validation passed:

- Focused query-order tests: 23 passed
- Full nautilus-okx test suite: 878 passed
- make format
- Full pre-commit suite, including clippy and docs

cjdsellers (changes requested):
Thanks for the follow-up. The three cases from my previous review are now fixed, and the focused regressions pass.

I found one adjacent conditional state that still needs handling. `has_regular_child` currently relies on `venue_order_ids().len() > 1`. If OKX triggers and fills while the child WebSocket event is entirely missed, the cache still contains only the parent `algoId`, so `QueryOrder` queries only the algo parent and remains at `Triggered` with zero filled quantity. There are also restart and external-adoption paths where the sole/current ID is already the child `ordId`; that ID is then sent as `algoId`, which OKX prioritizes, and the query returns no report.

Could you handle both single-ID states and add regressions for a completely missed child event and a cache whose only/current ID is the child? This is the ordinary single-child path, so it doesn't require recursive `ordIdList` or `subAlgoIdList` reconciliation.

Thanks again for sticking with the review and working through each round of feedback.

silarin:
Hi @cjdsellers,

I’ve pushed the follow-up at `24d180fd1`.

Both requested single-ID states are now handled:

- When the child WebSocket event is completely missed, `QueryOrder` queries the parent using `algoClOrdId`. If the response exposes the generated child `ordId`, it performs one exact regular-order lookup and selects the child’s more advanced fill state.
- When the cache’s sole/current venue ID is already the child `ordId`, that ID is not sent as `algoId`. The parent is queried using `algoClOrdId`, and the cached child ID is used for the regular-order lookup.
- A pre-trigger parent whose returned identity matches the cached ID does not cause an unnecessary regular-order lookup.
- The cached authoritative `ordId` is retained through final report selection for adopted external regular orders when the command venue ID is absent or stale.

The patch remains limited to the ordinary single-child path. It does not add recursive `ordIdList` or `subAlgoIdList` traversal or broader multi-child lifecycle reconciliation.

Validation after rebasing onto current `develop`:

- Focused query-order tests: 26 passed
- Full `nautilus-okx` suite: 881 passed
- `make format`: passed
- Strict all-target, all-feature `nautilus-okx` Clippy: passed
- Independent review: no unresolved findings
- Range-diff confirmed the three rebased commits are patch-identical

`make pre-commit` passes every hook except the non-Linux network Turmoil Clippy check at `crates/network/src/socket/client.rs:993`. I reproduced that same failure on a clean current `origin/develop` checkout at `409214a9d`; it is unrelated to this PR.

Could you take another look when convenient?

cjdsellers (approved):
@silarin, thanks for addressing the review feedback. I’ve had another look at the latest changes, and the cases from the last review are covered, so let’s proceed with this.

I’ll follow up locally after merge on a remaining edge case: have the `Algo` route follow the returned child `ordId` when the cached conditional has no venue ID, and add a regression test covering missed acceptance and child WebSocket updates.

PR #4740, opened 2026-08-13
Guard the domain of the f32 ln approximation

folknor:
`<f32 as BlackScholesReal>::ln` in `crates/model/src/data/black_scholes.rs` reconstructs a logarithm from raw IEEE-754 bits and assumes a positive, finite, normal input without checking any of the three. A follow-up to the exponent-range work in #4709, in the same file.

## Unchecked input domain

The method reads `exponent` from `bits >> 23` and rebuilds a mantissa from the fraction field. The sign bit is not discarded by that shift - it survives as an additional 256 in the exponent, so `ln(-1.0)` returns `256 * LN_2 = 177.44568` rather than NaN. Subnormals are read as normal mantissas with an implicit leading one they do not carry, so the smallest returns `-88.02969` against a correct `-103.27893`. The two signed zeros diverge, `+0.0` giving `-88.02969` and `-0.0` following the sign-bit path to `+89.415985`, where IEEE specifies `-inf` for both. Infinities and NaNs reconstruct finite values, `+inf` giving `128 * LN_2 = 88.72284`.

Every one of these returns a plausible finite number where the correct result is a special value or a different magnitude, so a caller has nothing to test for.

`ln` now classifies the raw bits before evaluating: zero magnitude returns `-inf`, NaN and negative inputs return NaN, `+inf` returns `+inf`, and positive subnormals are normalized by an exact multiplication by `2^23` and corrected by `23 * LN_2` afterwards. The zero test precedes the sign test so `-0.0` reaches `-inf` rather than NaN. Positive normal inputs keep the original statements, constants and `mul_add` sequence unchanged.

## Behaviour

Results for positive normal `ln` arguments are byte-identical. Pricing results may change where the logarithm argument falls outside that range - this is a property of the argument, not of `s` and `k`, since valid positive operands can still form a subnormal or infinite `s / k`. Both `compute_greeks` and `compute_iv_and_greeks` call this method through the generic parameter, so both are affected. The PR validates no Black-Scholes parameter and changes no constructor or API policy.

## Testing

`test_ln_special_values` pins both signed zeros, negative finite values, `-inf`, `+inf`, and NaNs with explicit payload bits. `test_ln_positive_subnormals` compares the smallest, a middle, and the largest subnormal against `f32::ln` within three ULPs; the smallest is what makes it differential, since the largest happens to return `-87.33655` and is close to correct unfixed. `test_compute_greeks_negative_strike_returns_nan_price` and its `compute_iv_and_greeks` counterpart pass a negative strike and assert `price` is NaN, which pins that both generic callers resolve to this method rather than to the inherent `f32::ln`. A negative strike is used rather than a zero one because `ln(s/0)` is `+inf`, and the kernel maps an infinite `d1` back to finite limiting values that the unfixed large finite `d1` also produces.

These four fail against the unfixed implementation. `test_ln_positive_normal_path_is_unchanged` compares against a frozen copy of the previous formula across all 254 normal exponent fields and eight mantissa patterns; it is a characterization test and passes either way.

### DIFF
--- a/crates/model/src/data/black_scholes.rs
+++ b/crates/model/src/data/black_scholes.rs
@@ -105,30 +105,12 @@ impl BlackScholesReal for f32 {
         // See: J.-M. Muller et al., "Handbook of Floating-Point Arithmetic", 2018, Section 10.2
         //      A. J. Salgado & S. M. Wise, "Classical Numerical Analysis", 2023, Chapter 10
         let bits = self.to_bits();
-        let magnitude = bits & 0x7fff_ffff;
-        if magnitude == 0 {
-            return Self::NEG_INFINITY;
-        }
 
-        let exponent_bits = magnitude & 0x7f80_0000;
-        let fraction_bits = magnitude & 0x007f_ffff;
-        if exponent_bits == 0x7f80_0000 && fraction_bits != 0 {
-            return Self::NAN;
-        }
-
-        if bits & 0x8000_0000 != 0 {
-            return Self::NAN;
-        }
-
-        if exponent_bits == 0x7f80_0000 {
-            return Self::INFINITY;
-        }
-
-        if exponent_bits == 0 {
-            // Scaling by 2^23 is exact and carries every positive subnormal into the
-            // normal range, so this recurses exactly once.
-            return <Self as BlackScholesReal>::ln(self * 8_388_608.0)
-                - 23.0 * std::f32::consts::LN_2;
+        // Positive normal bit patterns occupy one contiguous interval, so the ordinary
+        // path reaches the polynomial through a single range test. Everything else -
+        // zeros, negatives, subnormals, infinity, NaN - is handled out of line.
+        if !(0x0080_0000..0x7f80_0000).contains(&bits) {
+            return ln_f32_outside_normal_range(self);
         }
 
         let exponent = ((bits >> 23) as i32 - 127) as Self;
@@ -411,6 +393,39 @@ pub fn compute_iv_and_greeks<T: BlackScholesReal>(
     g_final.vol = vol;
 
     g_final
+}
+
+/// Returns `ln(value)` for the inputs outside the positive normal range: zeros, negatives,
+/// subnormals, infinity, and NaN.
+///
+/// Kept out of line because `<f32 as BlackScholesReal>::ln` is `#[inline(always)]`, so any
+/// classification left in its body is duplicated at every call site on the pricing path.
+#[cold]
+#[inline(never)]
+fn ln_f32_outside_normal_range(value: f32) -> f32 {
+    let bits = value.to_bits();
+    let magnitude = bits & 0x7fff_ffff;
+    if magnitude == 0 {
+        return f32::NEG_INFINITY;
+    }
+
+    let exponent_bits = magnitude & 0x7f80_0000;
+    let fraction_bits = magnitude & 0x007f_ffff;
+    if exponent_bits == 0x7f80_0000 && fraction_bits != 0 {
+        return f32::NAN;
+    }
+
+    if bits & 0x8000_0000 != 0 {
+        return f32::NAN;
+    }
+
+    if exponent_bits == 0x7f80_0000 {
+        return f32::INFINITY;
+    }
+
+    // Only positive subnormals remain. Scaling by 2^23 is exact and carries every one of
+    // them into the normal range, so this re-enters the ordinary path exactly once.
+    <f32 as BlackScholesReal>::ln(value * 8_388_608.0) - 23.0 * std::f32::consts::LN_2
 }
 
 // 4. UNIT TESTS
--- a/crates/model/src/data/black_scholes.rs
+++ b/crates/model/src/data/black_scholes.rs
@@ -105,6 +105,32 @@ impl BlackScholesReal for f32 {
         // See: J.-M. Muller et al., "Handbook of Floating-Point Arithmetic", 2018, Section 10.2
         //      A. J. Salgado & S. M. Wise, "Classical Numerical Analysis", 2023, Chapter 10
         let bits = self.to_bits();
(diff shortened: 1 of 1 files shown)


faysou:
Note that ifs should be avoided as much as possible for high performance code. Maybe it would actually be better to check inputs before calling the pricing kernels than making the pricing kernels slower.

faysou:
We could maybe use this library for more reliability around f32 special functions. https://docs.rs/micromath/latest/micromath/struct.F32.html

Putting ifs in special functions including in exp makes it incompatible with simd.

folknor:
@faysou Thank you - the normal-path cost is a fair objection. This is `#[inline(always)]`, so five sequential tests enlarge every call site. I should have caught this - my bad.

On SIMD: `BlackScholesReal::ln` is implemented per concrete type, so a vector type supplies its own masked `ln` rather than executing this one lane by lane. The body is already f32-specific through `to_bits`, `from_bits` and scalar integer arithmetic, so it could not be reused verbatim for a vector type with or without the branches. And `select` for `f32` is itself an `if`, so expressing the classification through `select`/`cmp_gt` would compile to the same branches.

micromath documents no special-value or subnormal behaviour for `ln` that I can see, states that it trades precision for speed and code size, and its `ln` carries an early return at `|x - 1| < EPSILON` plus a conditional inversion and a sign conditional. Adopting it would be an accuracy and behaviour change rather than a fix for this.

Validating before the kernels does not cover it either: `compute_greeks` and `compute_iv_and_greeks` are public, and valid positive `s` and `k` can still form a ratio that overflows to infinity or falls subnormal, which is the argument `ln` receives.

Positive-normal `f32` bit patterns occupy one contiguous interval, so the ordinary path can take a single predictable guard into the existing polynomial:

```rust
let bits = self.to_bits();
if (0x0080_0000..0x7f80_0000).contains(&bits) {
    // existing approximation, unchanged
}
// cold: zeros, negatives, subnormals, infinity, NaN
```

The polynomial is untouched inside the guard, and the subnormal case re-enters through it once.

1. Shall I rework the PR to that shape, with before-and-after benchmarks in `black_scholes_criterion.rs` rather than an asserted cost?
2. The merged `exp` has the same scalar shape. Do you want the one-guard form there as well, as its own measured PR?

faysou:
It's true that we can use different algorithms when simd will land one day. Yes it would be interesting to benchmark the used functions with and without the changes you are making. Maybe you could do baseline before the merged commit of today and with all the changes you are working on. Thank you

cjdsellers:
@folknor, I ran the existing Criterion benchmarks against both requested baselines.

Against the merged #4709 base, the current head regresses `compute_iv_and_greeks_f32` by 2.21%, with a 95% interval of 1.36% to 3.10%. `compute_greeks_f32` measured 1.44% slower, but Criterion classified that result within its noise threshold. Against the commit immediately before #4709, the combined changes measured approximately 18.5% and 28.3% slower respectively.

The numerical behavior in this PR looks sound, and the current CI passes.

I agree with @faysou's broader point here: future SIMD implementations can use different algorithms, but until SIMD support lands these scalar functions remain the active hot paths and need to stay reasonably optimized. Given the measured regression, can you rework `ln` around the single positive-normal guard you proposed and rerun both comparisons?

I think the `exp` guard should remain a separate measured follow-up, but the combined baseline confirms that it is worth addressing too.

folknor:
@cjdsellers Thank you for running those - reworked to the single positive-normal guard, and both comparisons are below.

`ln` now takes one range test over the raw bits. Positive normal `f32` patterns occupy the contiguous interval `[0x0080_0000, 0x7f80_0000)`, so that test admits exactly them and falls into the polynomial unchanged. Zeros, negatives, subnormals, infinity, and NaN move to a `#[cold]` `#[inline(never)]` function returning the same values in the same order, with the subnormal case re-entering the ordinary path once. The point is not only the branch count: `ln` is `#[inline(always)]`, so the five sequential tests and the subnormal recursion were being duplicated at every call site on the pricing path.

## Measurements

Criterion, 300 samples, 10s measurement time, each commit built and run in its own worktree.

Against the merged #4709 base (`8fa769d4eb`):

| | before | after |
| --- | --- | --- |
| `compute_iv_and_greeks_f32` | +11.32% [11.11, 11.54] | +2.14% [1.93, 2.33] |
| `compute_greeks_f32` | +0.41%, within noise | -22.26% [-22.44, -22.08] |

Against the commit before #4709 (`03062cce63`):

| | before | after |
| --- | --- | --- |
| `compute_iv_and_greeks_f32` | +19.39% [19.13, 19.64] | +9.53% [9.30, 9.77] |
| `compute_greeks_f32` | +7.16% [6.98, 7.32] | -17.04% [-17.22, -16.86] |

All p = 0.00.

The `compute_greeks_f32` improvement is larger than the change accounts for, and I have no mechanism for it. It beats the pre-#4709 code by 17% while doing strictly more work per call, which points at codegen around the smaller `ln` body rather than at the guard. It is not an accuracy trade: no test changed, and `test_ln_positive_normal_path_is_unchanged` still compares `to_bits()` against the reference formula for every normal exponent crossed with eight mantissa patterns.

faysou:
It would be interesting to know what happens in the iv computation that slows down things with the new approximations. Not sure it's possible to understand what happens. Or what accelerates compute_greeks but not compute_iv_and_greeks. Maybe outputting the code in ASM would shed some light. I'm just saying this in principle out of curiosity, it's not like I need the absolute top speed anyway (at least not now).

cjdsellers (approved):
Hey @folknor,

Thanks for the rework and measurements. I've had another look at the latest changes, and the single positive‑normal guard with the cold exceptional path looks good. I also checked every positive subnormal: the worst case was 1 ULP from `f32::ln`, with no monotonicity violations. The positive‑normal path remains bit‑identical.

I don't think the unexplained benchmark split needs to block this change. This looks good to merge.

PR #4744, opened 2026-08-13
Standardize risk engine order denied reasons

folknor:
A follow-up to the reason-code standardization in [`0301425975`](https://github.com/nautechsystems/nautilus_trader/commit/030142597520211cb199e2c897686efc90422772), which converted most of the risk engine's denials and left several branches emitting free-form text.

## The denials still rendered as prose

`RiskEngine::deny_order` takes `reason: &str`, so a caller may pass a rendered `OrderDeniedReason` or an arbitrary string. Eleven sites pass strings: both `check_price` conditions and their trigger-price variants, all three `check_quantity` conditions, two notional calculation failures, an initial-margin failure, a cumulative-margin overflow, a betting balance-lock failure, and the missing-market-price denial. Each reaches an `OrderDenied` whose reason carries no code, while `docs/concepts/execution.md` documents the codes as the source of truth for locally denied orders.

The clearest case is a condition already in the vocabulary. `check_orders_risk_for_account` denies a max-quantity breach as `QUANTITY_EXCEEDS_MAXIMUM: effective_quantity=15, max_quantity=10`; `check_quantity` denies the same condition as `quantity 15 invalid (> maximum trade size of 10)`.

`check_price` and `check_quantity` now return `Option<OrderDeniedReason>` and their callers render at the existing `deny_order` boundary. The min/max arms reuse `QuantityExceedsMaximum` and `QuantityBelowMinimum`; the calculation failures take one variant each, following `TrailingStopCalcFailed` and `QuantityConversionFailed`.

NB: the comment above `check_quantity`'s quote-quantity early return said the skipped bounds "will be checked in `check_orders_risk` using `effective_quantity`". They are not - `check_orders_risk_for_account` guards the same comparisons with `if !order.is_quote_quantity()` and skips them too. Corrected in passing.

## The price leg becomes a field

The trigger-price denial was `format!("trigger {risk_msg}")`, a prefix ahead of the message, which a typed reason cannot carry without putting prose before the code. The leg moves into the reason as `OrderPriceType::{Order, Trigger}`, so both render code-led: `PRICE_NOT_POSITIVE: price_type=TRIGGER, price=-0.1`. Four parallel variants would have forced consumers to match two codes for one rule.

## Whether `deny_order` should still take a string

With these branches converted, every caller of `deny_order`, `deny_order_list` and `deny_command` in this engine passes a rendered `OrderDeniedReason`. Narrowing those three to `&OrderDeniedReason` and rendering inside them would make a free-form risk-engine denial unrepresentable rather than merely absent. I left it out because it changes signatures beyond the branches at issue - say the word and I will add it here or file it separately.

`reject_modify_order` is not a candidate: closed orders, pending cancellations, missing instruments, halted trading and rate limits legitimately emit prose there.

## What this does not cover

Local denials outside the risk engine still emit free-form text - execution clients across ten adapters, and `crates/trading`'s market-exit path. Two of those, `MARKET_EXIT_IN_PROGRESS` and Interactive Brokers' `UNSUPPORTED_QUOTE_QUANTITY`, are formatted like codes without being in `OrderDeniedCode`. Those need per-adapter decisions and are untouched here.

Shared price and quantity failures also reach `OrderModifyRejected`, which now carries the same code-led text. Modify rejections are not standardized generally, so the documentation for that event is unchanged.

## Testing

No validation behaviour changes, so there is no behavioural differential to show: the tests pin the reason format, not which orders are denied. Existing risk-engine assertions moved to the new codes and still pin event type and count. `test_submit_order_when_invalid_trigger_price_then_denies` had its message assertion commented out; it is live again and pins `price_type=TRIGGER`. Three error branches that had no coverage now have it - initial-margin overflow, cumulative-margin overflow past `Money` bounds, and a betting balance-lock failure - each induced through fixture values with no production-code change. Check ordering is unchanged at every site, which is load-bearing: it is first-failure-wins, so a reordering would change which reason an order failing two rules receives.

### DIFF
--- a/crates/model/src/events/mod.rs
+++ b/crates/model/src/events/mod.rs
@@ -37,7 +37,7 @@ pub use crate::events::{
         canceled::OrderCanceled,
         canceled_batch::OrderCanceledBatch,
         denied::OrderDenied,
-        denied_reason::{OrderDeniedCode, OrderDeniedReason},
+        denied_reason::{OrderDeniedCode, OrderDeniedReason, OrderPriceType},
         emulated::OrderEmulated,
         expired::OrderExpired,
         fill_voided::OrderFillVoided,
--- a/crates/model/src/events/order/denied_reason.rs
+++ b/crates/model/src/events/order/denied_reason.rs
@@ -29,8 +29,18 @@ use thiserror::Error;
 use crate::{
     enums::{OrderSide, OrderType, TimeInForce, TrailingOffsetType},
     identifiers::{ClientId, InstrumentId, OrderListId, PositionId, Venue},
-    types::{Money, Quantity},
+    types::{Money, Price, Quantity},
 };
+
+/// The order price field being validated.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Display)]
+#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]
+pub enum OrderPriceType {
+    /// The order's execution price.
+    Order,
+    /// The order's trigger price.
+    Trigger,
+}
 
 /// A standardized reason an order was denied locally by the Nautilus system.
 ///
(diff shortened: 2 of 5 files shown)


cjdsellers (approved):
Hi @folknor,

Thanks for the PR. I've had another look at this, and the implementation looks good to merge.

I'm going to follow up locally with a broader standardization pass over the canonical denial taxonomy. That needs to cover the existing and new codes together, including full‑word naming, structured field names, the order‑price field terminology, and consistent invalid‑side classification, so it makes more sense for me to handle it as one maintainer‑owned change rather than ask you to extend this PR beyond its current scope.

Thanks again for finishing the remaining risk engine conversions.

PR #4749, opened 2026-08-14
Reset the inner moving average in AverageTrueRange

mkzung:
## Summary

reset() does not touch self.ma.

It clears previous_close, value, count, has_inputs and initialized, and self.ma is where the state actually lives: update_raw feeds it and then reads self.value = self.ma.value(). So the first values after a reset come from the previous run. At period 3 on five bars, the first run starts at 0.1000 and the replay starts at 0.2333, and it stays wrong for period - 1 values until the stale window has rolled out.

22 indicators in this crate hold another indicator inside. On develop this is the only one that does not reset it, and RelativeStrengthIndex already has test_reset_resets_inner_mas for the same thing.

## Type of change

## Testing

Added test_reset_resets_inner_ma, the same shape as rsi.rs.

544 tests pass in nautilus-indicators and make pre-commit is green over 61 hooks. Removing the one added line turns the new test red, and restoring it turns it green.

### DIFF
--- a/crates/indicators/src/volatility/atr.rs
+++ b/crates/indicators/src/volatility/atr.rs
@@ -84,6 +84,7 @@ impl Indicator for AverageTrueRange {
     }
 
     fn reset(&mut self) {
+        self.ma.reset();
         self.previous_close = 0.0;
         self.value = 0.0;
         self.count = 0;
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @mkzung,

Thanks for the PR. This is correct, and it also picks up `KeltnerChannel`, `VolatilityRatio`, and `Pressure`, since all three call `atr.reset()` from their own `reset()`. I went through the rest of the crate as well and `AverageTrueRange` was indeed the last one holding an inner indicator it wasn't resetting. This looks good to merge.

One adjacent thing I noticed while reading: `AverageTrueRange::new` passes `MovingAverageType::Simple` straight to `MovingAverageFactory::create` rather than the caller's `ma_type`, so a non-Simple `ma_type` gets stored and shown in `Display` but never actually used. Every other indicator in the crate passes it through. That changes output for anyone currently constructing an ATR with a non-Simple type, so it needs its own PR and I'll follow up on it separately.

PR #4750, opened 2026-08-14
Insert encoded query parameters before any URL fragment

folknor:
`HttpClient` builds its final request URL in `encode_url_params` (`crates/network/src/http/client.rs`) by concatenating the encoded query onto the end of the URL string, with no awareness of a fragment.

## The query lands inside the fragment

RFC 3986 orders the components `scheme://authority/path?query#fragment`, so everything after the first `#` belongs to the fragment. Appending to the end of the string therefore puts the encoded parameters inside it: `https://x/y#frag` with `a=b` becomes `https://x/y#frag?a=b`. A fragment is never sent to the server, so the parameters the caller supplied are dropped from the request and nothing reports an error. The fragment is also mutated in place - `frag` becomes `frag?a=b` - which is visible to anything that logs or returns the constructed URL, and an explicitly empty fragment `https://x/y#` acquires the fragment `?a=b`.

The separator test reads the wrong component for the same reason. `url.contains('?')` is true for a `?` that occurs only inside the fragment, which RFC 3986 section 3.5 permits, so `https://x/y#section?display=full` selects `&` despite having no query component at all.

The fix splits the URL once on the first literal `#`, inserts the query into the part preceding it, and reattaches the fragment unchanged. A data `#` in an earlier component must be percent-encoded as `%23`, so it cannot split. URLs without a fragment take the former construction exactly, and both `Cow::Borrowed` fast paths are unchanged.

No in-tree default or hard-coded adapter endpoint contains a fragment. Polymarket Gamma is the only adapter caller supplying this parameter shape, and its standard base URL is fragment-free. This therefore fixes an unenforced correctness property on the public `HttpClient` URL path rather than request corruption observed under an in-tree configuration; caller-supplied URLs and configurable adapter base URLs are outside the repository's control.

## Testing

`test_query_is_inserted_before_the_fragment` is parameterized over six URLs: no fragment, an existing query, a fragment, a fragment with an existing query, a fragment containing `?`, and an empty fragment. The four fragment-bearing cases fail against the unfixed function - `https://x/y#frag` yields `https://x/y#frag?a=b`, and `https://x/y?old=1#frag` yields `https://x/y?old=1#frag&a=b` - while the two without a fragment pass unchanged, pinning that this alters nothing for the ordinary shape. `test_url_is_borrowed_when_no_params_are_supplied` and `test_url_is_borrowed_when_params_are_empty` match on `Cow::Borrowed` so the zero-allocation paths cannot regress into allocating on every request.

These are pure string tests and live in their own module rather than the file's existing `#[cfg(target_os = "linux")]` module, which is gated for server-backed tests, so they run on every target. `socket/client.rs` and `websocket/client.rs` carry the same split.

### DIFF
--- a/crates/network/src/http/client.rs
+++ b/crates/network/src/http/client.rs
@@ -669,6 +669,7 @@ impl Default for InnerHttpClient {
 /// Returns `Cow::Borrowed` when no parameters need appending (zero-alloc fast path).
 /// Parameters can have multiple values per key (for doseq=True behavior).
 /// Preserves existing query strings in the URL by appending with '&' instead of '?'.
+/// The query is inserted before any fragment, which is preserved unchanged.
 fn encode_url_params<'a>(
     url: &'a str,
     params: Option<&HashMap<String, Vec<String>>>,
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. This looks good to merge: splitting on the first literal `#` and choosing the separator from the pre-fragment part only is the right reading of RFC 3986, and I appreciate the tests pinning both `Cow::Borrowed` fast paths alongside the fragment cases.

I checked the other URL-building paths too: the websocket client derives its resource from `url::Url` accessors, which already respect fragments, so no other site needs the same handling.

PR #4751, opened 2026-08-14
Reserve margin on absolute notional value

folknor:
A follow-up to the negative-price handling in #4725, which fixed the sibling path where a negative-price buy tripped the `locked.raw >= 0` assert in `base_calculate_balance_locked`.

## Margin is reserved on a signed notional

`try_notional_value` (`crates/model/src/instruments/mod.rs`) requires a positive price only on the inverse branch. The ordinary branch computes `quantity * multiplier * price` and carries the price's sign, so a negative price yields a negative notional. Both margin models then multiply that straight through: `StandardMarginModel` by `margin_init` / `margin_maint`, and `LeveragedMarginModel` by the same rates after dividing by leverage. The result is a negative margin requirement. `InstrumentClass::allows_negative_price` covers `Option`, `FuturesSpread` and `OptionSpread`, so a calendar spread quoted below zero reaches this path through the ordinary non-inverse branch.

Nothing downstream rejects it. `MarginBalance::new_checked` validates currency equality but not sign, and `recalculate_balance` clamps only when margin exceeds the total balance, so the negative requirement flows into `free = total - margin` and inflates the free balance - directly against the comment above that code stating locked margin must never be negative. The risk engine then checks orders against the inflated figure, and the negative requirement itself cannot exceed a positive free balance under its `margin_req > margin_free` comparison.

The fix takes the absolute notional in all four margin methods, before the leverage division and the rate multiplication. A requirement reserves against exposure magnitude, so a spread quoted at -2.00 reserves what the same spread quoted at +2.00 does.

The change is confined to the margin models deliberately. `try_notional_value` also feeds cash-account fill PnL, commission calculation and premium-instrument PnL, where the sign is meaningful; only a margin requirement cannot be negative, so the invariant belongs at the margin-policy boundary rather than in the shared helper. Directional netting is unaffected because it already happens earlier - positions are netted by signed quantity and converted to absolute magnitude before the margin model is called.

## Testing

`test_standard_margin_is_positive_for_a_negative_price` and `test_leveraged_margin_is_positive_for_a_negative_price` cover both models across initial and maintenance margin, using a `FuturesSpread` built with `margin_init` 0.01 and `margin_maint` 0.02. The rates are set explicitly because the existing spread stub leaves both at zero, which would let every assertion pass on a zero requirement.

Both assert exact decimal values rather than a sign, since a positivity-only assertion would also pass on an implementation that clamped to zero. The standard test additionally asserts that the negative price reserves exactly what the equivalent positive price reserves. Against the unfixed calculation the standard case yields -2.00 where 2 is required and the leveraged case -0.20 where 0.2 is required.

### DIFF
--- a/crates/model/src/accounts/margin_model.rs
+++ b/crates/model/src/accounts/margin_model.rs
@@ -164,8 +164,11 @@ impl MarginModel for StandardMarginModel {
     ) -> anyhow::Result<Money> {
         let use_quote = use_quote_for_inverse.unwrap_or(false);
         let notional = instrument.try_calculate_notional_value(quantity, price, Some(use_quote))?;
+        // Spreads and options may quote negative, which carries the sign into the notional.
+        // A requirement is a reserve against exposure magnitude, so take it on `abs`.
         let margin = notional
             .as_decimal()
+            .abs()
             .checked_mul(instrument.margin_init())
             .ok_or_else(|| anyhow::anyhow!("initial margin calculation overflow"))?;
         let currency = margin_currency(instrument, use_quote)?;
@@ -184,6 +187,7 @@ impl MarginModel for StandardMarginModel {
         let notional = instrument.try_calculate_notional_value(quantity, price, Some(use_quote))?;
         let margin = notional
             .as_decimal()
+            .abs()
             .checked_mul(instrument.margin_maint())
             .ok_or_else(|| anyhow::anyhow!("maintenance margin calculation overflow"))?;
         let currency = margin_currency(instrument, use_quote)?;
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the follow-up. Taking `abs` at the margin-model boundary and leaving the signed notional alone for valuation and PnL is the right call, and I confirmed the four margin methods are the only places the notional meets a margin rate. It's also good that the tests use non-zero rates and assert exact values, so a clamp-to-zero implementation couldn't pass.

This looks good to merge.

PR #4752, opened 2026-08-14
Skip a rehedge whose quantity rounds to zero

folknor:
## A rehedge quantity that rounds to zero aborts the strategy

`DeltaNeutralVol::check_rehedge` decides whether to hedge from an unrounded float and then builds the order quantity from a different representation. `should_rehedge` compares `portfolio_delta().abs()` against the configured `rehedge_delta_threshold`, while the order takes `Quantity::new(hedge_qty, size_precision)` with the hedge instrument's precision. Nothing relates the two, so a delta above the threshold can still quantize to zero: with whole-contract sizing and a threshold of 0.1, every delta below 0.5 passes the gate and yields a zero quantity. `OrderFactory::market` ends in `unwrap_or_else(|e| panic!("{e}"))` and `MarketOrder::new_checked` opens with `check_positive_quantity`, so the strategy panics, and release builds are `panic = "abort"`.

The default `rehedge_delta_threshold` of 0.5 does not reach this. Tightening the hedge tolerance below what the instrument can express does, which is an ordinary thing to configure.

The fix constructs the quantity once and returns before submitting when it is zero, logging the delta and precision at debug because a timer can meet the same unexpressible residual on every tick. Rounding up to one increment was rejected: it overhedges and can reverse the residual, so a desired sell of 0.4 contracts becoming 1.0 leaves roughly -0.6 rather than +0.4. That is a policy change requiring hysteresis or a minimum-hedge setting, not a repair.

The guard sits before the `hedge_pending` latch. Setting the latch and then skipping would leave it true with no order able to produce a terminal event, wedging the strategy against all later rehedges. The existing submission log also moves below the guard, so it no longer announces a submission on the skip path, and it now reports the quantity actually sent rather than the pre-quantization float.

## Testing

`test_rehedge_skips_quantity_rounded_to_zero` registers the configured hedge instrument as a `CryptoPerpetual` with `size_precision` 0, sets the threshold to 0.1 and the portfolio delta to 0.4, and drives the strategy through its timer handler. It asserts that `hedge_pending` remains clear, so a later rehedge at an expressible size is not blocked.

Against the unfixed code the test fails with `invalid 'Quantity' for 'quantity' not positive, was 0` from `factories/order.rs`. The instrument is cached with an explicit precision rather than relying on the missing-instrument fallback, so the test pins zero-quantity protection alone and does not depend on what that fallback happens to be.

### DIFF
--- a/crates/trading/src/examples/strategies/delta_neutral_vol/strategy.rs
+++ b/crates/trading/src/examples/strategies/delta_neutral_vol/strategy.rs
@@ -439,11 +439,6 @@ impl DeltaNeutralVol {
             OrderSide::Buy
         };
 
-        log::info!(
-            "Rehedging: portfolio_delta={delta:.4}, submitting {side:?} {hedge_qty:.4} on {}",
-            self.config.hedge_instrument_id,
-        );
-
         let hedge_id = self.config.hedge_instrument_id;
         let size_precision = {
             let cache = self.cache();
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. This looks good to merge. Placing the guard before the `hedge_pending` latch is the subtle part of this fix, and you're right that the other order would leave the strategy wedged with no order event able to clear it.

PR #4761, opened 2026-08-15
Allow verified emergency position exits under a trading halt

folknor:
Replaces the `validate_order_risk` entry point with the narrow gateway exception agreed on the thread.

## The HALTED gateway exception

`execution_gateway` denies every submit under `TradingState::Halted`, so an operator's closing order is denied exactly when a halt makes it necessary, and the only alternative was bypassing the risk engine. `handle_submit_order` now classifies a submit before the risk checks: `is_emergency_position_exit` requires the `emergency_exit=true` params intent, a reduce-only order with positive quantity, command/order instrument agreement, and an identified open position - `position_id` on the command, the cached order mapped to it, the position open on the same instrument, and an opposing side. The classification reaches the gateway as a private two-variant enum; under `Halted` a verified exit forwards to the submit throttler and the normal event path, while ordinary submits, generic reduce-only submits, and all order lists keep the `TRADING_HALTED` denial. `Active` and `Reducing` ignore the classification.

A new `PARAMS_EMERGENCY_EXIT` key carries the intent. The Binance futures adapter reads `PARAMS_CLOSE_POSITION` to set venue-side `closePosition` on conditional orders, and `is_full_position_exit_order` requires a non-reduce-only conditional type - a different venue feature, so that key cannot double for this one.

## The route must enforce reduce-only

`order.is_reduce_only()` is not evidence that the venue will refuse to cross flat: the Interactive Brokers client builds an ordinary venue order and never encodes the instruction, so a fill in flight can make the engine's cached-position read stale and the exit can open exposure under a halt. The risk engine cannot answer this itself, because it cannot see which client will be selected - an explicit but unregistered `command.client_id` falls through to account, venue, then default routing in `find_client_for_command`.

`ExecutionClient` gains a default-false `enforces_reduce_only(&SubmitOrder, &OrderAny)`, contracted on the downstream path preventing any execution which would increase or reverse the position once it reaches flat, whether by native venue enforcement or by matching-engine enforcement against execution-time position state. `ExecutionEngine::handle_submit_order` checks it beside the existing `ClientVenueMismatch` denial and denies `REDUCE_ONLY_NOT_ENFORCED` before reaching the client. `BacktestExecutionClient` and `SandboxExecutionClient` qualify when `use_reduce_only` is enabled, which gates every clamp in `OrderMatchingEngine`; `BinanceFuturesExecutionClient` qualifies in one-way mode, where the WS, HTTP and algo paths all send native `reduceOnly` - in hedge mode `reduce_only_param` returns `None`, since Binance rejects `reduceOnly` alongside `positionSide`. Every other client keeps the default, which records that this PR has not established eligibility rather than that the venue cannot support it.

Because `emergency_exit` is caller-supplied, the risk engine removes the key before the bypass return, classifies from the captured value, and re-inserts it only on the branch it admitted under `Halted`, so an emergency-marked order still submits unchanged under `Active`. Caller authentication remains deferred per the thread.

## The bound is the submitted quantity

`would_reduce_only` compares `leaves_qty`, while the execution client reconstructs the venue request from `order.quantity()`. A partially filled order with quantity 10, filled 6 and leaves 4 against an open position of 4 therefore qualified and then submitted 10, crossing flat by 6. Eligibility now also requires `order.quantity() <= position.quantity`, inside the emergency classifier only; the shared position lookup is extracted so `would_reduce_only`, the upstream reduce-only pre-check and the full-position exit path are unchanged.

A partially filled order rejected by this bound produces no `OrderDenied`, because `deny_order` returns early for an order past `Initialized`. It cannot reach the client, and a denial event would be invalid for that order state.

## Testing

`test_submit_emergency_exit_when_trading_halted_forwards_to_execution` pins the forwarded command's client order id, position id, reduce-only flag and restamped params; `test_submit_emergency_exit_when_over_rate_limit_denies` pins the throttler path (`RATE_LIMIT_EXCEEDED` on the 11th submit). `test_submit_emergency_exit_uses_submitted_quantity_bound` pins quantity 10, filled 6, leaves 4 against a position of 4, asserting `would_reduce_only` still holds so the case exercises the new bound, plus the boundary where submitted quantity equals the position. `test_submit_emergency_exit_requires_capable_route` asserts the denial reason and that nothing reaches an incapable client, and submission on a capable one. Two provenance tests pin the intent key absent when forwarded under `Active`, on bypass, and for order lists. A six-case matrix pins each near-miss denial reason, an order list carrying a qualifying leg stays denied, a verified exit over max-notional is denied by the normal check, and the three capability predicates have their own matrices including Binance hedge mode, a disabled backtest exchange, and a sandbox with `use_reduce_only` off. The three existing `HALTED` tests and the `test_submit_close_position_*` family are unchanged.

Reverting the quantity bound makes the partial-fill case forward, and disabling the route check makes the incapable case submit.

### DIFF
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -1335,6 +1335,23 @@ fn resolve_order_position_identity(
     Ok((position_side, venue_position_id))
 }
 
+fn binance_enforces_reduce_only(
+    is_hedge_mode: bool,
+    command: &SubmitOrder,
+    order: &OrderAny,
+) -> bool {
+    let close_position = command
+        .params
+        .as_ref()
+        .and_then(|params| params.get_bool(PARAMS_CLOSE_POSITION))
+        .unwrap_or(false);
+
+    order.is_reduce_only()
+        && !is_hedge_mode
+        && !close_position
+        && order_type_to_binance_futures(order.order_type()).is_ok()
+}
+
 fn validate_submit_position_id(
     submitted_position_id: Option<PositionId>,
     venue_position_id: Option<PositionId>,
@@ -1672,6 +1689,10 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
 
     fn get_account(&self) -> Option<AccountAny> {
         self.core.cache().account_owned(&self.core.account_id)
+    }
+
+    fn enforces_reduce_only(&self, command: &SubmitOrder, order: &OrderAny) -> bool {
+        binance_enforces_reduce_only(self.is_hedge_mode(), command, order)
     }
 
     async fn connect(&mut self) -> anyhow::Result<()> {
@@ -3801,10 +3822,12 @@ fn is_instrument_for_product(instrument: &InstrumentAny, product_type: BinancePr
 #[cfg(test)]
 mod tests {
     use nautilus_model::{
+        enums::OrderSide,
         instruments::stubs::{
             crypto_future_btcusdt, crypto_perpetual_ethusdt, currency_pair_btcusdt,
             perpetual_contract_eurusd, xbtusd_bitmex,
         },
+        orders::{Order, builder::OrderTestBuilder},
         types::Price,
     };
     use rstest::rstest;
@@ -3817,6 +3840,56 @@ mod tests {
             code,
             message: format!("test error {code}"),
         })
+    }
+
+    fn reduce_only_capability_command(order: &OrderAny, close_position: bool) -> SubmitOrder {
+        let params = close_position.then(|| {
+            let mut params = Params::new();
+            params.insert(PARAMS_CLOSE_POSITION.to_string(), true.into());
+            params
+        });
+
+        SubmitOrder::new(
+            order.trader_id(),
+            None,
+            order.strategy_id(),
+            order.instrument_id(),
+            order.client_order_id(),
+            order.init_event().clone(),
+            None,
+            None,
+            params,
+            UUID4::new(),
+            UnixNanos::default(),
+            None,
+        )
+    }
+
+    #[rstest]
+    #[case::one_way(false, OrderType::Market, true, false, true)]
+    #[case::hedge(true, OrderType::Market, true, false, false)]
+    #[case::not_reduce_only(false, OrderType::Market, false, false, false)]
+    #[case::close_position(false, OrderType::Market, true, true, false)]
+    #[case::unsupported_order_type(false, OrderType::MarketToLimit, true, false, false)]
+    fn test_binance_enforces_reduce_only_predicate(
+        #[case] is_hedge_mode: bool,
+        #[case] order_type: OrderType,
+        #[case] reduce_only: bool,
+        #[case] close_position: bool,
+        #[case] expected: bool,
+    ) {
+        let order = OrderTestBuilder::new(order_type)
+            .instrument_id(InstrumentId::from("BTCUSDT-PERP.BINANCE"))
+            .side(OrderSide::Sell)
+            .quantity(Quantity::from(1))
+            .reduce_only(reduce_only)
+            .build();
+        let command = reduce_only_capability_command(&order, close_position);
+
+        assert_eq!(
+            binance_enforces_reduce_only(is_hedge_mode, &command, &order),
+            expected
+        );
     }
 
     #[rstest]
--- a/crates/adapters/sandbox/src/execution.rs
+++ b/crates/adapters/sandbox/src/execution.rs
@@ -929,6 +929,12 @@ impl ExecutionClient for SandboxExecutionClient {
         self.cache.borrow().account_owned(&account_id)
     }
 
+    fn enforces_reduce_only(&self, _command: &SubmitOrder, order: &OrderAny) -> bool {
+        // Every matching engine is built from this config, and the engine clamps
+        // reduce-only fills against the position it holds at execution time.
+        order.is_reduce_only() && self.config.use_reduce_only
+    }
+
     fn generate_account_state(
         &self,
         balances: Vec<AccountBalance>,
--- a/crates/adapters/sandbox/tests/execution.rs
+++ b/crates/adapters/sandbox/tests/execution.rs
@@ -3693,3 +3693,37 @@ fn test_submit_order_through_exec_engine_no_reentrant_panic(
         "Expected order events through the exec event channel"
     );
 }
+
+#[rstest]
+#[case::enabled(true, true, true)]
+#[case::disabled(false, true, false)]
+#[case::not_reduce_only(true, false, false)]
+fn test_enforces_reduce_only_follows_config(
+    trader_id: TraderId,
+    account_id: AccountId,
+    venue: Venue,
+    instrument: InstrumentAny,
+    #[case] use_reduce_only: bool,
+    #[case] reduce_only: bool,
+    #[case] expected: bool,
+) {
+    let ctx = create_test_context_with(trader_id, account_id, venue, |config| {
+        config.use_reduce_only = use_reduce_only;
+    });
+    let order = OrderTestBuilder::new(OrderType::Market)
+        .instrument_id(instrument.id())
+        .side(OrderSide::Sell)
+        .quantity(Quantity::from("0.001"))
+        .reduce_only(reduce_only)
+        .build();
+    let command = SubmitOrder::from_order(
+        &order,
+        trader_id,
+        None,
+        None,
+        UUID4::new(),
+        UnixNanos::default(),
+    );
+
+    assert_eq!(ctx.client.enforces_reduce_only(&command, &order), expected);
+}
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -203,6 +203,10 @@ impl Debug for SimulatedExchange {
 }
 
 impl SimulatedExchange {
+    pub(crate) const fn use_reduce_only(&self) -> bool {
+        self.use_reduce_only
+    }
+
     /// Creates a new [`SimulatedExchange`] instance from a venue configuration.
     ///
     /// # Errors
--- a/crates/backtest/src/execution_client.rs
+++ b/crates/backtest/src/execution_client.rs
@@ -36,7 +36,7 @@ use nautilus_model::{
     enums::OmsType,
     events::OrderEventAny,
     identifiers::{AccountId, ClientId, ClientOrderId, TraderId, Venue},
-    orders::OrderAny,
+    orders::{Order, OrderAny},
     types::{AccountBalance, MarginBalance},
 };
 
@@ -161,6 +161,20 @@ impl ExecutionClient for BacktestExecutionClient {
 
     fn get_account(&self) -> Option<AccountAny> {
         self.cache.borrow().account_owned(&self.core.account_id)
+    }
+
+    fn enforces_reduce_only(&self, _command: &SubmitOrder, order: &OrderAny) -> bool {
+        order.is_reduce_only()
+            && self
+                .exchange
+                .upgrade()
+                .and_then(|exchange| {
+                    exchange
+                        .try_borrow()
+                        .ok()
+                        .map(|exchange| exchange.use_reduce_only())
+                })
+                .unwrap_or(false)
     }
 
     fn generate_account_state(
@@ -303,10 +317,11 @@ mod tests {
     use nautilus_core::UUID4;
     use nautilus_execution::models::latency::{LatencyModelHandle, StaticLatencyModel};
     use nautilus_model::{
-        enums::{AccountType, BookType, OmsType},
+        enums::{AccountType, BookType, OmsType, OrderSide, OrderType},
         identifiers::{InstrumentId, StrategyId},
+        orders::{Order, builder::OrderTestBuilder},
         stubs::TestDefault,
-        types::{Currency, Money},
+        types::{Currency, Money, Quantity},
     };
     use rstest::rstest;
 
@@ -314,6 +329,12 @@ mod tests {
     use crate::config::SimulatedVenueConfig;
 
     fn setup_client_with_latency() -> (BacktestExecutionClient, Rc<RefCell<SimulatedExchange>>) {
+        setup_client_with_reduce_only(true)
+    }
+
+    fn setup_client_with_reduce_only(
+        use_reduce_only: bool,
+    ) -> (BacktestExecutionClient, Rc<RefCell<SimulatedExchange>>) {
         let cache = Rc::new(RefCell::new(Cache::default()));
         let clock: Rc<RefCell<dyn Clock>> = Rc::new(RefCell::new(TestClock::new()));
         let latency_model = StaticLatencyModel::new(
@@ -329,6 +350,7 @@ mod tests {
             .book_type(BookType::L2_MBP)
             .starting_balances(vec![Money::new(1_000.0, Currency::USD())])
             .latency_model(LatencyModelHandle::new(latency_model))
+            .use_reduce_only(use_reduce_only)
             .build()
             .unwrap();
         let exchange = Rc::new(RefCell::new(
@@ -389,6 +411,71 @@ mod tests {
             .into();
 
         assert!(Rc::ptr_eq(&upgraded, &exchange));
+    }
+
+    #[rstest]
+    #[case::enabled(true, true, true)]
+    #[case::disabled(false, true, false)]
+    #[case::not_reduce_only(true, false, false)]
+    fn test_enforces_reduce_only_follows_exchange_config(
+        #[case] use_reduce_only: bool,
+        #[case] reduce_only: bool,
+        #[case] expected: bool,
+    ) {
+        let (client, _exchange) = setup_client_with_reduce_only(use_reduce_only);
+        let order = OrderTestBuilder::new(OrderType::Market)
+            .instrument_id(InstrumentId::from("AUD/USD.SIM"))
+            .side(OrderSide::Sell)
+            .quantity(Quantity::from(1))
+            .reduce_only(reduce_only)
+            .build();
+        let command = SubmitOrder::new(
+            TraderId::test_default(),
+            None,
+            StrategyId::test_default(),
+            order.instrument_id(),
+            order.client_order_id(),
+            order.init_event().clone(),
+            None,
+            None,
+            None,
+            UUID4::new(),
+            UnixNanos::default(),
+            None,
+        );
+
+        assert_eq!(client.enforces_reduce_only(&command, &order), expected);
+    }
+
+    #[rstest]
+    fn test_enforces_reduce_only_fails_closed_when_exchange_unavailable() {
+        let (client, exchange) = setup_client_with_reduce_only(true);
+        let order = OrderTestBuilder::new(OrderType::Market)
+            .instrument_id(InstrumentId::from("AUD/USD.SIM"))
+            .side(OrderSide::Sell)
+            .quantity(Quantity::from(1))
+            .reduce_only(true)
+            .build();
+        let command = SubmitOrder::new(
+            TraderId::test_default(),
+            None,
+            StrategyId::test_default(),
+            order.instrument_id(),
+            order.client_order_id(),
+            order.init_event().clone(),
+            None,
+            None,
+            None,
+            UUID4::new(),
+            UnixNanos::default(),
+            None,
+        );
+
+        let borrow = exchange.borrow_mut();
+        assert!(!client.enforces_reduce_only(&command, &order));
+        drop(borrow);
+        drop(exchange);
+        assert!(!client.enforces_reduce_only(&command, &order));
     }
 
     #[rstest]
--- a/crates/common/src/clients/execution.rs
+++ b/crates/common/src/clients/execution.rs
@@ -27,6 +27,7 @@ use nautilus_model::{
         AccountId, ClientId, ClientOrderId, InstrumentId, StrategyId, Venue, VenueOrderId,
     },
     instruments::InstrumentAny,
+    orders::OrderAny,
     reports::{ExecutionMassStatus, FillReport, OrderStatusReport, PositionStatusReport},
     types::{AccountBalance, MarginBalance, Money, Price, Quantity},
 };
@@ -72,6 +73,18 @@ pub trait ExecutionClient {
     /// the instrument's exchange venue.
     fn handles_order_venue(&self, venue: Venue) -> bool {
         self.venue() == venue
+    }
+
+    /// Returns whether this client, for this exact command and cached order, will use a
+    /// downstream execution path that prevents any execution which would increase or reverse the
+    /// position once it reaches flat.
+    ///
+    /// The guarantee may be native venue enforcement or matching-engine enforcement against
+    /// execution-time position state. Serializing a flag the venue ignores, selecting a nominal
+    /// closing side, relying on balance rejection, or checking a cached position before submission
+    /// is not sufficient because an in-flight fill can make that check stale.
+    fn enforces_reduce_only(&self, _command: &SubmitOrder, _order: &OrderAny) -> bool {
+        false
     }
 
     /// Returns whether a bulk position status report request provides complete coverage for the
--- a/crates/common/src/messages/execution/mod.rs
+++ b/crates/common/src/messages/execution/mod.rs
@@ -44,6 +44,9 @@ pub use self::{
 /// Parameter indicating that a conditional order should close the whole position at trigger time.
 pub const PARAMS_CLOSE_POSITION: &str = "close_position";
 
+/// Parameter indicating that a reduce-only order is an emergency position exit.
+pub const PARAMS_EMERGENCY_EXIT: &str = "emergency_exit";
+
 /// Execution report variants for reconciliation.
 #[derive(Clone, Debug, Display)]
 pub enum ExecutionReport {
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -48,7 +48,8 @@ use nautilus_common::{
         ExecutionReport,
         execution::{
             BatchCancelOrders, BatchModifyOrders, CancelAllOrders, CancelOrder, ModifyOrder,
-            QueryAccount, QueryOrder, SubmitOrder, SubmitOrderList, TradingCommand,
+            PARAMS_EMERGENCY_EXIT, QueryAccount, QueryOrder, SubmitOrder, SubmitOrderList,
+            TradingCommand,
         },
     },
     msgbus::{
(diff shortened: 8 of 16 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. The internals hold up well: the submission path keeps its denial reasons and fan-out intact, and sharing `check_order_trading_state` removes a duplicated `Reducing` predicate in `execution_gateway`.

One path to a risk decision keeps it deterministic and auditable, and those properties are worth preserving: it runs in response to a command, in a fixed order, and a denial leaves an `OrderDenied` on the bus. A synchronous entry point is a second route to the same decision with nothing holding the two in step, and it reports without leaving a record.

That leaves the question of what the flatten needs that a command doesn't already give it. Is it that you need the answer without the side effects, since `deny_order` inserts the order into the cache and publishes? If that's the requirement then I'd rather design for it directly, and we can discuss a potential direction separately.

folknor:
Thank you - and yes, but side-effect freedom is downstream of the constraint rather than the constraint itself.

## Why the flatten bypasses the engine today

The operator's flatten builds reduce-only closing orders and posts them to the execution engine queue - the same endpoint `send_to_execution` posts to, so an approved order and a flatten land in the same place. It skips the risk queue because of one arm:

`execution_gateway` under `TradingState::Halted` denies every `SubmitOrder` and every `SubmitOrderList` with `OrderDeniedReason::TradingHalted`. The order's flags are never read.

An operator flattens *because* something halted the account - a brake, a manual halt, a quarantine. Routing the flatten as an ordinary command therefore denies it exactly when it is wanted, and the operator's last resort becomes unavailable under the only conditions that produce it. What the bypass costs is everything else in the gate: price and quantity sanity and the max-notional bounds. A reduce-only close has no business being vetoed by `Halted`, and should still be checked for order sanity; today it is neither or both. #4724 and this PR were both attempts to get the sanity half back.

## The asymmetry

Cancels never enter the risk engine. `handle_command` matches `SubmitOrder`, `SubmitOrderList`, `ModifyOrder`, `ModifyOrders` and `QueryAccount`, and anything else falls through to `log::error!("Cannot handle command: {command}")`, so a cancel routes to execution under any trading state. Under `Halted` an operator may cancel every working order against a naked position but cannot submit the one reduce-only order that closes it.

`TradingState::Reducing`, the next arm of the same match as the halt, likewise denies only orders that increase exposure (`is_buy()` against `is_net_long`, `is_sell()` against `is_net_short`) and forwards the rest to the throttler.

And `check_orders_risk_for_account` exempts reduce-only three times over:

- the balance check, skipped when `is_reducing` - `order.is_reduce_only()` or a net-side test - logged "Position-reducing order skips balance check";
- the margin check, on the same predicate, logged "Position-reducing order skips margin check";
- the venue minimum notional check, `if !order.is_reduce_only()`, commented "Reduce-only orders may close residual positions below venue minimum", which you added in c59b27838a.

The engine carves reduce-only out of a check wherever the carve-out is correct. The trading-state check is the one that has not.

## A potentially less invasive alternative

A reduce-only exception on the `Halted` arm of `execution_gateway`, and nothing else:

1. Under `Halted`, a `SubmitOrder` whose order `is_reduce_only()` forwards to the throttler as it would under `Active`.
2. Everything else under `Halted` is denied as today - openers, and any `SubmitOrderList` carrying a non-reduce-only leg.
3. Key it on `is_reduce_only()` alone, not on the portfolio net-side test `Reducing` uses. Under a halt the explicit order flag is the narrower and more auditable predicate, and it consults no portfolio state that the halt may itself have been declared over.

The halt stays intact for ordinary strategy flow. What the flatten would then be checked against is price, quantity and the max-notional bounds; balance, margin and minimum notional stay skipped for it by the three carve-outs above, which is the intended behaviour for a position-reducing order rather than a shortfall. It also keeps one path to the decision: a command, a fixed order, an `OrderDenied` on the bus. If you prefer that direction, `validate_order_risk` is unnecessary and I will close this PR.

Is `Halted` meant as no new exposure, in which case reduce-only belongs on the allowed side and is consistent with the cancel path? Or as no discretionary submits at all - in which case, where is the operator's must-exit case meant to live?

cjdsellers:
Thanks for the write up @folknor.

I'll just pause on this one until after the next release (likely today but we'll see).

folknor:
Let me know when/if you want me to bring this PR up to speed with develop, please.

cjdsellers:
Hey @folknor,

I've had another look at this after the recent risk-engine changes. The implementation itself holds up, but I don't think we should ask you to rebase this shape yet.

The remaining issue is the policy boundary. `validate_order_risk` still returns `TRADING_HALTED`, so the flatten must either honor that result and remain blocked or special-case it before submitting outside the normal audited path. `HALTED` currently means that all submits are denied, while `REDUCING` is the state intended to allow non-increasing exposure.

I think the operator flow should use that existing distinction: enter `REDUCING`, submit the closing order through the normal risk path, then return to `HALTED`. If that transition cannot be made safely, the missing behavior should be an explicit operator-exit path through the risk engine, with the override represented in the command and its denial or submission events. The command-aware whole-position exit checks added in 8aa30f9aca are a useful precedent; an `&OrderAny` API cannot represent those command fields or venue capability.

Can the downstream flatten own the `HALTED -> REDUCING -> HALTED` transition? If not, what coordination is unavailable there? That answer will tell us whether to extend the state model or add an explicit operator-exit command.

folknor:
Thank you - taking your two questions in order.

## The flatten cannot own the transition

There is no route to it: `set_trading_state` is a `&mut self` method with no message-bus or command surface, so a component that reaches the node over the bus cannot perform even the first transition. Giving the flatten kernel access changes what it can reach, not whether the bracket is sound.

The bracket cannot be sequenced. In live mode a submit routes through the queued risk endpoint, which defers the command to a later event-loop iteration, so "enter `REDUCING`, submit, restore `HALTED`" restores the halt before the command reaches `execution_gateway`, and the order is denied `TradingHalted` anyway. The bracket must therefore stay open until the engine has processed the command - the gateway decision and the submit throttler's outcome are synchronous once it runs - but nothing on the bus acknowledges that disposition: a denial produces an event, while admission is observable only later via the execution engine.

The window is trader-wide: `trading_state` is a single field, so while the bracket is open every strategy's non-increasing submits and modifies pass. And the failure modes degrade the halt itself: a crash mid-bracket leaves the trader in `REDUCING` with nothing restoring `HALTED`, and `reset()` assigns `Active` directly without emitting `TradingStateChanged`.

Last, `REDUCING`'s predicate is the wrong one for this flow. Its gateway arm keys on `portfolio.is_net_long`/`is_net_short` - the bookkeeping an incident may have corrupted - a flat or mis-booked portfolio admits either side, and no quantity bound prevents crossing through flat. The exit needs a stricter predicate than the state it would borrow.

## The operator-exit command, mostly already built

8aa30f9aca provides most of the path: intent on the command's params, the venue allowlist, and `full_position_exit_reduces` checking side and quantity against the cached position, with the verified classification threaded through `check_order` and `check_orders_risk`.

What is missing is carrying that classification into `execution_gateway` so a verified exit forwards under `HALTED`, plus three design points: a distinct params key, since `is_full_position_exit_order` requires `!is_reduce_only()` and conditional order types; an explicit rule for `SubmitOrderList` (deny mixed lists); and the admitted/denied classification carried into the events, since `OrderDenied` does not preserve params.

That leaves one policy question:
- If `HALTED` means no new exposure, the exemption needs no caller restriction: an order that is reduce-only on a venue that enforces it, closing an identified open position, grants any caller only the ability to reduce - consistent with cancels already routing to execution under any trading state.
- If `HALTED` means no submits without privileged intervention, the same verification sits behind an authenticated operator command, and the remaining question is who holds that authority.

This supersedes `validate_order_risk`: the exit verification lives entirely in the command-only fields its contract excluded (`position_id`, execution-client capability, venue acceptance).

Happy to implement the gateway extension in either form.

cjdsellers:
Thanks @folknor, this answers the remaining question. Let's take the narrower gateway extension: `HALTED` should continue to deny ordinary submits, but allow a single order when the risk engine verifies it as an explicit emergency exit against an identified open position. That means venue-supported reduce-only semantics, the opposing side, and a quantity no greater than the open position. The order should then follow the normal sanity, max-notional, throttling, and event path. `REDUCING` retains its broader role for inferred reductions and reducing modifies, while generic reduce-only submits and order lists remain denied under `HALTED`.

This supersedes `validate_order_risk`. If you're able, could you bring the branch up to date with `develop` and replace that entry point with the narrow gateway exception, along with the corresponding documentation and focused coverage? The branch currently conflicts with `develop`, so the update and fresh CI need to pass before it can merge. We can revisit an authenticated caller boundary or a separate absolute-stop state later; neither needs to be defined in this PR 👍

folknor:
Thank you, I certainly will. Tomorrow!

folknor:
Thank you. Done as ruled: `validate_order_risk` is gone, replaced by the gateway exception, on a branch rebuilt against current develop; the description is updated to match.

The verification carries no venue allowlist: no per-venue reduce-only capability signal exists in the engine today, so the reduce-only flag is what it can check, and `RiskEngineConfig` is a pyclass, making a new config field a binding-surface change beyond this round. If you want a config gate on the exception, I can add it in a follow-up.

cjdsellers:
@folknor, thank you for the update. I've had another look at this, and two safety cases still need closing before it can merge.

The exception currently treats `order.is_reduce_only()` as proof that every downstream execution path will enforce reduce-only. That isn't true across the maintained adapters: Interactive Brokers builds an ordinary venue order without encoding that instruction. If an in-flight fill makes the cached position stale, the order can cross flat while `HALTED`. The exception needs to be limited to execution paths that guarantee reduce-only, with the normal halt denial everywhere else.

There is also a replay case for partially filled orders. The classifier uses `leaves_qty`, but the forwarded submit can use the original `quantity`; an order with 4 remaining against a position of 4 qualifies, then Binance can submit the original larger quantity. The eligibility check needs to bind to the exact quantity downstream will submit.

The normal checks, order-list denial, and focused coverage otherwise look sound.

folknor:
Thank you - both cases hold. The second has a direct answer; for the first I could not find one in the core, so the shapes are below for you to choose between.

## The submitted-quantity case

`would_reduce_only` compares `leaves_qty` against the position (`crates/model/src/orders/mod.rs:1453-1455`), while the execution client reconstructs the venue request from `order.quantity()` - Binance futures at `crates/adapters/binance/src/futures/execution.rs:509`, sent at `:592`. Your example is exactly right: quantity 10, filled 6, leaves 4, position 4 classifies and then submits 10.

The eligibility check binds to `order.quantity() <= position.quantity`, inside the emergency classifier only. `would_reduce_only` is correct for the question it answers and is left alone, so the upstream reduce-only pre-check and the full-position exit path are unchanged.

## The reduce-only enforcement case

I can't find a signal in the core by which the risk engine can establish that the selected execution path enforces reduce-only. `ExecutionClient`'s capability surface is `position_reconciliation_tolerance`, `handles_order_venue` and `provides_bulk_position_coverage` (`crates/common/src/clients/execution.rs:64-81`); none concerns order semantics. `RiskEngine` holds a clock, cache, portfolio, throttlers and config, and the clients live only in `ExecutionEngine.clients` alongside `routing_map` and `default_client_id`.

Interactive Brokers is not the only case. Betfair's `PlaceInstruction` and the Binance spot new-order request carry no reduce-only field, and Polymarket rejects the instruction outright (`crates/adapters/polymarket/src/execution/order_builder.rs:184`). Bybit, OKX, Hyperliquid, BitMEX, dYdX, Kraken futures, Deribit, Coinbase, Derive, Lighter and Architect encode it.

I see four possible paths forward.

### 1. Venue allowlist on `RiskEngineConfig`

An `emergency_exit_venues` set mirroring `full_position_exit_venues`, defaulting empty; eligibility requires the instrument venue to be listed. Smallest change and consistent with 8aa30f9aca, but it never inspects the client that receives the order. `BINANCE` names both execution clients - `crates/adapters/binance/src/factories.rs:174` and `:192` pass the same venue for the spot client, which cannot encode the instruction, and the USD-M client, which does - so one entry cannot separate them. `InteractiveBrokersExecutionClient::handles_order_venue` returns `true` for every venue (`crates/adapters/interactive_brokers/src/execution/core.rs:485`), so a listed venue can still route through it. The operator is attesting a property of the whole routing configuration, and a routing change invalidates the attestation without touching the risk config.

### 2. Execution-client capability, checked after client selection

A default-false capability on `ExecutionClient`, checked in `ExecutionEngine::handle_submit_order` beside the existing client-venue denial at `crates/execution/src/engine/mod.rs:2136-2146`. This inspects the client actually selected, including the account, venue and default fallbacks in `find_client_for_command`.

A client-wide boolean is probably too coarse - cash spot and Polymarket protect one side by balance rejection rather than by a general instruction - so `enforces_reduce_only(&self, order: &OrderAny) -> bool` may be the better signature. And it splits the authorization: the risk engine admits the submission past `HALTED`, the execution engine makes the final path-specific decision. Keying the downstream check on the `emergency_exit` parameter means a marked order is capability-checked outside `HALTED` too, since the execution engine does not see the trading state.

### 3. Resolve the route before the risk decision

The risk engine learns the resolved client and its capability, so the whole judgement stays in one component with one denial. A `ClientId -> capability` map is not sufficient - the risk engine would have to reproduce the full fallback order of `find_client_for_command`, and the route would have to be stable between the check and the submission, or the time-of-check race moves from the position quantity to the routing choice. It needs a shared routing registry, a pre-routing phase, or a route token carried through to execution.

### 4. Make reduce-only a contract every client meets

Require each `ExecutionClient` to submit through a path that guarantees reduce-only or to reject the order before submission, so no client silently drops the instruction. The risk engine could then trust `is_reduce_only()` with no allowlist and no query. Interactive Brokers and Betfair would have to reject: emulating the instruction by clamping quantity against a cached position before submitting reintroduces the race this finding is about, unlike `OrderMatchingEngine`, which clamps inside the matching loop against the position it owns (`crates/execution/src/matching_engine/mod.rs:4721`). This is a platform change across the in-tree clients and would alter behaviour for anyone whose reduce-only orders currently go out as ordinary orders, so it reads as separate work - but it is the only shape that removes the question rather than answering it.

Tell me which shape you want and I will proceed.

cjdsellers:
@folknor, thanks for laying out the options. The guarantee needs to be route-specific: the exception should stay denied unless the selected execution client and order path enforce reduce-only at the venue. A venue allowlist cannot establish that for shared venues or broker routing.

The quantity check also needs to use the quantity the client will submit rather than `leaves_qty`, with focused coverage for a partially filled order.
If you're able, go ahead with those two changes and a non-capable-route regression. The rest of the gateway checks and order-list behavior look sound.

folknor:
Thank you - both changes are in, as the route-specific shape.

## The route capability

`ExecutionClient` gains `enforces_reduce_only(&SubmitOrder, &OrderAny)`, defaulting false. Its contract is that the client will use a downstream path preventing any execution which would increase or reverse the position once it reaches flat, by native venue enforcement or by matching-engine enforcement against execution-time position state; serializing a flag the venue ignores, selecting a nominal closing side, relying on balance rejection, or checking a cached position before submission does not qualify. `ExecutionEngine::handle_submit_order` checks it beside the existing `ClientVenueMismatch` denial, so the decision sees the client `find_client_for_command` actually selected, and a route that does not qualify is denied `REDUCE_ONLY_NOT_ENFORCED`.

The risk engine keeps eligibility and the execution engine keeps the route, because the risk engine cannot see the registered client set: an explicit but unregistered `command.client_id` falls through to account, venue, then default routing, so anything it could key on would be an attestation about the whole routing graph rather than about the client that receives the order.

## The intent parameter is consumed, not forwarded

`emergency_exit` is caller-supplied, so keying the execution check on it directly would deny an emergency-marked order on an incapable route under `ACTIVE`, where today it submits. The risk engine now removes the key before the bypass return, classifies from the captured value, and re-inserts it only on the branch it admitted under `HALTED`. Order lists strip it too. Within the risk-to-execution pipeline the marker then records that the risk engine verified the order; a component publishing straight to execution can still forge it, which is the caller boundary you deferred.

## Which clients qualify

`BacktestExecutionClient` qualifies when its `SimulatedExchange` has `use_reduce_only` enabled, since that is what gates every clamp in `OrderMatchingEngine`. `BinanceFuturesExecutionClient` qualifies in one-way mode only: `reduce_only_param` returns `None` whenever `positionSide` is present, because Binance rejects the two together, so in hedge mode no path sends the instruction. This corrects what I said earlier about that client - it encodes `reduceOnly` in one-way mode across the WS, HTTP and algo paths, and in hedge mode across none of them.

Every other client stays at the default, including several that serialize a reduce-only field. That means this PR has not established their eligibility against the contract above, not that the venue cannot support it.

## The submitted quantity

Eligibility now requires `order.quantity() <= position.quantity` inside the emergency classifier. `would_reduce_only` is untouched, so the upstream reduce-only pre-check and the full-position exit path are unchanged; the shared position lookup was split into a helper both predicates call.

A partially filled order rejected by the new bound produces no `OrderDenied`, because `deny_order` returns early for an order past `Initialized`. The order cannot reach the client, and emitting a denial for a `PartiallyFilled` order would be wrong, so this is pre-existing risk-engine behaviour that the new fixture exposes rather than anything the narrowing introduces.

## Testing

`test_submit_emergency_exit_uses_submitted_quantity_bound` pins quantity 10, filled 6, leaves 4 against a position of 4, asserting `would_reduce_only` still holds so the case exercises the new bound, and a boundary case where submitted quantity equals the position. `test_submit_emergency_exit_requires_capable_route` asserts the denial reason and that nothing reaches the client on an incapable route, and submission on a capable one. Two provenance tests pin the key absent when forwarded under `ACTIVE`, on bypass, and for order lists, and present on the admitted `HALTED` branch. The capability predicates have their own matrices, including Binance hedge mode and a disabled backtest exchange.

Reverting the quantity bound makes the partial-fill case forward, and disabling the route check makes the incapable case submit.

cjdsellers:
@folknor, thanks for the latest update. The route-specific approach looks right, but one live path still needs wiring before this can merge. The `ExecutionClient` implementation in `crates/live/src/execution/client.rs` does not delegate `enforces_reduce_only` to the wrapped client, so it uses the default `false`. This means a capable `BinanceFuturesExecutionClient` is still denied with `REDUCE_ONLY_NOT_ENFORCED` in a live node.

Could you forward that capability through `LiveExecutionClient` and add focused coverage for a capable wrapped client? That looks like the remaining code blocker.

folknor:
Thank you - confirmed and fixed.

## The facade

`LiveExecutionClient` (`crates/live/src/execution/client.rs`) forwards every other capability method to the wrapped client - `is_connected`, `get_account`, `position_reconciliation_tolerance`, `handles_order_venue`, `provides_bulk_position_coverage` - and caches only the four identity values it reads in `new()`. `enforces_reduce_only` was the one trait method with a non-trivial default left unforwarded, so a capable `BinanceFuturesExecutionClient` answered `false` in a live node. It now delegates under the same short-lived `borrow()` as its neighbours.

The facade can attest on the wrapped client's behalf because it alters neither the command, the order, the route nor the adapter state: the same client answers the predicate and receives `submit_order`, and the borrow is released before the engine's later call. `LiveExecutionClient` is also the only wrapping `ExecutionClient` in the tree, so there is no second facade with the same omission.

## Testing

`test_enforces_reduce_only_forwards_to_wrapped_client` wraps `StubExecutionClient` in the facade and queries the capability through trait dispatch, in two cases. The capable case fails if the forwarding is removed, since the trait default returns `false`; the incapable case fails an unconditional `true`. Removing the method makes the capable case fail on `left: false, right: true`.

The engine-level pairing already exists in `test_submit_emergency_exit_requires_capable_route`, which asserts admission on a capable route and `REDUCE_ONLY_NOT_ENFORCED` on an incapable one; the missing seam was the delegation, and that is what the new test pins.

The branch is rebased onto current `develop`, with the rest of the diff unchanged.

cjdsellers:
@folknor, one external-client path still bypasses the route capability check. When an explicit `client_id` is configured in `external_clients`, `ExecutionEngine::execute_command` publishes the command and returns before `handle_submit_order`, so `enforces_reduce_only` is never checked.

This allows a verified emergency exit through `HALTED` without establishing that the external execution process enforces reduce-only. Could this path fail closed unless the external route has an explicit capability?

folknor:
Thank you - confirmed. The external route is now fail-closed.

## The bypass

`ExecutionEngine::execute_command` publishes to `commands.trading.<client_id>` and returns for any command whose `client_id` is in `external_clients`, before client selection and before `handle_submit_order`, so the `enforces_reduce_only` check never ran on that path. There is also nothing to run it against: an external route has no in-process `ExecutionClient`, so no object can answer the predicate.

A marked `SubmitOrder` routed to an external client is now denied with a new `REDUCE_ONLY_ENFORCEMENT_NOT_ESTABLISHED` rather than published. The new reason is distinct from `REDUCE_ONLY_NOT_ENFORCED` because it reports something weaker: not that a selected client answered no, but that enforcement was never established. Every other external command keeps its exact current path, and the guard inspects only `SubmitOrder` - `SubmitOrderList` never carries the marker, since `RiskEngine::handle_submit_order_list` consumes it unconditionally.

The denial resolves the order from the cache, falling back to `add_order_from_init` when it is absent, so a command whose order has not yet been cached is denied rather than silently dropped. Only a reconstruction failure suppresses without an `OrderDenied`, and that path logs an error naming the client order ID, the external client and the suppressed command.

## What this does not do

The patch answers "fail closed" but not "unless the external route has an explicit capability", because no channel exists for an external route to declare one. `external_clients` states where a command goes, not what the receiver enforces; the order's `reduce_only` flag is a request rather than an enforcement guarantee; and the `emergency_exit` parameter is risk-engine provenance. A configured allowlist of external client IDs that enforce reduce-only would close that, and it is a small change - but its shape is yours to specify: whether capability is static per client, conditional per order, or negotiated with the external process, and whether such an ID must also be a member of `external_clients`. Denial is the interim contract that closes the hole without committing the project to an API.

I am happy to implement the declaration in a follow-up once you say which shape you want.

## Testing

`test_external_client_submit_order_denied_only_when_emergency_exit` subscribes to the external topic and runs both cases: the marked command publishes nothing and leaves the order `Denied` with the new reason, while an unmarked one is published unchanged and stays `Initialized`. `test_external_client_emergency_exit_cache_miss_is_denied` starts with an empty cache and pins the reconstruction path.

Both assert the suppression and the denial, so removing the guard fails them on behaviour rather than on a log line: the command reaches the topic and the order is never denied.

cjdsellers:
@folknor, Thanks for closing the external-client path. I've had another look at the latest changes, and the safety cases now hold up. This looks good from my side. Failing closed for external execution is the right boundary here; an explicit capability declaration for external clients can stay separate :+1:.

**I found two small follow-ups**:

`SandboxExecutionClient` already routes through `OrderMatchingEngine` with `use_reduce_only`, but it inherits the default-false `enforces_reduce_only`, so emergency exits are always denied in sandbox trading. The execution docs also don't name the currently capable paths, Binance Futures in one-way mode and backtest with `use_reduce_only`, or clarify that `a single` means an individual submit rather than one exit allowed per halt. If/when you get a chance, could you follow those up so sandbox can exercise this path and the supported configuration is clear?

Thanks for sticking with the review process.

folknor:
Thank you - both follow-ups are in, and the branch is rebased onto current `develop`.

## Sandbox

`SandboxExecutionClient::enforces_reduce_only` answers `order.is_reduce_only() && config.use_reduce_only`. The predicate keys on the client config rather than the instrument's matching engine because engines are created on demand in `submit_order`, so the one for the order's instrument may not exist when the execution engine asks; every engine is built from that same config, and `OrderMatchingEngine` clamps reduce-only fills against the position it holds at fill time on both the market and limit paths, which is what the trait contract requires.

## Docs

The `HALTED` bullet in `docs/concepts/execution.md` now says each individual `SubmitOrder` is evaluated on its own with no per-halt limit, states the risk-engine conditions, and names the qualifying paths: Binance Futures in one-way position mode for a mapped order type without `close_position`, and backtest or sandbox with `use_reduce_only` enabled. Every other client, including external clients, is denied.

## Testing

`test_enforces_reduce_only_follows_config` in the sandbox crate is the same three-case matrix as the backtest client: enabled, disabled, and a non-reduce-only order. With the override removed the enabled case fails on `left: false, right: true`.

cjdsellers:
@folknor, thanks for those follow-ups. I've had another look at the latest changes, and this looks good to merge. The sandbox capability is correctly gated by `use_reduce_only`, the docs now make the supported execution paths clear, and the earlier quantity and route-enforcement safety cases still hold after the rebase 🙏

cjdsellers (approved):

cjdsellers:
Hey @folknor,

I’ve pushed the follow-up in [9da48e0399](https://github.com/nautechsystems/nautilus_trader/commit/9da48e0399dec9b7d60b1cd2fb67a973c260120f).

After applying the same reduce-only guarantee across the adapters, I concluded that the contract we agreed during review put this behavior on the wrong trading state. Your implementation carefully implemented that contract, including the fail-closed capability checks. Seeing the contract applied across the full adapter surface exposed the underlying problem: allowing an `emergency_exit` exception under `HALTED` turned it into a conditional trading mode and duplicated the purpose of `REDUCING`.

**The simpler model follows the original intent of the states:**

- `ACTIVE = 1`: normal trading. Submissions, modifications, cancellations, and queries operate normally.
- `REDUCING = 2`: controlled de-risking. Eligible individual orders with `reduce_only=true`, cancellations, and queries are allowed. Order lists and modifications are denied.
- `HALTED = 3`: trading is stopped. No new or modified orders are sent. Only cancellations and queries remain available.

This gives us one progressive axis of restriction: active, reducing, then halted. The numeric values now follow that progression, which is documented as a breaking ordinal change.

Under this model, a position exit during `REDUCING` is an ordinary Nautilus order with `reduce_only=true`. The risk engine verifies it against the cached open position. Each adapter must then either send a venue instruction that enforces reduce-only behavior or deny the order before transport. An equivalent instruction also satisfies the contract: for example, Binance receives `closePosition=true` without its incompatible `reduceOnly` field, while the Nautilus order still requires `reduce_only=true` to pass the risk engine. Any remaining venue-state failure, such as the position already being closed, is left for the venue to reject.

This removes the cross-layer machinery needed by the earlier design:

- The caller-facing `emergency_exit` parameter and its provenance handling.
- The `ExecutionClient::enforces_reduce_only` capability and its forwarding through live clients and stubs.
- The execution engine’s need to receive or reason about `TradingState`.
- The special `HALTED` exception and external-client path.
- The two enforcement-specific denial codes.

`TradingState` is now solely a risk-engine concern. If the risk engine is bypassed, trading-state enforcement is bypassed as well.

Measured across the same 16 file paths changed by this PR, the follow-up removes 1,471 lines and adds 453, for roughly 1,000 net fewer lines. That removes about 70% of this PR’s net 1,447-line expansion. The complete follow-up is still net +286 lines because it also applies the send-or-reject contract across the adapters and adds the corresponding tests and documentation.

Your implementation correctly met the contract we agreed. Applying that contract more broadly is what changed my view of the design. Thanks again for working through all the safety cases with me.

folknor:
That's the nature of development, sometimes you need to iterate and let things bake!

PR #4763, opened 2026-08-15
Defer backtest order events during immediate dispatch

folknor:
A `SimulatedVenueConfig` with `use_message_queue = false` aborts the backtest with a `BorrowMutError` on the first order it is given.

## The re-entrancy

`ExecutionEngine::register_msgbus_handlers` registers `exec_engine_execute` as `rc.borrow().execute(cmd)` and `exec_engine_process` as `rc.borrow_mut().process(&event)` against the same `Rc<RefCell<ExecutionEngine>>`. A submitted order therefore holds a shared borrow for the whole of `execute`, which reaches `BacktestExecutionClient::submit_order` and `exchange.send(...)`.

With `use_message_queue = false`, `send` dispatches synchronously into `process_trading_command` instead of buffering the command. The matching engine emits `OrderAccepted` or `OrderFilled` through `dispatch_order_event`, which with no handler installed publishes to `exec_engine_process` - taking a mutable borrow while the shared one is still live.

`BacktestExecutionClient` already defends against this for the one event it generates itself: `submit_order` pushes `OrderSubmitted` into a `queued_events` buffer drained later by the engine loop, with a comment naming the re-entrancy. The events the exchange generates had no equivalent.

The exchange now takes an optional order event handler that feeds that same buffer, installed by `BacktestEngine::add_venue` only when `use_message_queue` is false. Both dispatch sites use it: the matching engines, and the exchange's own `dispatch_order_event`, which generates `OrderUpdated` when modifying an order still in `Submitted` and previously published unconditionally. That second site is the one a matching-engine-only fix misses; with it left alone, an immediate `ModifyOrder` still panics while a submit test passes.

## Deferral is scoped to the borrow window

A flag is set across `process_trading_command` and cleared by a guard on drop, and the handler defers only while it is set. Events raised by market data, matching engine iteration, expiration, liquidation and open order loading do not run beneath the execution engine borrow, so they continue straight to `exec_engine_process` and keep their synchronous timing.

Left permanently installed, the handler would buffer those too, and a fill driven by a quote would reach the strategy after the data rather than before it. Since submitting an order in this mode panics today, that ordering is not history being preserved but semantics being chosen, and an immediate dispatch venue whose fills arrive at drain time is not immediate in the sense the option offers.

The two queued modes install no handler and are unchanged: the command is buffered, the matching engine runs after the borrow has ended, and both dispatch sites take their existing message bus path.

## Reachability

`use_message_queue` is a public `SimulatedVenueConfig` field with a builder setter, defaulted `true`, and passed through `BacktestEngine.add_venue` in Python. No backtest, example or test in the tree sets it false; the existing immediate-dispatch tests cover query commands being ignored and an instrument ID overflow, none of which routes an order through the execution engine. So this is a documented option that fails deterministically for anyone who selects it, rather than a defect in a configuration under test.

## Testing

Three tests in `crates/backtest/src/engine.rs`, all driving commands through the registered `exec_engine_execute` endpoint - calling the client directly does not reproduce the defect, because it never establishes the outstanding engine borrow.

`test_immediate_submit_defers_order_events` asserts the order is still `Initialized` with one event before draining, then `Submitted` ahead of `Accepted` after. `test_immediate_modify_submitted_order_defers_updated_event` covers the exchange-local path, asserting the quantity is unchanged and `Submitted` still last before draining, then `2.000` and a final `Updated`. `test_immediate_market_data_dispatches_fill_synchronously` covers the other half of the contract: a resting order filled by a quote is applied to the cache immediately, with no drain.

Each assertion was checked against a mutated implementation. Installing no handler makes both command tests panic with `RefCell already borrowed`. Reverting the exchange's dispatcher to publish unconditionally, with the matching engine handler still in place, fails only the modify test. Making the handler defer unconditionally fails only the market data test.

## A shared deferred event path

The message bus already protects commands against this class: `exec_engine_queue_execute` defers when a command is dispatching. Extending the same treatment to order events would remove the class rather than contain it per adapter, and would let the sandbox adapter and this one drop their separate handlers. It moves ordering and callback timing for every execution client, so it wants its own change with coverage across live, sandbox and backtest.

I am glad to take that on in whatever shape you prefer, or to leave it with you.

### DIFF
--- a/crates/backtest/src/engine.rs
+++ b/crates/backtest/src/engine.rs
@@ -282,6 +282,7 @@ impl BacktestEngine {
 
         let routing = Some(config.routing);
         let frozen_account = Some(config.frozen_account);
+        let use_message_queue = config.use_message_queue;
 
         let exchange =
             SimulatedExchange::new(config, self.kernel.cache.clone(), self.kernel.clock.clone())?;
@@ -298,6 +299,12 @@ impl BacktestEngine {
             routing,
             frozen_account,
         );
+
+        if !use_message_queue {
+            exchange
+                .borrow_mut()
+                .set_event_handler(exec_client.order_event_handler());
+        }
 
         exchange
             .borrow_mut()
@@ -2156,7 +2163,7 @@ mod tests {
         enums::Environment,
         messages::{
             data::{DataCommand, UnsubscribeCommand},
-            execution::{SubmitOrder, TradingCommand},
+            execution::{ModifyOrder, SubmitOrder, TradingCommand},
         },
         msgbus::{
             self, MessagingSwitchboard, TypedHandler,
@@ -2170,11 +2177,12 @@ mod tests {
             AccountType, BookType, MarketStatus, MarketStatusAction, OmsType, OrderSide,
             OrderStatus, OrderType, TriggerType,
         },
-        identifiers::{AccountId, ActorId, ClientId, PositionId, StrategyId, Venue},
+        events::OrderEventAny,
+        identifiers::{AccountId, ActorId, ClientId, ClientOrderId, PositionId, StrategyId, Venue},
         instruments::{
             CryptoPerpetual, Instrument, InstrumentAny, stubs::crypto_perpetual_ethusdt,
         },
-        orders::{Order, OrderAny, OrderTestBuilder},
+        orders::{Order, OrderAny, OrderTestBuilder, stubs::TestOrderEventStubs},
         types::{Money, Price, Quantity},
     };
     use nautilus_system::{KernelEventStore, RegisteredComponents};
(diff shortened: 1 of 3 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. The write-up made this straightforward to review: the re-entrancy analysis and catching the exchange's own `dispatch_order_event` site saved me a lot of digging, and the mutation work behind the tests shows.

I verified the deferral is scoped tightly: the guard exists only across the immediate branch of `send`, `process_trading_command` never re-enters `send`, and matching engines added after `add_venue` pick up the handler. This looks good to merge. I'll tidy the doc comments on the two private items at merge, so no further changes are needed.

I'll investigate the shared deferred event path in the message bus separately and ticket it, so nothing more is needed from you here.

PR #4764, opened 2026-08-15
Cache Binance Futures instruments from execution

### LINKED ISSUE #4734: [Binance] Futures execution client fetches instruments on connect but never caches them
As of `develop` @ `deff407e4`, the USD-M futures execution client's connect
path (`crates/adapters/binance/src/futures/execution.rs`, ~line 1355) requests
the instrument catalogue and discards it:

```rust
let _instruments = if self.core.instruments_initialized() {
    Vec::new()
} else {
    let instruments = self.http_client.request_instruments_with_config(…).await?;
    …
    self.core.set_instruments_initialized();
    instruments
};
```

The binding is `_`-prefixed and never read again, and nothing on this path
calls `cache_instruments` with the result. The **spot** execution client does
cache what it fetches (`spot/execution.rs:729` and `:1055`) — the
futures/spot asymmetry looks unintentional.

Consequence: with an execution client configured and no data client, the
shared cache contains no BINANCE instruments, and downstream consumers that
resolve instruments from the cache come up empty. See the companion issue
#4735 for how this surfaces during reconciliation (positions silently
dropped, reconciliation reported successful).

AtomicGlance:
## Summary

Binance Futures execution currently fetches instruments during connect, startup, and refresh, but does not publish the parsed definitions to Nautilus' live data route. In an execution-only setup, the shared cache can therefore remain empty and position or order reconciliation cannot resolve Binance symbols.

This change:

- publishes successfully loaded instruments as `DataEvent::Instrument` events;
- captures the sender before an async boundary so Tokio task migration cannot lose it;
- keeps the execution refresh path aligned with the initial load; and
- adds a focused regression test for the event publication helper.

Closes #4734

## Validation

- `git diff --check`
- GitHub Actions Rust formatting and tests

### DIFF
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -31,7 +31,12 @@ use dashmap::DashMap;
 use nautilus_common::{
     cache::fifo::FifoCache,
     clients::ExecutionClient,
-    live::{get_runtime, runner::get_exec_event_sender, task::TaskHandles},
+    live::{
+        get_runtime,
+        runner::{get_exec_event_sender, try_get_data_event_sender},
+        task::TaskHandles,
+    },
+    messages::DataEvent,
     messages::execution::{
         BatchCancelOrders, CancelAllOrders, CancelOrder, GenerateFillReports,
         GenerateFillReportsBuilder, GenerateOrderStatusReport, GenerateOrderStatusReports,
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @AtomicGlance,

Thanks for the PR, and for digging into this one.

I'm going to close it as it needs further design first. Publishing instrument definitions from an execution client isn't the direction we want to take: that route belongs to the data client, and routing execution through it makes the fix depend on when data events get drained during startup, which isn't the same on every startup path.

The underlying problem is that the Binance Futures execution client resolves instruments from the shared cache at all. Spot, BitMEX, and Bybit each keep their own parsed instrument cache from connect and resolve reports against that, and futures should do the same. I'll raise a dedicated bug ticket covering that, along with the silent drop in `generate_position_status_reports` and the `(8, 8)` precision fallback in `get_instrument_precision`, so it lands as one coherent change.

The investigation here was solid, it's just the wrong layer for the fix.

PR #4765, opened 2026-08-15
Fix Bybit open-only reports missing closed orders

### LINKED ISSUE #4730: Bybit: open-only consistency check can never resolve a lost cancel ack — realtime query omits openOnly, so terminal reports never arrive
## Confirmation

- [x] I've reviewed the source code (Python adapter and `crates/adapters/bybit/src/http/client.rs`) to confirm the behavior is not by design.
- [x] I've confirmed the code path is unchanged at v1.231.0 and on `develop`.

## Expected behavior

`LiveExecutionEngine._check_orders_consistency` in `open_check_open_only=True` mode applies real reports and deliberately refuses to act on absence — its own comment anticipates that the venue "may include recent closed orders with open query", which is what lets a cache-open order that the venue has actually closed heal from a **positive** terminal report (`_reconcile_order_reports`: `report.is_open != cache_open` → reconcile).

For that design to function on Bybit, the open-only report query needs to include recently closed orders.

## Actual behavior

For Bybit it never does. `request_order_status_reports(open_only=true)` (Rust, `crates/adapters/bybit/src/http/client.rs`, the `if open_only` branch) builds `BybitOpenOrdersParams` **without setting `open_only`**, and `/v5/order/realtime` with no `openOnly` param returns genuinely open orders only. Verified against the live venue (linear, UTA):

| query | result |
| --- | --- |
| `/v5/order/realtime` (no `openOnly` — what the client sends) | open orders only (`New`) |
| `openOnly=1` (`BybitOpenOnly::ClosedRecent` — exists in the codebase, unused here) | recent closed orders: `Cancelled`/`Filled`, observed spanning ~30h |
| `/v5/order/history` with `startTime` (what the `open_only=false` branch queries) | same terminal reports, bounded |

Consequence: an order stuck in `PENDING_CANCEL` because the cancel's ack was lost (transient WebSocket/API disruption — the cancel *succeeded* venue-side) produces **no report**, the open_only mode correctly refuses to resolve on absence, and the order is stuck until process restart. The venue holds the real `Cancelled` report the whole time — the check just never asks a question that returns it.

Production observation (live Bybit linear market maker, `inflight_check_interval_ms=0` due to #4655-class hazards): three such lost-ack episodes in four days, the worst leaving both quote sides held for 3.5 hours while `/v5/order/realtime?openOnly=1` had the `Cancelled` report available within one minute of the wedge starting.

## Suggested fix

In the Bybit client's `open_only=true` path, additionally query `ClosedRecent` (or reuse the bounded history leg the `open_only=false` branch already implements, honoring the command's `start`) and merge — so the engine's open_only mode receives positive terminal reports while its refuse-on-absence semantics stay untouched. We are running a vendored one-argument patch to this effect (routing the consistency check through the `open_only=false` query path) and the engine machinery downstream (`_reconcile_order_reports` + the `open_check_threshold_ms` recent-activity race guard) handles the widened report set as designed.

## Steps to reproduce

1. Live Bybit node, `open_check_interval_secs` set, `open_check_open_only=True` (default).
2. Get an order into `PENDING_CANCEL` whose venue-side cancel succeeded but whose ack never arrived (e.g. drop the private WS during a cancel burst).
3. Observe each consistency cycle: reports contain only open orders, the stuck order appears in the "not in venue's current response" DEBUG list, and the check returns without resolving — indefinitely.

## Specifications

- OS platform: Linux
- Python version: 3.12
- `nautilus_trader` version: 1.227.0 (verified unchanged at v1.231.0 and `develop`)
- Installed from: PyPI wheel
- Adapter/venue: Bybit (linear perpetuals, UTA)

NoiceHax:
# Pull Request

  and, if I used AI,
- [ ] This change is complete, locally validated, and ready for review, or a maintainer requested
  this draft
- [ ] I ran all relevant tests locally, no tests apply to this change, or I described an agreed
  limitation below

## Summary

The `open_only` branch of `request_order_status_reports` built `BybitOpenOrdersParams` without setting `open_only`, and `/v5/order/realtime` with no `openOnly` parameter returns genuinely open orders only. An order the venue had already cancelled, rejected or filled therefore produced no report at all. The live open-order consistency check resolves a stale cache entry from a positive terminal report and on purpose does not act on absence, so an order left in `PENDING_CANCEL` by a lost cancel ack could never heal, even though the venue held the `Cancelled` report the whole time.

The branch now makes a second pass with `openOnly=1` (`BybitOpenOnly::ClosedRecent`) for each order filter. The `seen_ids` set already in the branch drops orders that come back in both passes, and the `remaining_limit` bookkeeping still caps the total. Nothing in the engine changes. The Bybit execution client passes no limit and post-filters reports on `start`, so the widened set stays bounded by `open_check_lookback_mins`.

One judgement call worth flagging. This doubles the realtime calls per open check, from 2 to 4 for a single settle coin and from 4 to 8 for LINEAR without a symbol. The reporter is running a vendored patch that instead routes the consistency check through the `open_only=false` path, which already queries `/v5/order/history` with a bounded `start` and adds no realtime calls. I picked the in-branch `ClosedRecent` pass because it is the smaller change, stays inside the adapter, and leaves the engine alone, but the other shape is defensible and I am happy to switch if you prefer it.

## Related issues/PRs

Closes #4730

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not applicable.

## Documentation

The only doc change is three lines of rustdoc on `BybitHttpClient::request_order_status_reports`. The PyO3 wrapper `py_request_order_status_reports` and its stub are untouched, so no stub regeneration applies.

## Testing

`test_request_order_status_reports_open_only_includes_recently_closed` adds a mock router whose `/v5/order/realtime` handler serves a `Cancelled` order only when `openOnly=1` is present, matching the venue. It checks the terminal report comes back, that the open order returned by both passes appears exactly once, and that four realtime calls were made. Two existing tests had their hard-coded realtime query counts moved from 4 to 8 for the extra pass. Their report counts are unchanged because the shared handler returns the same orders for both passes and they are deduplicated.

Limitation, stated up front: I could not run anything locally. The machine I prepared this on has no Rust toolchain and no C toolchain, so `make format`, `make pre-commit`, and `cargo nextest run -p nautilus-bybit --test http` were all skipped. I reviewed the change by hand against the sibling `open_only=false` branch, which uses the same borrow pattern over `order_filters`, and against `get_open_orders`, which already calls `open_only` on this builder, but I have not seen it compile or the new test go green. Please treat CI as the first real signal here, and tell me if you would rather I hold this until I can validate it locally.

I used AI assistance while preparing this change.

### DIFF
--- a/crates/adapters/bybit/src/http/client.rs
+++ b/crates/adapters/bybit/src/http/client.rs
@@ -4223,6 +4223,10 @@ impl BybitHttpClient {
     ///
     /// Orders for instruments not currently loaded in cache will be skipped.
     ///
+    /// When `open_only` is true the realtime endpoint is queried twice per order filter,
+    /// once for currently open orders and once for recently closed orders, so terminal
+    /// reports are included.
+    ///
     /// # Errors
     ///
     /// Returns an error if:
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @NoiceHax,

Thanks for the PR, and for the write-up. The diagnosis matches what the code does, and keeping this inside the adapter is the right call.

## Bounding the closed pass

The `ClosedRecent` pass follows `nextPageCursor` until it comes back empty, and the execution client always passes `limit: None`, so nothing caps it. On an active account that's tens of extra realtime calls per open check, against a single global 10 req/s quota shared by all Bybit REST.

Those pages get discarded anyway, since `generate_order_status_reports` re-filters on `cmd.start`. Capping the closed pass at a single page keeps the healing behavior at the cost you intended.

**Two smaller follow-ups:**

- The doc comment should mention that `start` and `end` don't constrain the closed pass. It didn't matter when the branch only returned open orders, but the method is public and exposed through PyO3.
- The test handler comment says the venue echoes the still-open order back under `openOnly=1`. It returns closed records only, which is what the table in #4730 shows too. Keep the dedup coverage, just reword it.

## Before the next push

A green `pre-commit` is the minimum bar for review here. It's currently failing on `cargo fmt` over one hunk in `crates/adapters/bybit/tests/http.rs`, which gates the Rust test job, so there's no test signal on this yet.

Run `make format`, then `make pre-commit`, then `cargo nextest run -p nautilus-bybit --test http` for this change. That does mean getting a Rust toolchain up on your side, which is worth doing before the next push.

NoiceHax:
Addressed all the points.

The closed pass now stops after the first page per order filter, so the realtime calls stay bounded instead of following nextPageCursor to the end. The doc comment on request_order_status_reports now notes that start and end don't constrain the closed pass. The test handler comment is reworded: the venue returns closed records only, and the open order is kept in the response to keep the dedup coverage.

The cargo fmt hunk in tests/http.rs is fixed as well.

I still don't have a Rust toolchain on this machine, so I couldn't run make pre-commit or the http tests locally. Treating CI as the first real signal again.

cjdsellers (approved):
Hi @NoiceHax,

Thanks for addressing all three points.

This looks good from my side. I ran the checks from my earlier message against your latest push locally and they all pass. Once CI confirms, I'll merge this.

cjdsellers:
@NoiceHax, the latest push is green apart from the `Check generated file drift` step. The new doc comment on `request_order_status_reports` gets synced into the generated PyO3 wrapper and stubs, so the branch just needs a regeneration pass. Run `make py-stubs` from the repository root and commit the result; in CI it adds the five-line paragraph to `crates/adapters/bybit/src/python/http.rs`.

This should be the last thing needed for this PR to be merged.

cjdsellers:
Hi @NoiceHax,

I completed and landed the PR. The `ClosedRecent` second pass is as you wrote it, including the one-page cap. I regenerated the PyO3 rustdoc so the new comment syncs into the wrapper, tightened that comment a little, and had the mock return a closed-page cursor so the request-count assertion covers the one-page stop.

I also ran live mainnet smoke on the open-only reporter and the 60-minute open-check lookback. Recently cancelled orders came back as terminal reports, and the ones inside the lookback window are what the consistency check would use to heal a stale `PENDING_CANCEL`.

I credited you in the release notes, thanks again!

PR #4766, opened 2026-08-15
Fix panic on contract expiration for restored positions

### LINKED ISSUE #4721: Instrument expiration fails with `KeyError: TraderId` when positions are restored from a cache database
**Version:** 1.231.0 · **Environment:** sandbox/`SimulatedExchange`, Redis-backed
`CacheConfig(database=DatabaseConfig(type="redis"))`

### Summary

When a position is restored from a Redis-backed cache after a process restart, and its instrument
subsequently expires, `OrderMatchingEngine.check_instrument_expiration` raises an unhandled
`KeyError: TraderId(...)`. The position is never closed. Because `_expiration_processed` is set
**before** the work, every subsequent attempt returns early and appears to succeed, so the failure
is silent after the first occurrence.

### Mechanism (`backtest/engine.pyx`, 1.231.0)

1. `check_instrument_expiration` sets `self._expiration_processed = True` (`:5940`) before doing
   any work.
2. It constructs an `EXPIRATION-LEG` `MarketOrder` and calls `self.cache.add_order(...)` (`:5963`),
   which succeeds.
3. When `self._settlement_prices` contains the instrument, it calls `_generate_order_accepted`,
   which evaluates `account_id=order.account_id or self._account_ids[order.trader_id]` (`:8377`).
4. `self._account_ids` (`:4055`) is populated **only** in `process_order` (`:5055`), and there is
   one `OrderMatchingEngine` per instrument.
5. A freshly constructed `MarketOrder` has `account_id is None`, so the dict is indexed → `KeyError`.

### Why a cache database is the trigger

With an in-memory cache, a position can only exist if an order was processed by that matching
engine in the same process, so `_account_ids` is always populated. With a cache database, positions
are restored across a restart **without** their orders being re-processed, so an instrument that is
carried but not re-traded that session has an empty `_account_ids` when it expires.

### Reproduction

1. Run a sandbox node with a Redis-backed cache; open a position on an instrument with an
   expiration.
2. Restart the process so the position is restored from Redis.
3. Do not submit any order on that instrument.
4. Publish an `InstrumentClose` for that instrument, with `settlement_prices` set.

Expected: the position closes at the settlement price. Actual: `KeyError: TraderId(...)`, the
position stays open, and every retry returns early because the latch is already set.

### Suggested fix

Two independent hardenings:

* fall back rather than index — `self._account_ids.get(order.trader_id)` instead of `[...]`,
  so a missing entry is a handled case rather than an exception;
* set `_expiration_processed = True` only **after** the expiration work completes, so a failure is
  retryable instead of latching into a silent success.

Happy to open a PR if the maintainers prefer a particular shape.

NoiceHax:
# Pull Request

  and, if I used AI,
  this draft
- [ ] I ran all relevant tests locally, no tests apply to this change, or I described an agreed
  limitation below

## Summary

The issue points at v1 line numbers, but the same problem is in the v2 Rust matching engine. In `check_instrument_expiration` the synthetic EXPIRATION close order is built with `MarketOrder::new`, so it has no account ID of its own (mod.rs:2420-2485). `create_order_accepted` then falls back to `self.account_ids.get(&order.trader_id()).unwrap()` (mod.rs:6022), and `generate_order_filled` does the same. That index is only written by `process_order` and `liquidate_open_positions`, so a position loaded from a cache database after a restart has no entry for its trader and the unwrap panics.

`liquidate_open_positions` already gets this right. It collects `pos.account_id` (mod.rs:2561) and inserts it into `account_ids` before emitting any event (mod.rs:2630). This change applies the same pattern to the expiration path. Open positions are collected with their account ID, and that ID is put into `account_ids` before the accepted event is generated, which covers the later fill event too. Behaviour is unchanged for positions the engine has already seen, since the value written is the one `process_order` would have written.

The option expiry branch was already safe, so it is untouched. As a note for later, the three `.unwrap()` calls on `account_ids` at mod.rs:5982, 6022 and 6200 are all the same shape, and mod.rs:4963 shows the safe form. Hardening those is a separate change and is not in this PR.

## Related issues/PRs

Closes #4721

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not applicable.

## Documentation

## Testing

Added `test_check_instrument_expiration_with_restored_position_uses_position_account_id` in `crates/execution/tests/matching_engine.rs`. It builds the engine for an expiring futures contract, puts an open position straight into the cache without ever calling `process_order`, sets a settlement price, then drives the clock past expiry. It checks that the EXPIRATION accepted and filled events both carry the account ID from the position. The test panics on `develop` and passes with this change.

The existing `test_check_instrument_expiration_*` tests all submit an order through `process_order` first, which populates `account_ids` and hides the problem.

Run it with:

```
cargo nextest run -p nautilus-execution --test matching_engine --features high-precision,python -E 'test(check_instrument_expiration)'
```

### DIFF
--- a/crates/execution/src/matching_engine/mod.rs
+++ b/crates/execution/src/matching_engine/mod.rs
@@ -2417,7 +2417,14 @@ impl OrderMatchingEngine {
         self.cancel_open_orders_for_expiration();
 
         let instrument_id = self.instrument.id();
-        let positions: Vec<(TraderId, StrategyId, PositionId, OrderSide, Quantity)> = {
+        let positions: Vec<(
+            TraderId,
+            StrategyId,
+            AccountId,
+            PositionId,
+            OrderSide,
+            Quantity,
+        )> = {
             let cache = self.cache.borrow();
             cache
                 .positions_open(None, Some(&instrument_id), None, None, None)
@@ -2431,6 +2438,7 @@ impl OrderMatchingEngine {
                     (
                         pos.trader_id,
                         pos.strategy_id,
+                        pos.account_id,
                         pos.id,
                         closing_side,
                         pos.quantity,
@@ -2442,7 +2450,7 @@ impl OrderMatchingEngine {
         let ts_now = self.clock.borrow().timestamp_ns();
         let close_price_fallback = close.as_ref().map(|c| c.close_price);
 
-        for (trader_id, strategy_id, position_id, closing_side, quantity) in positions {
+        for (trader_id, strategy_id, account_id, position_id, closing_side, quantity) in positions {
             let client_order_id =
                 ClientOrderId::from(format!("EXPIRATION-{}-{}", self.venue, UUID4::new()).as_str());
             let mut order = OrderAny::Market(MarketOrder::new(
@@ -2482,6 +2490,10 @@ impl OrderMatchingEngine {
             }
 
             let venue_order_id = self.ids_generator.get_venue_order_id(&order).unwrap();
+            // The position may have been restored from a cache database without any
+            // order for this trader passing through the engine, so index the account
+            // ID from the position before emitting events.
+            self.account_ids.insert(trader_id, account_id);
             self.generate_order_accepted(&order, venue_order_id);
 
             let fill_price = self.settlement_price.or(close_price_fallback);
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @NoiceHax,

Thanks for the PR. The fix landed in 39f6118704 shortly before you opened this, closing #4721 with effectively the same change, so I'll close this one out.

PR #4767, opened 2026-08-15
Commit cached position state before persisting it

folknor:
`Cache::update_position` commits one representation of a position and rejects the other when the database write fails.

## The split transition

The method does three things in order: the open/closed index transition, the database write, then the replacement of the canonical position cell. The database call propagates with `?`, so an error returns with the index already moved and the value never written.

A failed close leaves the cache disagreeing with itself, and both halves are publicly readable: `is_position_closed(&id)` returns `true` while `position(&id)` returns a position whose `is_closed()` is `false`, and `positions_closed(...)` hands back that open position. A failed open-to-open update is quieter and worse - the index transition is a no-op, so nothing appears wrong, and the new quantity is discarded while the caller receives an error the execution engine only logs.

Moving the cell write ahead of the database call leaves the canonical value and its indexes describing one completed transition. The error is returned unchanged, so callers see exactly what they saw before.

This follows what the order path already does. `update_order` commits the canonical event transition first and logs a database failure afterwards, with a regression test requiring the applied event to survive persistence failure. A database error is a durability failure; it is not evidence that the fill or the close did not happen. Rolling the cache back instead would make memory and storage agree while both were knowingly wrong about the position, and `update_position` receives an already-applied `Position` from a caller that has not rolled anything back, so a cache-only rollback would move the inconsistency outward rather than remove it.

## The compensation that depended on the old timing

`SimulatedExchange` funding settlement adjusts every open position, and on a failed `update_position` restores the ones it already changed. It restored `adjusted_positions[..index]` - everything before the failure - which was right while a failed update committed nothing.

With the cell written first, the position whose update failed is committed too, and the exclusive bound leaves it adjusted. At index zero the rollback is empty, so a single-position venue keeps the adjustment outright, while settlement returns false, the account adjustment never runs, and no funding settlement or position event is published. The bound is now inclusive.

## Testing

`test_update_position_commits_canonical_state_when_database_update_fails` fails the database write on a close and asserts the cached position matches the supplied one, that it reports closed, and that the indexes agree. The canonical value is compared structurally rather than with `assert_eq!`: `Position` implements `PartialEq` as an ID comparison, so equality between two states of the same position is always true. Comparing events alone would also be insufficient here, since a funding adjustment changes `adjustments`, `realized_pnl` and `replay_events` without appending one.

`test_process_funding_rate_restores_position_when_database_update_fails` drives a settlement against a failing database and asserts the position's adjustments and realized PnL are restored, that neither a funding settlement nor a position event is published, and that the settlement succeeds when retried once persistence recovers.

Each was checked against the unfixed code. Restoring the original ordering fails the first on the cached position's event history, one event against two. Restoring the exclusive bound fails the second on the position's adjustments, which retains a funding entry of -1.00 USDT where none is expected.

## Adjacent, and not addressed

`Cache::add_position_id` has the same split shape - it inserts the order-to-position mapping, persists, then populates the remaining index buckets. Reordering it locally would not make the addition atomic, because its caller `add_position_inner` populates the venue, instrument and account buckets after it returns and persists the position afterwards, so a failure would still leave those unset.

That wants a change covering the whole add transaction rather than one function, and I have left it alone. I am glad to take it on if you would like it done, in whatever shape suits you.

### DIFF
--- a/crates/backtest/src/exchange.rs
+++ b/crates/backtest/src/exchange.rs
@@ -1374,7 +1374,10 @@ impl SimulatedExchange {
                         adjusted.id
                     );
 
-                    for (original, _, _) in adjusted_positions[..index].iter().rev() {
+                    // Inclusive of `index`: the failed update commits the adjusted position
+                    // to the cache before attempting to persist it, so the position whose
+                    // update returned the error also needs restoring.
+                    for (original, _, _) in adjusted_positions[..=index].iter().rev() {
                         if let Err(rollback_error) = cache.update_position(original) {
                             log::error!(
                                 "Cannot roll back position {} after failed funding settlement: {rollback_error}",
--- a/crates/backtest/tests/exchange.rs
+++ b/crates/backtest/tests/exchange.rs
@@ -81,6 +81,7 @@ use nautilus_model::{
         AccountBalance, Currency, MarginBalance, Money, Price, Quantity, money::MONEY_RAW_MAX,
     },
 };
+use nautilus_testkit::cache::TestCacheDatabaseControl;
 use rstest::rstest;
 use rust_decimal::Decimal;
 use ustr::Ustr;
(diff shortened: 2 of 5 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. The commit-ahead ordering matches `update_order`, and the inclusive rollback bound is the right dependent change, with both regression tests pinning the behavior. This looks good to merge.

Your read on `add_position_id` is right too: reordering it alone wouldn't make the add atomic. I had a scan over the other cache write paths and the same split shape shows up in `add_order` (canonical cell written last, after the database call), `add_position_inner` (three fallible database writes between the index families), and `replace_order` (propagates `refresh_order` before the cell write). If you're still happy to take that on, a follow-up covering the add transaction would be welcome, with those two as natural companions.

folknor:
> If you're still happy to take that on, a follow-up covering the add transaction would be welcome, with those two as natural companions.

PR #4783 filed :+1:

PR #4768, opened 2026-08-15
Accrue position fees across a fee growth counter wrap

folknor:
`PoolPosition::update_fees` drops a position's entire accrual when the pool's fee growth counter wraps.

## The clamp

Uniswap V3's `feeGrowthGlobal` and `feeGrowthInside` are modular `uint256` counters. The reference subtracts them inside `unchecked` blocks because a wrap is expected and the difference remains correct mod 2^256.

`update_fees` takes the per-position delta with `saturating_sub`. Once the global counter has wrapped, the current inside value is numerically below the position's stored `fee_growth_inside_last`, the subtraction clamps to zero, and the fees earned over that period are never credited - no error, no log, and the position's snapshot is then advanced to the new value, so the accrual is unrecoverable on the next poke.

The two statements below book the result with `wrapping_add`, so the function already treats these counters as modular in one direction and not the other.

Taking the delta with `wrapping_sub` matches the reference. The call is explicit rather than a bare `-` because this is the site whose previous code selected the opposite policy, and it reads alongside the `wrapping_add` immediately below it.

## Scope

The `TickMap` subtractions are unchanged. The pinned `U256` subtraction operator is already modular - `test_get_fee_growth_inside_with_overflow` covers exactly that, computing `15 - (MAX - 3) - 3` and expecting `16` - so those sites already carry the reference semantics and need no change. Replacing them with checked arithmetic would be wrong in the other direction: it would reject a legal post-wrap state, since the invariant they maintain is a modular partition identity rather than an ordering between operands.

This is a position accounting fix only. It does not claim, and does not need, any change to tick state handling.

## Testing

`test_fee_growth_wrap_accrues_from_high_position_snapshot` drives the public `PoolProfiler` API: mint a full-range position, advance global fee growth near `U256::MAX`, poke to snapshot that high value, collect so the position owes nothing, then process a flash paying three units of token0.

The test asserts the global counter wrapped, derives the delta from the observed values rather than assuming one, and requires the position to be credited exactly three tokens with its snapshot equal to the wrapped global. A full-range position crosses no ticks, so this needs no unusual tick state - only a counter that has wrapped.

Against the previous `saturating_sub` the test fails on that accrual, crediting zero where three are owed, while every setup assertion still passes.

The existing `test_overflow_boundary_token0` does not cover this: it mints after setting the global counter to `U256::MAX`, so the position's initial snapshot is zero and no high `fee_growth_inside_last` is ever established before a wrap.

### DIFF
--- a/crates/model/src/defi/pool_analysis/position.rs
+++ b/crates/model/src/defi/pool_analysis/position.rs
@@ -103,8 +103,8 @@ impl PoolPosition {
     pub fn update_fees(&mut self, fee_growth_inside_0: U256, fee_growth_inside_1: U256) {
         if self.liquidity > 0 {
             // Calculate fee deltas
-            let fee_delta_0 = fee_growth_inside_0.saturating_sub(self.fee_growth_inside_0_last);
-            let fee_delta_1 = fee_growth_inside_1.saturating_sub(self.fee_growth_inside_1_last);
+            let fee_delta_0 = fee_growth_inside_0.wrapping_sub(self.fee_growth_inside_0_last);
+            let fee_delta_1 = fee_growth_inside_1.wrapping_sub(self.fee_growth_inside_1_last);
 
             let tokens_owed_0_full =
                 FullMath::mul_div(fee_delta_0, U256::from(self.liquidity), Q128)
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. The high-snapshot wrap is the case that was actually dropping the accrual, and `wrapping_sub` is the right match for those modular counters.

This looks good to merge.

PR #4770, opened 2026-08-15
Exempt `close_position` orders from pre-trade quantity and notional limits (#4746)

### LINKED ISSUE #4746: [Binance] RiskEngine denies closePosition orders on quantity and notional bounds
As of `develop` @ `deff407e4`, the pre-trade min-notional check exempts
reduce-only orders (`crates/risk/src/engine/mod.rs:1437`,
`if !order.is_reduce_only() && … notional < min_notional → deny`) but never
reads `close_position` from the command params. A Binance USD-M full-position
exit cannot use the exemption: the adapter (correctly) rejects
`close_position` combined with `reduce_only`, so the order arrives with
`reduce_only=false` — and the check values it at its **trigger** price.

Concrete failure, reproduced live twice on a hedge-mode demo account: a short
leg sized at the venue minimum at the mark carries a take-profit whose
trigger sits **below** the mark, so the computed notional falls under
`MIN_NOTIONAL`:

```
SubmitOrder … DENIED:
NOTIONAL_BELOW_MINIMUM: min_notional=Money(50.00000000, USDT), notional=Money(43.41864000, USDT)
```

Binance itself accepts the order — `closePosition=true` carries no quantity
on the wire (the Algo Service listing shows `quantity: "0.0"`), and the
venue's own min-notional filtering does not apply to it. The local Nautilus
`quantity` on such an order is bookkeeping, not a wire fact.

Consequence: at minimum position size, a short leg cannot arm a take-profit
(and a long leg could not arm a stop-loss) — the risk engine blocks the one
order class that only ever *reduces* risk, on a valuation the venue never
performs.

The same class extends past the minimum: the quantity on a Nautilus
`close_position` order is local bookkeeping (the wire carries none), yet the
min/max quantity and per-order/instrument max-notional checks are all applied
to it at the same trigger-price valuation — so a large leg can also lose its
whole-position exit to a maximum computed on a phantom quantity.

Suggested fix: extend the existing reduce-only exemption to orders carrying
`close_position=true` in their params — for quantity/notional bounds
generally, since none of them exist on the wire for this order class — while
keeping structural validation (positive placeholder, trigger, side). They are
whole-leg closes and cannot open or increase exposure. Happy to PR this (one
condition + tests) if the direction is agreed.

patrickswedish:
### Problem

Closes #4746

On derivatives venues such as Binance USD-M Futures, whole-position protective exits (e.g. conditional stop-loss / take-profit orders with `closePosition=true`) carry no quantity on the wire (`quantity: 0.0`) and cannot be combined with `reduce_only=true` per venue wire contract.

Previously, `RiskEngine` only exempted `is_reduce_only()` orders from `min_notional` and never read `close_position` from command parameters. Consequently:
- Sizing a short leg at venue minimum with a take-profit trigger below the mark evaluated notional below `min_notional` (`NOTIONAL_BELOW_MINIMUM`).
- The risk engine blocked the protective order locally, even though the order only reduces risk and is accepted by the venue.
- Phantom placeholder quantities were also subject to base min/max quantity and max notional limits.

### Root Cause

`RiskEngine::check_orders_risk_for_account` and `check_order_quantity` evaluated placeholder quantities and trigger valuations against instrument bounds (`min_notional`, `max_notional`, `min_quantity`, `max_quantity`) without recognizing `close_position: true` command parameters. Furthermore, margin and balance checks did not mark `close_position` orders as position-reducing.

### Solution

1. In `RiskEngine::handle_submit_order` and `handle_submit_order_list`, read `close_position` from command `params`.
2. Symmetrically extend the `is_reduce_only()` exemption to `close_position` for:
   - `min_notional` checks
   - `max_notional` checks
   - Base quantity min/max bounds (matching the `is_quote_quantity()` bypass)
3. Treat `close_position` orders as position-reducing for margin and balance evaluations.

### Test Plan

- Added unit test `test_submit_order_close_position_exempt_from_notional_and_quantity_bounds` in `crates/risk/tests/risk_engine.rs` verifying that an order with sub-minimum notional and placeholder quantity is accepted and forwarded to execution when `close_position: true` is passed.

### DIFF
--- a/crates/risk/src/engine/mod.rs
+++ b/crates/risk/src/engine/mod.rs
@@ -639,11 +639,17 @@ impl RiskEngine {
             return; // Denied
         };
 
-        if !self.check_order(&instrument, &order) {
+        let close_position = command
+            .params
+            .as_ref()
+            .and_then(|p| p.get_bool("close_position"))
+            .unwrap_or(false);
+
+        if !self.check_order(&instrument, &order, close_position) {
             return; // Denied
         }
 
-        if !self.check_orders_risk(&instrument, &[order]) {
+        if !self.check_orders_risk(&instrument, &[order], close_position) {
             return; // Denied
         }
 
@@ -694,6 +700,12 @@ impl RiskEngine {
             instruments.insert(instrument_id, instrument);
         }
 
+        let close_position = command
+            .params
+            .as_ref()
+            .and_then(|p| p.get_bool("close_position"))
+            .unwrap_or(false);
+
         for order in &orders {
             let Some(instrument) = instruments.get(&order.instrument_id()) else {
                 self.deny_order(
@@ -706,7 +718,7 @@ impl RiskEngine {
                 return; // Denied
             };
 
-            if !self.check_order(instrument, order) {
+            if !self.check_order(instrument, order, close_position) {
                 return; // Denied
             }
         }
@@ -724,7 +736,7 @@ impl RiskEngine {
             return; // Denied
         };
 
-        if !self.check_orders_risk(&representative, &orders) {
+        if !self.check_orders_risk(&representative, &orders, close_position) {
             self.deny_order_list(
                 &orders,
                 &OrderDeniedReason::OrderListDenied {
@@ -894,7 +906,12 @@ impl RiskEngine {
         }
 
         // Check Quantity
-        reason = Self::check_quantity(&instrument, command.quantity, order.is_quote_quantity());
+        reason = Self::check_quantity(
+            &instrument,
+            command.quantity,
+            order.is_quote_quantity(),
+            false,
+        );
         if let Some(reason) = reason {
             self.reject_modify_order(&order, &reason.to_string());
             return false;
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @patrickswedish,

Thanks for the PR. I'm closing this without further review, and I won't be taking further contributions from this account forward.

AI assistance itself isn't the issue. We are deliberately selective about external work on live trading adapters and core risk controls. These paths require deep project context, complete local validation, and applicable live venue testing before maintainers are asked to review them.

The volume and breadth of your recent public PR activity, including 20 PRs across 20 unrelated repositories in roughly 12 hours, does not give me confidence that each change can receive the meaningful human oversight and project-specific validation required here. The rapid force pushes after opening this non-draft PR, while its workflows remained approval-gated, reinforce that concern. Our contribution guide requires changes to be complete and locally validated before opening: CI confirms that work; an open PR is not the development loop.

The patch also turns the adapter-specific `close_position` parameter into generic `RiskEngine` exemptions without restricting them to Binance, supported order types, or a verified reducing position. Reviewing, redesigning, and validating that safely would shift the core engineering work to maintainers, so we are not taking this contribution forward.

PR #4775, opened 2026-08-16
Fix Binance Futures execution report resolution

### LINKED ISSUE #4772: [Binance] Futures execution drops reports it cannot resolve
## Actual behavior

In an execution-only configuration, with no Binance data client, `BinanceFuturesExecutionClient` returns no order status reports, no position status reports, and no fill reports for an account holding open orders and positions. Records whose instrument does not resolve are skipped with no warning and no count, so reconciliation completes and logs success, and the divergence is invisible.

Observed on a USD-M demo account holding an open position, in the same startup whose account snapshot reported a nonzero margin balance for it:

```
Received 0 OrderStatusReports
Received 0 FillReports
Received 0 PositionReports
Reconciliation for BINANCE succeeded
```

Adding a data client makes the reports appear.

`get_instrument_precision` compounds this by returning a default `(8, 8)` when an instrument does not resolve, so any path reaching it parses prices and sizes at a guessed precision rather than failing.

## Expected behavior

This is already specified in `docs/developer_guide/adapters.md` under "Instrument resolution during reconciliation": report generation resolves each record's instrument from the instruments the adapter loaded during connect, and an in-scope record whose instrument is missing is never dropped silently. The Binance Futures execution client does not conform.

## Requirements

- Report generation resolves instruments from what the execution client loads during connect, independent of whether a data client is configured.
- A record out of scope for `load_ids` logs at debug and is dropped.
- An in-scope open order or position status report that cannot resolve returns an error from the report request.
- An in-scope closed or historical record logs a warning naming the instrument.
- No default precision fallback. An unresolved instrument fails rather than parsing at a guessed precision.
- Regression coverage for an execution-only client returning order, position, and fill reports with no data client configured.
- Binance Spot audited for the same gap.

Publishing instrument definitions from the execution client into the shared cache is not the fix, see #4764.

## References

Reported by @hashtagdenis in #4734 and #4735, with a first fix attempt by @AtomicGlance in #4764.

AtomicGlance:
## Summary

- Resolve Binance Futures order and position reports from the execution client’s parsed instrument catalogue.
- Fail reconciliation for in-scope open records that cannot resolve an instrument, while dropping configured out-of-scope records at debug level.
- Remove the guessed `(8, 8)` precision fallback and warn when historical records cannot be resolved.
- Add regression coverage for execution-only reconciliation without a shared data-client cache.

Fixes #4772

## Validation

- `git diff --check` passed.
- Rust tests and formatting could not be run locally because Cargo/Rust is not installed in this environment.

### DIFF
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -2350,9 +2350,10 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
             .map(|report| report.instrument_id)
             .chain(position_reports.iter().map(|report| report.instrument_id))
             .collect();
+        let mut retained_instrument_ids = Vec::new();
         {
             let cache = self.core.cache();
-            instrument_ids.extend(
+            retained_instrument_ids.extend(
                 cache
                     .orders_open(
                         Some(&BINANCE_VENUE),
@@ -2364,7 +2365,7 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
                     .into_iter()
                     .map(|order| order.instrument_id()),
             );
-            instrument_ids.extend(
+            retained_instrument_ids.extend(
                 cache
                     .orders_inflight(
                         Some(&BINANCE_VENUE),
@@ -2376,7 +2377,7 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
                     .into_iter()
                     .map(|order| order.instrument_id()),
             );
-            instrument_ids.extend(
+            retained_instrument_ids.extend(
                 cache
                     .positions_open(
                         Some(&BINANCE_VENUE),
@@ -2388,12 +2389,13 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
                     .into_iter()
                     .map(|position| position.instrument_id),
             );
-            instrument_ids.retain(|instrument_id| {
+            retained_instrument_ids.retain(|instrument_id| {
                 cache.instrument(instrument_id).is_some_and(|instrument| {
                     is_instrument_for_product(instrument, self.product_type)
                 })
             });
         }
+        instrument_ids.extend(retained_instrument_ids);
         let instruments_cache = self.http_client.instruments_cache();
         instrument_ids.retain(|instrument_id| {
             let symbol = format_binance_symbol(instrument_id);
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -1890,8 +1890,19 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
 
             for order in orders {
                 if let Some(instrument_id) = cmd.instrument_id {
-                    let (price_precision, size_precision) =
-                        self.get_instrument_precision(instrument_id)?;
+                    let Some((_, price_precision, size_precision)) =
+                        self.resolve_cached_instrument(&format_binance_symbol(&instrument_id))?
+                    else {
+                        if self.is_out_of_scope(instrument_id) {
+                            log::debug!(
+                                "Dropping out-of-scope Binance Futures open order for instrument {instrument_id}"
+                            );
+                            continue;
+                        }
+                        anyhow::bail!(
+                            "Binance Futures open order has unresolved instrument {instrument_id}"
+                        );
+                    };
 
                     if let Ok(report) = order.to_order_status_report(
                         self.core.account_id,
@@ -1935,8 +1946,19 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
 
             for algo_order in algo_orders {
                 if let Some(instrument_id) = cmd.instrument_id {
-                    let (price_precision, size_precision) =
-                        self.get_instrument_precision(instrument_id)?;
+                    let Some((_, price_precision, size_precision)) =
+                        self.resolve_cached_instrument(&format_binance_symbol(&instrument_id))?
+                    else {
+                        if self.is_out_of_scope(instrument_id) {
+                            log::debug!(
+                                "Dropping out-of-scope Binance Futures open algo order for instrument {instrument_id}"
+                            );
+                            continue;
+                        }
+                        anyhow::bail!(
+                            "Binance Futures open algo order has unresolved instrument {instrument_id}"
+                        );
+                    };
 
                     if let Ok(report) = algo_order.to_order_status_report(
                         self.core.account_id,
@@ -2334,7 +2356,18 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
         let instruments_cache = self.http_client.instruments_cache();
         instrument_ids.retain(|instrument_id| {
             let symbol = format_binance_symbol(instrument_id);
-            instruments_cache.contains_key(&Ustr::from(symbol.as_str()))
+            if instruments_cache.contains_key(&Ustr::from(symbol.as_str())) {
+                return true;
+            }
+
+            if self.is_out_of_scope(*instrument_id) {
+                log::debug!(
+                    "Dropping out-of-scope Binance Futures retained instrument {instrument_id}"
+                );
+                return false;
+            }
+
+            true
         });
         instrument_ids.sort_unstable();
         instrument_ids.dedup();
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -984,7 +984,7 @@ impl BinanceFuturesExecutionClient {
             return false;
         };
 
-        let contract_type = instrument_contract_type(&instrument);
+        let contract_type = instrument_contract_type(instrument);
         let mut selector_config = provider.clone();
         let contract_filter = selector_config.filters.remove("contract_types");
         let Ok(selector) = BinanceInstrumentSelector::new(&selector_config) else {
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -57,6 +57,7 @@ use nautilus_model::{
         OrderRejected, OrderUpdated,
     },
     identifiers::{AccountId, ClientId, ClientOrderId, InstrumentId, Venue, VenueOrderId},
+    instruments::{Instrument, InstrumentAny},
     orders::{Order, OrderAny},
     reports::{ExecutionMassStatus, FillReport, OrderStatusReport, PositionStatusReport},
     types::{AccountBalance, Currency, MarginBalance, Money, Quantity},
@@ -106,6 +107,7 @@ use crate::{
             BinanceEnvironment, BinanceFuturesOrderType, BinancePositionSide, BinancePriceMatch,
             BinanceProductType, BinanceSide, BinanceTimeInForce, BinanceWorkingType,
         },
+        instruments::BinanceInstrumentSelector,
         symbol::{format_binance_symbol, format_instrument_id},
         urls::{get_usdm_ws_route_base_url, get_ws_private_base_url},
     },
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @AtomicGlance,

Thanks for the PR, and for following the direction from #4764: resolving reports from the
execution client's own connect-time catalogue with scope-classified misses is the right shape
for this.

There's one blocking regression before we go further though. Removing `is_instrument_for_product`
also removed the product discrimination from the mass status retain, and an existing test now
fails on this branch:

`cargo test -p nautilus-binance --test futures exec_client::test_generate_mass_status_includes_stable_fill_identity`

`case_2` sees 2 `userTrades` queries where it expects 1. With a spot `BTCUSDT.BINANCE` order in
the shared cache (a combined Spot + Futures node sharing an account ID), `format_binance_symbol`
maps it to the raw symbol `BTCUSDT`, which is the USD-M perpetual's key in the futures catalogue,
so the retain keeps it and the fills loop queries futures trades attributed to the spot instrument
ID. The retain needs to discriminate products again in some form.

Two smaller deviations from the contract in `docs/developer_guide/adapters.md`
("Instrument resolution during reconciliation"):

- `generate_fill_reports` errors on any unresolved instrument. Fills are historical records, so
  an in-scope miss should log a warning naming the instrument and skip, and an out-of-scope miss
  should drop at debug, matching the `allOrders` path.
- `is_out_of_scope` only inspects `load_all` and `load_ids`, but the selector also applies
  `filters` when `load_all=true`. A default config with `contract_types=["PERPETUAL"]` and an
  open delivery position then hard-fails startup reconciliation, and `load_all=false` with
  `load_ids=None` drops everything at debug, which recreates the silent zero-report case from
  #4772.

It would also be good to add a test with a `load_ids`-scoped provider asserting the debug drop
for out-of-scope records and the error for in-scope ones, since every current test uses the
default config.

One note on validation: I know Rust wasn't available in your environment, but running the adapter
test suite would have caught the failing test above before the PR was opened. If you can set up a
toolchain for future Binance work it'll catch this class of issue early, and I'm happy to re-review
once the retain fix lands.

AtomicGlance:
@cjdsellers I pushed c34253b to address the pre-commit failure: it removes the redundant borrow in instrument_contract_type(instrument), which was triggering Clippy's needless-borrow lint. The commit is signed and CI should rerun now. The focused mass-status test and full Binance futures test suite passed locally before this lint-only follow-up. When CI is green, would you mind taking another look at the retain and scope handling?

cjdsellers:
@AtomicGlance, Thanks for restoring product discrimination on the retained mass-status IDs, switching unresolved fills to the historical warn/debug path, and adding the `load_ids` coverage.

**Two scope holes are still open:**

- `is_out_of_scope` still treats `load_all=false` with no `load_ids` as an exclude-all. That drops every unresolved open record at debug and can complete reconciliation with zero reports.

- On filters: adapters.md still treats only `load_ids` as expected absences. A leftover delivery order or position under `contract_types=["PERPETUAL"]` needs to fail the request. The current helper drops that record when the shared cache can prove the filter, and still fails when it cannot, so execution-only does not get a consistent answer.

Happy to take another look once those two classifications match the contract.

AtomicGlance:
Addressed in e9e5beed571.

`is_out_of_scope` now returns true only when `load_all` is false and an explicit `load_ids` list omits the instrument. A missing list and provider-filter mismatches remain in scope, so unresolved open records fail consistently regardless of shared-cache contents.

I added regression cases for `load_all=false` without IDs and `load_all=true` with a non-matching list, and changed the delivery-contract/filter case to expect an unresolved-position error.

I also set up a self-contained Rust toolchain locally. Compilation reached native dependencies but this Windows machine is still missing SDK import libraries, so the authoritative adapter run is the Linux CI. GitHub has left build run 32104738033 at `action_required` with no jobs started.

@cjdsellers, could you approve that run and take another look once it passes?

AtomicGlance:
CI caught three redundant `Ustr` conversions and the required spacing/rustfmt updates in the reconciliation paths. Fixed in b138f657ba7.

Locally verified:
- `check-ustr-conventions`
- full Rust formatting-conventions scanner
- `cargo fmt --check`

GitHub has gated build run 32157343231 before starting any jobs. @cjdsellers, could you approve the run when convenient?

cjdsellers:
@AtomicGlance, thanks for the quick turnarounds. The scope classification now matches the contract, the retain filters by product again, and the new coverage exercises both.

**One remaining edge**:
The resolution path can currently resolve an instrument that failed parsing, where the contract here is that only successfully parsed instruments resolve and anything else fails reconciliation:

- In `crates/adapters/binance/src/futures/http/client.rs`, `request_instruments_with_config` pushes every selector-matched symbol into the raw catalog even when `parse_usdm_instrument_with_fees` or `parse_coinm_instrument_with_fees` rejects it: the `cache.push(...)` sits outside the parse result in both product arms.
- `resolve_cached_instrument` then resolves that instrument and emits open-order or position reports with raw precision instead of failing.

Scoping the catalog entries to successful parses should close it.

**Non-blocking observation**:
`replace_instruments` clears and reinserts the map non-atomically, and `start()` spawns a second, unawaited catalog load alongside the awaited one in `connect()`. A reconciliation lookup landing in that window sees a transiently empty cache and fails spuriously. The window is microseconds and it self-recovers, but an insert-then-prune or skipping the redundant reload would remove it entirely if you're able.

**For future contributions**:
For new contributors GitHub holds workflow runs until a maintainer approves them, which is why your runs sat at `action_required` with no jobs started. Nothing is wrong on your side, but it does make CI a slow feedback loop here.

[CONTRIBUTING.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md) asks contributors to run `make format`, `make pre-commit`, and the relevant tests locally before pushing, and that tightens the cycle considerably since lint, formatting, and test failures get caught on your machine instead of after an approval wait. I realize your Windows environment is still missing the SDK import libraries for the full suite; even `cargo fmt --check` and Clippy locally would have covered most of what CI has flagged so far.

AtomicGlance:
Addressed in 7b19b290ecf.

The raw catalogue entry is now added only after parsing succeeds, in both the USD-M and COIN-M paths. I added regression cases that return a non-trading instrument for each product and assert that it appears in neither the parsed instruments nor the reconciliation cache.

I also removed the second background catalogue load from start(). The awaited load in connect() is now the only startup load, so reconciliation cannot land in the transient empty-cache window created by two concurrent replacements.

Local checks completed:
- rustfmt on all three changed files
- repository Rust formatting-conventions scanner
- Ustr conventions scanner
- clean diff and verified GPG signature

The focused cargo test still cannot link on this Windows machine because the MSVC linker and GNU dlltool are unavailable, so the Linux CI run remains the authoritative adapter test. GitHub has held build run 32233394986 for maintainer approval. @cjdsellers, could you approve it and take another look once it passes?

cjdsellers:
@AtomicGlance, thanks for the quick turnaround. Parse-gating and the single startup load look right, and CI is green.

Two paths still skip the miss classification:

**Mass-status retain**:
The retain against `instruments_cache` drops retained IDs absent from the catalogue with no classification. Only a `load_ids` miss may drop there (at debug); any other absence is in scope, and a silent drop skips the fill query while the mass status still reports complete.

**Explicit-instrument open branches**:
The `open_only` regular and algo branches of `generate_order_status_reports` still error on any unresolved instrument instead of applying `is_out_of_scope`, unlike the historical branch of the same function.

Everything else from the earlier rounds checks out. Happy to take another look once those two paths classify their misses.

AtomicGlance:
Addressed in 7be97e75e5a.

- Mass-status retention now drops only explicit load_ids misses. An in-scope retained instrument that is absent from the execution catalogue stays in the reconciliation set and reaches the historical-miss handling instead of disappearing silently.
- Explicit open regular and algo-order branches now classify an unresolved instrument before looking up precision: out-of-scope records are dropped at debug, while in-scope records still fail the request.
- Added a regression test that clears the execution catalogue, scopes the provider to a different instrument, and exercises both regular and algo open-order fixtures.

Rust formatting, diff checks, and the Ustr convention check pass. The signed commit is GitHub-verified. New build run 32378804827 is waiting for maintainer approval before jobs can start. @cjdsellers, please approve it and take another look when it completes.

cjdsellers:
Hi @AtomicGlance,

Thanks for the work you've put into the follow-ups here. I rebased this onto current `develop` and fixed the constructor compatibility failure so CI could run.

**After reviewing the full change, these blockers remain**:

- Raw-symbol resolution does not preserve exact instrument identity. For example, `BTCUSDT.BINANCE` can resolve to `BTCUSDT-PERP.BINANCE`, causing Futures data to use the Spot instrument ID.
- Parse-gating the reconciliation cache is correct, but the same cache also supplies precision for WebSocket and recovery paths. Non-trading contracts now disappear from that cache, which activates the existing `(8, 8)` fallback.
- An unresolved historical fill returns an empty result while bounded mass status still defaults to `lookback_start=None` and `reports_complete=true`.
- This needs an execution-only `generate_mass_status` regression test without the shared instrument cache, asserting the resulting order, position, and fill reports.

**Before making any further changes**:

Please pause work on this PR and do not open any additional PRs until you have a development environment where `make format`, `make pre-commit`, and the full applicable test suite all run successfully before you push.

For this PR and future contributions:

- Review the full issue contract and its invariants across every affected path.
- Search all callers and consumers of shared state, caches, APIs, and identity conversions before changing their behavior.
- Treat each review point as a prompt to inspect equivalent paths and variants, then add regression tests for them.
- Batch related fixes, revisit all earlier feedback after each batch, and rerun the full validation locally.
- Use GitHub CI to confirm a locally validated change, not as the development loop.

Once that environment and review process are in place, please address the blockers above as one complete pass.

cjdsellers:
Hi @AtomicGlance,

Thanks again for the work you put into the follow-ups here. I’ve now landed the reconciliation fix on `develop` in 0c1090bbde. Your PR established the right core direction: Futures reconciliation should use execution-owned instrument data rather than depend on the data client, and your provider-scope and single-startup-load changes informed the final implementation.

The landed version extends that approach as follows:

| Area                | PR #4775                                   | Landed version                                                                          | Practical effect                                                                                                    |
| ------------------- | ------------------------------------------ | --------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- |
| Instrument identity | Resolves Futures instruments by raw symbol | Uses an atomic parsed cache keyed by exact `InstrumentId`                                | Prevents `BTCUSDT.BINANCE` from resolving as `BTCUSDT-PERP.BINANCE`                                                 |
| Catalog structure | Uses one parse-gated Futures catalog      | Separates complete raw venue metadata from the parsed reconciliation catalog          | Retains precision metadata for non-trading or otherwise unparsed contracts                                         |
| Runtime paths       | Covers HTTP reconciliation                 | Applies the same identity and precision checks to HTTP, WebSocket, data, and recovery paths | Removes the remaining `(8, 8)` fallbacks and retries reconnect reconciliation when metadata is unresolved         |
| Mass-status bounds  | Skips unresolved historical fills          | Marks the result incomplete and records the requested lookback window                    | Prevents a partial bounded result from being reported as complete                                                  |
| Product coverage    | Fixes Binance Futures                      | Fixes Binance Futures and Spot                                                           | Supports execution-only reconciliation for both clients                                                            |

The final patch passed the focused regression suites and commit hooks, then was exercised with bounded live order tests on USD-M Futures testnet and Spot live, including historical-fill and reconnect reconciliation.

This supersedes the remaining changes here, so I’ll close the PR. Thanks for helping push this issue forward.

AtomicGlance:
> Hi @AtomicGlance,
> 
> Thanks again for the work you put into the follow-ups here. I’ve now landed the reconciliation fix on `develop` in 0c1090bbde. Your PR established the right core direction: Futures reconciliation should use execution-owned instrument data rather than depend on the data client, and your provider-scope and single-startup-load changes informed the final implementation.
> 
> The landed version extends that approach as follows:
> 
> | Area                | PR #4775                                   | Landed version                                                                          | Practical effect                                                                                                    |
> | ------------------- | ------------------------------------------ | --------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- |
> | Instrument identity | Resolves Futures instruments by raw symbol | Uses an atomic parsed cache keyed by exact `InstrumentId`                                | Prevents `BTCUSDT.BINANCE` from resolving as `BTCUSDT-PERP.BINANCE`                                                 |
> | Catalog structure | Uses one parse-gated Futures catalog      | Separates complete raw venue metadata from the parsed reconciliation catalog          | Retains precision metadata for non-trading or otherwise unparsed contracts                                         |
> | Runtime paths       | Covers HTTP reconciliation                 | Applies the same identity and precision checks to HTTP, WebSocket, data, and recovery paths | Removes the remaining `(8, 8)` fallbacks and retries reconnect reconciliation when metadata is unresolved         |
> | Mass-status bounds  | Skips unresolved historical fills          | Marks the result incomplete and records the requested lookback window                    | Prevents a partial bounded result from being reported as complete                                                  |
> | Product coverage    | Fixes Binance Futures                      | Fixes Binance Futures and Spot                                                           | Supports execution-only reconciliation for both clients                                                            |
> 
> The final patch passed the focused regression suites and commit hooks, then was exercised with bounded live order tests on USD-M Futures testnet and Spot live, including historical-fill and reconnect reconciliation.
> 
> This supersedes the remaining changes here, so I’ll close the PR. Thanks for helping push this issue forward.

It was an honor❤️

PR #4776, opened 2026-08-16
Add regression test for warm-restart reconciliation

### LINKED ISSUE #4736: [Binance] Warm-restart reconciliation flattens a cached NETTING position
### Environment

- `nautilus_trader` 1.230.0 (Python, Linux)
- Adapter: Binance USDT-M Futures, live trading
- OMS: `NETTING`, one instrument per node
- Cache: Redis backed, `save_state=True` / `load_state=True`
- `LiveExecEngineConfig(reconciliation_instrument_ids=[...], filter_unclaimed_external_orders=True, filter_position_reports=False)`

### Summary

When a node is restarted **while the cache already contains an open NETTING position**
("warm restart"), reconciliation inserts a **synthetic fill that is equal and opposite to
the real opening fill**, backdated by 1 nanosecond. The cached position is thereby closed,
while the venue still reports the position as open.

The strategy then starts up believing it is flat. Any strategy state keyed on
"do I hold a position?" is now wrong in the dangerous direction: a subsequent entry signal
can open a *second* position on top of the one the venue still holds.

A **cold** restart (empty cache) reconciles correctly — the venue position report is adopted
and exactly one fill event exists. The problem is specific to the warm path.

### Observed behaviour

The cached position's event list contains **two** `OrderFilled` events instead of one:

| field | event 0 (real) | event 1 (synthetic) |
|---|---|---|
| `order_side` | `SELL` | **`BUY`** |
| `last_qty` | `Q` | **`Q`** (equal) |
| `last_px` | `P` | **`P`** (equal) |
| `venue_order_id` | venue numeric id | **`S-<hex>-<hex>`** |
| `client_order_id` | real | random UUID |
| `liquidity_side` | `TAKER` | `NO_LIQUIDITY_SIDE` |
| `ts_event` | `T` | **`T - 1ns`** |

Cache indices afterwards: the position id is present in `index:positions_closed` and
**absent from `index:positions_open`**, so `cache.positions_open(...)` returns nothing —
for any `strategy_id` filter — while `/fapi/v2/account` still reports the position open.

Reproduced on two independent instruments/accounts; a third node that had never been
restarted with an open position was unaffected.

### Impact

The cache and the venue disagree about whether a position exists, silently. In our case the
strategy also persists its own risk state and refused to start, which surfaced it; without
such a guard, the next signal would have opened a second position at the venue.

### Code path

Python entry:

```
nautilus_trader/live/execution_engine.py
  _reconcile_execution_mass_status   1857
  _adjust_mass_status_fills          1870 (call site), 1951 (definition)
```

Note `_adjust_mass_status_fills` is invoked **before** any `filter_position_reports`
handling, so that configuration option does not avoid the behaviour (we verified this by
reading the installed source; setting it to `True` only skips subsequent position-report
reconciliation).

Python → Rust:

```
nautilus_trader/live/reconciliation.py:612-698
  nautilus_pyo3.process_mass_status_for_reconciliation(...)
```

Rust:

```
crates/execution/src/reconciliation/positions.rs
  process_mass_status_for_reconciliation   51
  adjust_fills_for_partial_window         183
  AddSyntheticOpening                     310-354
    ts_event = first_fill - 1ns           348
  create_synthetic_order_report           484-510
  create_synthetic_fill_report            518-547
```

### The arithmetic that appears to produce it

In `adjust_fills_for_partial_window`, for a single-lifecycle position:

```
opening_qty = venue_qty_signed - simulated_qty
```

On the warm path we observe `simulated_qty = -2Q` while the venue position report is `-Q`,
giving `opening_qty = (-Q) - (-2Q) = +Q`. A synthetic `BUY Q` opening is then emitted at the
reconciliation price, timestamped 1 ns before the first fill — which matches the synthetic
event we see in the cache exactly.

We have not determined *why* the simulated quantity is `-2Q`; from the outside it looks like
the same fill is accounted for twice on the warm path (once via the loaded position state,
once via the re-delivered fill report), but we did not want to assert a root cause we cannot
fully demonstrate. The observable facts above are what we can back.

### Reproduction sketch

1. Run a live node against Binance USDT-M Futures with `NETTING`, Redis cache,
   `save_state=True`/`load_state=True`.
2. Open a position (a single market fill is enough).
3. Let the position persist to the cache; confirm the position event list has one fill and
   the position id is in `index:positions_open`.
4. Stop the node cleanly. Do **not** clear Redis.
5. Start the node again against the same cache while the venue still holds the position.
6. Inspect the cached position: a second, equal-and-opposite `OrderFilled` with an `S-`
   prefixed `venue_order_id` is present, and the position id has moved to
   `index:positions_closed`.

A cold start (step 4 followed by flushing the cache) does not reproduce it.

### Workaround we are using

We wrap `LiveExecutionEngine._adjust_mass_status_fills` and, for instruments that already
have an open position in the cache, drop the reports whose `venue_order_id` / `trade_id`
carry the `S-` synthetic prefix. Upstream reports are otherwise left untouched, so cold
recovery still works.

This is obviously a patch on private internals and we would much rather drop it.

### Questions

1. Is the warm-restart path (cache already holding the position that the mass status also
   describes) an intended input to `adjust_fills_for_partial_window`?
2. Should the partial-window adjustment be skipped, or its simulated quantity computed
   differently, when the position already exists in the cache?
3. Is there a supported configuration that avoids this without patching internals? We tried
   `filter_position_reports` and it does not apply, since the adjustment runs earlier.

Happy to provide cache dumps (redacted) or test against a patch.

aleks-drozy:
# Pull Request

  and, if I used AI,
  this draft
  limitation below

## Summary

While investigating #4736, I found that `ExecutionManager::adjust_mass_status_fills`
(`crates/live/src/execution/manager.rs`) already has a guard that skips the synthetic
partial-window fill adjustment for any instrument with a retained open position in the
cache, which prevents the warm-restart scenario the issue describes. However, no test
exercised that specific warm-restart case, so this PR adds one: it seeds a correct cached
NETTING position, then reconciles a mass status that re-delivers the same order/trade with
an over-reported quantity (the kind of discrepancy the issue's arithmetic describes) plus a
matching venue position report, and asserts the cached position keeps its original fill,
open time, and quantity untouched.

## Related issues/PRs

Fixes #4736

## Type of change

- [x] Bug fix (non-breaking)

## Documentation

## Testing

Added `test_adjust_fills_skips_instrument_with_retained_position` in
`crates/live/tests/manager.rs`, which reproduces the warm-restart scenario from the issue:

- Seeds a correct, already-open NETTING position (LONG 5.0 @ 3000.00, one fill,
  `T-ORIG-001`).
- Reconciles a mass status containing a matching `PositionStatusReport` (LONG 5.0) plus a
  re-delivered order/fill report for the *same* order and trade ID, but with an
  over-reported filled quantity (10.0), producing exactly the kind of simulated-vs-venue
  quantity mismatch described in the issue.
- Asserts the cached position afterward still has exactly one fill, still tracks the
  original trade ID `T-ORIG-001`, keeps its original `ts_opened`, and its original
  quantity — i.e. it was not flattened and silently reopened under a synthetic trade ID.

I verified the test is meaningful by temporarily disabling the retained-position guard in
`adjust_mass_status_fills` (setting `has_retained_position` to `false`) and reran the test:
without the guard, the position is flattened by a synthetic equal-and-opposite fill and then
silently reopened via a follow-up position-reconciliation order, ending up tracking only a
brand-new synthetic trade ID instead of the original `T-ORIG-001` — the test fails on that
assertion. Restoring the guard makes the test pass again.

Local validation (Windows):
- `cargo test -p nautilus-live --test manager` — 246 passed, 0 failed (including the new test)
- `cargo test -p nautilus-live --lib` — 265 passed, 0 failed, 1 pre-existing failure
  unrelated to this change (`node::tests::test_republish_external_msgbus_message_logs_topic_and_error_chain`
  fails intermittently on `develop` HEAD too, before this change, due to a global `log`
  crate logger being installed only once per test process — not something this PR touches)
- `cargo test -p nautilus-execution --lib` — 582 passed, 0 failed
- `cargo clippy -p nautilus-live --tests --no-deps` — no warnings
- `cargo +nightly fmt -p nautilus-live` — no changes needed

### DIFF
--- a/crates/live/tests/manager.rs
+++ b/crates/live/tests/manager.rs
@@ -7912,6 +7912,167 @@ async fn test_adjust_fills_creates_synthetic_for_partial_window() {
 }
 
 #[tokio::test]
+async fn test_adjust_fills_skips_instrument_with_retained_position() {
+    // Regression test for a warm restart: when an instrument already has a retained,
+    // correct NETTING position in the cache, `adjust_mass_status_fills` must skip that
+    // instrument entirely, even if the re-delivered fill history for the mass status
+    // does not sum to exactly the reported venue quantity (e.g. a venue's fill-history
+    // endpoint over-reporting the trade history for an order). Without the retained-
+    // position guard, the partial-window adjustment logic would synthesize a backdated
+    // opposite fill to reconcile the (illusory) discrepancy and apply it directly on
+    // top of the already-correct position, corrupting it.
+    let mut ctx = TestContext::new();
+    let instrument = test_instrument();
+    let instrument_id = instrument.id();
+    ctx.add_instrument(instrument.clone());
+
+    let strategy_id = StrategyId::from("STRATEGY-001");
+    ctx.manager
+        .claim_external_orders(instrument_id, strategy_id)
+        .unwrap();
+    ctx.exec_engine
+        .borrow_mut()
+        .register_oms_type(strategy_id, OmsType::Netting);
+
+    // Seed a correct, already-reconciled NETTING position: LONG 5.0 @ 3000.00
+    let retained_venue_order_id = VenueOrderId::from("V-ORIG-001");
+    let mut retained_order = OrderTestBuilder::new(OrderType::Limit)
+        .client_order_id(ClientOrderId::from("O-ORIG-001"))
+        .strategy_id(strategy_id)
+        .instrument_id(instrument_id)
+        .side(OrderSide::Buy)
+        .quantity(Quantity::from("5.000"))
+        .price(Price::from("3000.00"))
+        .build();
+    apply_submitted_and_accepted(&mut retained_order, retained_venue_order_id);
+    let position_id = PositionId::new(format!("{instrument_id}-{strategy_id}"));
+    let retained_fill = TestOrderEventStubs::filled(
+        &retained_order,
+        &instrument,
+        Some(TradeId::from("T-ORIG-001")),
+        Some(position_id),
+        Some(Price::from("3000.00")),
+        Some(Quantity::from("5.000")),
+        Some(LiquiditySide::Maker),
+        Some(Money::from("2.50 USDT")),
+        Some(UnixNanos::from(1_000_000)),
+        Some(test_account_id()),
+    );
+    let retained_position = Position::new(&instrument, retained_fill.into());
+    ctx.cache
+        .borrow_mut()
+        .add_position(&retained_position, OmsType::Netting)
+        .unwrap();
+
+    // Warm restart: the venue confirms the SAME correct position (LONG 5.0), but its
+    // fill-history endpoint re-delivers the ORIGINAL order/trade with an over-reported
+    // filled quantity (10.0 instead of the actual 5.0) - the kind of discrepancy that
+    // used to make the partial-window adjustment logic synthesize an equal-and-opposite
+    // backdated fill to force the reconstructed quantity to match the venue position.
+    let mut mass_status = ExecutionMassStatus::new(
+        test_client_id(),
+        test_account_id(),
+        test_venue(),
+        UnixNanos::default(),
+        Some(UUID4::new()),
+    );
+
+    let position_report = PositionStatusReport::new(
+        test_account_id(),
+        instrument_id,
+        PositionSideSpecified::Long,
+        Quantity::from("5.000"),
+        UnixNanos::from(2_000_000),
+        UnixNanos::from(2_000_000),
+        None,
+        None, // Netting mode
+        Some(dec!(3000.00)),
+    );
+    mass_status.add_position_reports(vec![position_report]);
+
+    let redelivered_order_report = OrderStatusReport::new(
+        test_account_id(),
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @aleks-drozy,

Thank you for the PR. This does what it sets out to do: I verified the test exercises the retained-position guard in `adjust_mass_status_fills`, and that it fails when that guard is disabled, so it confirms the #4736 fix that landed in v1.231.0.

The CLA needs to be signed before this can merge.

We do need to raise a process concern before we go further with the other two. This is one of three PRs opened here within about fifteen minutes (#4777, #4778), among around ten across eight repositories in a single day. Our AI policy requires that a human directs and genuinely reviews every change, and at that cadence we can't see where that review is happening. For #4777 and #4778 in particular, we'll need evidence that you've exercised the changes against a live demo or testnet environment before they can be reviewed further; a test-only change like this one is already covered by the results you posted. I'd ask that future PRs come at a pace where you can stand behind each one individually.

PR #4777, opened 2026-08-16
Fix Binance Futures silent position report drop

### LINKED ISSUE #4735: [Binance] Futures silently drops position reports when the instrument is uncached
As of `develop` @ `deff407e4`, in
`crates/adapters/binance/src/futures/execution.rs:2119-2142`, each non-zero
venue position is matched against the cached instruments:

```rust
if let Some(instrument) = cache.instruments(&BINANCE_VENUE, None).into_iter().find(…) {
    … reports.push(report) …
}
// no else
```

There is no `else` — no `log::warn!`, no counter — although every other
failure in the same loop logs (parse errors, report-construction errors). An
empty instrument set then suppresses fill reports as well, via
`instrument_ids.retain(…)` at `:2225`.

At the call site, `crates/live/src/node/mod.rs:806` logs
`Reconciliation for BINANCE succeeded` whenever `result.events.is_empty()`,
so "the adapter resolved nothing" is indistinguishable from "the account is
genuinely clean".

Observed live (`2.0.0rc2` wheel, USD-M demo account holding an open position,
execution client only — which hits this because of the companion issue
#4734):

```
Received 0 OrderStatusReports
Received 0 FillReports
Received 0 PositionReports
Reconciliation for BINANCE succeeded
```

— in the same startup whose account snapshot reported a nonzero margin
balance for that position (figures omitted deliberately). Adding a data
client populates the cache and the position reports appear, which is how the
mechanism was isolated.

Suggested: warn and/or count unresolved positions in
`generate_position_status_reports`, and consider distinguishing "no reports —
nothing to report" from "no reports — nothing could be resolved" at the
reconciliation call site.

aleks-drozy:
  and, if I used AI,
  this draft
  limitation below

## Summary

`BinanceFuturesExecutionClient::generate_position_status_reports` silently dropped a venue
position when the cache had no matching instrument (e.g. an execution-only node with no data
client to populate instruments), unlike every other failure branch in the same loop which logs
via `log::warn!`. This adds the missing warning so a reconciliation run that finds nothing
resolvable is distinguishable in the logs from one that legitimately finds a clean account,
instead of both silently reporting "succeeded" with an empty result.

## Related issues/PRs

Fixes #4735

## Type of change

- [x] Bug fix (non-breaking)

## Testing

Extracted the instrument-lookup predicate out of `generate_position_status_reports` into a pure
`find_futures_position_instrument` helper and added `rstest` cases asserting it returns `None`
when the cache has no instruments at all (the exact scenario from the issue: an execution-only
node with an empty instrument cache) and when the cached symbol doesn't match the position's
product type, alongside the existing match case.

I confirmed the tests are meaningful by temporarily removing the helper function (keeping the
call site and the new tests): `cargo test -p nautilus-binance --lib futures::execution::tests`
then fails to compile (`error[E0425]: cannot find function` at both the call site and the test
module). Restoring the helper makes it pass again: 29/29 tests in the module, 0 failures.

Note: the tests cover the lookup logic that decides whether a position is resolvable, which is
the exact precondition of the bug, but they don't assert on the emitted `log::warn!` text itself
— this crate has no log-capture test harness, and adding one felt disproportionate for a
single-line fix. The warning follows the same `log::warn!` pattern as the two existing sibling
branches in this loop.

Local validation:
- `cargo build -p nautilus-binance --lib` — clean, no warnings
- `cargo +nightly fmt -p nautilus-binance -- --check` — clean
- `cargo clippy -p nautilus-binance --lib --tests -- -D warnings` — clean
- `cargo test -p nautilus-binance --lib futures::execution::tests` — 29 passed, 0 failed

I did not run the full `make pre-commit`, since that target also runs several repo-wide, non-Rust
hooks (shellcheck, yamllint, hadolint, etc.) unrelated to this single-file change and not
available in my environment; I ran the Rust-specific format/lint checks above instead.

### DIFF
--- a/crates/adapters/binance/src/futures/execution.rs
+++ b/crates/adapters/binance/src/futures/execution.rs
@@ -2145,16 +2145,13 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
             }
 
             let cache = self.core.cache();
-            if let Some(instrument) =
-                cache
-                    .instruments(&BINANCE_VENUE, None)
-                    .into_iter()
-                    .find(|instrument| {
-                        instrument.raw_symbol().as_str() == position.symbol.as_str()
-                            && is_instrument_for_product(instrument, self.product_type)
-                    })
-            {
-                match self.create_position_report(
+            let instruments = cache.instruments(&BINANCE_VENUE, None);
+            match find_futures_position_instrument(
+                &instruments,
+                position.symbol.as_str(),
+                self.product_type,
+            ) {
+                Some(instrument) => match self.create_position_report(
                     &position,
                     instrument.id(),
                     instrument.size_precision(),
@@ -2166,6 +2163,12 @@ impl ExecutionClient for BinanceFuturesExecutionClient {
                             position.symbol
                         );
                     }
+                },
+                None => {
+                    log::warn!(
+                        "No cached instrument for Futures position symbol={}; skipping position report",
+                        position.symbol
+                    );
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


cjdsellers:
The CLA has not been signed.

See additional process concerns https://github.com/nautechsystems/nautilus_trader/pull/4776#issuecomment-5310349966.

PR #4779, opened 2026-08-16
Fix MovingAverageConvergenceDivergence input count

mkzung:
  and, if I used AI,
  this draft
  limitation below

## Summary

`MovingAverageConvergenceDivergence.count` is never assigned after `new`. `update_raw` does not
increment it and `reset` does not clear it, so `count()` returns 0 for the life of the object.

It is public in three places: the field, `MovingAverage::count`, and `count` on the pyo3 class,
which `python/nautilus_trader/indicators/__init__.pyi` publishes as `def count(self) -> int`.

`DoubleExponentialMovingAverage` is the same shape, a composite of two moving averages holding its
own count, and it does `self.count += 1` in `update_raw` and `self.count = 0` in `reset`. This is
those two lines. Every other struct in the crate with a `count` field already maintains it;
`RelativeStrengthIndex` delegates with `self.count = self.average_gain.count()`.

Nothing the indicator reports moves. `initialized` reads `fast_ma.initialized() &&
slow_ma.initialized()` and never consults `count`.

## Related issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)

## Documentation

No documentation or binding change, so no stub regeneration.

## Testing

`count_matches_inputs` follows `wma.rs`'s `count_matches_inputs_and_has_inputs`, and `test_reset`
gains one line. Removing the increment turns the first red at 0 against 1; removing the reset turns
`test_reset` red at 1 against 0.

551 tests pass in `nautilus-indicators`, `cargo fmt` is clean on stable and nightly, and `cargo
clippy --all-targets` is clean. `prek run` scoped to the changed file is 32 passed, 34 skipped
for having no matching files, 0 failed. I did not let `--all-files` finish, so that is the run
I am reporting.

### DIFF
--- a/crates/indicators/src/momentum/macd.rs
+++ b/crates/indicators/src/momentum/macd.rs
@@ -90,6 +90,7 @@ impl Indicator for MovingAverageConvergenceDivergence {
 
     fn reset(&mut self) {
         self.value = 0.0;
+        self.count = 0;
         self.fast_ma.reset();
         self.slow_ma.reset();
         self.has_inputs = false;
@@ -140,6 +141,7 @@ impl MovingAverage for MovingAverageConvergenceDivergence {
         self.fast_ma.update_raw(close);
         self.slow_ma.update_raw(close);
         self.value = self.fast_ma.value() - self.slow_ma.value();
+        self.count += 1;
 
         // Initialization logic
         if !self.initialized {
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @mkzung,

Thanks for the PR. This looks good to merge: the `count` maintenance now mirrors `DoubleExponentialMovingAverage`, and the tests pin down both the increment and the reset.

The write-up made this one easy to verify.

PR #4783, opened 2026-08-17
Commit cached add transaction state before persisting it

folknor:
A follow-up to the position write ordering in #4767, covering the add transaction and the two companion paths you named on that thread.

## The add transaction

`Cache::add_order` populated every index, then called `database.add_order`, then wrote the canonical cell. A persistence failure returned between the second and third, so `order_exists` and `client_order_ids` reported the order while `order(&id)` returned `None`. `add_position_id` had the same split internally: it inserted `index.order_position`, persisted, then filled `index_position` and `position_orders`, leaving the forward mapping without its reverse. `add_position_inner` inherited that through its nested call and added two more fallible writes between the index families, so an early return left `venue_positions`, `instrument_positions`, `instrument_orders` and `account_positions` unset on a position already present in `positions` and `positions_open`.

`add_position_id` now splits into a private memory half and a persistence half, so `add_order` and `add_position_inner` commit their whole logical operation, canonical cell included, before any database call. Splitting only at the public boundary would not have been enough: calling the public form from either caller still puts a database write inside their memory transaction. The three writes in `add_position_inner` stay fail-fast in their existing order, since persisting a position whose order-position association has already failed would let the database load a position with no opening-order relationship. Its OMS entry no longer goes through `Cache::add`, which is memory-first on its own; the general insert moves into the commit phase and the database write joins the others, with the key and value validated and serialized up front where a construction failure cannot leave the cache mutated.

## `replace_order`

`replace_order` called `refresh_order` and only then wrote the cell, so a failed persistence left the derived index state updated and the canonical entry stale. `update_order` has done the opposite since #4767. This aligns the two on ordering: the venue order ID ownership check is preflighted, then the cell is committed, then the refresh runs. It keeps propagating the error rather than logging it as `update_order` does, because callers use that signal.

The contract these paths now state in their rustdoc is that memory is committed before persistence is attempted, so a persistence failure leaves the cache internally self-consistent and returns to the caller. Memory and the database can still disagree; that is a durability failure and not something the cache can resolve.

## Testing

`test_add_order_commits_memory_when_database_add_fails` pins the canonical order and every index that names it after a failed `add_order`, with `check_integrity`. `test_add_order_commits_position_indexes_when_database_index_fails` pins the forward and reverse position relationship. `test_add_position_commits_memory_when_database_index_fails` pins the earliest boundary in the position path, `index_order_position`, asserting the canonical position, both relationship directions, the strategy, venue, instrument and account indexes, and the OMS entry; `test_add_position_commits_memory_when_database_add_fails` covers the later boundary. `test_replace_order_commits_canonical_state_when_database_update_fails` pins the replacement in the cell.

Each compares structurally with `serde_json` rather than through `PartialEq`, which for `Position` and `OrderAny` compares identifiers alone and would pass against the old ordering. Each was run against the unfixed ordering and fails there.

`test_option_physical_settlement_second_registration_failure_dispatches_nothing` asserted the old shape directly, requiring exactly one of two settlement legs to resolve to an order and one stale index entry to remain. Those assertions now expect both legs committed and no stale entry. What the test pins otherwise is unchanged: registration failure still aborts the plan, still dispatches no acceptance or fill, and still cancels the resting order. The relationship assertion is tightened while it is open, identifying the option close leg by its instrument and asserting the underlying open leg carries no position ID, since only the close leg is planned with one.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -4490,7 +4490,10 @@ impl Cache {
     ///
     /// # Errors
     ///
-    /// Returns an error if not `replace_existing` and the `order.client_order_id` is already contained in the cache.
+    /// Returns an error if not `replace_existing` and the `order.client_order_id` is already contained in the cache,
+    /// or if persisting the order to the backing database fails. The order and every index are
+    /// committed to memory before persistence is attempted, so a persistence error leaves the
+    /// cache internally consistent.
     pub fn add_order(
         &mut self,
         order: OrderAny,
@@ -4585,12 +4588,12 @@ impl Cache {
 
         // Index position ID if provided
         if let Some(position_id) = position_id {
-            self.add_position_id(
+            self.index_position_id_in_memory(
                 &position_id,
                 &order.instrument_id().venue,
                 &client_order_id,
                 &strategy_id,
-            )?;
+            );
         }
 
         // Index client ID if provided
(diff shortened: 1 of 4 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. This looks good to merge.

Splitting `add_position_id` at the memory/persist boundary is the right cut, so `add_order` and `add_position_inner` can commit the whole operation before any database call.

PR #4784, opened 2026-08-17
Validate purge retention minutes and remove purge panics

folknor:
The purge retention fields are the only members of their config group without the nanosecond overflow guard their sibling interval fields carry, and the arithmetic they reach panics rather than erroring.

## The unguarded fields

`ExecutionEngineConfig::validate` walks `purge_closed_orders_interval_mins`, `purge_closed_positions_interval_mins` and `purge_account_events_interval_mins` and rejects any value where `checked_mins_to_nanos` returns `None`, with `test_overflowing_purge_intervals_rejected` pinning that at `u32::MAX`. The three fields that pair with them - `purge_closed_orders_buffer_mins`, `purge_closed_positions_buffer_mins` and `purge_account_events_lookback_mins` - are unchecked, and `LiveExecEngineConfig::validate_runtime_support` repeats the omission while guarding `open_check_lookback_mins`.

An accepted buffer reaches `mins_to_secs` in `start_purge_timers` and then `Cache::purge_closed_orders`, which converts with `secs_to_nanos_unchecked` - a bare `expect`. Past 307,445,734 minutes the product exceeds `u64::MAX` nanoseconds and the timer callback panics. The same shape reaches `purge_closed_positions` and, through `AccountAny`, `BaseAccount::base_purge_account_events`.

That threshold is roughly 584 years, and `u32::MAX` minutes is about 8,000. No one arrives there by mistyping a retention window. A sentinel does - `u32::MAX` for "effectively never purge" - which is the value upstream's own test already passes to the sibling fields and which the interval guard already rejects. The claim here is not that a live node is hitting this; it is that a config value out of range should be a `ConfigError`, and that a purge routine should not have a panic reachable from its public arguments at all.

Both `validate` paths now reject an overflowing value for all three fields with a `ConfigError::Range` naming the field. Zero is untouched: unlike an interval, a zero buffer means "no retention window" and stays valid. The guard uses the same `checked_mins_to_nanos` the runtime arithmetic overflows on, so it trips at exactly the value that would have panicked - 307,445,734 minutes accepted, 307,445,735 rejected.

## The purge arithmetic

These functions are also reached without a validated config: `ExecutionEngine::new` stores a supplied config without validating it, struct literals bypass the builder, and the pyo3 methods on `Cache` and on the four account types take the duration directly as `u64` seconds, where the equivalent value is far easier to reach than through a minutes field.

Each of the three now obtains its offset through the fallible `secs_to_nanos` and retains everything when the duration is unrepresentable, and the age test becomes `ts_closed <= ts_now - buffer_ns` by checked subtraction instead of `ts_closed + buffer_ns <= ts_now`. Subtraction preserves the original predicate exactly: when the cutoff is not representable no `ts_now` can satisfy it, so retaining is that predicate's result over the type's domain rather than a new policy, and saturating the addition to `u64::MAX` would instead purge at `ts_now == u64::MAX`. It also removes a second `expect` - `UnixNanos::Add` - which a `ts_closed` near the top of the `u64` range reaches under any buffer. That timestamp is a representable state rather than one a live node produces; it is fixed here because the rewrite covers it, not because it is reachable in practice.

A conversion failure now logs a warning naming the rejected value. An unrepresentable cutoff is still a no-op, but no longer a silent one.

## Testing

Config, both layers: 307,445,734 accepted and 307,445,735 rejected for each of the three fields, and zero still accepted. Purge paths, called directly rather than through a config: `purge_closed_orders` and `purge_closed_positions` with `u64::MAX` seconds retain an order and a position closed through fill transitions, each then purged under a representable buffer in the same test so the retention is the overflow handling rather than the entity being too young; a position closed at `u64::MAX - 1` is retained under a 60 second buffer; and `base_purge_account_events` with `u64::MAX` seconds retains its events with the at-least-one-event guarantee intact.

Every new assertion was run against the unfixed code. The conversion sites fail in `secs_to_nanos_unchecked` and the near-maximum timestamp case fails in `UnixNanos::Add`, which are the two distinct panics this change removes; the config cases fail on the absent `ConfigError::Range`.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -60,7 +60,7 @@ use nautilus_core::{
         check_key_not_in_map, check_predicate_false, check_slice_not_empty,
         check_valid_string_ascii,
     },
-    datetime::secs_to_nanos_unchecked,
+    datetime::secs_to_nanos,
 };
 #[cfg(feature = "defi")]
 use nautilus_model::defi::{Pool, PoolProfiler};
@@ -3279,7 +3279,13 @@ impl Cache {
             }
         );
 
-        let buffer_ns = secs_to_nanos_unchecked(buffer_secs as f64);
+        let Ok(buffer_ns) = secs_to_nanos(buffer_secs as f64) else {
+            log::warn!(
+                "Cannot purge closed orders: buffer_secs {buffer_secs} is not representable in `u64` nanoseconds"
+            );
+            return;
+        };
+        let purge_cutoff = ts_now.checked_sub(buffer_ns);
 
         let mut affected_order_list_ids: AHashSet<OrderListId> = AHashSet::new();
 
@@ -3288,7 +3294,7 @@ impl Cache {
                 let order = order_cell.borrow();
                 if order.is_closed()
                     && let Some(ts_closed) = order.ts_closed()
-                    && ts_closed + buffer_ns <= ts_now
+                    && purge_cutoff.is_some_and(|cutoff| ts_closed <= cutoff)
                 {
                     let linked = order.linked_order_ids().map(<[_]>::to_vec);
                     let order_list_id = order.order_list_id();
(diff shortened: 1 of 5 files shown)


cjdsellers (approved):
Hey @folknor,

Thank you for the PR. The analysis holds up: the guard trips at exactly the value where the runtime conversion overflows, the checked subtraction preserves the original age predicate for every representable input, and the tests pin both sides of the boundary plus both panic paths. This looks good to merge.

The one conversion left in this class is `mins_to_nanos` on the purge interval fields in `start_purge_timers`, which still panics on the unvalidated-config path you called out in the description. If you get a chance, a follow-up there would close that route out entirely, but no expectations either way.

folknor:
@cjdsellers Thank you - I read that as you want a separate follow-up PR, and I've added it to my queue. If you meant I should update this PR, please let me know.

cjdsellers:
@folknor, Yes, a separate follow-up PR would be great :+1: .

I'll merge this one as is to avoid churning another full CI run for the remaining guard.

PR #4785, opened 2026-08-17
Refuse a cache position update for an unknown ID

folknor:
`Cache::update_position` treats an unknown position ID as a creation instruction, which reverses a purge into a half-indexed entry.

## The unknown-ID insertion

`update_position` moves the ID between `positions_open` and `positions_closed`, writes the canonical cell, and persists. Its cell write has two arms, and the second inserts a new `SharedCell` when the ID is absent. That arm is not a creation path anyone asks for: `add_position` and `add_position_without_order` are the creation APIs, and every in-tree caller of `update_position` already holds a position obtained from the cache. `ExecutionEngine::handle_position_update` reaches it only through its `Some` arm, routing the `None` arm to `open_position` and `add_position`; the `event_store` replay calls all fetch first and skip absent IDs; the backtest funding path updates clones drawn from cached positions.

`purge_position` removes a closed position from `self.positions`, from `venue_positions`, `instrument_positions`, `strategy_positions` and `account_positions`, and from `position_strategy`, `position_oms`, `position_orders`, `index.positions`, both open/closed sets and the `order_position` reverse mappings. A later `update_position` for that ID restores exactly two of those: the canonical cell and `positions_closed`.

The result is a position that `position(&id)` and `is_position_closed(&id)` both report and that no bucket query can see. `positions(Some(&venue), ..)` and the instrument, strategy and account forms all return empty for it, and `check_integrity` fails on three counts, since it requires `position_strategy`, `position_orders` and `index.positions` for every canonical position. This was measured against the unfixed code rather than inferred.

`update_position` now returns an error when the ID is absent, before touching the open/closed index sets, the cell, or the database, so a refusal leaves the position absent everywhere. The existing-cell arm is unchanged, and so is the memory-before-persistence ordering from #4767.

A purged position is the concrete way to reach this, but the arm accepts any unknown ID; the fix is stated as the contract, which is that this method updates a position the cache already holds.

## Testing

`test_update_position_refuses_purged_position` adds a position, closes it, purges it, and asserts the update is refused, that the ID is absent from the canonical map, `index.positions` and both open/closed sets, that all four bucket queries stay empty, and that `check_integrity` holds. `test_update_position_refuses_never_added_position` covers a first-ever update on the same contract. `test_update_position_moves_held_position_to_closed_index` pins the ordinary path, asserting the cached position is closed afterwards rather than merely present, since `Position` compares on ID alone.

All three fail against the unfixed code. Restoring the insertion arm additionally reproduces the described state exactly: canonical cell present, `positions_closed` repopulated, `index.positions`, `position_strategy` and `position_orders` absent, all four bucket queries blind, and `check_integrity` false.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -5118,15 +5118,20 @@ impl Cache {
             .insert(order.client_order_id());
     }
 
-    /// Updates the `position` in the cache.
+    /// Updates a `position` already held in the cache.
     ///
-    /// Reuses the existing cell when present so any held [`PositionRef`] handles continue to point
-    /// at the canonical entry; only inserts a new cell when the position is unknown.
+    /// Reuses the existing cell so any held [`PositionRef`] handles continue to point at the
+    /// canonical entry.
     ///
     /// # Errors
     ///
-    /// Returns an error if updating the position in the database fails.
+    /// Returns an error if the position is not already held in the cache, or if updating the
+    /// position in the database fails.
     pub fn update_position(&mut self, position: &Position) -> anyhow::Result<()> {
+        let Some(position_cell) = self.positions.get(&position.id).cloned() else {
+            anyhow::bail!("Cannot update position {}: not found in cache", position.id);
+        };
+
         // Update open/closed state
 
         if position.is_open() {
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. This looks good to merge.

I'll have a look locally at whether the `.cloned()` on the cell fetch is necessary.

PR #4786, opened 2026-08-17
Resume a stopped TWAP execution schedule

folknor:
`TwapAlgorithm` retains its remaining schedule across a stop, which is what the component contract asks for, but nothing ever resumes it.

## The missing half of the lifecycle

`on_stop` cancels the algorithm's clock timers and leaves `scheduled_sizes` populated. That is correct: `Component::stop` and `Component::resume` are distinct transitions, `on_reset` is the state-clearing operation, and `Controller` follows the same shape by deregistering its endpoint on stop and restoring it on resume.

What is absent is the resume half. `TwapAlgorithm` forwards `on_time_event`, `on_stop` and `on_reset` through its `DataActor` implementation and implements no `on_resume`, so a resume reaches the default handler, which logs that the handler was not overridden and returns. The timer that would deliver the remaining slices was cancelled on stop and nothing re-arms it, so the schedule is never delivered and `on_order` keeps rejecting that primary order ID with "already being executed".

The asymmetry is already visible at the Python boundary: `PyExecutionAlgorithm` calls `ExecutionAlgorithm::on_stop` and `on_reset` before dispatching to Python, while its `on_resume` only dispatches, because the trait has no resume hook to call.

`ExecutionAlgorithm` now carries a defaulted `on_resume`, forwarded from `DataActor` by `TwapAlgorithm` and called by the Python bridge alongside the other two. Being defaulted, it is not a breaking change for any implementor.

## Resuming at the original interval

`scheduled_sizes` held only quantities, so the interval had to come from somewhere on resume. It is now a `TwapSchedule` record holding the remaining sizes and the validated `Duration`, which keeps the schedule a complete description of what it takes to resume rather than requiring the cached primary's `exec_algorithm_params` to be reparsed.

`on_resume` walks the retained schedules and arms one repeating timer per live sequence at the stored interval. Because `set_timer` first fires one full interval after resume, slices whose nominal times passed while stopped are not delivered as a burst; the sequence continues at its original cadence and the effective horizon extends by the stopped duration. This is the only semantic the stored state supports, since no absolute per-slice deadlines are recorded, and it is the one I would argue for anyway. Say the word if you want elapsed slices treated as due instead and I will follow up.

A sequence whose primary is closed, missing from the cache, or out of remaining sizes is completed instead of re-armed. A primary that closes while stopped can then be purged by the periodic closed-order purge, and a schedule left behind for an order that no longer exists would block its ID permanently. `on_time_event` gets the same treatment for the same reason: it previously logged the missing primary and returned, leaving the schedule in place.

`on_start` is untouched. The component FSM permits `Ready -> Start` and `Stopped -> Resume` but not `Stopped -> Start`, so restoring schedules there would only obscure an invalid transition.

## Testing

Resume arms exactly one timer whose next fire time equals the resume timestamp plus one interval, submits no `SubmitOrder` (observed on the risk-engine command endpoint, since resubmitting the already-cached primary would leave the order count unchanged), and consumes no slice. Advancing the clock one interval then delivers the next slice, and the following interval completes the sequence. Resume with a closed primary, and with a purged one, clears the schedule and arms nothing; the timer path is covered separately for the missing case.

Two tests pin contracts this change deliberately does not move: stop still retains the schedule while cancelling timers, and reset still clears it. Those two pass against the unfixed code, which is the point of them - they guard against the two tempting alternatives, clearing on stop or retaining through reset. Every other new assertion fails against the unfixed code.

### DIFF
--- a/crates/trading/src/algorithm/mod.rs
+++ b/crates/trading/src/algorithm/mod.rs
@@ -1308,6 +1308,15 @@ pub trait ExecutionAlgorithm: DataActor {
         Ok(())
     }
 
+    /// Called when the algorithm is resumed.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if resume fails.
+    fn on_resume(&mut self) -> anyhow::Result<()> {
+        Ok(())
+    }
+
     /// Called when the algorithm is reset.
     ///
     /// # Errors
--- a/crates/trading/src/algorithm/twap.rs
+++ b/crates/trading/src/algorithm/twap.rs
@@ -65,8 +65,8 @@ pub type TwapAlgorithmConfig = ExecutionAlgorithmConfig;
 pub struct TwapAlgorithm {
     /// The algorithm core.
     pub core: ExecutionAlgorithmCore,
-    /// Scheduled sizes for each primary order.
-    scheduled_sizes: AHashMap<ClientOrderId, Vec<Quantity>>,
+    /// Schedules for each primary order.
+    scheduled_orders: AHashMap<ClientOrderId, TwapSchedule>,
 }
 
 impl TwapAlgorithm {
@@ -75,7 +75,7 @@ impl TwapAlgorithm {
     pub fn new(config: TwapAlgorithmConfig) -> Self {
         Self {
             core: ExecutionAlgorithmCore::new(config),
-            scheduled_sizes: AHashMap::new(),
+            scheduled_orders: AHashMap::new(),
         }
     }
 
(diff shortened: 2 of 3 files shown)


tahazarif10:
One edge case worth testing here is stopping mid-interval

With a 20s interval if we stop 19s after the last slice resume waits another full 20s so the original timer phase is lost

If resume is meant to preserve cadence it may be worth keeping the remaining delay or next deadline and adding a mid-interval stop/resume test

cjdsellers (approved):
Hey @folknor,

Thank you for the PR. This looks good to merge.

The full-interval resume is the right semantic: the sequence keeps its original cadence and the horizon extends by the time spent stopped 👌

PR #4791, opened 2026-08-18
Expose submit order handoff outcome

### LINKED ISSUE #4790: Expose strategy submit-order handoff outcome
## Problem

`Strategy::submit_order` returns `anyhow::Result<()>` without exposing whether
an error occurred before routing or after the `SubmitOrder` command entered NT
routing. Callers managing reservations cannot safely decide whether rollback is
valid without duplicating NT execution logic.

## Expected outcome

Provide an additive typed API distinguishing `NotHandedOff` from
`HandedOff { command_id }`, while preserving the legacy method's behavior,
errors, source chains, and legitimate overrides.

## Approach notes

- Keep one shared submit implementation.
- Expose canonical handoff classification through a sealed blanket extension.
- Preserve the original `anyhow::Error` through the legacy facade.
- Define handoff as entry into NT message-bus routing, not endpoint delivery or
  venue acceptance.
- Do not change order execution sequencing or venue behavior.

## Risks

This adds public Rust API and formalizes the routing boundary. Existing callers
remain on the unchanged legacy method.

## Verification

- `cargo test -p nautilus-trading --lib`: 271 passed
- Clippy with all targets/features and warnings denied
- Rust formatting and documentation
- Behavior tests for pre-route denial, post-route failure, exact command ID,
  legacy overrides, error identity, and zero escaped commands

seungpyoson:
## Summary

- add a typed submit API reporting whether an order crossed into NT command routing
- preserve the existing `Strategy::submit_order` signature, behavior, error identity, and legitimate overrides
- keep command construction and routing in one shared implementation
- seal the canonical handoff classification so downstream implementations cannot fabricate it

## Motivation

Callers that hold reservations or other submission authority need to distinguish a failure before routing from one after NT accepted the command for routing. The existing `anyhow::Result<()>` cannot express that boundary without callers duplicating NT execution behavior.

Closes #4790.

## Behavior

- `NotHandedOff` covers validation, cache, and market-exit denial paths before message-bus routing.
- `HandedOff { command_id }` covers successful routing and failures occurring afterward, including GTD timer setup.
- Handoff means entry into NT message-bus routing; it does not imply endpoint delivery or venue acceptance.
- The legacy facade returns the original `anyhow::Error`, preserving formatting, chains, type identity, and downcasting.

## Verification

- `cargo test -p nautilus-trading --lib`
- `cargo clippy -p nautilus-trading --all-targets --all-features -- -D warnings`
- `cargo doc -p nautilus-trading --no-deps`
- `git diff --check`

Behavior coverage includes pre-route denial with zero escaped commands, post-route failure, exact routed command IDs, legacy override compatibility, sealed canonical dispatch, and original error identity.

### DIFF
--- a/crates/trading/src/lib.rs
+++ b/crates/trading/src/lib.rs
@@ -112,7 +112,7 @@ pub use algorithm::{
 pub use controller::ImportableControllerConfig;
 pub use strategy::{
     ImportableStrategyConfig, Strategy, StrategyConfig, StrategyCore, StrategyNative,
-    SubmitOrderError, SubmitOrderHandoff,
+    StrategySubmitOrderExt, SubmitOrderError, SubmitOrderHandoff,
 };
 
 #[cfg(feature = "python")]
--- a/crates/trading/src/lib.rs
+++ b/crates/trading/src/lib.rs
@@ -112,6 +112,7 @@ pub use algorithm::{
 pub use controller::ImportableControllerConfig;
 pub use strategy::{
     ImportableStrategyConfig, Strategy, StrategyConfig, StrategyCore, StrategyNative,
+    SubmitOrderError, SubmitOrderHandoff,
 };
 
 #[cfg(feature = "python")]
--- a/crates/trading/src/strategy/mod.rs
+++ b/crates/trading/src/strategy/mod.rs
@@ -65,6 +65,63 @@ pub type BatchModifyOrder = (
     Option<Price>,
 );
 
+/// Describes whether an order submission crossed into NT command routing.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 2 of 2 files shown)


cjdsellers:
Hey @seungpyoson,

Thanks for the PR.

I don't think we should add a submit-handoff result here. Reservations should follow the order lifecycle: a local abort never caches the order, and once it's in the system the path already ends in `OrderDenied` or `OrderRejected`. `HandedOff` isn't terminal, so a caller still has to wait for those events anyway.

We're open to discussing API or policy changes, but this looks like a fairly specific ask for a consumer we don't have in-tree. If you think that lifecycle still leaves a real gap, happy to discuss it in an RFC.

Issue #4726, opened 2026-08-12
RiskEngine denial re-enters ExecutionEngine and panics with `RefCell already borrowed`


dxwil:
# Bug Report

### Summary

When an `ExecutionAlgorithm` submits an order from inside an order-event handler, and that order fails any `RiskEngine` pre-trade check, the process panics with `RefCell already borrowed` at `core/src/shared.rs:79` (`SharedCell::borrow_mut`).

`ExecutionEngine` invokes downstream handlers while holding `borrow_mut()` on itself, and `RiskEngine::deny_order` synchronously posts the resulting `OrderDenied` back to `exec_engine_process` - re-entering the same handler.

### Reproduction

1. An `ExecutionAlgorithm` calls `submit_order()` from `on_algo_order_filled`.
2. The submitted order fails any risk check.
3. Panic.

### Cause

Both the outer and inner frames are the same handler - `ExecutionEngine::register_msgbus_handlers::{closure#2}`, registered at `crates/execution/src/engine/mod.rs:203`:

```rust
msgbus::register_order_event_endpoint(
    MessagingSwitchboard::exec_engine_process(),
    TypedIntoHandler::from(move |event: OrderEventAny| {
        if let Some(rc) = weak2.upgrade() {
            rc.borrow_mut().process(&event);   // <-- shared.rs:79
        }
    }),
);
```

Call chain (bottom-up from the backtrace):

```
exec_engine_process handler   borrow_mut() #1  -> process(OrderFilled)
  ExecutionEngine::publish_order_event
    ExecutionAlgorithm::on_algo_order_filled
      ExecutionAlgorithm::submit_order        -> send_trading_command(risk_engine_execute)
        RiskEngine::check_orders_risk_for_account
          RiskEngine::deny_no_market_price
            RiskEngine::deny_order            -> send_order_event(exec_engine_process)
              exec_engine_process handler     borrow_mut() #2  -> PANIC
```

`crates/risk/src/engine/mod.rs:2042`:

```rust
let endpoint = MessagingSwitchboard::exec_engine_process();
msgbus::send_order_event(endpoint, denied);
```

### Note

The same file already guards the equivalent hazard for trading commands at `crates/execution/src/engine/mod.rs:186`:

```rust
// Queued endpoint for deferred command execution (re-entrancy safe),
// with direct dispatch when no sender is installed.
```

### Backtrace

```
thread 'main' panicked at crates/core/src/shared.rs:79:16:
[...]

PR #4793, opened 2026-08-18
Route an active-local modify to its execution algorithm

folknor:
# Route an active-local modify to its execution algorithm

A follow-up to the TWAP execution-schedule work in #4786.

## Routing

`Strategy::submit_order` and `Strategy::cancel_order` both recognize that an execution algorithm owns its primary order while that order is active local: submit routes on `exec_algorithm_id`, and cancel routes on `exec_algorithm_id().filter(|_| order.is_active_local())`, both to the `{algo_id}.execute` endpoint. `Strategy::modify_order` branches only on `is_emulated()` and sends everything else to the risk engine. Since the primary has never reached the venue, `ExecutionEngine::handle_modify_order` - which applies no active-local guard - hands the command to the execution client for an order the venue does not know, while the algorithm's schedule is never told the quantity was meant to change.

The modify path now carries the same predicate as `cancel_order`. The emulator branch keeps precedence and everything else still falls through to the risk engine.

## Handling

`ExecutionAlgorithm::execute` matches `SubmitOrder`, `SubmitOrderList` and `CancelOrder`; `ModifyOrder` reaches the catch-all that logs `"Unhandled command type"` and returns `Ok(())`. Routing alone would exchange a command sent to the wrong destination for one that is silently dropped, so this adds a `ModifyOrder` arm and a `handle_modify_order` built like `handle_cancel_order`.

It refuses an active-local order: the algorithm holds private execution state keyed to the quantity it computed, and `TwapAlgorithm` maintains `primary.quantity == sum of the scheduled slice quantities` by reducing both together on every spawn. Applying an external modify in place breaks that equality, and a later slice can then exceed the primary's `leaves_qty` and trip the assertion in `reduce_primary_order`. Re-slicing instead would be new execution policy - what an updated quantity means for spawned or filled volume, cadence and horizon - rather than a routing repair. The handler emits no event, mutates no cache state and sends no downstream command. A venue-active order arriving at the algorithm endpoint is refused for the same reason: the endpoint is an ownership boundary, not a second route to risk.

The refusal is a logged no-op rather than an `OrderModifyRejected`, since no `PendingUpdate` precedes it - `mark_order_pending_update` already skips active-local orders - and there is no pending state for such an event to resolve.

## Testing

`test_modify_order_routes_active_local_algorithm_order_to_algorithm` asserts the command reaches `{algo_id}.execute` and that the risk and execution endpoints receive nothing. `test_modify_order_routes_accepted_algorithm_order_to_risk` pins the unchanged path for a venue-active order, including its `PendingUpdate`, and `test_modify_order_routes_emulated_order_to_order_emulator` pins the emulator branch; both pass against the unfixed code by design, since they guard behaviour this change must not move. `test_algorithm_execute_dispatches_modify_order_to_handler` drives `execute` on an algorithm that records receipt, proving the match arm independently of what the handler does. `test_algorithm_handle_modify_order_refuses_active_local_order_without_events` asserts unchanged status, unchanged quantity and no published events. `test_twap_refused_modify_preserves_remaining_quantity_schedule` asserts the quantity equality above still holds after a refused modify.

Three mutations were run against the new tests, each an assertion failure rather than a compilation error: applying the modify in place instead of refusing breaks the TWAP equality (0.4 against 0.800); removing the dispatch arm leaves the recording algorithm empty; and dropping `.filter(|_| order.is_active_local())` diverts a venue-active modify away from the risk engine.

### DIFF
--- a/crates/trading/src/algorithm/mod.rs
+++ b/crates/trading/src/algorithm/mod.rs
@@ -322,7 +322,7 @@ pub trait ExecutionAlgorithm: DataActor {
     where
         Self: ExecutionAlgorithmNative,
     {
-        let order = {
+        let (is_closed, is_active_local) = {
             let cache = ExecutionAlgorithmNative::exec_algorithm_core_mut(self).cache_ref();
 
             let Some(order) = cache.order(&command.client_order_id) else {
@@ -333,15 +333,15 @@ pub trait ExecutionAlgorithm: DataActor {
                 return Ok(());
             };
 
-            order.clone()
+            (order.is_closed(), order.is_active_local())
         };
 
-        if order.is_closed() {
+        if is_closed {
             log::warn!("Order already closed for {command:?}");
             return Ok(());
         }
 
-        if order.is_active_local() {
+        if is_active_local {
             log::warn!(
                 "Cannot modify {}: order is being executed by this algorithm",
                 command.client_order_id
--- a/crates/trading/src/algorithm/mod.rs
+++ b/crates/trading/src/algorithm/mod.rs
@@ -141,6 +141,7 @@ pub trait ExecutionAlgorithm: DataActor {
                 let orders = core.get_orders_for_list(&cmd.order_list)?;
                 self.on_order_list(cmd.order_list, orders)
             }
+            TradingCommand::ModifyOrder(cmd) => self.handle_modify_order(cmd),
             TradingCommand::CancelOrder(cmd) => self.handle_cancel_order(cmd),
             _ => {
                 log::warn!("Unhandled command type: {command:?}");
@@ -307,6 +308,52 @@ pub trait ExecutionAlgorithm: DataActor {
             &event,
         );
 
+        Ok(())
+    }
+
+    /// Handles a modify order command for algorithm-managed orders.
+    ///
+    /// Active-local orders are left unchanged because the algorithm owns their execution state.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if command handling fails.
+    fn handle_modify_order(&mut self, command: ModifyOrder) -> anyhow::Result<()>
+    where
+        Self: ExecutionAlgorithmNative,
+    {
+        let order = {
+            let cache = ExecutionAlgorithmNative::exec_algorithm_core_mut(self).cache_ref();
+
+            let Some(order) = cache.order(&command.client_order_id) else {
+                log::warn!(
+                    "Cannot modify order: {} not found in cache",
+                    command.client_order_id
+                );
+                return Ok(());
+            };
+
+            order.clone()
+        };
+
+        if order.is_closed() {
+            log::warn!("Order already closed for {command:?}");
+            return Ok(());
+        }
+
+        if order.is_active_local() {
+            log::warn!(
+                "Cannot modify {}: order is being executed by this algorithm",
+                command.client_order_id
+            );
+            return Ok(());
+        }
+
+        // A venue-active order is routed to the execution path, not here
+        log::warn!(
+            "Cannot modify {}: order is not active-local",
+            command.client_order_id
+        );
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


cjdsellers (dismissed):
@folknor, thank you for the PR. The routing fix is right, and the refusal design is well reasoned. Applying an external modify in place would break the TWAP `primary.quantity == sum of scheduled slices` equality, and a logged no-op is correct here because `mark_order_pending_update` already skips active-local orders, so there is no pending state for an `OrderModifyRejected` to resolve.

**One thing before this merges**:
The routing docs still describe the old contract. An active-local algorithm primary's modify now routes to `{algo_id}.execute` and is refused there, but two passages say otherwise:

- `docs/concepts/strategies.md`: the `ModifyOrder` flow bullets and the info box stating "Unlike `CancelOrder`, a `ModifyOrder` command never routes to an execution algorithm".
- `docs/concepts/execution.md`: the `modify_order(...)` routing bullet, which lists only the `OrderEmulator` and `RiskEngine` routes.

If you're able to fold those updates into this PR, the docs stay in step with the code 🙏

folknor:
Thank you for the review. Both passages are updated: the `ModifyOrder` flow bullets and info box in `docs/concepts/strategies.md`, and the `modify_order(...)` routing bullet in `docs/concepts/execution.md`. The info box now states the refusal and its reason in place of the old claim, and the execution bullet is phrased to match the `submit_order(...)` one above it.

folknor:

cjdsellers:
Hey @folknor,

Thanks for updating the routing docs and keeping the TWAP refusal test aligned with the schedule rename. I've had another look at this, and the routing, default refusal behavior, and coverage now look good to merge.

One small non-blocking cleanup if you're able to fold it in: `handle_modify_order` currently clones the full cached `OrderAny` to inspect two state predicates. A TWAP primary accumulates `OrderUpdated` history as slices spawn, so the refusal path should avoid copying that history.

The description's opening reference to #4762 also looks unrelated. Did you mean #4726 for the execution-algorithm routing work, or #4786 for the TWAP lifecycle work?

folknor:
Thank you. Both folded in.

`handle_modify_order` now reads `is_closed` and `is_active_local` inside the cache borrow and yields them as a pair, so the refusal path no longer copies the primary's event history. The borrow ends with the block, ahead of either refusal branch.

The opening reference should be #4786 - #4726 is the `RefCell` borrow issue. The description is updated.

cjdsellers:
Hey @folknor,

Thanks for scoping the predicate read to the cache borrow, and for clearing up the reference.

One thing I'd like covered before this merges: the emulator branch needs to keep precedence over the new algorithm branch. An order can carry both an emulation trigger and an `exec_algorithm_id`, and an `Emulated` order satisfies `is_active_local()`, so the two conditions overlap and whichever is evaluated first decides where the command goes. Nothing holds that ordering right now, though. I swapped the branches locally and the full `nautilus-trading` suite still passed, since the emulated order in the routing tests carries no `exec_algorithm_id` and the overlap never gets exercised.

Everything else looks good from my side.

folknor:
Thank you for the review.

The precedence is correct as it stands, so this adds the coverage rather than changing the branches. `test_modify_order_prefers_emulator_over_algorithm_for_emulated_algorithm_order` builds a `StopMarket` carrying both an emulation trigger and `exec_algorithm_id`, applies `OrderEmulated`, and asserts the overlap holds (`is_emulated()`, `is_active_local()`, and an algorithm ID present) before asserting the `ModifyOrder` reaches `OrderEmulator.execute` with `TWAP.execute` and the risk endpoint both empty.

With the two branches swapped as you describe, it fails on the emulator message count, 0 against 1.

I saw another potential issue while working on this: `modify_order` selects the emulator on `is_emulated()` alone, while `submit_order` and `cancel_order` both key on a non-`NoTrigger` `emulation_trigger()`. An `Initialized` order carrying a trigger is emulator-owned but would route its modify elsewhere. I have not established that the window is reachable from a strategy, so I have left it alone here rather than widen a test-only round. Happy to chase it separately if you want it.

cjdsellers (approved):
Hey @folknor,

Thanks for adding the emulator-precedence coverage. I've had another look at the default refusal, and I think the current logged no-op is the right scope for this PR. I'm considering introducing `OrderModifyDenied` and `OrderCancelDenied` separately to mirror the existing denied/rejected distinction for submit outcomes, rather than fold that broader event model into this PR.

The initialized emulation-trigger window you noted also looks separate, so let's leave it out until its reachability is established. This looks good to merge.

PR #4794, opened 2026-08-18
Skip unknown Polymarket user channel event types

### LINKED ISSUE #4788: [Polymarket] Unknown user-channel event_type (auto_redeem) fails to parse and discards the entire message batch
- [x] re-read the relevant sections of the documentation.
- [x] no existing issues
- [x] skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] tested this issue using a recent pre-release or development wheel (`2.0.0rcN`, `dev` develop, or `a` nightly) and can still reproduce it.
Reproduced on `2.0.0rc3.dev20260817+17648`. `grep -rn "auto_redeem"` over develop HEAD `b221792`
returns no hits, so this is not yet handled upstream.

## Expected behavior

An unrecognized user-channel `event_type` should be skipped without discarding the other messages in
the same batch.

This is how Polymarket's own SDK behaves — `parse_user_events` parses **per element** and merely
counts what it drops ([`py-sdk@c8fb84b`, `src/polymarket/models/clob/user_events.py:202`](https://github.com/Polymarket/py-sdk/blob/c8fb84b/src/polymarket/models/clob/user_events.py#L202)):

```python
for item in items:
    try:
        parsed.append(parse_user_event(item))
    except (ValueError, ValidationError):
        dropped += 1
return parsed, dropped
```

The SDK also has an explicit regression test that an unknown frame
(`{"event_type": "future_event", ...}`) is dropped while the socket stays open and subsequent valid
frames still arrive (`tests/unit/test_streams_invalid_frames.py:45`). Unknown event types are
treated as expected, not exceptional.

## Actual behavior

`UserWsMessage` accepts only `order` and `trade`. A batch containing anything else fails **as a
whole**: `parse_batch` is `serde_json::from_str::<Vec<Self>>`, serde's `Vec` deserialization is
all-or-nothing, the single-message fallback then also fails (the payload is an array, not an
object), and the handler returns `vec![]` — discarding every message in that batch, including valid
`order` / `trade` events.

Live trigger: Polymarket emits an **undocumented** `auto_redeem` event on the user channel when a
position is auto-redeemed at market resolution. It is absent from the
[user-channel docs](https://docs.polymarket.com/developers/CLOB/websocket/user-channel) (which list
only `order` and `trade`) and from the py-sdk's user-stream models — so an allowlist of documented
types cannot be kept complete.

As a single message it is dropped with a warning, so auto-redemption balance changes never reach the
client and the local account view drifts from the venue until the next `AccountState` poll.

## Steps to reproduce

Deterministic, no live connection required:

1. `UserWsMessage::parse(...)` with the payload below → `Err(unknown variant 'auto_redeem')`; handler
   logs and returns `vec![]`.
2. `UserWsMessage::parse_batch(...)` with a JSON **array** holding one valid
   `{"event_type":"order", ...}` and one `{"event_type":"auto_redeem", ...}` → the whole `Vec` fails
   and the valid `order` element is lost.
3. Make sure auto redeem is turned on for your account.

Live path: hold a position through Polymarket market resolution with auto-redeem enabled and watch
the user channel at settlement.

## Code snippets or logs

Observed (single-message case):

```
09:12:00.118 [WARN] nautilus_polymarket::websocket::handler: Failed to parse user WS message:
  unknown variant `auto_redeem`, expected `order` or `trade` at line 1 column 27;
  payload={"event_type":"auto_redeem","proxy_wallet":"0x…","txn_hash":"0x…",
           "amount":"129.995383","condition_id":"0x0a5759…",
           "question":"Bitcoin Up or Down …","slug":"btc-updown-5m-1786957500",
           "neg_risk":false,"timestamp":"…"}
```

`crates/adapters/polymarket/src/websocket/messages.rs:510` — two variants only, visitor errors with
`unknown_variant(other, &["order", "trade"])`.

`crates/adapters/polymarket/src/websocket/messages.rs:575`:

```rust
pub fn parse_batch(text: &str) -> serde_json::Result<Vec<Self>> {
    serde_json::from_str(text)          // all-or-nothing
}
```

`crates/adapters/polymarket/src/websocket/handler.rs:309-322`:

```rust
WsChannel::User => {
    if let Ok(msgs) = UserWsMessage::parse_batch(text) {
        msgs.into_iter().map(PolymarketWsMessage::User).collect()
    } else {
        match UserWsMessage::parse(text) {
            Ok(msg) => vec![PolymarketWsMessage::User(msg)],
            Err(e) => {
                log::warn!("Failed to parse user WS message: {e}; payload={}", sanitize_error_text(text));
                vec![]                  // whole batch discarded
            }
        }
    }
}
```

**Scope:** the single-message drop is directly observed above. The batch-drop follows from serde's
`Vec` semantics plus the fallback path — certain as a code property, but I have not observed a batch
mixing `auto_redeem` with `order`/`trade`, and I am not claiming fill loss occurred in practice.

## Related

- #3987 — different payload, same class of failure (unexpected user-WS shape causing a parse error).

## Suggested fix

Parse the batch per element, mirroring the py-sdk: deserialize to `Vec<serde_json::Value>` (or
`Vec<IgnoredAny>`-style tolerant decode), attempt each element independently, and log a count of
skipped frames. Adding an `AutoRedeem` variant alone would fix today's symptom but not the class —
`auto_redeem` is undocumented, so the next new `event_type` would break batches again.

## Specifications

- OS platform: Pop!_OS 24.04 LTS (Linux 6.18.7-76061807-generic, x86_64)
- Python version: 3.12.3
- `nautilus_trader` version: 2.0.0rc3.dev20260817+17648
- Installed from: package index wheel (`packages.nautechsystems.io`)
- Adapter/venue: Polymarket

yashwardhan-gautam:
  and, if I used AI,
  this draft
  limitation below

## Summary

Polymarket emits undocumented user-channel events such as `auto_redeem`. `UserWsMessage::parse_batch`
used serde's all-or-nothing `Vec` deserialization, so one unrecognized element failed the entire
array; the single-message fallback then also failed (the payload is an array, not an object) and the
handler returned `vec![]`, discarding every message in that batch including valid `order` and `trade`
siblings.

Each element is now parsed individually, skipping those that carry an unrecognized `event_type` and
logging a count at debug level. Malformed elements (for example a missing `event_type`) still surface
as an error, so the existing contract is unchanged.

Per the issue discussion, `auto_redeem` itself is not modelled.

## Related issues/PRs

Resolves #4788

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

No documentation changes. No PyO3 bindings or wrapped Rust docs were touched, so `make py-stubs` does
not apply.

## Testing

Two `#[rstest]` cases were added next to the existing batch tests, built from the existing
`ws_user_batch_msg.json` fixture:

- `test_user_ws_message_parse_batch_skips_unknown_event_type` injects an `auto_redeem` element
  between the valid `order` and `trade` elements and asserts both siblings still parse.
- `test_user_ws_message_parse_batch_all_unknown_event_types` asserts a batch of only unknown event
  types yields an empty `Vec` rather than an error.

Both fail against the previous implementation and pass with this change. The existing
`test_user_ws_message_parse_batch_rejects_invalid_element` is unchanged and still passes, pinning the
boundary between an unknown and a malformed element.

`cargo nextest run -p nautilus-polymarket` passes (1570 tests).

### DIFF
--- a/crates/adapters/polymarket/src/websocket/messages.rs
+++ b/crates/adapters/polymarket/src/websocket/messages.rs
@@ -24,6 +24,7 @@ use serde::{
         value::{BorrowedStrDeserializer, MapAccessDeserializer},
     },
 };
+use serde_json::value::RawValue;
 use ustr::Ustr;
 
 use crate::common::{
@@ -569,11 +570,38 @@ impl UserWsMessage {
 
     /// Parses a batch of user-channel JSON messages.
     ///
+    /// Elements carrying an unrecognized `event_type` are skipped so that a single unknown
+    /// message cannot discard the valid `order` and `trade` messages batched alongside it.
+    ///
     /// # Errors
     ///
     /// Returns [`serde_json::Error`] when `text` is not a valid user-message batch.
     pub fn parse_batch(text: &str) -> serde_json::Result<Vec<Self>> {
-        serde_json::from_str(text)
+        /// Reads only the tag, to classify an element before deserializing it.
+        #[derive(Deserialize)]
+        struct EventTypeTag {
+            event_type: Option<String>,
+        }
+
+        // Elements stay raw so the derived impl parses each one and rejects a duplicated
+        // `event_type`; `serde_json::Value` would silently keep the last occurrence.
+        let elements: Vec<&RawValue> = serde_json::from_str(text)?;
+        let mut messages = Vec::with_capacity(elements.len());
+        let mut skipped = 0usize;
+
+        for element in elements {
+            let tag: EventTypeTag = serde_json::from_str(element.get())?;
+            match tag.event_type.as_deref() {
+                Some(event_type) if !matches!(event_type, "order" | "trade") => skipped += 1,
+                _ => messages.push(serde_json::from_str(element.get())?),
+            }
+        }
+
+        if skipped > 0 {
+            log::debug!("Skipped {skipped} user WS message(s) with an unrecognized event_type");
+        }
+
+        Ok(messages)
     }
 
     fn parse_reordered(text: &str) -> serde_json::Result<Self> {
@@ -682,6 +710,26 @@ mod tests {
     fn load_text(filename: &str) -> String {
         let path = format!("test_data/{filename}");
         std::fs::read_to_string(path).expect("Failed to read test data")
+    }
+
+    /// An `auto_redeem` user-channel event as observed from the venue, which is undocumented and
+    /// not modelled by [`UserWsMessage`].
+    fn auto_redeem_element() -> serde_json::Value {
+        serde_json::json!({
+            "event_type": "auto_redeem",
+            "proxy_wallet": "0x0000000000000000000000000000000000000000",
+            "txn_hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+            "amount": "70",
+            "condition_id": "0xb88862256916cb0c82e72667bd43f3e6cfe94cd7d5cda0bf763ec82394021e07",
+            "question": "Bitcoin Up or Down - August 17, 9:35AM-9:40AM ET",
+            "slug": "btc-updown-5m-1786973700",
+            "neg_risk": false,
+            "timestamp": "1786974414920",
+            "position_id": "",
+            "outcome_index": 0,
+            "legs": 0,
+            "owner": "00000000-0000-0000-0000-000000000000",
+        })
     }
 
     #[rstest]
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @yashwardhan-gautam,

Thanks for the PR. This looks good from my side. The CLA needs to be signed before this can merge.

I noticed `parse_batch` now goes through `serde_json::Value`, which keeps the last `event_type` when the key is duplicated, so that element can parse or be skipped instead of erroring the way `parse` still does. If you're able, it's worth keeping that rejection so the malformed-element contract stays the same.

cjdsellers:
@yashwardhan-gautam, CLA-assistant is waiting on the commit email ygautam@deepsystems.com to be added to your GitHub account (https://github.com/settings/emails). Once that's linked, the signature should go through.

yashwardhan-gautam:
> @yashwardhan-gautam, CLA-assistant is waiting on the commit email [ygautam@deepsystems.com](mailto:ygautam@deepsystems.com) to be added to your GitHub account (https://github.com/settings/emails). Once that's linked, the signature should go through.

amended the author to use my personal email

yashwardhan-gautam:
> Hi @yashwardhan-gautam,
> 
> Thanks for the PR. This looks good from my side. The CLA needs to be signed before this can merge.
> 
> I noticed `parse_batch` now goes through `serde_json::Value`, which keeps the last `event_type` when the key is duplicated, so that element can parse or be skipped instead of erroring the way `parse` still does. If you're able, it's worth keeping that rejection so the malformed-element contract stays the same.

makes sense,  elements are now held as `Vec<&RawValue>` and each is deserialized by
the derived impl, so a duplicate is rejected exactly as before. The tag is read via
a small `#[derive(Deserialize)]` struct with `event_type: Option<String>`, which also
rejects duplicates.

cjdsellers (approved):
@yashwardhan-gautam, thanks for the quick rework. I've had another look at the latest changes: the duplicated `event_type` rejection is intact and pinned by the new test, and skipping unrecognized event types now keeps the valid batch elements as agreed in #4788.

This looks good from my side. I'll merge once the remaining CI goes green 👌

Issue #4782, opened 2026-08-17
[RFC] Preserve execution-client source on runtime reports


xxxxxx-oss:
## Before you proceed

- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I have no other RFC open. Please keep to one at a time so each proposal gets the attention it
      deserves.

## Context

Runtime `OrderStatusReport` and `FillReport` values currently enter the live runner as
`ExecutionEvent::Report(ExecutionReport)`. The event carries the report, but not the `ClientId` of
the execution client that emitted it.

This becomes ambiguous in a multi-client node. I reproduced the following case using only the
official `StubExecutionClient` and runtime event sender on `develop`:

- client A has the instrument venue as its primary venue;
- routing client B has a different primary venue and also handles the instrument venue;
- B's runtime path emits an unknown `OrderStatusReport` for its account;
- the engine materializes an external order, stores no client origin, and registers it with A via
  the venue/default fallback.

The focused regression test observes:

```text
actual:   (cache origin = None, routed client = VENUE-A)
expected: (cache origin = ROUTING-B, routed client = ROUTING-B)
```

An unknown `FillReport` reaches the same external-order materialization seam. Reproduced on `develop` at `510b5937` using only the official stub clients. The relevant runtime-report path was unchanged on current `develop` when this RFC was filed.

PR #4781 addresses a separate startup reconciliation path, where the node already knows which
client returned each `ExecutionMassStatus`. It does not change runtime report transport.

The desired runtime invariants are:

1. preserve the emitting execution client's trusted `ClientId` through dispatch;
2. validate the source client, account, and claimed order venue before mutating state;
3. write the source `ClientId` once as the cache origin when materializing an unknown order;
4. register that order with the exact source client rather than venue/default routing;
5. reject source/origin mismatches before applying a report;
6. apply the same source rule to late reports for known orders and to both order and fill paths.

## Considerations

The source should be transport metadata stamped at a trusted execution-client or emitter boundary.
It should not be inferred from `account_id` or instrument venue, and it should not become venue
evidence inside `OrderStatusReport` or `FillReport`.

There is a compatibility decision to make. `ExecutionEvent` is a public Rust enum and is not
`#[non_exhaustive]`; changing the `Report` payload or adding a variant affects downstream exhaustive
matches and all existing emitters/direct senders. An internal adapter-bound sourced channel avoids
changing report model types, but requires an emitter migration and a clear policy for legacy
unsourced reports.

[...]

PR #4797, opened 2026-08-18
Persist resolved execution client origins

Tickwhiz:
- [ ] I have read and followed
  and, if I used AI,
- [ ] I understand and can explain every submitted change and all information in this PR description

## Summary

Persist the execution client selected after routing and validation as a write-once cache origin
before an order reaches outbound transport. This prevents venue-routed orders without an explicit
`client_id` from remaining unowned and provides the prerequisite for strict runtime report source
validation in #4782. Order-list claims are validated and enqueued as one batch so conflicts or
persistence admission failures cannot partially update in-memory origins or reach transport.

## Related issues/PRs

Related to #4782

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Problem

An order can enter the cache before the execution engine resolves its final routing or default
client. If the command does not carry an explicit `client_id`, the selected client is not recorded
as the order's origin. This leaves later report validation without a trusted cached origin and can
cause subsequent multi-client routing to fall back to venue or default selection.

Order lists also require a batch contract: claiming each child independently could leave a
partially updated origin index when a later child conflicts or persistence cannot accept the
complete update.

## Solution

- Add `Cache::claim_order_clients` with write-once semantics: an unclaimed order is assigned,
  a matching claim is idempotent, and a conflicting or missing order is rejected.
- Validate the complete claim batch and enqueue its persistence operation before changing any
  in-memory origin.
- Claim the actual client selected after routing and venue validation, but before own-book or
  client transport side effects, for both single orders and order lists.
- Deny the submission without calling the execution client when a claim conflicts or the
  persistence command cannot be enqueued.
- Persist built-in Redis claims with one multi-field hash operation and PostgreSQL claims in one
  transaction, without changing the database schema.
- Record the new cache mutation in the EventStore recovery inventory without claiming that current
  command payloads can reconstruct a client selected only at runtime.

The asynchronous cache backends guarantee that the batch persistence command is accepted before
outbound transport. They retain their existing asynchronous failure model and do not claim durable
completion before transport.

## Stack and merge order

This is Stage 1 of the implementation agreed in #4782. It must land before the mandatory runtime
execution-report source migration because strict origin validation requires every newly submitted
order to have the resolved execution client recorded.

## Breaking change details

The new `CacheDatabaseAdapter::index_order_clients` method has a default implementation, so existing
third-party implementations remain Rust source-compatible. The default rejects non-empty claims,
however, because silently omitting origin persistence would violate the source-validation contract.

A custom cache database used for outbound live execution must therefore implement atomic batch
order-client indexing. Otherwise, affected submissions fail closed before transport. No migration
is required when no cache database is configured, and the built-in Redis and PostgreSQL adapters
implement the new operation.

## Documentation

- [x] Documentation changes follow the style guide (no documentation changed)
- [x] PyO3 stubs are unchanged because no bindings or wrapped Rust documentation changed

## Testing

- [ ] No logic changed

Validation covered:

- write-once, idempotent, conflicting, duplicate, and missing-order cache claims;
- venue-routed single-order origin persistence;
- atomic order-list claims and no-transport failure paths;
- persistence enqueue failures leaving the in-memory origin index unchanged;
- Redis and PostgreSQL persistence and reload behavior;
- the full relevant `common`, `execution`, `infrastructure`, and `event_store` Rust test suites;
- formatting, workspace pre-commit checks, Clippy, and diff validation.

### DIFF
--- a/crates/common/src/cache/database.rs
+++ b/crates/common/src/cache/database.rs
@@ -492,6 +492,19 @@ pub trait CacheDatabaseAdapter {
         position_id: PositionId,
     ) -> anyhow::Result<()>;
 
+    /// Indexes order-client mappings as one batch operation.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if batch order-client indexing is unsupported or cannot be enqueued.
+    fn index_order_clients(&self, claims: &[(ClientOrderId, ClientId)]) -> anyhow::Result<()> {
+        if claims.is_empty() {
+            return Ok(());
+        }
+
+        anyhow::bail!("Batch order-client indexing is not supported by this cache database")
+    }
+
     /// Updates actor state in the cache.
     ///
     /// # Errors
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -4657,6 +4657,73 @@ impl Cache {
         Ok(())
     }
 
+    /// Claims the execution-client origin for one or more cached orders.
+    ///
+    /// Claims are write-once: an unclaimed order is assigned to `client_id`, a matching existing
+    /// claim is idempotent, and a conflicting claim is rejected. The complete batch is validated
+    /// and its persistence command is successfully enqueued before any in-memory index is
+    /// changed.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if an order is not cached, an order is already claimed by another client,
+    /// the same order has conflicting claims in the batch, or persistence cannot be enqueued.
+    pub fn claim_order_clients(
+        &mut self,
+        claims: &[(ClientOrderId, ClientId)],
+    ) -> anyhow::Result<()> {
+        let mut requested = AHashMap::with_capacity(claims.len());
+        let mut ordered_claims = Vec::with_capacity(claims.len());
+
+        for (client_order_id, client_id) in claims {
+            if let Some(existing_client_id) = requested.get(client_order_id) {
+                if existing_client_id != client_id {
+                    anyhow::bail!(
+                        "Conflicting execution client claims for {client_order_id}: \
+                         {existing_client_id} and {client_id}"
+                    );
+                }
+                continue;
+            }
+
+            requested.insert(*client_order_id, *client_id);
+            ordered_claims.push((*client_order_id, *client_id));
+        }
+
+        let mut pending_claims = Vec::with_capacity(ordered_claims.len());
+        for (client_order_id, client_id) in ordered_claims {
+            if !self.orders.contains_key(&client_order_id) {
+                return Err(OrderLookupError::not_found(client_order_id).into());
+            }
+
+            match self.index.order_client.get(&client_order_id) {
+                Some(existing_client_id) if *existing_client_id == client_id => {}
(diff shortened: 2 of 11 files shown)


cjdsellers:
Hey @xxxxxx-oss,

Thank you for the PR. I've had an initial pass over this and the general shape looks good, with no immediate issues beyond one small test gap noted below. We'll hold the deeper review until after the next release, which is imminent.

**Applicability**:
This closes the unowned-origin gap for venue-routed orders and is the prerequisite for the strict report source validation in #4782. The added surface stays proportionate, reusing the existing async persistence queues, the Redis hash index, and the `order_event` client column with no schema change.

**Follow-ups**:
The Postgres conflict branch in `index_order_clients` (rejecting a claim when persisted rows already name a different client) isn't exercised, only the missing-order rollback is. If you get a chance, a test covering that rollback would lock in the intended behavior, but it can also land after merge.

xxxxxx-oss:
Thanks @cjdsellers — I added the requested PostgreSQL conflict rollback coverage in `020bcada35`.

The new `test_index_order_clients_conflict_rolls_back_batch` first updates an unclaimed order within the batch, then encounters an order already persisted with a different client. It verifies that the transaction rolls back the earlier update while preserving the existing client origin.

Validated on Linux with the full PostgreSQL integration test target (28/28), the focused infrastructure Clippy check, and nightly rustfmt. CI has been retriggered for the updated head.

xxxxxx-oss:
The refreshed CI run was cancelled because the pre-commit job remained in Common setup for six hours; the Run pre-commit step never started. Contributor permissions cannot rerun upstream workflows. Could someone with repository permissions please rerun the failed job?

cjdsellers (approved):

cjdsellers:
Hey @xxxxxx-oss,

Thanks again for the execution‑client origin work, including the PostgreSQL rollback coverage. I've followed this up on `develop` in 86eebbe014 because strict startup validation had no compatibility path for cache state written before origins were persisted.

The write‑once binding remains the direction we want. This is a tactical compatibility patch: missing or conflicting origins now produce one aggregated warning per condition, reconciliation continues, and any recorded origin remains unchanged. The conflict warning also states that this will become a startup error in a future release. This avoids blocking existing nodes while we prepare the migration path needed before strict enforcement.

I've also documented when origins are assigned and the asynchronous persistence guarantee, so the binding itself and the temporary startup policy are clear. The details are in [Execution‑client origins](https://github.com/nautechsystems/nautilus_trader/blob/develop/docs/concepts/reconciliation.md#execution-client-origins).

xxxxxx-oss:
Thanks for the follow-up and for documenting the migration policy. I reviewed `86eebbe` and
the execution-client origin documentation, and incorporated that boundary into the dependent
Stage 2 draft in #4818.

Startup mass-status preserves the aggregated warn-and-continue behavior for legacy missing or
conflicting origins without rewriting any recorded origin, while strict source and origin
validation remains enforced for runtime sourced reports. The documentation update in #4818 keeps
that temporary startup policy intact while describing the mandatory source-bearing runtime path.

PR #4799, opened 2026-08-18
Fix convert_stream_to_data silently skipping FeatherWriter custom data (#4607)

changshenhan:
Fixes the silent no-op where `convert_stream_to_data` converts nothing for custom data written by `FeatherWriter` (issue #4607).

## Problem

Two coordinated causes left the custom-data conversion branch unreachable, so conversion returned `Ok(())` while a subsequent `query_custom_data_dynamic` returned zero records:

1. **The `custom/` marker was destroyed.** `to_snake_case("custom/RustTestHashMapCustomData")` produced `custom_rust_test_hash_map_custom_data`, so `is_supported_stream_data_type` — which checks `starts_with("custom/")` — rejected it, and `convert_feather_batches_to_parquet` never reached its `strip_prefix("custom/")` branch.

2. **File discovery searched the wrong prefix.** `list_feather_files` looked beneath `{data_name}/`, but `FeatherWriter::get_writer_path_custom` physically writes beneath `data/custom/{TypeName}/...`.

The failure is silent by design — `convert_stream_to_data` deliberately treats an empty file list as a successful no-op — so no error surfaced.

## What

In `convert_stream_to_data` (`crates/persistence/src/backend/catalog.rs`):

- **Preserve the marker**: a `custom/`-prefixed `data_cls` keeps the `custom/{TypeName}` form instead of passing through `to_snake_case`.
- **Map to the physical layout**: custom data is discovered beneath `data/custom/{TypeName}/...`, matching the FeatherWriter's actual output.

The conversion side (`convert_feather_batches_to_parquet`, `identifier_from_batch_or_path`, `make_path_custom_data`) already handled the `custom/` layout correctly — it was simply never reached.

## Tests

New `test_convert_stream_to_data_converts_custom_data_written_by_feather_writer`: writes six `RustTestHashMapCustomData` records (three per identifier) through a real `FeatherWriter`, converts the stream, and asserts all six are queryable via `query_custom_data_dynamic`. This is the exact repro flow from the issue.

Note on the CLA: happy to sign the [Contributor License Agreement](CLA.md) via CLA Assistant on first PR.

### DIFF
--- a/crates/persistence/src/backend/catalog.rs
+++ b/crates/persistence/src/backend/catalog.rs
@@ -3947,13 +3947,29 @@ impl ParquetDataCatalog {
             return Ok(());
         }
 
-        // Convert data class name to filename (e.g., "quotes" -> "quotes")
-        // The data_cls should already be in the correct format (snake_case)
-        let data_name = to_snake_case(data_cls);
+        // Preserve the `custom/` marker that `is_supported_stream_data_type`
+        // checks: `custom/RustTestHashMapCustomData` must not become
+        // `custom_rust_test_hash_map_custom_data`, or the custom branch in
+        // `convert_feather_batches_to_parquet` is never reached and custom
+        // streams silently convert nothing.
+        let data_name = if let Some(type_name) = data_cls.strip_prefix("custom/") {
+            format!("custom/{type_name}")
+        } else {
+            to_snake_case(data_cls)
+        };
+
+        // `FeatherWriter` physically writes custom data beneath
+        // `data/custom/{TypeName}/...` (see `get_writer_path_custom`), while
+        // the logical name keeps the `custom/` marker, so the discovery prefix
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @changshenhan,

Thanks for the PR. The diagnosis is spot on: both causes check out, and preserving the `custom/` marker while discovering under `data/custom/...` is the same fix I confirmed on #4607.

I'm going to close this without merging though. We're about to land a new data catalog which replaces this persistence path, so we're not investing in or landing fixes to the current one. I deferred this same change on #4607 for that reason, and custom stream conversion will be verified as part of that work. In the meantime the limitation is documented, and writing custom data directly with `ParquetDataCatalog.write_custom_data` is the supported path.

I appreciate the effort here, and the thorough write-up and regression test made this quick to verify. Contributions will be very welcome once the new catalog lands.

PR #4800, opened 2026-08-19
Harden Polymarket pagination progress

seungpyoson:
## Summary

- Replace six endpoint-owned Polymarket HTTP pagination loops with one crate-private operation driver that owns request position and validates progress before accumulation.
- Reject CLOB cursor stalls and cycles, plus repeated full Data API pages using canonical semantic fingerprints.
- Preserve Gamma offset/cap behavior and Data API trade limit/offset-ceiling behavior while validating the actual cursors and offsets sent on the wire.

## Problem

Pagination policy was duplicated across CLOB, Gamma, and Data API operations. That made progress validation, cursor/offset ownership, accumulation ordering, and endpoint-specific stopping behavior easy to drift independently.

## Design

- `Paginator::run` owns fetch sequencing and makes pagination failures discard private reducer state rather than return an accumulated prefix.
- Cursor protocols validate missing, stalled, repeated, and cyclic cursors.
- Offset protocols detect identical full pages using order-independent, multiplicity-preserving fingerprints with canonical decimal encoding.
- Data API trade rows are checked against the requested condition before admission.
- Reducers retain Gamma windowing and trade-tick whole-operation sorting, filtering, ID assignment, limits, and venue-ceiling behavior.
- `WireExhausted` records only observed wire termination; it is not a source-completeness proof.

## Scope and non-goals

This is pagination progress and custody hardening only. It does not claim a stable venue snapshot, solve mutable-offset overlap or source completeness, add persistence or reconciliation authority, impose a new arbitrary global pagination bound, or expand the public API.

## Verification coverage

- CLOB missing/stalled/cyclic cursors and recurrence of the initial cursor.
- Data API cursor/offset transmission, repeated semantic pages, condition scope, decimal-scale canonicalization, and multiplicity preservation.
- Gamma offsets, caps, terminal behavior, and progress-error precedence.
- Data API trade sorting, filtering, IDs, result limits, and local/remote offset-ceiling behavior.
- Adapter integration tests, formatting, Clippy with warnings denied, and repository pre-commit checks were exercised during fork review.

### DIFF
--- a/crates/adapters/polymarket/src/http/clob.rs
+++ b/crates/adapters/polymarket/src/http/clob.rs
@@ -15,7 +15,10 @@
 
 //! Provides the HTTP client for the Polymarket CLOB REST API.
 
-use std::{collections::HashMap, result::Result as StdResult, str::from_utf8, sync::Arc};
+use std::{
+    collections::HashMap, convert::Infallible, result::Result as StdResult, str::from_utf8,
+    sync::Arc,
+};
 
 use nautilus_core::{
     consts::NAUTILUS_USER_AGENT,
@@ -45,6 +48,7 @@ use crate::{
             ClobBookResponse, ClobMarketResponse, FeeRateResponse, PolymarketOpenOrder,
             PolymarketOrder, PolymarketTradeReport, TickSizeResponse,
         },
+        pagination::{CollectAll, Completion, CursorProtocol, FetchOutcome, Paginator},
         query::{
             BalanceAllowance, BatchCancelResponse, CancelMarketOrdersParams, CancelResponse,
             ClobVersionResponse, GetBalanceAllowanceParams, GetOrdersParams, GetTradesParams,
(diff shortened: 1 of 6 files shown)


seungpyoson:
Exact-head review record

- Head: `f455da3e467cbc26ecbf22f516141a842afedae2`
- Merge base: `2297c36fa3f7c7816d161b95b4fad97f3b97ef58`
- Scope: 3 commits, 6 files; no production change after source-reviewed head `552aedc066ee471188330154dc632db962ccc2a9` (the final commit adds only fingerprint regression tests).
- Fresh `git diff --check upstream/develop...HEAD`: passed.
- Fresh `git show --check HEAD`: passed.
- Merge simulation against current upstream `develop` (`210b453afaae546b64a3fa937a0931fbf38013ca`): conflict-free.
- Source-identical head `552aedc06`: full `nautilus-polymarket` suite reported 1,601 passed / 1 ignored; HTTP integration target 230 passed / 1 ignored; Clippy with warnings denied, formatting, and pre-commit passed.
- Final tests-only delta: focused `request_trade_ticks_` group passed 21 tests; mutation checks independently demonstrated failure when Decimal normalization was removed and when descriptor multiplicity was deduplicated.
- Two independent read-only adversarial reviews of exact head `f455da3e4` returned `APPROVE` with no findings.

The full suite was not rerun after the tests-only final commit because the local runner quota rejected that command; this limitation is disclosed rather than inferred away. Advisory upstream CI is left to evaluate the exact PR head.

seungpyoson:
@cjdsellers  CI formatting is fixed in `d4fae6219`; no functional code changed.

Before expanding this PR's reviewed scope, I would like maintainer guidance on four edge cases raised during review:

1. Trade-page fingerprints intentionally exclude `proxyWallet` because it does not affect `TradeTick` output. Should wallet identity nevertheless participate in pagination progress?
2. Condition IDs are compared case-insensitively, but transaction hashes enter fingerprints as received. Does the Data API guarantee canonical hash casing, or should the adapter normalize hashes before fingerprinting?
3. The `/positions` API documents `offset <= 10,000`. A very large account could make the paginator request beyond that. Should this fail closed, be reduced with server-side filtering, or use another explicit outcome? Returning partial positions would be unsafe for reconciliation.
4. The remote historical-offset test proves 429 classification, while a separate local-ceiling test proves non-empty prefix preservation. Would you prefer a combined regression test where the remote ceiling occurs after successful pages?

I have not changed behavior for these items. Should any block this PR, or should they be handled in a separate focused issue or PR?

cjdsellers (approved):
Hey @seungpyoson,

Thank you for the PR, and for framing these edge cases so precisely. I've had a thorough look at the change and this looks good to merge. On your questions: none of them block, and 3 is the only one I'd definitely take further.

1. Keep `proxyWallet` out of the fingerprint. Progress should track the economic content of the page, and including wallet identity would let a repeated page escape detection whenever only the metadata moved, which is exactly what your restamped-metadata test pins down.
2. Hashes entering as received can only weaken detection, never produce a false repeated-page error, so this is safe as-is. I don't know whether the Data API guarantees canonical casing; if it ever turns out to vary, normalizing before fingerprinting is a cheap follow-up.
3. Agreed on both counts for `/positions`: partial positions would be unsafe for reconciliation, and fail closed is the right outcome. The old loop had the same unbounded behavior, so this isn't a regression from your change. Let's handle it in a separate focused issue.
4. The split tests each prove one property cleanly, so nothing more needed here. A combined test where the remote ceiling follows successful pages would lock in the prefix-preservation ordering, if you're able to add one in a follow-up.

I'll merge this as-is and we can pick up 3, and optionally 4, from there.

seungpyoson:
Thanks — this resolves all four points. I opened #4810 to track the `/positions` offset ceiling as a focused fail-closed change.

I’ll leave `proxyWallet` exclusion and hash handling unchanged. I also won’t treat the combined historical-offset test as required work; the split coverage remains as reviewed.

PR #4801, opened 2026-08-19
Suppress the data command for a duplicate actor subscription

folknor:
# Suppress the data command for a duplicate actor subscription

`DataActorCore` guards every subscription family against subscribing to the same topic twice, but the guard's result never reaches the caller that sends the command.

## The discarded guard

`add_quote_subscription` returns early with a warning when `quote_handlers` already holds the topic, and it returns `()`. Its caller `subscribe_quotes` cannot tell the difference, so it builds a `SubscribeQuotes` and calls `send_data_cmd` regardless. A second subscribe to a topic already subscribed therefore registers no second handler, logs the duplicate, and still sends a subscribe command downstream. The shape repeats across every family: 20 `add_*_subscription` helpers and 22 command-producing callers, including the six DeFi callers in `crates/common/src/defi/data_actor.rs`, whose helpers live in `actor/data_actor.rs` while their `DataCommand::DefiSubscribe` sends do not.

The redundant command is absorbed downstream rather than acted on - `DataClientAdapter` calls `track_subscribe`, which returns `false` for an instrument already in its tracking set and never reaches the underlying client. What remains is command construction, bus traffic, engine dispatch and an inflated `command_count`, from an actor that has already decided the operation is a duplicate.

Each helper now returns whether it added a handler, and each caller that sends a command returns when it did not. `subscribe_queue_state` and `subscribe_socket_state` ignore the result, since they subscribe locally and send nothing. `subscribe_signal` is unchanged: its inline guard already returns before any later action.

`OrderEmulator::subscribe_quotes_for_instrument` is the existing form of this in the tree, returning before `send_data_command` for a duplicate.

The unsubscribe helpers have the same structure, but their contract differs - most ignore a missing handler silently, and the engine applies subscriber counts and book ownership rules on that path. They are left for separate work rather than half-converted here.

## Testing

`test_duplicate_book_depth10_subscription_sends_one_command` and `test_duplicate_block_subscription_sends_one_command` capture the data command endpoint, subscribe twice, and assert exactly one command of the expected type reaches it; the second covers the DeFi callers, which live in a different file and compile only under the `defi` feature. Both assertions fail against the unfixed code, which emits two commands.

The existing `test_duplicate_book_depth10_subscription_delivers_once` and `test_duplicate_subscribe_custom_data` are unchanged and still pass - they pin the handler side of the same guard, which this does not move.

### DIFF
--- a/crates/common/src/actor/data_actor.rs
+++ b/crates/common/src/actor/data_actor.rs
@@ -3035,26 +3035,27 @@ impl Debug for DataActorCore {
 }
 
 impl DataActorCore {
-    /// Adds a subscription handler for the `topic`.
+    /// Adds a subscription handler for the `topic` and returns whether it was added.
     ///
-    //// Logs a warning if the actor is already subscribed to the topic.
+    /// Logs a warning if the actor is already subscribed to the topic.
     pub(crate) fn add_subscription_any(
         &mut self,
         topic: MStr<Topic>,
         handler: ShareableMessageHandler,
         priority: Option<u32>,
-    ) {
+    ) -> bool {
         let pattern: MStr<Pattern> = topic.into();
         if self.topic_handlers.contains_key(&pattern) {
             log::warn!(
                 "Actor {} attempted duplicate subscription to topic '{topic}'",
                 self.actor_id,
             );
-            return;
+            return false;
         }
 
         self.topic_handlers.insert(pattern, handler.clone());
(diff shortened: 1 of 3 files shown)


folknor:
CI caught a regression this introduces, and the reasoning I gave in the description is wrong.

My mistake, and the PR is withdrawn.

PR #4802, opened 2026-08-19
Reject a fill whose instrument does not match its position

folknor:
# Reject a fill whose instrument does not match its position

`Position::new` validates that the opening fill belongs to the instrument, but every later fill reaches `Position::apply` with no identity check at all.

## The routing

`determine_position_id` prefers the position ID cached against the fill's `client_order_id` over the one the fill carries, warning on a mismatch. `handle_position_update` then fetches the position by that ID alone, and `update_position` calls `position.apply(fill)`. `OrderAny::apply` guards `client_order_id` and `strategy_id`, but nothing on this path compares the instrument, so a fill for one instrument applied to a position for another is priced with that position's precision, multiplier, currencies, and PnL rules, and its quantity and realized PnL move accordingly.

The engine now rejects the fill where it can still do nothing: after the effective position ID is assigned and before the cached order is updated, the fill is forwarded to the portfolio, the update/flip/reopen branch is chosen, or any event is published. `handle_leg_fill_without_order` carries the same guard, since an orderless leg fill reaches `handle_position_update` without passing through the order-associated path, and in `HEDGING` the supplied position ID is accepted when no same-instrument correlation is found.

Only `instrument_id` is compared. A position's instrument never changes, so there is no legitimate mismatch. `account_id` and `strategy_id` both have one and are deliberately excluded: netting position IDs are `{instrument_id}-{strategy_id}`, so two accounts trading one instrument under one strategy share a position ID; and external order claims can be handed to a successor strategy while the predecessor's positions stay cached, so a later venue fill legitimately carries the new strategy against them. Rejecting on either would drop venue-confirmed fills. Both exclusions are recorded on the function and pinned by tests.

The fill is dropped rather than rerouted - an unroutable fill has no correct destination, but nothing then records it against any position. Direct `Position::apply` callers also stay unguarded: event-store replay, Redis and Postgres reconstruction, and the Python wrapper. Closing those means giving `Position::apply` an observable outcome so callers stop building events after a rejection, which changes every Rust caller and the pyo3 binding, so it is not attempted here.

If you would rather have that shape, or would rather this were rejected earlier or later on the path, I am happy to redo it.

## Testing

`test_fill_with_foreign_instrument_is_rejected` asserts the order stays `Accepted` at zero filled quantity, the position keeps its quantity and trade IDs, and no order or position event is published. `test_leg_fill_without_order_with_foreign_instrument_is_rejected` covers the orderless path: an XCME futures leg fill carrying a cached AUD/USD position ID leaves that position untouched and creates no second position. Both fail against the unfixed code - the second with the FX position's quantity moved from 1 to 2 by the futures fill.

`test_fill_with_foreign_account_or_strategy_is_still_applied` pins the two exclusions from the other side, asserting a foreign account and a foreign strategy are both applied; restoring either comparison fails it. `test_fill_with_matching_position_identity_is_unchanged` and `test_oversized_foreign_fill_does_not_flip_position` cover the unchanged success path and the flip branch.

### DIFF
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -2684,6 +2684,10 @@ impl ExecutionEngine {
                 let mut fill = fill.clone();
                 fill.position_id = Some(position_id);
 
+                if !self.validate_fill_for_position(position_id, &fill) {
+                    return;
+                }
+
                 let validation = if apply_position {
                     self.validate_fill_for_order(&order_before_fill, &fill)
                 } else {
@@ -2846,6 +2850,11 @@ impl ExecutionEngine {
         let oms_type = self.determine_oms_type(&fill);
         let position_id = self.determine_leg_fill_position_id(&fill, oms_type);
         fill.position_id = Some(position_id);
+
+        if !self.validate_fill_for_position(position_id, &fill) {
+            return;
+        }
+
         let duplicate_position_fill = self.position_contains_trade_id(position_id, fill.trade_id);
 
         let event = OrderEventAny::Filled(fill.clone());
@@ -3100,6 +3109,35 @@ impl ExecutionEngine {
         }
 
         position_id
+    }
+
+    /// Returns whether `fill` may be applied to the position assigned to `position_id`.
+    ///
+    /// Only `instrument_id` is compared. A position's instrument never changes, and a fill
+    /// for another instrument would be priced with this position's precision, multiplier,
+    /// currencies, and PnL rules.
+    ///
+    /// `account_id` and `strategy_id` are deliberately NOT compared, because each has a
+    /// legitimate mismatch path. Netting position IDs are `{instrument_id}-{strategy_id}`,
+    /// so two accounts trading one instrument under one strategy share a position ID.
+    /// External order claims can be handed to a successor strategy while the predecessor's
+    /// positions stay cached, so a later venue fill can carry the new strategy against them.
+    fn validate_fill_for_position(&self, position_id: PositionId, fill: &OrderFilled) -> bool {
+        let Some(position) = self.cache.borrow().position_owned(&position_id) else {
+            return true;
+        };
+
+        if position.instrument_id != fill.instrument_id {
+            log::error!(
+                "Cannot apply fill {} to position {position_id}: instrument_id mismatch, expected={}, received={}",
+                fill.trade_id,
+                position.instrument_id,
+                fill.instrument_id
+            );
+            return false;
+        }
+
+        true
     }
 
     fn determine_hedging_position_id(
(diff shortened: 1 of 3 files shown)


folknor:

cjdsellers:
Hi @folknor,

Thanks for the PR. I found one blocking execution-algorithm path:

- For a HEDGING child fill without a cached position mapping, `determine_position_id` can accept the supplied foreign position ID and assign it to the primary order and cache index before `validate_fill_for_position` runs. The fill is then rejected, but the primary and its siblings remain routed to that foreign position, so later valid fills are rejected as well. The rejection needs to leave both the primary order and the position index unchanged.

I also found two smaller points:

- `validate_fill_for_position` uses `position_owned` only to read `instrument_id`, which clones the position's growing fill and replay history on every fill. This check should avoid cloning the full position.
- The execution docs should state that an instrument-mismatched fill is dropped with an error log while the order remains non-terminal.

folknor:
Thank you for the review. All three points are addressed.

## The rejection no longer mutates routing

You were right about the ordering. `determine_position_id` resolved an ID and then propagated it to the primary and the cache index, so after a rejection every sibling resolving through that primary took the cached-mapping early return, resolved to the same rejected ID, and was rejected in turn.

The function is now split into resolve, validate, then propagate, returning `Option<PositionId>`; `None` means nothing was written. Validation gates on the *resolved* ID rather than on `fill.position_id`, because `determine_hedging_position_id` also adopts an ID from an already stamped sibling via `orders_for_exec_spawn` - so a check of the incoming fill alone would leave that route open. Both the cached-mapping branch and the freshly resolved branch pass through it.

Undoing the writes after the fact is not available: `Cache::add_position_id` has no removal counterpart, it touches `order_position`, `position_orders` and the position indexes plus the backing database, and the primary-order write and the cache write are not atomic with each other.

The leg-fill-without-order site is unchanged. `determine_leg_fill_position_id` only reads the index and selects an ID - it never stamps an order or calls `add_position_id` - and its validation already runs before `handle_position_update` and before publication.

## The validation no longer copies the position

Changed to `position_ref`, which reads `instrument_id` through a borrow.

## Docs

`docs/concepts/execution.md` now states that a fill resolving to a cached position for a different instrument is dropped with an error log, and that the order remains non-terminal so a later valid fill applies.

## Testing

`test_exec_algorithm_rejected_foreign_fill_does_not_route_spawn` builds a HEDGING execution-algorithm spawn on GBPUSD, caches a position on AUDUSD, and sends a child fill carrying the AUDUSD position ID. It asserts the primary's `position_id()` and the cache mapping for the primary are both still unset, the child is still `Accepted`, and the foreign position did not take the trade ID. It then sends a valid fill for the same child and asserts it routes the primary, fills the child, and opens a position on GBPUSD - that second half is what pins the stranding rather than only the immediate write.

Against the unfixed code the test fails at the first assertion with `Some("P-FOREIGN-INSTRUMENT")` against `None`.

cjdsellers (approved):
Hey @folknor,

Thanks for the careful follow-up here. I've had another look at the latest changes, including the execution-algorithm rejection and the later valid sibling fill. The blocker and both follow-ups are addressed, and this looks good to merge :ok_hand:

PR #4803, opened 2026-08-19
Report a strategy left running after backtest end

folknor:
# Report a strategy left running after backtest end

`BacktestEngine::end` can return with a managed-stop strategy still `RUNNING`, and nothing reports it.

## The mechanism

`end_with_result` performs its only timer flush, then calls `stop_trader()`. That reaches `Trader::stop_components`, which invokes each strategy's stop wrapper before `Component::stop`. With `manage_stop` enabled the strategy is still `Running` at that point, so `Strategy::stop` sets `pending_stop`, calls `market_exit()`, and returns `false` - meaning stop is deferred until the exit completes - and `Component::stop` is never called. `market_exit` passes its own `Running` guard and schedules the recurring retry timer.

Nothing after that fires timer callbacks. The command drain and `settle_venues` do not, and the flush already ran. So the exit is never evaluated once, and the run ends with the strategy `Running`, `pending_stop`, `is_exiting`. This happens with no open orders or positions, because completion is only discovered by the first timer callback.

## Why this only reports it

A second flush at the same boundary fires nothing: `market_exit` schedules at terminal clock plus interval, which is past `flush_ts` by construction. Extending the horizon would fire every other component timer in the added window - bars, GTD alerts, actor and strategy periodics - and reverse the graceful-stop cap at `last_ns` that exists to suppress exactly that. Moving `stop_trader()` before the flush is also unsafe, since actors and execution algorithms cancel their timers during stop, which is what the flush comment says it must precede.

So the change is one observation: after the stop sequence, any registered strategy whose state resolves to `Running` is reported at ERROR. The message states the observed state and no cause, because a strategy can also be left `Running` when an earlier component stop failure makes `Trader::stop_components` return before reaching it - it propagates with `?`, and the kernel logs and swallows the result. Attributing the managed exit would be wrong on that path.

Control flow is unchanged: `end_with_result` returns exactly what it returned before in every case.

## Scope

Completing an exit initiated at `end_ns` would mean allowing managed exits simulated time beyond the requested end - a bounded phase between requesting exits and stopping components, which needs a lifecycle seam that does not exist today. That is a separate decision and is not part of this change.

The `manage_stop` section of the strategy guide now records the boundary: the first completion check falls past the requested end, so the strategy can end the run `RUNNING` and be reported at `ERROR` while already flat.

## Testing

`test_end_reports_strategy_stranded_by_managed_stop` runs a `manage_stop` strategy holding a position to completion, asserts the run returns `Ok`, and asserts the strategy is still `Running` afterwards - the stranding as an executable fact rather than a code trace. There is no differential claim here: this changes no behaviour, so that assertion holds against the unmodified code as well, which is the point of it.

`test_end_reports_no_cleanly_stopped_strategies` covers both negative cases - a `manage_stop` disabled strategy stops normally and is not reported, and an engine with no strategies reports nothing.

### DIFF
--- a/crates/backtest/src/engine.rs
+++ b/crates/backtest/src/engine.rs
@@ -30,8 +30,8 @@ use nautilus_common::{
     actor::{DataActor, DataActorNative},
     cache::Cache,
     clock::{Clock, TestClock},
-    component::Component,
-    enums::LogColor,
+    component::{Component, component_state},
+    enums::{ComponentState, LogColor},
     log_info,
     logging::{
         logging_clock_set_realtime_mode, logging_clock_set_static_mode,
@@ -54,7 +54,7 @@ use nautilus_model::{
     accounts::{Account, AccountAny},
     data::{Data, HasTsInit},
     enums::{AccountType, AggregationSource, BookType},
-    identifiers::{AccountId, ClientId, InstrumentId, TraderId, Venue},
+    identifiers::{AccountId, ClientId, InstrumentId, StrategyId, TraderId, Venue},
     instruments::{Instrument, InstrumentAny},
     position::Position,
     types::Price,
@@ -971,6 +971,12 @@ impl BacktestEngine {
 
         self.settle_venues(ts_now);
 
+        for strategy_id in self.running_strategy_ids() {
+            log::error!(
+                "Strategy {strategy_id} is still RUNNING after the backtest end sequence; its stop did not complete",
+            );
+        }
+
         let save_result = self.kernel.save_trader_state();
         self.kernel.portfolio.borrow_mut().finalize_equity_curve();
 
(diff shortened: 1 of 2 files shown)


folknor:

cjdsellers:
Hey @folknor,

Thanks for the PR and the careful analysis here. Let's keep this scoped to reporting the stranded state. I do not think this PR should advance simulated time beyond the requested end or add a grace phase; that would need a separate lifecycle decision.

This looks good from my side. If you're able, could you add a short note to the `manage_stop` section of the strategy guide explaining the backtest boundary? The first exit check falls beyond the requested horizon, so the strategy can remain `RUNNING` and be reported at ERROR even when already flat.

folknor:
Thank you - scoped to the report, with no time advance and no grace phase.

The `manage_stop` section of the strategy guide now carries the boundary: `market_exit()` schedules its first completion check `market_exit_interval_ms` after the current time, which falls past the requested end and after the engine's final timer flush, so no callback runs to observe completion and the strategy ends the run `RUNNING` and reported at `ERROR` while already flat.

I also replaced the description's open question with a scope note pointing at that lifecycle decision as separate work.

cjdsellers (approved):
Hey @folknor,

Thanks for adding the `manage_stop` boundary note and tightening the scope in the description. I've had another look at the latest changes, and this looks good to merge.

PR #4804, opened 2026-08-19
Remove remaining purge timer conversion panics

folknor:
A follow-up to the purge retention validation in #4784, which closed the unvalidated-config route for the retention fields only.

## The interval conversion

`ExecutionEngine::start_purge_timers` converts each of the three purge intervals through `mins_to_nanos`, which is `checked_mins_to_nanos(...).expect(...)` and panics above 307,445,734 minutes. `ExecutionEngineConfig::validate` rejects such a value, but `ExecutionEngine::new` stores the supplied config without validating it and the fields are public, so a struct literal or a direct `new` carries an unvalidated `Option<u32>` straight into the timer setup.

Each timer now converts with `checked_mins_to_nanos` and, on overflow, logs an error naming the field and the rejected value and skips that timer alone.

## The timer registration

A second panic sits in the same function for intervals the conversion accepts. `validate_and_prepare_timer` computes the first event time as `start_time_ns.checked_add(interval_ns)` and returns an error on overflow; `start_time_ns` defaults to the current clock, so an interval below the conversion ceiling but near it fails registration on any nonzero clock, and `set_timer_ns(...).expect(...)` turned that into an abort.

The three `expect` calls become an error log naming the timer and the underlying error, again skipping only that timer. `start_purge_timers` still returns `()`: one invalid purge timer should not stop the other two or the execution clients from starting.

Behaviour is unchanged for every interval that previously registered - the zero-interval and already-registered guards, the buffer conversion, the callbacks and the timer arguments are all untouched.

## Testing

`test_start_purge_timers_overflowing_interval_skipped` configures `u32::MAX` minutes for the closed-orders interval and asserts that timer is absent while the other two register. `test_start_purge_timers_first_event_time_overflow_skipped` uses 307,445,734 minutes - the largest input that converts - on a clock set to 1,000,000,000,000 ns, which exceeds the remaining headroom, and pins the same partial-registration outcome. Both use `TestClock` with an explicitly set time, so neither depends on wall-clock time.

Both new tests panic against the unfixed implementation, at the conversion and at the removed `expect` respectively - the two distinct routes this change closes.

### DIFF
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -63,7 +63,7 @@ use nautilus_common::{
 };
 use nautilus_core::{
     UUID4, UnixNanos, WeakCell,
-    datetime::{mins_to_nanos, mins_to_secs, secs_to_nanos},
+    datetime::{checked_mins_to_nanos, mins_to_secs, secs_to_nanos},
 };
 use nautilus_model::{
     accounts::Account,
@@ -808,10 +808,6 @@ impl ExecutionEngine {
     }
 
     /// Starts the purge timers if configured.
-    #[expect(
-        clippy::missing_panics_doc,
-        reason = "timer registration is not expected to fail"
-    )]
     pub fn start_purge_timers(&mut self) {
         if let Some(interval_mins) = self
             .config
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for following up on #4784 with this. Skipping just the misconfigured timer while the other two and the clients still start is the right behavior, and the two tests pin each route well.

Longer term, we can validate these fields further upstream, at engine or kernel construction, so this path is rarely if ever hit. That also keeps config errors off the panic route across the Python boundary, where they should surface as raised exceptions instead. That's a separate change though - no need to expand this PR.

This looks good to merge.

PR #4809, opened 2026-08-20
Validate execution manager lookback minutes at construction

folknor:
A follow-up to the purge-timer conversion work in #4784 and #4804, taking up the suggestion on that thread to validate these fields at construction rather than at the conversion site.

## Unvalidated lookback minutes reach a panicking conversion

`ExecutionManagerConfig` carries no validation of any kind - its only inherent method is `with_trader_id` - and `ExecutionManager::new` stores the config verbatim. Two of its public `Option<u64>` fields then reach conversions that panic on overflow: `open_check_lookback_mins` reaches `mins_to_nanos` in the open-order report check, which panics above 307,445,734 minutes, and `lookback_mins` reaches `Duration::from_mins`, which panics above 307,445,734,561,825,860.

The supported node route is unaffected: `LiveExecEngineConfig::validate` already rejects an unrepresentable value with `checked_mins_to_nanos`, and the conversion into `ExecutionManagerConfig` widens `Option<u32>` to `Option<u64>`. What remains exposed is direct construction of `ExecutionManagerConfig`, which is public with public fields.

`ExecutionManagerConfig::validate` now collects every offending field through `ConfigErrorCollector`, and `ExecutionManager::new` enforces it by returning `ConfigResult<Self>`. Unlike `ExecutionEngineConfig` and `RiskEngineConfig`, which validate inside a derived-builder `build()`, `ExecutionManagerConfig` has no builder and is commonly constructed directly, so the constructor is the only boundary covering every construction path. This does change a public constructor signature; both call sites, in `LiveNode::new` and `LiveNodeBuilder`, already return `anyhow::Result` and propagate with `?`.

`position_check_lookback_mins` is deliberately not validated: it has no read site, so there is no target representation whose bound could honestly be imposed. Its eventual consumer should introduce the constraint.

## Testing

`test_new_validates_open_check_lookback_mins_boundaries` and `test_new_validates_reconciliation_lookback_mins_boundaries` pin the maximum valid value and the first invalid value for each field, asserting the typed `ConfigError::Range` and its field name. `test_new_reports_every_invalid_lookback_field` sets both fields out of range and asserts a `ConfigError::Multiple` carrying both field names in declaration order.

Each test was checked against a mutation that removes the property it claims. Discarding the validation result in the constructor, which leaves the code compiling, fails both boundary tests by assertion. Making the second field check conditional on no earlier error, recreating fail-fast, fails only the aggregation test, with `Range` in place of `Multiple`.

### DIFF
--- a/crates/live/src/execution/manager.rs
+++ b/crates/live/src/execution/manager.rs
@@ -27,6 +27,7 @@ use nautilus_common::{
     cache::Cache,
     clients::{DEFAULT_POSITION_RECONCILIATION_TOLERANCE, ExecutionClient},
     clock::Clock,
+    config::{ConfigError, ConfigErrorCollector, ConfigResult},
     enums::{LogColor, LogLevel},
     live::dst,
     log_info,
@@ -44,7 +45,7 @@ use nautilus_common::{
 };
 use nautilus_core::{
     UUID4, UnixNanos,
-    datetime::{mins_to_nanos, mins_to_secs},
+    datetime::{checked_mins_to_nanos, checked_mins_to_secs, mins_to_nanos, mins_to_secs},
 };
 use nautilus_execution::{
     engine::ExecutionEngine,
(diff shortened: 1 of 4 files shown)


folknor:
@cjdsellers I noticed I can no longer ask for review - when a PR is ready should I simply do nothing?

That's fine by me, I just want to know which workflow you prefer.

cjdsellers:
@folknor, noted - unsure what might have happened there. I'm continuously monitoring the repo including PRs, so if you wait then I will eventually circle around to a review :pray:

cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I've had a look through this, and putting the guard in `ExecutionManager::new` closes the direct-construction panic path cleanly. The bounds match the downstream conversions, both node paths propagate the new `ConfigResult<Self>`, and the tests pin the boundary and aggregation behavior. This looks good to merge 👍

PR #4818, opened 2026-08-20
Preserve execution client source for Bybit reports

Tickwhiz:
# Pull Request

- [ ] I have read and followed
  and, if I used AI,
- [ ] I understand and can explain every submitted change and all information in this PR description
  this draft
  limitation below

## Summary

Add an adapter-scoped `use_sourced_execution_events` opt-in for Bybit, fixed at client construction
and defaulting to `false`. When enabled, the complete Bybit `ExecutionEvent` stream uses a
source-bound ingress; legacy and sourced lanes are serviced fairly with per-lane FIFO,
exactly-one-lane delivery, and no fallback to the legacy lane. Only sourced `OrderStatusReport` and
`FillReport` values use strict source, account, handled-venue, cached-origin, and order-ID validation
with exact-source external-order registration; other variants retain legacy semantics, while the
public event shape, existing sender/emitter APIs, raw topics, and capture payloads remain unchanged.

## Related issues/PRs

Related to #4782.

Builds on #4797 and implements the Bybit pilot requested in the maintainer's
[scope review](https://github.com/nautechsystems/nautilus_trader/pull/4818#issuecomment-5385114946),
with the transport contract confirmed in the
[follow-up](https://github.com/nautechsystems/nautilus_trader/pull/4818#issuecomment-5392028856).

The pre-split reference implementation remains available on the
[`source-runtime-execution-reports-reference`](https://github.com/xxxxxx-oss/nautilus_trader/tree/source-runtime-execution-reports-reference)
branch as requested.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not applicable. The opt-in defaults to `false`. Existing `ExecutionEvent` variants, sender/emitter
method signatures, the seven-field `AsyncRunnerChannels` shape, `take_channels(self)`, raw report
topics, bare capture payloads, and legacy-client behavior remain unchanged.

## Documentation

No documentation files changed. The generated Bybit Python stub includes the additive opt-in
configuration.

## Testing

Ubuntu validation completed across the pilot implementation; the final submitted patch also passed
the full pre-commit suite:

- Execution engine integration tests: 262 passed; execution library tests: 598 passed.
- Focused live sourced-ingress tests: 14 passed; Bybit library tests: 513 passed.
- Common message-bus tests: 62 passed.
- Bybit Python config and binding tests passed with Python 3.12, 3.13, and 3.14.
- `make py-stubs` passed and produced the committed stub.
- `UV_PYTHON=3.13 make pre-commit` passed in full.

Coverage includes the original two-client routing case for both order and fill reports,
exact-source external-order materialization and registration, the strict rejection matrix,
post-publication re-entrant validation, preflight-before-deduplication, database persistence failure
after an in-memory cache commit, per-lane FIFO, fair progress, startup/pending preservation, and
closed sourced-lane behavior without legacy fallback. Pull-request CI independently revalidates the
submitted head.

### DIFF
--- a/crates/adapters/bybit/src/config.rs
+++ b/crates/adapters/bybit/src/config.rs
@@ -230,6 +230,9 @@ pub struct BybitExecutionClientConfig {
     /// coverage as unavailable.
     #[builder(default)]
     pub use_spot_position_reports: bool,
+    /// Whether to route this client's execution events through the source-bound ingress.
+    #[builder(default)]
+    pub use_sourced_execution_events: bool,
     /// Whether to automatically repay SPOT margin borrows after BUY orders tracked by
     /// this client and reported on the standard `execution` channel (not `execution.fast`)
     /// fully fill.
@@ -262,6 +265,7 @@ nautilus_core::impl_pyo3_config_getters!(BybitExecutionClientConfig {
     recv_window_ms: u64,
     account_id: Option<AccountId>,
     use_spot_position_reports: bool,
+    use_sourced_execution_events: bool,
     auto_repay_spot_borrows: bool,
     margin_mode: Option<BybitMarginMode>,
     transport_backend: TransportBackend,
@@ -436,6 +440,7 @@ mod tests {
         assert_eq!(config.product_types, vec![BybitProductType::Linear]);
         assert_eq!(config.http_timeout_secs, 60);
         assert_eq!(config.heartbeat_interval_secs, 20);
+        assert!(!config.use_sourced_execution_events);
     }
 
     #[rstest]
@@ -525,6 +530,18 @@ http_timeout_secs = 45
             expected.heartbeat_interval_secs,
         );
         assert_eq!(config.recv_window_ms, expected.recv_window_ms);
+        assert_eq!(
+            config.use_sourced_execution_events,
+            expected.use_sourced_execution_events,
+        );
         assert_eq!(config.transport_backend, expected.transport_backend);
+    }
+
+    #[rstest]
+    fn test_exec_config_toml_enables_sourced_execution_events() {
+        let config: BybitExecutionClientConfig =
+            toml::from_str("use_sourced_execution_events = true").unwrap();
+
+        assert!(config.use_sourced_execution_events);
     }
 }
--- a/crates/adapters/bybit/src/execution.rs
+++ b/crates/adapters/bybit/src/execution.rs
@@ -874,9 +874,14 @@ impl ExecutionClient for BybitExecutionClient {
                     account_state.balances.len()
                 );
             }
-            self.emitter
-                .try_send_bootstrap_account_state(account_state)
-                .context("failed to emit Bybit bootstrap account state")?;
+
+            if self.use_sourced_execution_events {
+                self.emitter
+                    .try_send_bootstrap_account_state(account_state)
+                    .context("failed to emit Bybit bootstrap account state")?;
+            } else {
+                self.emitter.send_account_state(account_state);
+            }
 
             self.await_account_registered(30.0).await?;
 
--- a/crates/adapters/bybit/src/execution.rs
+++ b/crates/adapters/bybit/src/execution.rs
@@ -28,11 +28,15 @@ use futures_util::{StreamExt, pin_mut};
 use nautilus_common::{
     clients::ExecutionClient,
     live::runner::get_exec_event_sender,
-    messages::execution::{
-        BatchCancelOrders, CancelAllOrders, CancelOrder, GenerateFillReports,
-        GenerateFillReportsBuilder, GenerateOrderStatusReport, GenerateOrderStatusReports,
-        GenerateOrderStatusReportsBuilder, GeneratePositionStatusReports, ModifyOrder,
-        QueryAccount, QueryOrder, SubmitOrder, SubmitOrderList,
+    messages::{
+        ExecutionEvent,
+        execution::{
+            BatchCancelOrders, CancelAllOrders, CancelOrder, GenerateFillReports,
+            GenerateFillReportsBuilder, GenerateOrderStatusReport, GenerateOrderStatusReports,
+            GenerateOrderStatusReportsBuilder, GeneratePositionStatusReports,
+            GeneratePositionStatusReportsBuilder, ModifyOrder, QueryAccount, QueryOrder,
+            SubmitOrder, SubmitOrderList,
+        },
     },
 };
 use nautilus_core::{
@@ -43,6 +47,7 @@ use nautilus_core::{
 use nautilus_live::{
     ExecutionClientCore, ExecutionEventEmitter, SocketControl,
     execution::failure::CommandFailure,
+    runner::{SourcedExecutionEventSink, get_sourced_exec_event_sink},
     task::{TaskGroup, TaskGroupGuard},
 };
 use nautilus_model::{
@@ -99,6 +104,7 @@ pub struct BybitExecutionClient {
     core: ExecutionClientCore,
     clock: &'static AtomicTime,
     config: BybitExecutionClientConfig,
+    use_sourced_execution_events: bool,
     emitter: ExecutionEventEmitter,
     http_client: BybitHttpClient,
     ws_private: BybitWebSocketClient,
@@ -176,6 +182,7 @@ impl BybitExecutionClient {
         }
         ws_trade.set_recv_window_ms(config.recv_window_ms);
 
+        let use_sourced_execution_events = config.use_sourced_execution_events;
         let clock = get_atomic_clock_realtime();
         let emitter = ExecutionEventEmitter::new(
             clock,
@@ -192,6 +199,7 @@ impl BybitExecutionClient {
             core,
             clock,
             config,
+            use_sourced_execution_events,
             emitter,
             http_client,
             ws_private,
@@ -366,6 +374,23 @@ impl BybitExecutionClient {
         }
 
         Ok(reports)
+    }
+
+    fn bind_execution_event_target(&mut self) {
+        self.bind_execution_event_target_with(get_exec_event_sender, get_sourced_exec_event_sink);
+    }
+
+    fn bind_execution_event_target_with(
+        &mut self,
+        legacy_sender: impl FnOnce() -> tokio::sync::mpsc::UnboundedSender<ExecutionEvent>,
+        sourced_sink: impl FnOnce(ClientId) -> SourcedExecutionEventSink,
+    ) {
+        if self.use_sourced_execution_events {
+            self.emitter
+                .set_sourced_sink(sourced_sink(self.core.client_id));
+        } else {
+            self.emitter.set_sender(legacy_sender());
+        }
     }
 
     fn resolve_position_idx(
@@ -947,8 +972,7 @@ impl ExecutionClient for BybitExecutionClient {
             return Ok(());
         }
 
-        let sender = get_exec_event_sender();
-        self.emitter.set_sender(sender);
+        self.bind_execution_event_target();
         self.core.set_started();
 
         let http_client = self.http_client.clone();
@@ -2249,7 +2273,11 @@ impl BybitExecutionClient {
 
 #[cfg(test)]
 mod tests {
-    use std::{cell::RefCell, rc::Rc, time::Duration};
+    use std::{
+        cell::{Cell, RefCell},
+        rc::Rc,
+        time::Duration,
+    };
 
     use nautilus_common::{
         cache::Cache,
@@ -2260,7 +2288,7 @@ mod tests {
         },
     };
     use nautilus_core::{Params, UUID4};
-    use nautilus_live::ExecutionClientCore;
+    use nautilus_live::{ExecutionClientCore, runner::AsyncRunner};
     use nautilus_model::{
         enums::{AccountType, OrderSide, OrderStatus},
         events::OrderEventAny,
@@ -2276,7 +2304,9 @@ mod tests {
         enums::BybitMarketUnit,
     };
 
-    fn test_execution_client() -> (BybitExecutionClient, Rc<RefCell<Cache>>) {
+    fn test_execution_client_with_sourced_events(
+        use_sourced_execution_events: bool,
+    ) -> (BybitExecutionClient, Rc<RefCell<Cache>>) {
         let cache = Rc::new(RefCell::new(Cache::default()));
         let core = ExecutionClientCore::new(
             TraderId::from("TESTER-001"),
@@ -2291,10 +2321,44 @@ mod tests {
         let config = BybitExecutionClientConfig {
             api_key: Some("test_key".to_string()),
             api_secret: Some("test_secret".to_string()),
+            use_sourced_execution_events,
             ..Default::default()
         };
 
         (BybitExecutionClient::new(core, config).unwrap(), cache)
+    }
+
+    fn test_execution_client() -> (BybitExecutionClient, Rc<RefCell<Cache>>) {
+        test_execution_client_with_sourced_events(false)
+    }
+
+    #[rstest]
+    #[case(false)]
+    #[case(true)]
+    fn test_sourced_execution_event_setting_is_frozen_at_construction(#[case] enabled: bool) {
+        let (mut client, _cache) = test_execution_client_with_sourced_events(enabled);
+        let runner = AsyncRunner::new();
+        runner.bind_senders();
+        let legacy_calls = Cell::new(0);
+        let sourced_calls = Cell::new(0);
+
+        assert_eq!(client.use_sourced_execution_events, enabled);
+        client.config.use_sourced_execution_events = !enabled;
+        assert_eq!(client.use_sourced_execution_events, enabled);
+        client.bind_execution_event_target_with(
+            || {
+                legacy_calls.set(legacy_calls.get() + 1);
+                get_exec_event_sender()
+            },
+            |client_id| {
+                sourced_calls.set(sourced_calls.get() + 1);
+                assert_eq!(client_id, *BYBIT_CLIENT_ID);
+                get_sourced_exec_event_sink(client_id)
+            },
+        );
+
+        assert_eq!(legacy_calls.get(), (!enabled) as usize);
+        assert_eq!(sourced_calls.get(), enabled as usize);
     }
 
     async fn wait_for_spawned_tasks(client: &BybitExecutionClient) {
--- a/crates/adapters/bybit/src/execution.rs
+++ b/crates/adapters/bybit/src/execution.rs
@@ -28,15 +28,12 @@ use futures_util::{StreamExt, pin_mut};
 use nautilus_common::{
     clients::ExecutionClient,
     live::runner::get_exec_event_sender,
-    messages::{
-        ExecutionEvent,
-        execution::{
-            BatchCancelOrders, CancelAllOrders, CancelOrder, GenerateFillReports,
-            GenerateFillReportsBuilder, GenerateOrderStatusReport, GenerateOrderStatusReports,
-            GenerateOrderStatusReportsBuilder, GeneratePositionStatusReports,
-            GeneratePositionStatusReportsBuilder, ModifyOrder, QueryAccount, QueryOrder,
-            SubmitOrder, SubmitOrderList,
-        },
+    messages::execution::{
+        BatchCancelOrders, CancelAllOrders, CancelOrder, GenerateFillReports,
+        GenerateFillReportsBuilder, GenerateOrderStatusReport, GenerateOrderStatusReports,
+        GenerateOrderStatusReportsBuilder, GeneratePositionStatusReports,
+        GeneratePositionStatusReportsBuilder, ModifyOrder, QueryAccount, QueryOrder, SubmitOrder,
+        SubmitOrderList,
     },
 };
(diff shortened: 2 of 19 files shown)


cjdsellers:
Hey @xxxxxx-oss,

Thanks for carrying the second stage through and documenting the validation in detail. I've had a look at the overall shape, and the scope is too large to review confidently or merge as one change. It combines a public runtime-report API migration, changes across 16 adapters, fail-closed engine validation, live-node ordering, external-order routing, and EventStore capture and replay. That is more than 6,000 changed lines across 60 files, with several independently sensitive runtime contracts.

We're also moving toward stricter validation through explicit opt-in configuration. Existing users should retain the current runtime behavior and Rust API by default during that rollout. A config flag alone does not make changing `ExecutionEvent::Report(ExecutionReport)` to `ExecutionEvent::Report(SourcedExecutionReport)` compatible: the pilot also needs an additive transport boundary so existing custom senders and exhaustive matches continue to compile.

**Preferred sequence**:
1. Pilot source-bound `OrderStatusReport` and `FillReport` handling on one representative adapter, behind an opt-in config that defaults to the current behavior. Keep the existing public event shape and capture behavior, and cover the original two-client routing case plus exact-source external-order registration.
1. Add sourced EventStore capture and replay for that pilot path in a separate PR, while retaining the existing payloads by default.
1. Extend the opt-in validation to `OrderWithFills` and position reports, including the no-partial-mutation contract.
1. Handle runtime and startup mass status separately, including the temporary legacy-cache compatibility policy.
1. Migrate the remaining adapters in small batches after the shared behavior and configuration semantics are settled.
1. Consider changing defaults or deprecating the legacy path later as a separate compatibility decision.

When you get a chance, could you reduce this PR to the first pilot and move the remaining work into follow-up PRs in that order? The current branch can remain as a useful reference implementation while the work is split.

xxxxxx-oss:
Thanks, this sequencing makes sense. I will keep the current full implementation reachable from a reference branch and reduce #4818 itself to the first pilot.

I propose Bybit as the representative adapter because its untracked order and execution paths already have focused offline coverage for both `OrderStatusReport` and `FillReport`. The pilot would be behind an adapter-scoped opt-in that defaults to the current behavior, retain `ExecutionEvent::Report(ExecutionReport)` and the existing capture behavior unchanged, and scope source-bound validation and exact-source external-order registration to those two report types and the original two-client routing case.

Before rewriting the branch, I would like to confirm one transport detail. Should the additive source-bound boundary carry the pilot adapter’s entire `ExecutionEvent` stream through a separate ingress, preserving per-client FIFO between order events and reports while leaving ordering relative to legacy-client ingress undefined? Or must sourced and legacy inputs share one globally ordered ingress?

cjdsellers:
Thanks @xxxxxx-oss, Bybit works for the pilot.

Use the separate adapter-bound ingress for the entire Bybit `ExecutionEvent` stream. The pilot does not need a defined global order relative to legacy-client ingress, but the runner must service both lanes fairly so neither can starve. The sourced sink should bind the client ID once, and each event should enter exactly one lane without falling back to legacy handling if sourced delivery fails.

The rest of the proposed scope looks right: an adapter-scoped opt-in fixed at client construction and defaulting off, unchanged `ExecutionEvent` and existing sender/emitter APIs, and unchanged raw topics and bare capture payloads. Strict sourced handling should cover only `OrderStatusReport` and `FillReport`, while retaining the full source, account, handled-venue, cached-origin, and order-ID preflight for both, plus exact-source external-order registration. Other report variants should retain explicit legacy semantics for this pilot.

The pilot guarantee is limited to reports arriving through the opted-in Bybit path. Legacy reports and sourced capture/replay remain follow-up work.

xxxxxx-oss:
@cjdsellers Thanks for the detailed direction. #4818 has now been reduced to the requested Bybit
pilot and rebased onto `develop`. The pre-split reference implementation remains available on the
[`source-runtime-execution-reports-reference`](https://github.com/xxxxxx-oss/nautilus_trader/tree/source-runtime-execution-reports-reference)
branch.

The submitted head keeps the opt-in adapter-scoped, fixed at construction, and default-off. It uses
the separate source-bound ingress for the complete Bybit `ExecutionEvent` stream, with fair per-lane
FIFO service, exactly-one-lane delivery, and no legacy fallback. Strict source, account,
handled-venue, cached-origin, and order-ID validation plus exact-source external-order registration
apply only to sourced `OrderStatusReport` and `FillReport` values; other variants retain legacy
semantics. The public event shape, existing sender/emitter APIs, `AsyncRunnerChannels`, raw topics,
and bare capture payloads remain unchanged.

The PR description now reflects the pilot scope, and all CI checks for the submitted head pass,
including Python 3.12/3.13/3.14 builds, Rust tests, and pre-commit.

cjdsellers:
Hey @xxxxxx-oss,

Thanks for narrowing this to the Bybit pilot. The separate source-bound lane matches the migration path we discussed: opted-in Bybit events enter that lane exactly once, while the legacy channel remains the default until changing or deprecating it is considered separately.

I also checked this against #4844. The two branches merge cleanly, and the changes are compatible: #4844 affects SPOT `PositionStatusReport` generation, while this pilot applies strict sourced handling only to `OrderStatusReport` and `FillReport`. Position reports keep the existing reconciliation semantics here.

I found one startup problem that needs fixing before this can merge. During `connect_exec_phase`, `drive_with_event_buffering` buffers every sourced execution event. Bybit `connect()` sends its initial `AccountState` and then waits for that account to be registered in the cache. With `use_sourced_execution_events = true` and an empty account cache, that state remains buffered until `connect()` returns, so the client reaches its 30-second timeout instead. The Bybit connection tests seed the account cache before connecting and do not cover this path.

Could you update the startup handling so a source-bound `ExecutionEvent::Account` can register the account while the execution client is connecting, and add an empty-cache regression test? The buffering and strict handling for sourced order and fill reports should remain unchanged.

xxxxxx-oss:
Updated `connect_exec_phase` so source-bound `ExecutionEvent::Account` can register while the execution client is connecting.

Added an empty-account-cache regression test.

Sourced order and fill report buffering and strict handling remain unchanged.

cjdsellers:
@xxxxxx-oss, thanks for the startup update. I've had another look at the latest changes, and that path now does what we need: the initial sourced `AccountState` can register while the execution client connects, while sourced order and fill reports remain buffered.

I found three remaining issues before this is ready:

- `use_sourced_execution_events` is inserted before the existing `auto_repay_spot_borrows`, `margin_mode`, and `transport_backend` Python constructor arguments. Existing positional callers now bind those values differently, so the new option needs to preserve all existing positional bindings.
- Public integrations using `AsyncRunner::take_channels()` drop the private sourced receiver. An opted-in Bybit client can still resolve the TLS sender, but every subsequent event fails on the closed sourced lane without fallback. The sourced lane needs to remain serviceable through the public integrated-runner path while preserving the existing seven-field channel shape.
- Strict sourced preflight clones both cached order resolutions, including their event histories. The live path runs three preflights per report, so a tracked report can cause up to six full order clones. This should use borrowed orders or copy only the fields needed for validation.

Could you also add `use_sourced_execution_events` to the Bybit execution configuration table when you address these?

xxxxxx-oss:
@cjdsellers Thanks for the review. The latest head (`df6dbc31d2`) addresses the three remaining issues:

- `use_sourced_execution_events` now follows all existing Python constructor arguments, with regression coverage for the legacy positional bindings.
- `AsyncRunner::take_channels()` now retains the private sourced ingress, and `AsyncRunner::handle_next_exec_event()` services both execution lanes while preserving the existing seven-field `AsyncRunnerChannels` shape. The integrated path is covered for sourced account and strict report dispatch, closure, and draining.
- Strict sourced preflight now borrows cached orders through `order_ref()` instead of cloning their event histories.

I also added `use_sourced_execution_events` to the Bybit execution configuration table. The startup behavior remains unchanged: sourced `AccountState` is handled during client connection, while sourced order and fill reports remain buffered.

cjdsellers:
@xxxxxx-oss, thanks for the latest updates. The earlier startup and compatibility concerns are addressed, but one ordering issue remains.

`BybitExecutionClient::connect` starts private stream processing and subscribes to order and execution updates before it requests and emits the initial `AccountState`. During `connect_exec_phase`, sourced non-account events received in that window are buffered, while the later account event is dispatched immediately. This lets the account event overtake earlier events in the same sourced lane, so the implementation does not preserve the documented per-lane FIFO guarantee.

Could you preserve FIFO through bootstrap account registration and update the regression coverage to include an earlier sourced event, the bootstrap account, and a later sourced event?

There are also two smaller test-only production seams to clean up: `bind_execution_event_target_with` only injects senders for its test, and `StrictSourcedReportResult` is only consumed by test assertions. Their behavior can be checked through the real runner channels, engine counts, and cache state without adding those production surfaces.

xxxxxx-oss:
@cjdsellers Thanks for the precise review. The latest head [`7ba3edc4f0`](https://github.com/nautechsystems/nautilus_trader/commit/7ba3edc4f070e9b75039ef378736b220f1f0ff18) addresses the remaining items:

- During execution-client bootstrap, a sourced `AccountState` now drains the already-received prefix of the same sourced lane before it is dispatched, preserving `earlier event -> AccountState -> later event` while still allowing account registration to complete during `connect()`.
- The regression test exercises that exact sequence through the real runner path and verifies dispatch order, cache registration, no duplicate dispatch, and an empty pending queue.
- `bind_execution_event_target_with` and `StrictSourcedReportResult` have been removed. The replacement tests use the real runner channels and observable engine, cache, and registration state.

All required CI checks pass on the submitted head.

cjdsellers:
@xxxxxx-oss, thanks for the update. The latest change preserves the `earlier event -> AccountState -> later event` sequence I asked for, but tracing that sequence through startup exposes another constraint: the bootstrap `AccountState` cannot be processed as an ordinary FIFO event.

`ExecutionEngine::connect()` holds `exec_engine.borrow_mut()` across the client `connect().await`. Draining an earlier order or report before the bootstrap account therefore re-enters the same `ExecutionEngine` while that borrow is live and panics with `already mutably borrowed`.

I think the clean boundary is to treat the initial HTTP `AccountState` as an explicitly identified, source-bound bootstrap barrier. Normal sourced events should remain buffered until the connect borrow is released:

```text
While `ExecutionEngine::connect()` holds its mutable borrow:

    subscribe to private streams
        |
        +-- runtime event E1 ----------> buffer E1
        |
        +-- bootstrap AccountState ---> register account once
        |
        +-- runtime event E2 ----------> buffer E2
        |
        `-- client connect returns

After the mutable engine borrow is released:

    buffered E1
        |
        v
    `LiveNode::process_sourced_exec_event()`
        |
        v
    buffered E2
        |
        v
    steady-state sourced FIFO processing
```

This requires the following behavior:

- Retain the current subscribe-before-snapshot order so private updates cannot fall into an observation gap.
- Distinguish the requested bootstrap snapshot from later wallet `AccountState` events using an internal source-bound marker, connection-scoped envelope, or equivalent design that leaves the public `ExecutionEvent` shape unchanged.
- Apply only that bootstrap account during `connect()`. Do not drain buffered order, fill, position, or later account events while the execution-engine borrow is live.
- After `connect()` returns, drain buffered sourced events through `LiveNode::process_sourced_exec_event()` in their original FIFO order. Sending them directly through `AsyncRunner` skips recent-fill suppression, execution-manager observation, and closed-order cleanup.
- Keep one receiver owner across the bootstrap-to-steady-state transition so newly arriving events cannot overtake the pending prefix.

The regression test needs to exercise the real `SystemKernel::connect_exec_clients()` borrow and registered execution-engine handlers. The current test proves channel ordering with a stub report endpoint, but it cannot reproduce the production `RefCell` conflict or verify the execution-manager state.

Could you rework the startup path around that bootstrap barrier and replace the stub coverage with the real engine path?

xxxxxx-oss:
@cjdsellers Thanks for the detailed guidance. The latest head `69c9407c81` reworks startup around the requested source-bound bootstrap account barrier.

- Bybit retains subscribe-before-snapshot ordering. Only the explicitly marked initial HTTP `AccountState` is applied during `connect()`; all runtime sourced events, including later account updates, remain buffered while the execution-engine mutable borrow is held.
- After `connect()` returns and the borrow is released, buffered sourced events are drained through `LiveNode::process_sourced_exec_event()` in their original FIFO order. The same receiver remains owned across the bootstrap-to-steady-state transition, so newly arriving events cannot overtake the pending prefix.
- The regression now exercises the real kernel `connect_exec_clients()` path with registered execution-engine handlers, covering bootstrap registration, the live borrow boundary, FIFO ordering, execution-manager and cache state, and the transition to steady-state processing.

The branch has also been rebased onto the current `develop`, with the emitter conflict reconciled against the shared-target behavior introduced by #4874. All required CI checks pass on the submitted head.

cjdsellers:
@xxxxxx-oss, thanks for reworking the startup path around the bootstrap barrier and replacing the stub coverage with the real kernel regression. I've had another look at the latest changes, and this looks good from my side. For the sourced lane, only the marked bootstrap `AccountState` is processed while `ExecutionEngine::connect()` holds its mutable borrow, and the buffered runtime events retain FIFO through `LiveNode::process_sourced_exec_event()` after that borrow is released.

The branch now conflicts with the latest `develop` in `crates/adapters/bybit/src/execution.rs` and `crates/live/src/runner.rs`, so those conflicts need to be resolved and the relevant checks rerun before this can merge 👌.

When you get a chance, it would be ideal if you could provide live validation evidence for the exercised paths on Bybit live.

xxxxxx-oss:
@cjdsellers Thanks for the review. I resolved the conflicts with the current `develop` in #4818 at [`a45a25c`](https://github.com/nautechsystems/nautilus_trader/commit/a45a25c34f76c744f9813d40c0422dd550622316). GitHub now reports the PR as mergeable, and the relevant [CI run](https://github.com/nautechsystems/nautilus_trader/actions/runs/33258833781) passed, including the Python 3.12/3.13/3.14 builds, Rust tests, script tests, pre-commit, and security audit.

I also completed venue-connected validation of this exact head using the Bybit Testnet V5 REST and private WebSocket APIs. This validation used Testnet, not Mainnet.

### Validation conditions

- Bybit Linear perpetual account in one-way position mode.
- `use_sourced_execution_events=true`.
- Nautilus started with an empty account cache and no retained test position or active test order.
- Orders were submitted, amended, and canceled through Bybit outside Nautilus, so the validation exercised inbound source-bound reports rather than Nautilus order submission.
- The Nautilus execution engine generated zero outbound order commands throughout the scenarios.

### Startup/bootstrap path

Starting from the empty account cache:

- The order, execution, position, and wallet private-stream subscriptions were each acknowledged exactly once.
- The marked initial HTTP `AccountState` registered successfully through the source-bound bootstrap barrier.
- The node reached `Running` and completed a clean shutdown.
- Runtime sourced events remained on the normal buffered path.
- The startup run produced no errors or warnings.

### Strict-origin negative path

The strict-origin failure path was exercised both before and after external cancellation:

- Both mismatched reports were rejected by sourced preflight.
- Rejection occurred before raw order/fill report publication.
- The cached execution-client origin and order state remained unchanged.
- No external order was materialized or re-registered against the wrong execution client.
- Nautilus generated zero order commands.
- The venue finished with no active test order and a flat position.

### Positive order/fill path

Three externally controlled orders were exercised:

1. A PostOnly Limit probe followed `submit -> New -> cancel -> Cancelled` and had zero executions.
2. A non-PostOnly GTC Limit opening order followed:
   `submit -> New -> raw Accepted -> reconciliation Initialized -> strict sourced preflight -> cache materialization -> amend -> Filled`.
3. Only after the opening order was filled, a reduce-only GTC Limit closing order followed the same sequence with `reduce_only=true` and ended `Filled`.

For both the opening and closing orders:

- The raw Accepted marker, full source/identity diagnostic, and cache-materialized marker each occurred exactly once.
- The published `OrderInitialized` carried reconciliation provenance.
- The client order identity was derived from the venue order identity.
- The cache contained both client-to-venue and venue-to-client indexes.
- The cached execution-client origin was exactly `BYBIT`.
- Account, instrument, strategy, order identity, order type, status, and reduce-only state matched the source report.
- Materialization additionally required the corresponding raw Accepted report.
- Amend prices were tick-aligned, selected inside the current Bybit dynamic price limits, and confirmed from terminal venue state.
- The REST execution set was non-empty and internally unique.
- The raw `FillReport`, REST execution, cached trade-ID, and filled-event trade-ID sets were equal.
- The opening and closing execution sets were disjoint.
- No duplicate fills or overfill were observed.

### Observed results

```text
startup.running=1
startup.bootstrap_account_registered=true
startup.subscription.order=1
startup.subscription.execution=1
startup.subscription.position=1
startup.subscription.wallet=1

negative.strict_origin_rejection.before_cancel=1
negative.strict_origin_rejection.after_cancel=1
negative.raw_order_reports=0
negative.raw_fill_reports=0
negative.origin_unchanged=true
negative.cache_unchanged=true

positive.post_only.cancelled=true
positive.post_only.execution_count=0

positive.open.raw_accepted=1
positive.open.reconciliation_initialized=1
positive.open.full_diagnostic=1
positive.open.materialized=1
positive.open.execution_count=1

positive.close.raw_accepted=1
positive.close.reconciliation_initialized=1
positive.close.full_diagnostic=1
positive.close.materialized=1
positive.close.execution_count=1

positive.exact_origin=true
positive.bidirectional_indexes=true
positive.materialized_before_amend=true
positive.amend_terminal_prices_confirmed=true
positive.trade_sets_match=true
positive.duplicate_fills=0
positive.scenario_pass=1

execution_engine.order_commands=0
final.venue.active_orders=0
final.venue.flat=true
final.cache.flat=true
harness.errors=0
harness.warnings=0
```

A final fresh REST query independently confirmed zero active orders and a flat position at Bybit, matching the Nautilus cache.

cjdsellers:
@xxxxxx-oss, I've had another look at the latest changes. Thanks for adding the Testnet results. **I found two remaining issues:**

- With `use_sourced_execution_events=false`, `connect()` now propagates failures from `try_send_bootstrap_account_state`. The previous legacy path logged this send failure and continued, so callers with a pre-registered account could connect before `start()`. The latest test changes add `start()` to accommodate the new requirement. The default-off path needs to preserve its existing behavior.
- During sourced startup, a wallet `AccountState` can arrive before the HTTP bootstrap. The node buffers that wallet state, immediately applies the later bootstrap, then replays the earlier wallet state after connection. Account updates do not reject older timestamps, so the earlier wallet balances can overwrite the bootstrap values until another update arrives. This needs coverage where the wallet state precedes the bootstrap and uses distinct balances and timestamps.

The sourced report routing and strict source validation otherwise match the agreed Bybit pilot.

xxxxxx-oss:
@cjdsellers Thanks for the review. The latest rebased head [`e9e39e1ee8`](https://github.com/nautechsystems/nautilus_trader/commit/e9e39e1ee817d1f8262262aea842e7f1e1625eee) addresses both remaining issues.

- With `use_sourced_execution_events=false`, `connect()` now uses the legacy `send_account_state` path, so a send failure is logged without aborting the connection. The pre-`start()` lifecycle coverage for pre-registered accounts has been restored, while the sourced bootstrap path remains fail-closed.
- Before dispatching the marked HTTP bootstrap, startup now removes only buffered runtime `AccountState` values from the same execution client and account whose `ts_event` is strictly older than the bootstrap. The real-kernel regression sends the wallet state first with 100 USDT, followed by the bootstrap with 200 USDT, and verifies that the bootstrap remains authoritative. A later wallet update with 300 USDT is still applied normally. Additional coverage confirms that different sources, different accounts, and equal or newer timestamps remain queued in FIFO order.

The branch has now been rebased onto `develop` at `5d5c21e24a`, and the two conflicts introduced by the later task-lifecycle and snapshot-test changes have been resolved without changing the agreed Bybit pilot scope.

Local Ubuntu validation passed on the submitted head, including all 40 Bybit execution-client integration tests, all 395 `nautilus-live` tests under the official nextest process-isolated run, all 267 execution-engine tests, and all 249 non-ignored matching-engine tests, with one existing test remaining ignored. The relevant Clippy checks with `-D warnings`, pinned nightly formatting check, and applicable pre-commit hooks also passed.

The pushed head is now available for CI and review.

cjdsellers:
@xxxxxx-oss, thank you for all the thoughtful follow-ups and for carrying this through several substantial revisions, including the detailed Testnet validation. I'm going to close this PR here and take the remaining work forward as maintainer-owned work after the next RC wheel release later this week. The implementation and tests you've developed here will form the basis of that work, and I'll make sure your contribution is attributed in the resulting implementation and credited in the release notes when it lands.

The remaining startup issue needs an architectural decision rather than another bounded code fix. The REST wallet snapshot and WebSocket wallet updates do not expose a shared sequence boundary that we can safely rely on, so the exact synchronization guarantee and venue-specific trade-offs need to be settled on the maintainer side. I don't think it would be fair or productive to ask you for another open-ended revision.

cjdsellers:
Hi @xxxxxx-oss,

Quick note: have not forgotten, just bedding down some execution standardizations before circling back to this.

Tickwhiz:
@cjdsellers Thanks for the update and for taking over the remaining work. I appreciate the time and care you've put into reviewing the revisions and clarifying the startup synchronization constraints. I'm glad the implementation and tests from this PR can provide a useful basis for the subsequent work. Thanks again for keeping me informed.

PR #4819, opened 2026-08-21
Prevent L2/L3 matching against a deleted touch

SoYuCry:
## Summary

Synchronize the matching core from the authoritative L2/L3 order book **before matching** so deleting the final level on one side cannot leave a stale touch active during the same iteration.

Polymarket's valid one-sided books made this easier to observe, but the invariant and fix are generic to L2/L3 matching.

## Problem

`iterate` previously copied a book touch into the matching core only when `best_bid_price()` / `best_ask_price()` returned `Some(price)`. If a delta changed a side from `Some(price)` to `None`, the old core price remained until the synchronization at the end of the iteration.

That end-of-iteration synchronization restores the final state, but it is too late: resting orders and stops may already have been processed against the deleted touch during that iteration.

## Change

For non-trade L2/L3 iterations, assign the complete option values before matching:

```rust
self.core.bid = self.book.best_bid_price();
self.core.ask = self.book.best_ask_price();
```

This propagates both present and empty sides. Existing L1 quote/trade behavior is unchanged.

The earlier post-trade restoration change has been removed because `iterate` already restores both core touches from the book at the end of every pass.

## Regression coverage

The regression test creates the pre-match stale-touch window:

1. Submit a resting buy at 1500 with an empty L2 book.
2. Pause the market.
3. Add an ask at 1500; matching is blocked, while end-of-iteration sync records the ask in the core.
4. Reopen the market without iterating.
5. Delete the final ask; this invokes one iteration.
6. Assert the resting buy is not filled from the deleted ask and remains open.

The production implementation before this patch enters the matching pass with the stale `core.ask = 1500`; the patch clears it to `None` before matching.

## Verification

- `rustfmt --check` on both changed Rust files.
- `git diff --check`.
- Targeted local Rust build was attempted, but this Windows host lacks the MSVC `link.exe`; the pushed CI run provides the executable regression result.

### DIFF
--- a/crates/execution/src/matching_engine/mod.rs
+++ b/crates/execution/src/matching_engine/mod.rs
@@ -3704,12 +3704,20 @@ impl OrderMatchingEngine {
         // where `process_trade_tick` overrides both sides to the trade
         // price; without it the override is undone here.
         if aggressor_side == AggressorSide::NoAggressor && self.last_trade_size.is_none() {
-            if let Some(bid) = self.book.best_bid_price() {
-                self.core.set_bid_raw(bid);
-            }
+            if self.book_type == BookType::L1_MBP {
+                if let Some(bid) = self.book.best_bid_price() {
+                    self.core.set_bid_raw(bid);
+                }
 
-            if let Some(ask) = self.book.best_ask_price() {
-                self.core.set_ask_raw(ask);
+                if let Some(ask) = self.book.best_ask_price() {
+                    self.core.set_ask_raw(ask);
+                }
+            } else {
+                // L2/L3 books are authoritative. Assigning the complete options
+                // propagates an empty side before matching and prevents fills
+                // or triggers from a stale touch.
+                self.core.bid = self.book.best_bid_price();
+                self.core.ask = self.book.best_ask_price();
             }
         }
 
--- a/crates/execution/tests/matching_engine.rs
+++ b/crates/execution/tests/matching_engine.rs
@@ -1506,6 +1506,70 @@ fn test_process_limit_post_only_order_that_would_be_a_taker(
 }
 
 #[rstest]
+fn test_l2_delete_final_ask_does_not_fill_resting_bid_from_stale_touch(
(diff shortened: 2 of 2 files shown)


cjdsellers:
Hi @SoYuCry,

Thanks for the PR. I ran the two new tests against the production code from before this patch and both still pass. `iterate` already restores `core.bid` and `core.ask` from the book at the end of each pass, and in the trade case the old `price_raw < ask.raw` guard is false when both prices are 1500, so neither test exercises the new pre-match clearing behavior.

The regression coverage needs a case with a resting order or stop in the matching core while a delta removes the final opposing level, asserting that the same iteration does not trigger or fill from the deleted touch. That test should fail before the production change and pass with this patch.

SoYuCry:
Thanks, you're right: the previous tests only checked post-iteration state, which the existing end-of-pass synchronization already restores. I've replaced them with a resting-order regression that exercises the same-iteration pre-match window: while paused, an opposing touch is introduced; after reopening, deleting that final level must not fill the resting order from the stale core touch. I also narrowed the production change to complete L2/L3 Option synchronization before matching and removed the redundant post-trade restoration change. L1 behavior remains unchanged.

cjdsellers:
@SoYuCry, thanks for updating this. I ran the replacement test against the production code from before this patch, and it still passes. The stale `core.ask` sends the resting limit order into `fill_limit_order`, but that path then reads the now-empty book, produces no fills, and leaves the order open, so every current assertion also passes without the fix.

A stop-limit variant does expose the regression: deleting the final ask emits `OrderTriggered` before the patch and remains untriggered with this change. Could you update the committed regression to cover that state change, or another observable effect that fails without the production change?

cjdsellers:
@SoYuCry, I've added the stop-limit regression and had another look over the latest changes. Everything we discussed is now addressed, and this looks good from my side. Thanks again for working through the feedback and narrowing the fix 👍

cjdsellers (approved):

PR #4823, opened 2026-08-21
Fix Bybit order book depth validation

### LINKED ISSUE #4820: [Bybit] Order book depth validation rejects 1000 and accepts a 500 the venue no longer serves
# Bug Report

## Confirmation

## Expected behavior

The Bybit data client should accept the order book depths the venue actually serves.
Per the [Bybit V5 WebSocket orderbook documentation](https://bybit-exchange.github.io/docs/v5/websocket/public/orderbook),
those are:

| Category         | Levels served                                      |
| ---------------- | -------------------------------------------------- |
| linear / inverse | 1 (10ms), 50 (20ms), 200 (100ms), **1000** (200ms)  |
| spot             | 1 (10ms), 50 (20ms), 200 (100ms), **1000** (200ms)  |
| option           | 25 (20ms), 100 (100ms)                             |

The [V5 changelog](https://bybit-exchange.github.io/docs/changelog/v5) records the transition:

- **2025-08-14** — "Add level 1000 orderbooks for spot and futures"
- **2025-09-11** — "Remove level 500 orderbooks for futures"
- **2025-11-25** — "The push frequency of Orderbook Level 1000 data is changed from 300ms to 200ms"

So `orderbook.500.*` was valid on futures at one point and was withdrawn in September 2025,
while `orderbook.1000.*` is now the deepest tier available on spot, linear and inverse.

## Actual behavior

`BybitDataClient::subscribe_book_deltas` validates depth against `1 | 50 | 200 | 500`
— rejecting the deepest tier the venue really offers, and admitting one it no longer serves.

`crates/adapters/bybit/src/data.rs:876-882`:

```rust
let depth = cmd
    .depth
    .map_or(BYBIT_DEFAULT_ORDERBOOK_DEPTH, |d| d.get() as u32);

if !matches!(depth, 1 | 50 | 200 | 500) {
    anyhow::bail!("invalid depth {depth}; valid values are 1, 50, 200, or 500");
}
```

**Both failure modes are silent, which is what makes this costly to diagnose:**

1. `depth = 1000` — the deepest tier the venue serves — is rejected locally. The subscription
   never reaches the exchange.
2. `depth = 500` passes validation and the subscription is sent. The venue answers with an
   error, which surfaces only as a WARN log; `Subscribed ... order book deltas` is still
   logged at INFO, and **no deltas ever arrive**. Nothing is raised to the caller.

Either way the strategy silently runs on the 50-level default
(`BYBIT_DEFAULT_ORDERBOOK_DEPTH`, `crates/adapters/bybit/src/common/consts.rs:51`) or on no
book at all.

**Real-world impact.** A strategy of ours sums resting book volume between the touch and its
own limit price (~1.2% below the touch). It was reading a 50-level book that only spans 0.21%
below the touch, undercounting that sum by roughly 5x. Its thin-book compensation term
therefore never saw the volume that would have switched it off, ran away to its ceiling in
~165s, and parked the quote ~14% below the market. No exception, no error — just a strategy
behaving inexplicably until we traced it to the depth tier.

## Steps to reproduce

1. Configure a Bybit data client (linear, spot or inverse).
2. Subscribe to order book deltas with `depth=1000` → the client bails locally with
   `invalid depth 1000; valid values are 1, 50, 200, or 500`.
3. Subscribe with `depth=500` instead → validation passes, `Subscribed ... order book deltas`
   is logged, and no deltas ever arrive.

The venue behavior can be confirmed without the adapter using the standalone script below.

## Code snippets or logs

Probing `wss://stream.bybit.com` directly (standard library only, no dependencies).
Measured 2026-08-20 and re-confirmed 2026-08-21:

| product type | symbol  | 1  | 50 | 200 | 500          | 1000   |
| ------------ | ------- | -- | -- | --- | ------------ | ------ |
| linear       | ACEUSDT | ok | ok | ok  | **rejected** | **ok** |
| linear       | BTCUSDT | ok | ok | ok  | **rejected** | **ok** |
| linear       | ETHUSDT | ok | ok | ok  | **rejected** | **ok** |
| spot         | BTCUSDT | ok | ok | ok  | **rejected** | **ok** |
| inverse      | BTCUSD  | ok | ok | ok  | **rejected** | **ok** |

Venue responses for depth 500:

- linear / inverse: `error:handler not found,topic:orderbook.500.BTCUSDT`
- spot: `Invalid topic :[orderbook.500.BTCUSDT]`

I did not probe the `option` category (see the note at the end).

Snapshot span actually delivered (ACEUSDT linear, same moment): 50 levels span 0.21% below
the touch, 200 levels span 1.15%, 1000 levels span 9.77%. Cumulative bid volume down to a
given distance from the touch:

| summed down to | depth 50 | depth 200 | depth 1000    |
| -------------- | -------- | --------- | ------------- |
| -1.18%         | 39,181   | 152,709   | 171,765       |
| -4.00%         | 35,027   | 130,504   | **361,601**   |
| -15.00%        | 30,112   | 143,505   | **1,045,152** |

At depth 1000 the sum grows with the window; at 50/200 it is flat, because the price being
summed to already lies outside the span those tiers deliver.

<details>
<summary>Standalone reproduction script (standard library only)</summary>

```bash
python3 probe.py ACEUSDT                        # linear; probes 1/50/200/500/1000
BYBIT_WS_PATH=/v5/public/spot    python3 probe.py BTCUSDT
BYBIT_WS_PATH=/v5/public/inverse python3 probe.py BTCUSD
```

```python
import json, os, socket, ssl, base64, struct, sys, time

HOST = "stream.bybit.com"
PATH = os.environ.get("BYBIT_WS_PATH", "/v5/public/linear")
SYMBOL = sys.argv[1] if len(sys.argv) > 1 else "ACEUSDT"
TIERS = [int(t) for t in (sys.argv[2:] or ["1", "50", "200", "500", "1000"])]

def frame(payload: bytes) -> bytes:
    mask = os.urandom(4)
    masked = bytes(b ^ mask[i % 4] for i, b in enumerate(payload))
    n = len(payload)
    header = (struct.pack("!BB", 0x81, 0x80 | n) if n < 126
              else struct.pack("!BBH", 0x81, 0x80 | 126, n))
    return header + mask + masked

def read_frames(sock, budget):
    buf, out = b"", []
    # An accepted subscription streams continuously, so recv never times out —
    # the loop has to be bounded by wall time, not by idleness.
    deadline = time.monotonic() + budget
    try:
        while time.monotonic() < deadline:
            sock.settimeout(max(0.1, deadline - time.monotonic()))
            chunk = sock.recv(65536)
            if not chunk:
                break
            buf += chunk
            while len(buf) >= 2:
                b1, b2 = buf[0], buf[1]
                length, offset = b2 & 0x7F, 2
                if length == 126:
                    if len(buf) < 4: break
                    length, offset = struct.unpack("!H", buf[2:4])[0], 4
                elif length == 127:
                    if len(buf) < 10: break
                    length, offset = struct.unpack("!Q", buf[2:10])[0], 10
                if len(buf) < offset + length: break
                payload, buf = buf[offset:offset + length], buf[offset + length:]
                if b1 & 0x0F == 0x1:
                    out.append(payload.decode("utf-8", "replace"))
    except socket.timeout:
        pass
    return out

ctx = ssl.create_default_context()
sock = ctx.wrap_socket(socket.create_connection((HOST, 443), timeout=15),
                       server_hostname=HOST)
key = base64.b64encode(os.urandom(16)).decode()
sock.send((f"GET {PATH} HTTP/1.1\r\nHost: {HOST}\r\nUpgrade: websocket\r\n"
           f"Connection: Upgrade\r\nSec-WebSocket-Key: {key}\r\n"
           f"Sec-WebSocket-Version: 13\r\n\r\n").encode())
if "101" not in sock.recv(4096).decode("utf-8", "replace").split("\r\n")[0]:
    print("handshake failed"); sys.exit(1)
print(f"connected to {HOST}{PATH}; probing {SYMBOL}")

for tier in TIERS:
    topic = f"orderbook.{tier}.{SYMBOL}"
    sock.send(frame(json.dumps({"op": "subscribe", "args": [topic]}).encode()))
    verdict, levels = "no response", None
    for m in read_frames(sock, 4.0):
        try: d = json.loads(m)
        except Exception: continue
        if d.get("op") == "subscribe":
            verdict = "ACCEPTED" if d.get("success") else f"REJECTED ({d.get('ret_msg')})"
        if d.get("topic") == topic and d.get("data"):
            book = d["data"]
            levels = (len(book.get("b", [])), len(book.get("a", [])))
    print(f"  depth {tier:>5}: {verdict}" + (f"   snapshot levels bid/ask={levels}" if levels else ""))
    if verdict.startswith("ACCEPTED"):
        sock.send(frame(json.dumps({"op": "unsubscribe", "args": [topic]}).encode()))
        read_frames(sock, 1.5)
sock.close()
```

Sample output (linear / BTCUSDT, 2026-08-21):

```
connected to stream.bybit.com/v5/public/linear; probing BTCUSDT
  depth     1: ACCEPTED   snapshot levels bid/ask=(1, 1)
  depth    50: ACCEPTED   snapshot levels bid/ask=(3, 6)
  depth   200: ACCEPTED   snapshot levels bid/ask=(97, 92)
  depth   500: REJECTED (error:handler not found,topic:orderbook.500.BTCUSDT)
  depth  1000: ACCEPTED   snapshot levels bid/ask=(231, 229)
```

</details>

## Specifications

- OS platform: Linux 6.17
- Python version: 3.13
- `nautilus_trader` version: `develop` @ `2114cf6f76` (also reproduced on 0.61.0 and 0.62.0 —
  the list is identical in all three)
- Adapter/venue: Bybit

## Additional notes

**This is the only place the tier list appears**, and it has never matched the venue.
`crates/adapters/bybit/src/data.rs:880` is the sole occurrence; `git log -S` traces it to
`2a68f04ea3` "Add Bybit initial data and execution clients in Rust" (2026-02-01), which is
five months *after* Bybit removed level 500. No test covers the validation in either
direction, which is likely why it went unnoticed.

**The HTTP path is a separate, and correct, matter.** `request_orderbook_snapshot`
(`crates/adapters/bybit/src/http/client.rs:3966`) clamps to 200 / 25 / 500 for
spot / option / linear-inverse, matching the REST `GET /v5/market/orderbook` limits. The REST
limits genuinely differ from the WebSocket tiers, so 500 there is not a bug and should not be
changed alongside this.

**Possible related gap for the `option` category.** `subscribe_book_deltas` applies the same
tier list regardless of product type, so an option book subscription can only ever go out at
1, 50, 200 or 500 — none of which the venue serves for options (25 and 100 per the docs
above). With the default depth of 50 that would fail the same silent way. I have not tested
this against the venue and do not want to assert it; flagging it in case a maintainer wants
it covered by the same fix or split into its own issue.

**Suggested fix.** Correct the accepted set to `1 | 50 | 200 | 1000`, lift it into a named
constant next to `BYBIT_DEFAULT_ORDERBOOK_DEPTH` (as Deribit's `DERIBIT_BOOK_VALID_DEPTHS`
and Binance's `BINANCE_BOOK_DEPTHS` already do), and add a test pinning both directions
(1000 accepted, 500 rejected). I'm happy to open a PR for this if that's welcome.

**Possible follow-up, not proposed for this fix.** A subscription the venue rejects currently
produces only a WARN and leaves the client believing it is subscribed. Surfacing that to the
caller would have turned this from a multi-hour investigation into an immediate error, but
it's a design change and belongs in its own discussion.

Centaur-pub:
  and, if I used AI,
  this draft
  limitation below

## Summary

`BybitDataClient::subscribe_book_deltas` validated the requested order book depth against
`1 | 50 | 200 | 500`, but Bybit's WebSocket `orderbook` topic serves `1 | 50 | 200 | 1000`
for the spot, linear and inverse categories. The check therefore rejected the deepest tier the
venue offers and accepted one it removed on 2025-09-11. Both failure modes were silent: depth
1000 was rejected locally and never reached the exchange, while depth 500 produced a
subscription the venue answered with an error that surfaced only as a WARN, leaving the client
logged as subscribed with no deltas arriving.

This corrects the accepted set, lifts it into `BYBIT_BOOK_DEPTHS` next to the existing
`BYBIT_DEFAULT_ORDERBOOK_DEPTH` following the `BINANCE_BOOK_DEPTHS` and
`DERIBIT_BOOK_VALID_DEPTHS` precedent, and pins both directions with tests. Users who
previously passed `depth=500` now receive an error at subscribe time rather than a
subscription that silently delivers nothing.

Deliberately left out of scope, as discussed in the issue: the REST depth limits in
`http/client.rs` (spot 200 / option 25 / linear-inverse 500 are the correct
`GET /v5/market/orderbook` limits and genuinely differ from the WebSocket tiers), the
`option` category which the venue serves at 25/100 and this validation cannot express, and
propagating venue subscription errors to the caller.

## Related issues/PRs

Closes #4820

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not a breaking change. Depth 500 passed validation before this change but never produced data,
because the venue rejected the resulting subscription; it now fails fast with an accurate
error instead.

## Documentation

No documentation changes. `docs/integrations/bybit.md` does not document depth tiers, and the
new constant is not exposed through PyO3, so no stub regeneration is required. The REST depth
limits documented on `BybitHttpClient::request_orderbook_snapshot` are unaffected and remain
correct.

## Testing

`test_validate_orderbook_depth_accepts_1000` and `test_validate_orderbook_depth_rejects_500`
pin the corrected set from both sides, the second also asserting the user-facing error text.
Both were confirmed non-vacuous by temporarily restoring the old list, against which they
fail.

The venue behavior was verified directly against `wss://stream.bybit.com` across the linear,
spot and inverse categories before filing the issue; the table, the venue's rejection messages
and a standalone reproduction script are in #4820.

Validation run locally: `make format`, `make pre-commit`,
`cargo clippy -p nautilus-bybit --all-targets -- -D warnings`, and `cargo nextest run -p
nautilus-bybit`.

### DIFF
--- a/crates/adapters/bybit/src/common/consts.rs
+++ b/crates/adapters/bybit/src/common/consts.rs
@@ -49,3 +49,9 @@ pub const BYBIT_WS_TESTNET_PRIVATE_URL: &str = "wss://stream-testnet.bybit.com/v
 pub const BYBIT_WS_TOPIC_DELIMITER: char = '.';
 
 pub const BYBIT_DEFAULT_ORDERBOOK_DEPTH: u32 = 50;
+
+/// Valid WebSocket order book depths for Bybit spot, linear and inverse.
+///
+/// See <https://bybit-exchange.github.io/docs/v5/websocket/public/orderbook>
+/// for further details.
+pub const BYBIT_BOOK_DEPTHS: [u32; 4] = [1, 50, 200, 1000];
--- a/crates/adapters/bybit/src/data.rs
+++ b/crates/adapters/bybit/src/data.rs
@@ -64,7 +64,7 @@ use ustr::Ustr;
 
 use crate::{
     common::{
-        consts::{BYBIT_DEFAULT_ORDERBOOK_DEPTH, BYBIT_VENUE},
+        consts::{BYBIT_BOOK_DEPTHS, BYBIT_DEFAULT_ORDERBOOK_DEPTH, BYBIT_VENUE},
         enums::BybitProductType,
         instruments::diff_and_emit_instruments,
         parse::{extract_raw_symbol, make_bybit_symbol},
@@ -325,6 +325,14 @@ fn send_data(sender: &tokio::sync::mpsc::UnboundedSender<DataEvent>, data: Data)
     if let Err(e) = sender.send(DataEvent::Data(data)) {
         log::error!("Failed to emit data event: {e}");
     }
+}
+
+fn validate_orderbook_depth(depth: u32) -> anyhow::Result<()> {
+    if !BYBIT_BOOK_DEPTHS.contains(&depth) {
+        anyhow::bail!("invalid depth {depth}; valid values are {BYBIT_BOOK_DEPTHS:?}");
+    }
+
+    Ok(())
 }
 
 /// Cached funding state per symbol: (funding_rate, next_funding_time, funding_interval_hour).
@@ -877,9 +885,7 @@ impl DataClient for BybitDataClient {
             .depth
             .map_or(BYBIT_DEFAULT_ORDERBOOK_DEPTH, |d| d.get() as u32);
 
-        if !matches!(depth, 1 | 50 | 200 | 500) {
-            anyhow::bail!("invalid depth {depth}; valid values are 1, 50, 200, or 500");
-        }
+        validate_orderbook_depth(depth)?;
 
         let instrument_id = cmd.instrument_id;
         let product_type = self
@@ -1965,7 +1971,7 @@ mod tests {
     use rstest::rstest;
     use ustr::Ustr;
 
-    use super::handle_ws_message;
+    use super::{handle_ws_message, validate_orderbook_depth};
     use crate::{
         common::{
             enums::BybitProductType,
@@ -2037,6 +2043,21 @@ mod tests {
             Arc::new(AtomicSet::new()),
             Arc::new(AtomicMap::new()),
         )
+    }
+
+    #[rstest]
+    fn test_validate_orderbook_depth_accepts_1000() {
+        assert!(validate_orderbook_depth(1000).is_ok());
+    }
+
+    #[rstest]
+    fn test_validate_orderbook_depth_rejects_500() {
+        let e = validate_orderbook_depth(500).unwrap_err();
+
+        assert_eq!(
+            e.to_string(),
+            "invalid depth 500; valid values are [1, 50, 200, 1000]"
+        );
     }
 
     #[rstest]


cjdsellers (approved):
Hi @Centaur-pub,

Thanks for the PR and for validating the venue behavior so thoroughly. I’ve had a look through this, and it looks good from my side. The focused change leaves the separate REST limits and option handling untouched, and the tests cover both sides of the correction.

PR #4835, opened 2026-08-24
Fix dYdX retry tests to exercise the operation timeout

folknor:
Both dYdX tests covering the retry operation timeout pass without ever reaching it, so a regression in that timeout would go undetected.

## The tests fail on configuration before the request is issued

`ExponentialBackoff::new` rejects a zero initial delay (`crates/network/src/backoff.rs:126`), and `RetryManager::execute_with_retry_inner_delay` constructs the backoff before its retry loop, unconditionally - `max_retries: 0` does not bypass it (`crates/network/src/retry.rs:206-217`). A `RetryConfig` carrying `initial_delay_ms: 0` therefore returns `RetryError::InvalidConfiguration` in microseconds, which the dYdX adapter maps to `DydxHttpError::ValidationError`, and the operation closure never runs.

`test_request_timeout_short` (`crates/adapters/dydx/tests/http.rs`) and `test_http_timeout_respects_configuration_and_does_not_block` (`crates/adapters/dydx/src/http/client.rs`) both build that config and then assert only `result.is_err()` and a loose wall-clock bound. The configuration error satisfies both. Neither test reaches its slow route, and neither exercises `operation_timeout_ms`. They complete in 0.28s and 0.05s against nominal timeouts of 1000ms and 500ms.

The predicate landed in `19de9b697b` (May 2025); both tests were written six months later, so they have never enforced what they are named for.

Both configs now use `initial_delay_ms: 1`, `max_delay_ms: 1`. With zero retries the delay is never calculated or slept: the operation timeout produces the error, `should_retry` returns true, and `attempt >= max_retries` returns immediately.

`test_invalid_configuration_reason` (`crates/network/src/retry.rs`) also sets `initial_delay_ms: 0` and is left untouched - it is the test for that error.

## The assertions could not distinguish a timeout from a setup failure

`assert!(result.is_err())` is what allowed the defect to persist, so both tests now assert the concrete failure: `DydxHttpError::HttpClientError` carrying the payload built from `RetryError::OperationTimeout { timeout_ms }`, which separates an operation timeout from a `ValidationError` and from a transport error. Each slow route also records entry, so the tests prove the request reached the server rather than timing out ahead of dispatch.

`test_request_timeout_short` additionally passed `1` as `timeout_secs` while its operation timeout was 1000ms; once the request is actually issued those two timers race, so the client timeout is now 60s and only the operation timeout can produce the failure. The `client.rs` test gained a readiness handshake against a fast route, since binding a listener makes the port connectable before the serve task reaches its accept loop.

## Testing

No production code changes; the diff is test-only.

Restoring `initial_delay_ms: 0` fails both tests on the error assertion with `Expected operation timeout, received Err(ValidationError("Invalid configuration: delay_initial must be non-zero"))`, and the handler-entry assertion reports the route was never entered.

The wall-clock bounds are kept as secondary guards at 3s and 5s against 5s and 10s handlers, loose enough to tolerate a loaded runner since the error assertion carries the claim. One limitation: the adapter erases `RetryError` into a string, so the tests match on the message rather than the variant. That message is unique to the operation-timeout path, but it is a string comparison rather than a type assertion.

### DIFF
--- a/crates/adapters/dydx/src/http/client.rs
+++ b/crates/adapters/dydx/src/http/client.rs
@@ -1785,7 +1785,13 @@ impl DydxHttpClient {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    };
+
     use axum::{Router, routing::get};
+    use nautilus_common::testing::wait_until_async;
     use nautilus_model::identifiers::Symbol;
     use rstest::rstest;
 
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. I checked both timeout paths and the retry error mapping, and the tests now distinguish `RetryError::OperationTimeout` from invalid configuration and the HTTP client timeout while proving the requests reached the server. This looks good to merge.

cjdsellers (approved):

PR #4844, opened 2026-08-24
Repair Bybit SPOT position reports from wallet balances

folknor:
The SPOT position-report option has been disconnected since the execution client was ported to Rust in `2a68f04ea3`, and reconnecting it exposes a wallet-balance fan-out that has to be removed in the same change.

## The option never reaches the HTTP client

`BybitExecutionClientConfig::use_spot_position_reports` is documented and exposed through pyo3, but nothing copies it into `BybitHttpClient`'s `use_spot_position_reports` atomic. The only writers of that atomic are the pyo3 setter and one test, so setting the option on an execution client has no effect. The Python client this replaced called `set_use_spot_position_reports` during construction, and the Kraken execution client still threads its equivalent through to the spot client. `BybitExecutionClient::new` now does the same.

## SPOT is skipped one layer above the fallback written for it

`generate_position_status_reports` skipped SPOT in both its scoped and unscoped branches, commented "positions API only supports derivatives". That is true of `/v5/position/list`, and `request_position_status_reports` already branches on SPOT and routes to a wallet-balance fallback for exactly that reason. The skip severed the feature from the code implementing it. Both skips are removed; the execution client forwards every configured product type and the HTTP client keeps ownership of product policy.

## An unscoped wallet balance cannot name an instrument

`generate_spot_position_reports_from_wallet` keys balances by coin. Given an `instrument_id` it emitted one report for the instrument the caller named, which is sound. Without one it walked the instrument cache, kept every `-SPOT` symbol, and pushed the full coin balance to every instrument whose base currency matched, undivided. One 1.0 BTC holding became simultaneous 1.0-BTC positions on `BTCUSDT-SPOT`, `BTCUSDC-SPOT` and every other cached BTC pair, so an account holding one coin reported N positions of the full amount.

An unscoped SPOT request now returns no reports, before any wallet request is issued, and the helper takes `InstrumentId` rather than `Option<InstrumentId>` so the unscoped case cannot reach it. A unified-wallet coin balance carries no pair identity and `PositionStatusReport` requires one, so silence is the only answer that does not invent the attribution.

Reporting pairs with prior orders or fills proves historical interaction rather than current provenance. Picking the single cached pair for a base depends on what happened to be loaded. Splitting the balance replaces inflation with an arbitrary allocation.

A configured quote currency, as Kraken takes, settles by convention what the venue response does not say, at the cost of a config field. That one is yours to decide.

Happy to reshape this.

The Kraken counterpart avoids the multiplicative fan-out through its quote filter but rests on the same assumption that a held coin was acquired through the configured pair. Left alone here.

`use_spot_position_reports` defaults false and stays false, so a client that does not set it sees no change.

## Testing

`test_exec_client_scoped_spot_position_reports_follow_config` covers the restored path end to end, asserting one report when the option is enabled and none when it is not. It fails against the unfixed code, and reverting either the constructor wiring or the SPOT forwarding alone makes it fail again - the two links are AND-gated, so no black-box test can separate them further.

`test_unscoped_spot_position_reports_are_empty_without_wallet_request` caches `ETHUSDT-SPOT` and `ETHUSDC-SPOT` against a single ETH balance and asserts both an empty result and zero wallet requests. Against the unfixed code it produces two reports each carrying the whole balance.

`test_spot_position_report_short_from_borrowed_balance` moves to a scoped request with its assertions unchanged, so the `spot_borrow` subtraction and SHORT side from `5634b4636d` stay pinned on the branch that still supports them. It passed before this change and still passes; it pins unchanged behaviour rather than acting as a regression test.

### DIFF
--- a/crates/adapters/bybit/src/config.rs
+++ b/crates/adapters/bybit/src/config.rs
@@ -224,9 +224,10 @@ pub struct BybitExecutionClientConfig {
     pub recv_window_ms: u64,
     /// Optional account identifier to associate with the execution client.
     pub account_id: Option<AccountId>,
-    /// Whether scoped SPOT position requests derive positions from the wallet balance for the
-    /// requested pair. Bulk requests fail because wallet coin balances cannot be attributed to a
-    /// pair and therefore cannot establish that no position exists.
+    /// Whether scoped execution-client SPOT position requests derive positions from wallet
+    /// balances. The HTTP client rejects enabled unscoped SPOT requests because balances cannot be
+    /// attributed to pairs. The execution client omits SPOT from bulk requests and reports its bulk
+    /// coverage as unavailable.
     #[builder(default)]
     pub use_spot_position_reports: bool,
     /// Whether to automatically repay SPOT margin borrows after BUY orders tracked by
--- a/crates/adapters/bybit/src/config.rs
+++ b/crates/adapters/bybit/src/config.rs
@@ -224,7 +224,9 @@ pub struct BybitExecutionClientConfig {
     pub recv_window_ms: u64,
     /// Optional account identifier to associate with the execution client.
     pub account_id: Option<AccountId>,
-    /// Whether to generate position reports from wallet balances for SPOT positions.
+    /// Whether scoped SPOT position requests derive positions from the wallet balance for the
+    /// requested pair. Bulk requests fail because wallet coin balances cannot be attributed to a
+    /// pair and therefore cannot establish that no position exists.
     #[builder(default)]
     pub use_spot_position_reports: bool,
     /// Whether to automatically repay SPOT margin borrows after BUY orders tracked by
--- a/crates/adapters/bybit/src/execution.rs
+++ b/crates/adapters/bybit/src/execution.rs
@@ -285,6 +285,27 @@ impl BybitExecutionClient {
         !matches!(product_type, BybitProductType::Spot)
     }
 
+    async fn generate_bulk_position_status_reports(
+        &self,
+        product_types: Vec<BybitProductType>,
+    ) -> anyhow::Result<Vec<PositionStatusReport>> {
+        let mut reports = Vec::new();
+
+        for product_type in product_types {
+            if !Self::provides_bulk_position_coverage_for_product_type(product_type) {
+                continue;
+            }
+
+            let mut fetched = self
+                .http_client
+                .request_position_status_reports(self.core.account_id, product_type, None)
+                .await?;
+            reports.append(&mut fetched);
+        }
+
+        Ok(reports)
+    }
+
     fn resolve_position_idx(
         &self,
         instrument_id: InstrumentId,
@@ -606,9 +627,16 @@ impl ExecutionClient for BybitExecutionClient {
 
     fn provides_bulk_position_coverage(&self, instrument_id: InstrumentId) -> bool {
         // Resolve the suffix directly rather than through `get_product_type_for_instrument`, whose
-        // Linear fallback would claim coverage for an identifier this adapter cannot classify.
-        BybitProductType::from_suffix(instrument_id.symbol.as_str())
-            .is_some_and(Self::provides_bulk_position_coverage_for_product_type)
+        // Linear fallback would claim coverage for an identifier this adapter cannot classify. A
+        // recognized derivative suffix is still uncovered when its product type is unconfigured,
+        // since the bulk loop only queries `product_types()`.
+        let Some(product_type) = BybitProductType::from_suffix(instrument_id.symbol.as_str())
+        else {
+            return false;
+        };
+
+        self.product_types().contains(&product_type)
+            && Self::provides_bulk_position_coverage_for_product_type(product_type)
     }
 
     async fn connect(&mut self) -> anyhow::Result<()> {
@@ -1073,34 +1101,19 @@ impl ExecutionClient for BybitExecutionClient {
         &self,
         cmd: &GeneratePositionStatusReports,
     ) -> anyhow::Result<Vec<PositionStatusReport>> {
-        let mut reports = Vec::new();
-
         if let Some(instrument_id) = cmd.instrument_id {
             let product_type = self.get_product_type_for_instrument(instrument_id);
-            let mut fetched = self
-                .http_client
+            self.http_client
                 .request_position_status_reports(
                     self.core.account_id,
                     product_type,
                     Some(instrument_id),
                 )
-                .await?;
-            reports.append(&mut fetched);
+                .await
         } else {
-            for product_type in self.product_types() {
-                if !Self::provides_bulk_position_coverage_for_product_type(product_type) {
-                    continue;
-                }
-
-                let mut fetched = self
-                    .http_client
-                    .request_position_status_reports(self.core.account_id, product_type, None)
-                    .await?;
-                reports.append(&mut fetched);
-            }
+            self.generate_bulk_position_status_reports(self.product_types())
+                .await
         }
-
-        Ok(reports)
     }
 
     async fn generate_mass_status(
@@ -1130,7 +1143,6 @@ impl ExecutionClient for BybitExecutionClient {
             .map_err(|e| anyhow::anyhow!("{e}"))?;
 
         let position_reports_fut = async {
-            let mut reports = Vec::new();
             let product_types = self.product_types();
             let skip_spot = product_types.iter().any(|product_type| {
                 !Self::provides_bulk_position_coverage_for_product_type(*product_type)
@@ -1142,19 +1154,8 @@ impl ExecutionClient for BybitExecutionClient {
                 );
             }
 
-            for product_type in product_types {
-                if !Self::provides_bulk_position_coverage_for_product_type(product_type) {
-                    continue;
-                }
-
-                let mut fetched = self
-                    .http_client
-                    .request_position_status_reports(self.core.account_id, product_type, None)
-                    .await?;
-                reports.append(&mut fetched);
-            }
-
-            anyhow::Ok(reports)
+            self.generate_bulk_position_status_reports(product_types)
+                .await
         };
 
         let (order_reports, fill_reports, position_reports) = tokio::try_join!(
--- a/crates/adapters/bybit/src/execution.rs
+++ b/crates/adapters/bybit/src/execution.rs
@@ -279,6 +279,12 @@ impl BybitExecutionClient {
         })
     }
 
+    const fn provides_bulk_position_coverage_for_product_type(
+        product_type: BybitProductType,
+    ) -> bool {
+        !matches!(product_type, BybitProductType::Spot)
+    }
+
     fn resolve_position_idx(
         &self,
         instrument_id: InstrumentId,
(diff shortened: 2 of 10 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR. The fan-out diagnosis and scoped path look sound, but the enabled unscoped behavior is not safe at the reconciliation boundary.

`ExecutionManager` builds periodic position-report requests with `instrument_id=None`. With SPOT reports enabled, this path returns `Ok([])`, so the Bybit client is recorded as successfully queried. For an open cached SPOT position, reconciliation treats that empty set as venue quantity zero and can generate a synthetic closing fill. The shipped Bybit execution tester enables both `use_spot_position_reports` and periodic position checks, so this is a supported path.

Could you make the unscoped path preserve unknown coverage rather than return a successful empty set, and add a regression test showing that an open SPOT position is not flattened when pair attribution is unavailable? The config and integration docs should also state the final scoped and bulk behavior.

folknor:
Thank you.

Fixed. The unscoped path now returns an error rather than a successful empty set, so `ExecutionManager` records the client in `failed_clients` and skips reconciliation for its position keys instead of reading absence as flat.

## Where the error is raised

`BybitHttpClient::request_position_status_reports` fails for an enabled unscoped SPOT request before any wallet fetch. `BybitExecutionClient::generate_position_status_reports` preflights the same condition ahead of its product-type loop, so the outcome does not depend on where SPOT falls in the configured ordering and no derivative request is issued only to be discarded by the `?`.

## Startup reconciliation needed an exception

Raising the error from `generate_position_status_reports` alone would have failed node startup: `generate_mass_status` builds an unscoped position command and drives it through `tokio::try_join!`, so the error discards the order and fill reports too, and `perform_startup_reconciliation` treats a mass-status error as fatal.

The two paths draw opposite inferences from a missing report. Periodic reconciliation walks the cached open positions and treats a key with no venue report as flat. Startup is report-driven throughout - `reconcile_mass_status` and `adjust_mass_status_fills` both iterate `mass_status.position_reports()`, and nothing walks cached positions looking for missing ones - so an omitted SPOT report flattens nothing there. `generate_mass_status` therefore skips only SPOT and keeps the derivative reports, which is the pre-PR startup behaviour.

## The cost, and a question

With the option enabled alongside a derivative product type, every periodic position cycle now skips reconciliation for the whole Bybit client, derivatives included. `generate_position_status_reports` returns `Result<Vec<PositionStatusReport>>` and the manager's coverage bookkeeping is per client, so nothing can express that the derivative reports are authoritative while SPOT coverage is unknown.

If you would rather keep derivative coverage, one option is to group open cached SPOT instruments by base currency, use the scoped wallet path where a base has exactly one open pair, and fail only where a base is ambiguous. I did not implement it because it changes what a bulk request means, which is your call.

Separately, and pre-existing: `ExecutionMassStatus::new` defaults `reports_complete` to `true` and the Bybit client never calls `set_report_window`, so a mass status omitting SPOT overstates its own completeness. Left alone here, since changing it would alter startup fill projection through `order_only_venue_order_ids`.

## Testing

`test_unscoped_spot_position_reports_fail_without_wallet_request` pins the error and that no wallet request is issued. `test_exec_client_mixed_unscoped_spot_reports_fail_before_derivative_request` pins the preflight with no position request. `test_exec_client_mass_status_omits_spot_and_preserves_derivative_positions` pins mass status succeeding with the linear position report present. `test_position_reconciliation_preserves_spot_position_when_client_query_fails` seeds an open cached SPOT position, reports the responsible client as queried and failed, and asserts no synthetic closing fill - that arm had no coverage, every existing test passes an empty `failed_clients`.

Each new assertion fails against the unfixed code: restoring `Ok(Vec::new())` fails the error assertion; removing the preflight issues two derivative requests before failing; forcing the mass-status skip off reproduces the startup failure; and reporting the client as succeeded rather than failed produces the synthetic closing fill.

Scoped SPOT behaviour is unchanged. The config rustdoc and the integration table row now state both the scoped and bulk behaviour.

cjdsellers:
@folknor, thanks for laying out the trade-off so clearly. I think the mixed-product consequence is too broad for this to merge as-is.

`ExecutionManager` sends one bulk request per client. With SPOT enabled, the preflight in `generate_position_status_reports` returns before the product loop, so LINEAR and INVERSE reports are never requested. The runner then marks the whole Bybit client as failed, and reconciliation skips every derivative key assigned to it. The previous bulk path skipped SPOT while keeping derivative checks, and mixed SPOT/derivative clients are a documented configuration.

Could you keep derivative position reconciliation active while preserving unknown coverage for SPOT? If the current client-level coverage model cannot represent that split, I think we need to settle the coverage representation before this merges.

One smaller API point: `BybitHttpClient::request_position_status_reports` and its Python binding now intentionally fail for enabled unscoped SPOT requests, but their `# Errors` docs still mention only request failure. Could you document that condition on both methods?

folknor:
Thank you.

Fixed, and it did need the coverage representation rather than a change inside the adapter. A client's bulk request can now cover some of its position keys and not others.

## The representation

`ExecutionClient` gains a defaulted `provides_bulk_position_coverage(instrument_id) -> bool`, alongside `handles_order_venue`. It answers whether an absent report from a bulk request is evidence the position is flat, not whether the client can report on the instrument at all - Bybit can produce a scoped SPOT report, it just cannot establish that a cached SPOT pair missing from a bulk response is closed. The default is `true`, so no other adapter changes and the report-generation signature is untouched.

`resolve_position_report_client_coverage` consults it per key, and `ReportClientCoverage` gains an `Unavailable(clients)` variant between `Resolved` and `Unresolved`. `Unavailable` keeps the full responsible-client set rather than dropping the incapable client, which would let the remainder look like complete coverage. `reconcile_position_reports` only consults coverage when the venue reports for a key are empty, so a positive report still reconciles under every variant. Open-order reconciliation never constructs `Unavailable` and treats it with `Unresolved`.

Bybit then skips SPOT inside the bulk product-type loop instead of failing ahead of it, so LINEAR and INVERSE are requested and returned as before, the client is no longer recorded as failed, and only the SPOT keys are held as unknown.

## The override is independent of `use_spot_position_reports`

With the flag off, `request_position_status_reports` returns `Ok(Vec::new())` for SPOT, which asserts flat just as quietly. Tying the capability to the flag would encode the claim that absence is evidence in that case. It also fails closed for an instrument ID carrying no recognized product-type suffix, rather than inheriting the `Linear` fallback the order-placement paths use - the shared cache can hold restored or legacy IDs, and a missed reconciliation is cheaper than a synthetic closing fill.

Both `# Errors` sections now state the enabled-unscoped-SPOT condition, in the Rust client and the Python binding.

## Testing

`test_exec_client_mixed_unscoped_position_reports_omit_spot` asserts a post-connect delta of two position requests - the unscoped LINEAR path queries once per settle coin - and zero wallet balance requests, which is what shows SPOT was omitted. `test_exec_client_bulk_position_coverage_by_product_type` pins LINEAR, INVERSE, SPOT and a suffix-less ID. `test_prepare_position_report_check_resolves_mixed_bulk_coverage` pins one client resolving a derivative key to `Resolved` and a SPOT key to `Unavailable`. `test_position_reconciliation_preserves_unavailable_spot_coverage` reconciles both keys with no reports and asserts the derivative fill is generated while the SPOT position is left open.

Each new assertion fails against the previous revision: disarming the coverage predicate produces the synthetic SPOT closing fill, and disarming the SPOT skip in the bulk loop moves the wallet request count to one.

`test_position_reconciliation_preserves_spot_position_when_client_query_fails` is unchanged and still pins the transient case, where a responsible client fails its query outright.

folknor:
@cjdsellers Let me know if you want to re-trigger the CI here to pick up the "performance" CI fixes and rerun that. Otherwise CI is in fact green.

cjdsellers:
@folknor, thanks for working through the coverage split. The per-instrument representation is the right shape, but it still does not reach the live periodic path.

`LiveNode::start_position_report_check` passes `LiveExecutionClient` facades into `prepare_position_report_check`. The facade forwards the neighboring routing and tolerance methods, but it does not forward `provides_bulk_position_coverage`, so it inherits the default `true`. A cached Bybit SPOT position is therefore still recorded as covered; the bulk request omits SPOT, and reconciliation can synthesize the same closing fill this revision is intended to prevent. Could you forward the predicate through `LiveExecutionClient` and add a regression test that exercises the facade or node path?

There is a second coverage edge after that is fixed: the Bybit override returns `true` for any derivative suffix, even when that product type is absent from `self.product_types()`. For example, a Spot-only client with a restored LINEAR position does not query LINEAR but still declares complete coverage, so absence can again be read as flat. The predicate needs to include the configured product types, including the empty-list Linear default.

Two smaller points: `generate_mass_status` now duplicates the unscoped position-report product loop, and the `use_spot_position_reports` config rustdoc says bulk requests fail while the execution client actually succeeds and omits SPOT.

folknor:
Thank you.

Fixed.

## The facade

`LiveExecutionClient` forwards every other `ExecutionClient` method, including the two defaulted ones beside it. Round 2 added a third defaulted method and did not forward it, so `LiveNode::start_position_report_check` - which casts these facades to `&dyn ExecutionClient` before handing them to `prepare_position_report_check` - took the default `true` for every client. The forward is now in place.

Rust gives no warning when an implementation inherits a newly added default method, so nothing prevents the next one being missed the same way. Making capability predicates required trait methods would prevent it, at the cost of a change to every adapter implementing `ExecutionClient`. I am happy to do that separately if you want it.

## The configured product types

`provides_bulk_position_coverage` now requires the resolved product type to be in `product_types()` as well as bulk-eligible, so a Spot-only client no longer claims coverage for a restored LINEAR position. A suffix-less identifier stays uncovered even under the empty-list Linear default: the bulk loop does query the Linear category there, but the adapter cannot classify that particular cached instrument as Linear.

## The duplicated loop

`generate_bulk_position_status_reports` now holds the only bulk product-type loop, and takes the effective set as a parameter rather than calling `product_types()` itself, so the predicate and the request loop are driven by one value rather than by two calls that agree because the config is immutable. `generate_position_status_reports` keeps its scoped branch; `generate_mass_status` computes the set, emits the SPOT-unavailable warning, then calls the same helper. The warning stays at the mass-status call site, so the periodic path does not begin warning every cycle.

The `use_spot_position_reports` rustdoc now separates the HTTP client's rejection of an enabled unscoped SPOT request from the execution client's omission of SPOT.

## One consequence

A Spot-only Bybit client now provides no bulk position coverage for any key, so nothing it owns is reconciled from a bulk cycle. That follows from the client querying nothing in bulk, but it is a behaviour change for that configuration.

## Testing

`test_prepare_position_report_check_uses_live_client_bulk_coverage` wraps the stub in a `LiveExecutionClient` and drives the manager through `&dyn ExecutionClient`. It lives in-crate because the facade is `pub(crate)`. Removing the forward turns the SPOT key from `Unavailable` to `Resolved`, reproducing the synthetic closing fill.

`test_exec_client_bulk_position_coverage_by_product_type` is now a sixteen-case cross product of four effective configurations against LINEAR, INVERSE, SPOT and a suffix-less identifier. Reverting the predicate to its suffix-only form fails exactly the five over-claiming cases, including the Spot-only LINEAR case you named.

`test_position_reconciliation_preserves_unavailable_spot_coverage` is unchanged and still drives the concrete client through reconciliation end to end.

cjdsellers (approved):
@folknor, I've had another look at the latest changes, and this looks good from my side. Thanks for the final follow-up: `LiveExecutionClient` now forwards the coverage capability, and the configured product-type check preserves unknown coverage for product types the client does not query.

I don't think making the capability predicates required trait methods needs to be part of this PR; the facade regression test pins the contract here.

PR #4847, opened 2026-08-25
Warn on an open order excluded by the lookback window

folknor:
A follow-up to the missing-order resolution work in #4479.

## An open order outside the lookback window is excluded without a trace

`ExecutionManager::reconcile_open_order_reports` resolves cached open orders the venue did not report. When `open_check_open_only` is `false`, the candidate list is filtered to `o.ts_last() >= cutoff`, where `cutoff` is `now - open_check_lookback_mins`. The filter is correct: absence from a bounded bulk response is not evidence of absence at the venue, and inferring "missing" from it would manufacture false findings.

What the filter also does is remove those orders from the targeted ladder. An excluded order never reaches `prepare_missing_order_query`, so it never becomes a `targeted_candidate` and never receives the per-order query that does not depend on a time window. The exclusion is silent: every other exclusion in that loop warns - unresolved coverage, responsible clients not queried, responsible clients failed - and each is deduplicated through `missing_order_coverage_warnings`. The age exclusion produced no log at any level, so an order that stops being reconciled leaves no evidence that it stopped.

This reaches restart-inherited orders in practice. An order the strategy still expects has its `ts_last()` refreshed by ongoing activity; residue adopted at startup carries no such expectation, ages past the window, and then falls out of reconciliation permanently.

## Warn once per exclusion episode

An order excluded by the cutoff and absent from `venue_reported_ids` now emits a single warning naming it and stating that absence from the bulk response cannot be treated as evidence. The positive-report loop is not age-filtered, so an old order the venue does return is already reconciled normally and would be described as skipped without that second condition.

Tracking is a sibling `IndexSet`, not the existing `missing_order_coverage_warnings`, which is cleared when coverage becomes resolvable - sharing it would let either diagnostic suppress or re-arm the other. It is cleared on a positive report through both identification paths, in `clear_recon_tracking`, and pruned against the open-order snapshot in `prepare_open_order_report_check` alongside its two siblings. An order that ages out, recovers and ages out again warns again.

The candidate list is unchanged: every order at or above the cutoff is pushed in the same order as the previous `filter`. No events, no status changes, no new venue requests.

## Why the targeted ladder is not extended

The obvious repair is to let age-excluded orders reach the targeted ladder, since a targeted query names one `client_order_id` and is not time-bounded. It rests on an assumption the tree does not support. An empty targeted response drives `resolve_missing_order`, which marks an `Accepted` or `Submitted` order `REJECTED` and a `PartiallyFilled` order `CANCELED` after retries, and no adapter documents whether an empty targeted order-status response means the order is absent at the venue or only that it is outside what that adapter can answer. Under the second reading, extending the ladder closes live resting orders.

If you want those orders resolved, one potential solution would be to query them and treat the result as positive evidence only - reconcile a matching report, treat `None` as permanently inconclusive rather than as absence. That needs an explicit split between absence-authoritative and positive-evidence-only queries, so the semantics are your call.

## Testing

`test_open_check_lookback_exclusion_warns_once_without_reconciliation_actions` drives two consecutive cycles against an empty bulk response and asserts no events, no targeted queries, the order still `Accepted`, the retry, query-recency and targeted-query state not advanced, and the warning set holding exactly one entry across both cycles.

`test_open_check_lookback_warning_clears_at_boundary_and_rearms` moves the same order across the cutoff and back, asserting it leaves the warning set at the boundary and re-warns after aging out again. It pins the unchanged selection boundary by asserting the boundary order lands in `missing_order_coverage_warnings`, which only happens if it traversed the candidate loop.

`test_positive_report_clears_old_order_lookback_warning_without_reinserting_it` and `test_venue_order_id_mapped_report_clears_old_order_lookback_warning` cover both identification paths, the latter with `client_order_id` unset so the report resolves through the cache's venue order ID mapping.

The new assertions fail against the unfixed implementation: removing the clear on the at-or-above-cutoff branch fails the re-arm assertion, and removing the `venue_reported_ids` condition fails the positive-report test by re-inserting an order the venue had just reported.


cjdsellers (dismissed):
Hey @folknor,

Thanks for the PR. This looks good to merge, and keeping the change diagnostic-only rather than extending the targeted-query ladder looks right.

**One small non-blocking test detail** if you get a chance: the new `GenerateOrderStatusReports::new` fixtures put the cutoff in `end`, while production puts it in `start` and mirrors it in `OpenOrderReportCheck.start`. Could you align those constructors so the fixtures preserve the production invariant?

folknor:
Thank you.

Fixed. The four `GenerateOrderStatusReports::new` fixtures now put the cutoff in `start` and leave `end` as `None`, matching `prepare_open_order_report_check` and the `OpenOrderReportCheck.start` each one already mirrors.

No assertion changed: `reconcile_open_order_reports` reads `check.start`, so the command's own bounds were never consulted. The fixtures now carry the production invariant rather than a shape that happened not to be read.

cjdsellers (approved):
Hey @folknor,

Thanks for the follow-up. I've had another look at the latest changes, and the updated fixtures now preserve the production `start` and `end` bounds. This looks good to merge.

PR #4849, opened 2026-08-25
Await betfair stream lifecycle transitions in tests

folknor:
A follow-up to the betfair test determinism work in #4837.

## The stream tests wait on a wall-clock deadline instead of on the event

`test_market_subscription_lifecycle_degrades_and_recovers` intermittently fails with "Timeout waiting for condition after 2.0s" at its first wait for `StreamLifecycleState::Degraded`. Every wait in `tests/stream_client.rs` polls a predicate every 100ms against a 2s budget, and asserts on elapsed wall-clock time.

`#[tokio::test]` without a `flavor` builds a current-thread runtime, so the test body, the mock server task and the socket read and write tasks are a single schedulable unit. When a parallel build competes for the machine, that unit can lose the CPU for seconds at a stretch and the budget expires while nothing has gone wrong: the passing path completes the whole test in under a second, and the failing path is the same work displaced past the deadline. Widening the budget only moves the threshold.

The adapter itself is not implicated. The 503 frame travels from the mock server's write through the socket read loop into the betfair message handler, where `ProtocolLifecycle::on_change` stores `Degraded` unconditionally and returns; the predicate is a plain atomic load. No channel, worker or timer sits between decoding the frame and storing the state, neither mutex on the path is held across an await, and the 503 branch returns before invoking the data callback. The intended ordering also rules out a missed transient: the server blocks until the test signals recovery, which it only does after observing `Degraded`.

## Wait on the transition

`ProtocolLifecycle`'s three lifecycle fields become a `LifecycleState` holding the existing `AtomicU8` alongside a `watch::Sender<()>` used purely as a change notifier. `set` stores the state with `Release` then advances the notifier; `get` is the previous `Acquire` load. The atomic remains the single authoritative representation, and because the channel payload is `()` there is no second copy of the state to drift.

The lifecycle read on the market and order message paths stays a plain atomic load, so the per-frame cost is unchanged.

`wait_for_authentication_state`, `wait_for_market_subscription_state` and `wait_for_order_subscription_state` subscribe before the first check and then loop, which closes the gap between observing the state and awaiting the next change: a transition landing in that window has already advanced the notifier, so `changed()` returns immediately. The tests await these instead of polling.

The waits carry no deadline of their own. Each logical phase is wrapped in a 15 second bound that reports the expected and observed state, so a real regression fails with a diagnosis rather than hanging until the harness intervenes. That bound is a circuit breaker rather than synchronization: it sits far above the scheduling gaps that caused the flake, and below the harness slow-test thresholds so the diagnostic can actually fire.

The `tokio::time::timeout` calls around real futures are unchanged. They already await an event, and their bounds are what turn a failure into a failure rather than a hang.

## Testing

The converted file's assertions are unchanged; only the waiting is different. `test_market_subscription_lifecycle_degrades_and_recovers` passes in 0.32s.

Suppressing the notification in `LifecycleState::set` while leaving the atomic store in place hangs the test rather than passing it, which establishes that the tests observe the transition through the notifier and are not succeeding because the expected state happened to be present at the first check.

`tests/exec_client.rs` and `tests/data_client.rs` use the same polling pattern and are left for a follow-up. Their execution-client waits need separate treatment: `is_reconciling` can reuse the generation already published by `ReconciliationGate`, while `is_connected` is a composite predicate with a side effect on the success path and wants its own design rather than a watched mirror.

### DIFF
--- a/crates/adapters/betfair/src/stream/client.rs
+++ b/crates/adapters/betfair/src/stream/client.rs
@@ -99,14 +99,38 @@ impl StreamLifecycleState {
 }
 
 #[derive(Debug)]
+struct LifecycleState {
+    value: AtomicU8,
+    changed: watch::Sender<()>,
+}
+
+impl LifecycleState {
+    fn new(state: StreamLifecycleState) -> Self {
+        Self {
+            value: AtomicU8::new(state as u8),
+            changed: watch::channel(()).0,
+        }
+    }
+
+    fn get(&self) -> StreamLifecycleState {
+        StreamLifecycleState::from_atomic(&self.value)
+    }
+
+    fn set(&self, state: StreamLifecycleState) {
+        self.value.store(state as u8, Ordering::Release);
+        self.changed.send_replace(());
+    }
+}
+
+#[derive(Debug)]
 struct ProtocolLifecycle {
     transport_connected: AtomicBool,
-    authenticated: AtomicU8,
-    market: AtomicU8,
+    authenticated: LifecycleState,
+    market: LifecycleState,
     market_was_current: AtomicBool,
     market_requires_image: AtomicBool,
     market_image_tainted: AtomicBool,
-    order: AtomicU8,
+    order: LifecycleState,
     order_was_current: AtomicBool,
     order_requires_image: AtomicBool,
     order_image_tainted: AtomicBool,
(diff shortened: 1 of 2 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. The new lifecycle wait methods are currently only used by our tests, so their wider use remains somewhat speculative. Even so, I think this is a reasonable refinement: it keeps state reads lock-free and makes lifecycle waiting event-driven.

It also fixes `test_market_subscription_lifecycle_degrades_and_recovers`, which is currently flaky in CI due to its short polling deadline. On that basis, this looks good to merge :ok_hand:

Issue #4564, opened 2026-07-24
[Interactive Brokers] Detect duplicate live orders after modify during connectivity loss


davidgreyme:
## Bug report

**Adapter:** Interactive Brokers
**Version:** `1.231.0a20260709` (develop nightly), Python 3.13, Linux (Docker), gnzsnz `ib-gateway` (paper)

### Summary

An order modified with a normal in-place `placeOrder` (same IB `orderId`) **while the TWS↔IBKR connectivity was degraded/lost** was resolved by IB as a **new working order with a new `permId`, while the original `permId` stayed live** — two real broker orders from one modify. The adapter cannot represent "one client order → two broker incarnations", so:

1. On reconnect, IB's open-order snapshots replayed **both** incarnations; `_on_open_order` overwrites the cached `venue_order_id` last-writer-wins (`venue_order_id_modified=True`), and the **stale original** incarnation arrived last, leaving the engine tracking the wrong order.
2. A later `cancel_order` resolved and cancelled only that one incarnation. IB confirmed `OrderCanceled` — but the *modified* incarnation kept working at the broker, invisible to the engine.
3. Six hours later the surviving incarnation **filled** (4 contracts, a protective stop → an unintended position flip). The fill report carried the surviving `permId`, so `Order.apply` raised `ValueError('self.venue_order_id' ... was not equal to 'event.venue_order_id' ...)` against the CANCELED order, and the engine fell back to inferred-fill NET reconciliation (which healed positions correctly — but the harm was the orphaned order executing at all).

### Timeline (UTC, 2026-07-24, paper account, MESU6.CME)

```
06:10:03  submit STOP_MARKET BUY 4 @ 7463.75 GTC  -> OrderAccepted venue_order_id=PERM-253949451 (IB orderId=187)
06:14:08  ERROR InteractiveBrokersClient: Not connected (code 504)
06:14:13  ERROR Client failed to initialize; connection timeout
06:15:08  ModifyOrder trigger 7463.75 -> 7443.75; adapter logs exactly ONE
          "Placing 136287796088976: 187,0,0: STP BUY 4@ GTC"  (in-place, same orderId=187)
06:15:14+ modify never confirms; engine in-flight check queries every ~2s
06:30:05  ERROR Connectivity between IBKR and TWS has been lost (code 1100); watchdog reconnect
[...]

PR #4852, opened 2026-08-25
Fix IBKR restored-order cancel routing

gapurov:
  and, if I used AI,
  this draft
  limitation below

## Summary

Canceling an IBKR order restored with a `PERM-*` venue ID resolves the current raw IB order ID, but
the adapter did not rebuild the tracking maps used by status callbacks. Because `CancelOrder` actor
IDs identify the requester and can differ from the target order owner, this change resolves and
validates the cached target before spawning, then caches its trader, strategy, and instrument
identity before sending the cancel. The command client and venue IDs still drive broker resolution,
and the raw-to-client route is published last so `PendingCancel` and `Cancelled` callbacks see
complete target identity.

## Related issues/PRs

Related to #4564.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not applicable.

## Documentation

No documentation change is required because this corrects internal adapter routing and changes no
public API.

## Testing

Local validation completed:

- `make format`
- `make pre-commit`
- `cargo test -p nautilus-interactive-brokers`

The adapter suite passed 414 unit tests, one connection test, and doc tests. Regression coverage
includes same-strategy and cross-strategy requesters, target trader and strategy routing,
instrument and venue identity conflicts, `PendingCancel` and `Cancelled` callbacks, and terminal
tracking cleanup.

### DIFF
--- a/crates/adapters/interactive_brokers/src/execution/core.rs
+++ b/crates/adapters/interactive_brokers/src/execution/core.rs
@@ -1617,6 +1617,7 @@ impl ExecutionClient for InteractiveBrokersExecutionClient {
         let client = self.ib_client.as_ref().context("IB client not connected")?;
 
         let order_id_map = Arc::clone(&self.order_id_map);
+        let venue_order_id_map = Arc::clone(&self.venue_order_id_map);
         let instrument_id_map = Arc::clone(&self.instrument_id_map);
         let trader_id_map = Arc::clone(&self.trader_id_map);
         let strategy_id_map = Arc::clone(&self.strategy_id_map);
@@ -1632,6 +1633,7 @@ impl ExecutionClient for InteractiveBrokersExecutionClient {
                 &cmd,
                 &client_clone,
                 &order_id_map,
+                &venue_order_id_map,
                 &instrument_id_map,
                 &trader_id_map,
                 &strategy_id_map,
@@ -2056,6 +2058,7 @@ impl InteractiveBrokersExecutionClient {
         cmd: &CancelOrder,
         client: &Arc<Client>,
         order_id_map: &Arc<Mutex<AHashMap<ClientOrderId, i32>>>,
+        venue_order_id_map: &Arc<Mutex<AHashMap<i32, ClientOrderId>>>,
         instrument_id_map: &Arc<Mutex<AHashMap<i32, InstrumentId>>>,
         trader_id_map: &Arc<Mutex<AHashMap<i32, TraderId>>>,
         strategy_id_map: &Arc<Mutex<AHashMap<i32, StrategyId>>>,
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @gapurov,

Thanks for the PR. I've rebased the branch onto the latest `develop` and started CI. I found three cancel-routing cases that need to be addressed before this can merge:

- When a different strategy requests the cancel, definitive failure paths build `OrderCancelRejected` with the requester identity. The target is already `PendingCancel` under its owning strategy, so it cannot apply that event and return to its previous state. The rejection needs to use the target order identity.
- Publishing `venue_order_id_map` last only protects a new raw route. If recovery has already populated that route without the actor mappings, `handle_order_status` can observe incomplete identity and discard the update. The update needs equivalent protection when the route already exists.
- `OrderPendingCancel` uses the raw IB order ID instead of the cached `PERM-*` ID, which makes the venue identity inconsistent across restored-order events.

Could you address these cases when you get a chance? The regression tests also need to cover the cross-strategy rejection path, the pre-existing route case, and the pending-cancel venue ID before this can merge.

gapurov:
@cjdsellers 

Thanks for the review. I pushed an update that addresses these three cases:
- Cancel rejections now use the cached target order identity, including cross-strategy requests.
- Cancel tracking blocks callback lookup while it installs all identity mappings, including when the raw route already exists.
- OrderPendingCancel now keeps the authoritative cached `PERM-*` venue order ID.

cjdsellers:
Hi @gapurov,

Thanks for the follow-up. I've had another look at the latest changes. The three cancel-routing cases from my earlier review are fixed, and the regression coverage now exercises the cross-strategy rejection, pre-existing route, and `PERM-*` pending-cancel paths. This looks good from my side.

I noticed two small non-blocking follow-ups: `venue_map.remove(&ib_order_id)` is redundant while the venue-map guard is held and can delete a valid route if a later identity-map lock is poisoned, and the new venue-first lock ordering rule should be documented so the instrument-map/venue-map inversion isn't reintroduced. If you're able, would you be open to folding those in? Could you also validate the restored-order path against an IBKR paper session if/when you get the chance?

cjdsellers (approved):

cjdsellers:
Hi @gapurov,

I've taken a closer look at the two code follow-ups I mentioned and I'm withdrawing them. The existing `venue_map.remove(&ib_order_id)` is necessary because it hides a pre-existing route while the supporting identity maps are rebuilt, so a later lock failure cannot expose partial identity. The lock-order point would only lead to a comment-only change, which isn't worth a standalone diff. The merged implementation is right as it stands 👍

PR #4813, opened 2026-08-20
Expose matching-engine config on sandbox execution

graceyangfan:
# Pull Request

**NautilusTrader can execute live trades involving real capital. Pull requests are held to a very
high standard for correctness, reliability, testing, clarity, and maintainability.**

- [ ] A maintainer agreed on the problem and approach in an issue, or this is a small,
  self‑contained fix that does not need prior discussion
- [x] I have read and followed
  [CONTRIBUTING.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md)
  and, if I used AI,
  [AI_POLICY.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/AI_POLICY.md)
- [x] I understand and can explain every submitted change and all information in this PR description
- [x] This change is complete, locally validated, and ready for review, or a maintainer requested
  this draft
- [x] I ran `make format`, then ran `make pre-commit` locally and confirmed it passed, or I
  described an agreed limitation below
- [x] I ran all relevant tests locally, no tests apply to this change, or I described an agreed
  limitation below
- [x] I have not modified `RELEASES.md` (maintainers keep it current to avoid merge conflicts)

## Summary

Sandbox already uses `OrderMatchingEngine` and already routes `TradeTick`s into it, but
`SandboxExecutionClientConfig` did not expose the matching flags backtest venues already have.
Paper trading therefore always used `queue_position=false`, `liquidity_consumption=false`, and a
default fill model, so LIMIT fills on a print were more optimistic than a queue-aware backtest of
the same trades.

This PR wires those existing knobs onto sandbox config. Defaults stay off, so current sandbox
runs do not change.

Review follow-up: the sandbox client now owns one `FillModelHandle` and clones that handle into
each matching engine, matching `SimulatedExchange`. Cloning `FillModelAny` per instrument rebuilt
the probabilistic RNG from `random_seed`. `crates/backtest/src/python/config.rs` now imports
`fill_model_any_to_pyobject` from the execution crate. Latest `develop` is merged; factory tests
keep both the custom fee-model coverage and the matching-knob tests.

## Related issues/PRs

- Closes #4812
- Related to #1677 (sandbox latency; not in this PR)
- Related to #4806 (Python fee-model converter width; fee models already land on develop)
- Related to #4631 (closed: land matching pieces on the existing engine, no new realism layer)

## Type of change

- [x] New feature (non-breaking)
- [x] Documentation update

## How to review

[...]

Issue #1677, opened 2024-05-29
Latency modeling for sandbox/real-time


cjdsellers:
# Feature Request

A `use_message_queue` config option was added to `SimulatedExchange` as a temporary work around for processing trading commands in real-time cbfc7035aa62df6cdf6c45cf3621cf3d0eea4faf to accommodate the "sandbox mode".

This means that currently when trading commands are submitted to the sandbox execution client (and onto its internal simulated exchange) there is practically no latency in either direction. The commands will be processed and events generated at local platform latency (microseconds) which is relatively unrealistic for most venue and setups.

A future feature could leverage the `LatencyModel` to extend to real-time, potentially by using a timer to "delay" command processing and events being generated - to achieve to desired latency model spec.

TheLegendAli:
yup we need something life that, for my backtest I'm implementing few latencies:

- latency that I receive the data

- latency for nautilus to do its magic

- latency for exchange to receive order from nautilus

I'm manual using "timer" to add delay. But would be nice if we can automate it

ChrisAB:
Was wondering if this is still needed and if I could pick it up? @TheLegendAli

TheLegendAli:
i think so. I would use it everyday :)

ChrisAB:
According to https://github.com/nkaz001/hftbacktest/blob/master/docs/latency_models.rst
This seems like two PRs instead of 1?
My question is, currently, response latency is missing in latency model. Should this be added? Should this also take base_latency into account or no? Cause if it does, I think it will break existing configurations as it will add base_latency into every call.
Or should this issue only focus on actually adding latency model to sandbox?

TheLegendAli:
Hi Chris,

Sorry for the late reply, I was away for a bit. The latency model is everything I asked for and then some — really nice work.

To your questions:

**Response latency:** I don't need it personally, but I'd assume others will, so worth including.

**base_latency:** my instinct is that it's a one-way figure, which would mean it applies to both legs. If I read you correctly, this will break the current config? If so, would it be cleaner to just split it into two explicit parameters?

1. a `base latency` for sending orders
2. a separate `response_latency` for the fill coming back

**Sandbox only:** yes, i would imagine we only need it in sandbox.

Let me know if I answered your questions

Thanks again for picking this up.

ChrisAB:
No problem, I was also away for quite a bit. I also took a look at the discussion created by vedjaw.

[...]

Issue #3732, opened 2026-03-18
Bug: Rust `ExecutionEngine` panics with `RefCell already borrowed` when `SandboxExecutionClient` submits orders


linimin:
# Bug Report

Use this template only for issues that fit the **Bug** definition.

| Term                          | Definition |
|-------------------------------|------------|
| **Bug**                       | Behavior that contradicts the platform's documented or intended design as per code, docs, or specs. (i.e., the implementation is incorrect.) |
| **Expectation&nbsp;mismatch** | Behavior that follows the platform's documented or intended design but differs from what you expected. (i.e., the design/spec might be the problem.) |
| **Enhancement request**       | A request for new functionality or behavior that is not implied by existing design. (i.e., *"It would be great if the platform could…"*) |

**Note:**

- Submitting this issue automatically applies the `bug` label.
- `bug`-labeled issues are triaged with higher priority because they require corrective implementation work.
- **Expectation mismatches** and design-level concerns should be opened as [Discussions](https://github.com/nautechsystems/nautilus_trader/discussions), or RFCs instead, where they can be validated and discussed to consensus before any work is scheduled.
- The absence of a feature is typically not an expectation mismatch, and should be filed as an enhancement request.

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I've tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

Checking a recent development wheel can save time because the issue may already have been fixed.
You can install a development wheel by running:

```bash
pip install -U nautilus_trader --pre --index-url https://packages.nautechsystems.io/simple
```

See the [development-wheels](https://github.com/nautechsystems/nautilus_trader#development-wheels) section for more details.

### Expected Behavior

When a Rust strategy calls `submit_order()` (or `cancel_order()`, `modify_order()`) inside a sandbox (async runner) environment, the order should be processed through the full order lifecycle (Submitted → Accepted → Filled/Rejected) without crashing, just as it does in backtest mode and live mode.

### Actual Behavior

The process immediately panics with **`RefCell already borrowed`** at `crates/core/src/shared.rs:79:16` whenever any Rust strategy submits an order through the `SandboxExecutionClient` in an async runner context.

This happens because:

[...]

PR #4865, opened 2026-08-26
Add inbound latency modeling to sandbox execution client

ChrisAB:
- [ ] A maintainer agreed on the problem and approach in an issue, or this is a small,
  and, if I used AI,
  this draft
  limitation below

## Summary

`SandboxExecutionClientConfig` gains an optional `latency_model` that defers submit, modify, and
cancel commands by the matching leg of the existing `StaticLatencyModel` before they reach the
matching engine, so a sandbox run reproduces a venue's command latency instead of local process
latency. Venue-generated events (accepts, fills, cancels, expirations) still dispatch synchronously,
so this covers the command half of #1677 only. The default `latency_model=None` is byte-identical to
current behavior.

## Related issues/PRs

Related to #1677. Builds on #4813, which added the `fill_model` field and the runtime-only serde
guard pattern that `latency_model` follows.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

None. The field defaults to `None`, which leaves every existing code path unchanged.

## Documentation

## Testing

`cargo test -p nautilus-sandbox --features python` passes (97 tests), as does
`cargo test -p nautilus-sandbox -p nautilus-execution -p nautilus-backtest` (1482).
`make format` and `make pre-commit` pass. Python: 12 tests in
`python/tests/unit/adapters/sandbox`.

Roughly 70% of the diff is tests: 2,000 lines in `crates/adapters/sandbox/tests/execution.rs`
covering leg selection, FIFO ordering under equal due times, timer-driven and data-driven drains,
the same-pass hold-back, commands overtaking their own submit, unknown-instrument rejection, and
`stop()` / `reset()` teardown.

## Commits

1. **Move latency model converters to execution crate** — pure relocation of
   `pyobject_to_latency_model_any` and `latency_model_any_to_pyobject` out of the backtest crate,
   mirroring where #4813 put the fill-model converters. No behavior change; reviewable on its own.
2. **Add inbound latency modeling to sandbox execution client** — the feature and its tests.

## Design notes

**Leg mapping** follows the model's existing semantics: submit and submit-list use the insert leg,
modify and batch-modify the update leg, cancel / batch-cancel / cancel-all the delete leg.
`query_order` and `query_account` are not deferred, since they are local cache reads.

**Due time is keyed off arrival at the client** (`now + leg`), not `command.ts_init()`. This is an
intentional difference from backtest `generate_inflight_command`: in the sandbox the two can differ
by real wall-clock time spent upstream.

**FIFO within a due timestamp** comes from a monotonic sequence tie-break rather than the backtest's
per-due-timestamp counter map. The map is safe in the backtest because `exchange.rs::process` drains
every due entry in one pass; the hold-back below breaks that precondition here.

**The drain is timer-driven**, from a `LiveClock` alert armed at the earliest queued due time, so a
deferred command lands even when no further market data arrives. Incoming market data also drains
anything already due, which tightens ordering against ticks under timer jitter.

**Immediate and deferred paths share one code path.** The venue-side body of each affected
`ExecutionClient` method moved into an `apply_*` helper on `SandboxInner`; the client method now
only chooses "apply now" or "enqueue". This is most of the churn in `execution.rs` and is relocation
rather than new logic. `OrderSubmitted` is still dispatched immediately on submit as the client's
"sent to venue" record.

Deferral creates two problems that the change has to solve:

- **A command can overtake the submit that would have created its matching engine.** Sandbox engines
  are created lazily on first submit, unlike the backtest exchange, which pre-registers one per
  instrument and can therefore `assert!`. Left alone, a deferred cancel finds no engine and is
  discarded with no event, stranding the order in `PENDING_CANCEL`, the sandbox returns `Ok(None)`
  from `generate_order_status_report`, so nothing resolves it later. The deferred path therefore
  builds the engine from the cached instrument so the venue itself issues the rejection, and
  synthesizes the rejection directly when the instrument is absent and no engine can be built.
  Either way the command is rejected, never silently dropped.
- **Two commands for the same order falling due in one drain pass.** In a live node the sandbox's
  venue events go onto the exec channel (`start()` installs an `event_handler`), so they reach the
  cache only after the runner drains it. The backtest has no such handler, so its cache is current
  between commands in a pass and ours is not. The drain applies at most one command per order per
  pass and retries the rest ~1ms later, with a floor so a later enqueue cannot re-arm the alert onto
  the held-back command's elapsed due time.

**Teardown.** `stop()` cancels the alert and rejects each command still in flight
(`OrderRejected` / `OrderModifyRejected` / `OrderCancelRejected`). A discarded `CancelAllOrders` is
dropped silently instead: neither `Strategy::cancel_all_orders` nor
`ExecutionEngine::handle_cancel_all_orders` marks any order `PENDING_CANCEL` for it, and the order
FSM has a single `CancelRejected` transition, `(PendingCancel, CancelRejected)`, so there is no valid
rejection to synthesize. `reset()` clears the queue without rejecting.

## Open questions

1. **Modify overtaking its own submit.** Backtest `process_modify_submitted_order` amends an order
   still in flight in that case. I did not port it: here the modify reaches the venue and is
   rejected, as a real venue would for an order it has not received. Porting keeps backtest/sandbox
   parity; not porting keeps the deferred path a straight route to the engine and avoids
   synthesizing venue events in the client. I would rather you picked this one.
2. **The same-pass hold-back** is the least obvious part of the change. Is one command per order per
   pass reasonable, or would you rather the sandbox not batch a drain at all?


cjdsellers:
Hi @ChrisAB,

Thank you for the PR and for working through the ordering and teardown cases in detail. I think this is worth revising rather than replacing wholesale. The config plumbing, latency-leg mapping, client-arrival timestamps, stable FIFO ordering, converter move, and much of the behavioral test coverage should remain useful.

Keeping response latency separate is the right scope for this PR. A modify that overtakes its submit should also receive the realistic venue rejection rather than using the backtest amendment behavior.

One central part needs a different design before this can merge: the same-order settlement barrier.

The matching engine sends execution events to the execution channel, and the cache applies them on a later runner turn. The current ~1 ms hold-back only makes that cache update likely to happen first. If the runner is delayed, the retry and the earlier execution event can both be ready, and time-event priority can apply the next command against stale cache state. Starting the retry window from the callback time would narrow this race, but it would not establish the required ordering.

The invariant I want is: after a delayed command reaches the sandbox, every execution event it emits synchronously must be processed into the cache before the next arrived command for that client is applied. Runner serialization or an explicit completion barrier can provide that guarantee; an elapsed-time retry cannot.

**The revised design also needs to preserve these boundaries:**

- A command released from the latency queue is applied once and does not enter the latency queue again.
- `(due_ns, seq)` defines ordering within one client, including commands created by callbacks. This feature does not need to introduce ordering across separate clients.
- Stop distinguishes commands still in transit from commands already released for processing, and unwinds multiple pending commands for the same order in an order the FSM can apply.
- `latency_model=None` preserves the existing immediate behavior.
- Due commands and market data have one explicit ordering point rather than relying on elapsed time or independent drain hooks.

The timed hold-back, retry floor, and associated tests should be replaced with this deterministic serialization contract.

**There are also a few smaller changes:**

- Append `latency_model` after the existing Python constructor parameters so positional calls remain compatible.
- Add the missing `Money` and `Venue` imports to the documentation example.
- Avoid warning logs for expected zero-latency or scheduled-timer replacement paths if the same timer mechanism remains.

Would you be interested in revising the PR around that contract? If you’d rather hand off the runner integration, that’s also fine. I can continue the implementation on this branch while keeping your commits and authorship intact, so the work does not get duplicated.

ChrisAB:
> Would you be interested in revising the PR around that contract? If you’d rather hand off the runner integration, that’s also fine. I can continue the implementation on this branch while keeping your commits and authorship intact, so the work does not get duplicated.

I'll take the revision. Most of the functionality stays the same so there's no reason not to in my opinion.

> Keeping response latency separate is the right scope for this PR. A modify that overtakes its submit should also receive the realistic venue rejection rather than using the backtest amendment behavior.

I agree, response latency should be a smaller PR anyways.

> The invariant I want is: after a delayed command reaches the sandbox, every execution event it emits synchronously must be processed into the cache before the next arrived command for that client is applied. Runner serialization or an explicit completion barrier can provide that guarantee; an elapsed-time retry cannot.

Agreed and something I might have also missed while making the PR:
1. Retry floor is calculated from `event.ts_event` rather than callback time, so if the runner is delayed for 1ms, the retry is already elapsed when it's set. That means `arm_inbound_alert_at` fires immediately.
2. `drain_due_now` doesn't care about the retry floor. Any quote, trade, bar or delta arriving after a pass re-drains and applies the command.

> The timed hold-back, retry floor, and associated tests should be replaced with this deterministic serialization contract.

For this, I plan to follow the implementation already done in `BacktestExecutionClient::drain_queued_events` and `BacktestEngine::drain_command_queues`, which means apply one due command, flush its events synchronously to
`exec_engine_process()`, then take the next. That drops `COMMAND_SETTLE_RETRY_NS`, the retry floor, `AppliedThisPass`, and `for_each_order_id` entirely. 

> - Due commands and market data have one explicit ordering point rather than relying on elapsed time or independent drain hooks.

Do you mean:

- **(a)** the timer alert becomes the sole release point and the data handlers stop draining, or
- **(b)** the data handlers stay the release point as a defined contract, with the timer only waking an otherwise-idle queue?

I lean (b), but let me know.

> - Append `latency_model` after the existing Python constructor parameters so positional calls remain compatible.

Will move it to the end of the `#[pyo3(signature = ...)]` list after `price_protection_points`.

> - Add the missing `Money` and `Venue` imports to the documentation example.

Will do.

> - Avoid warning logs for expected zero-latency or scheduled-timer replacement paths if the same timer mechanism remains.

Got it, but I won't touch the `clock.rs` warnings themselves since that path is shared. Tell me if you want me to proceed otherwise.

ChrisAB:
@cjdsellers 

Pushed the revision. The hold-back, retry floor, `AppliedThisPass` and `for_each_order_id` are gone; `drain_inbound` applies one due command, flushes its events to `exec_engine_process()`, then takes the next. Stop, the Python signature and the docs imports are as discussed.

Some things that were by-products of your comments:

`OrderSubmitted` took the execution channel while the engine's `OrderAccepted` went through the barrier, so a runner lagged past the inbound leg applied `Accepted` first and `(Accepted, Submitted)` had no transition (the event was dropped and `ts_submitted` never set) With a latency model configured the client no longer uses the execution channel for order events. `latency_model=None` is unchanged.

I added an `OrderEventDispatchGuard` in `crates/common/src/runner.rs`, mirroring `trading_cmd_is_dispatching`. That's 44 lines outside this PR's area; happy to split it out, though it would need to land first.

A zero-leg command queued behind one still in flight arms at a `due_ns` that `LiveClock` has already passed. Keeping the later alert delays release on an idle queue, clamping forward doesn't survive a live clock, and draining from the command path costs nothing but makes command arrival a release trigger. The last reads best to me, but it's your call. Which would you prefer?

cjdsellers:
@ChrisAB, thanks for the latest round of updates. I've had another look at the revised settlement path, and there are two remaining contracts to address:

- Around `EventRouter::flush`, a latency-enabled command can re-enter `exec_engine_process` while `exec_engine_execute` still holds the engine borrow. No event should re-enter the engine during that borrow, while `OrderSubmitted` must still settle before venue events.
- Around `order_event_is_dispatching()`, a command queued from an external order-event callback can be left without a wake-up. Every callback-created command needs eventual release with `OrderSubmitted` ordered before its venue events.

The PR title also needs an imperative verb before squash merge, and the three Unicode arrows in the added test comments should use `->`.

ChrisAB:
@cjdsellers

Both contracts are addressed. There's a scope question at the end.
PS: Sorry for the force pushes, I bundled everything into second commit because this rework basically changed the same stuff. Also deleted some tests and more comments as they were a bit useless.

> Around `order_event_is_dispatching()`, a command queued from an external order-event callback can be left without a wake-up.

Three changes:

- `enqueue` keys on a sandbox-local `draining` flag instead
- `flush` drains its parked events before the incoming batch
- `defer_or_apply` enqueues rather than applying inline while either dispatch is active.

> Around `EventRouter::flush`, a latency-enabled command can re-enter `exec_engine_process` while `exec_engine_execute` still holds the engine borrow.

This is a regression of #3732. The regression test for it says the fix was to route sandbox events through the async runner channel so they're processed after the borrow is released.
I added `ExecCommandDispatchGuard` / `exec_command_is_dispatching()` in `crates/common/src/runner.rs`, raised in the `exec_engine_execute` handler immediately before `rc.borrow()`, mirroring `OrderEventDispatchGuard`.

That's a second guard outside this PR's area, on top of the `OrderEventDispatchGuard`. While tracing it I enumerated every place the engine's `RefCell` is borrowed in a handler or callback:

`exec_engine_execute`,  `exec_engine_queue_execute`, `exec_engine_process`, `exec_engine_reconcile_execution_report`, `subscribe_venue_instruments`

All of the above is mitigation. The underlying problem is that `exec_engine_execute` holds a borrow across the synchronous call into the client. #3732 was closed by having one client avoid the borrow. Narrowing the borrow so it doesn't span the client call would make all five sites safe by construction and let both guards be deleted, but I felt like this would be a bit out-of-scope, although I can submit a PR with this more concise fix after.

Also, title has an imperative verb now, and the three arrows are `->`.

cjdsellers:
@ChrisAB, thanks for the latest update. I've had another look. The config plumbing, latency-leg mapping, and most of the tests look useful, but six execution paths still need revision.

**Submission timing:**
`OrderSubmitted` remains deferred on the production command path. Each command enters under `ExecCommandDispatchGuard`, so `EventRouter::flush` parks the event until a later drain. This also defers the zero-latency case. The submission event needs to settle before the inbound delay starts without re-entering the engine.

**Event ordering:**
Market-data events enter the asynchronous execution channel, while inbound timer events go directly to `exec_engine_process`. Since the runner prioritizes time events, a later command response can overtake an earlier fill waiting in the channel. Both sources need one ordering point.

**Shutdown:**
`SandboxExecutionClient::stop` flushes rejections while `ExecutionEngine::stop` still holds the engine's mutable borrow. A queued command can therefore re-enter `exec_engine_process` and panic. Shutdown needs to preserve the rejection behavior without re-entering the engine.

**Reset:**
The queue cleanup is implemented in the inherent `reset(&self)`, but `ExecutionEngine::reset` calls the default `ExecutionClient::reset`. The production reset path therefore leaves the queue and alert alive. That trait path needs to clear both.

**Default behavior:**
`apply_submit_order_list` now rejects a leg with a missing instrument when `latency_model=None`; the previous immediate path skipped it. Keep this rejection limited to the deferred path so the default behavior remains unchanged.

**Submit-list instruments:**
Both rejection paths use the list's representative `instrument_id`. For a list containing different instruments, this assigns the wrong instrument and message-bus topic to some rejections. Each rejection needs the rejected order's instrument.

Once these paths are addressed, the inbound latency model should fit the sandbox execution path cleanly.

ChrisAB:
I believed I adressed all points and also made sure to not have a regression related to your earlier comments (Which happened in the last push)

**Event ordering:**
With a latency model the client no longer installs the execution-channel sink.

**Shutdown:**
`stop_clients` enters the guard around `client.stop()` and processes the captured rejections with the `&mut self` it already holds.

**Reset:**
The trait `reset` delegates to the inherent one, so the production path clears the queue and cancels the alert. That also means `ExecutionEngine::reset` now resets the sandbox's matching engines and reseeds its balances. ***Say if you'd rather the trait impl clear only the queue and the alert.***

**Default behavior:**
Only the deferred path rejects an unresolved leg. The immediate path skips it as before.

**Submit-list instruments:**
Rejections are keyed per order on `(client_order_id, instrument_id)`, so each carries its own instrument and topic.

Two more changes from the final state:

- **Zero-latency warnings:** A zero-leg command is now applied inline unless a queued command is already due or it was issued from a callback during a drain.
- **Market-data path:** now captures and flushes after the borrow is released, as the drain does, and `on_instrument` takes the same route. The immediate path bypasses this and is unchanged.

The engine-side change is one guard type over the five borrow sites, a no-op for adapters on the execution channel.

I also folded overlapping tests and trimmed some comments.

cjdsellers:
@ChrisAB, thanks for addressing the latest points. Some matching-engine fixes have now landed on `develop` that change the settlement requirement I asked you to implement earlier. The engine now accounts for pending order updates and fills before their events reach the cache, including successive modifications and reduce-only fills within a batch.

I think this gives the PR a simpler path. The inbound latency configuration, command queue, and shutdown handling remain useful. Could you rebase and check whether the latency-enabled path can retain the existing execution channel, removing the synchronous settlement machinery where it is no longer needed? I realise that machinery followed my earlier request; the matching-engine fixes address the stale-state problem at a different level.

The important behavior to preserve is deterministic command arrival ordering, ordered event delivery, and eventual release of callback-created commands, without re-entering an active engine borrow. The latency tests should verify those outcomes against the updated matching engine. If any case still requires synchronous settlement, that would help identify the remaining requirement.

If the synchronous route remains, there is still a callback-ordering case to cover: when a quote emits fills A and B, and A's callback publishes another quote that emits C, the current flush can deliver A, C, B.

I'm happy to retain the full trait reset, since execution-engine reset also clears the cache. The rebase also needs to accommodate the recent `DurationNanos` changes.

ChrisAB:
Managed to rebase on develop, and the latency-based path now uses the execution channel.

Removed:
- `crates/common/src/runner.rs`: `ExecEngineDispatchGuard`, `exec_engine_is_dispatching`, `capture_order_event`, their thread-locals and test.
- `crates/execution/src/engine/mod.rs`: the guard at the borrow sites, `process_captured`, the `stop_clients` guard, the `on_instrument` guard. `stop` is back to develop's shape.
- `crates/adapters/sandbox/src/execution.rs`: `EventRouter` with `flush` and `release`, `EventSink`, `EventCaptureGuard`, `DrainPassGuard` and the `draining` flag, `with_settlement_barrier`, `process_with_barrier`, the dispatching check in `drain_inbound`, the capture-and-flush in `stop`, and the `start()` condition that skipped the channel sink when a latency model was set.
- Tests: `EngineStandIn`, `install_engine_stand_in`, `LatencyHarness`, `setup_latency_harness`, `flushed_at_barrier`, and the seven tests that asserted the barrier or the guards.

The PR no longer touches `crates/common` or the execution engine.

A, C, B: That reordering should be fixed now. A and B are on the channel before the runner processes A. A's callback runs during that processing, outside any client borrow. Its quote reaches the sandbox data handler, which drains and processes it, so C is appended after B. `test_callback_published_quote_keeps_channel_order` asserts A, B, C on the published events.

Synchronous settlement: No remaining case requires it.

cjdsellers:
@ChrisAB, thanks for the rebase and simplification. One cancel-all case still needs attention.

With an existing matching engine, a 3-second insert leg and a 1-second delete leg let `CancelAllOrders` cancel a submit that is still in transit. `process_cancel_all` includes cached inflight orders and emits `OrderCanceled` even when the order has no matching-core entry. The latency-enabled path needs to limit cancellation to orders the venue has received, leaving later-arriving submits unaffected. Could you cover that ordering in a regression test as well?

cjdsellers:
@ChrisAB, thanks for the cancel-all fix. The exclusion and the two regression tests cover the direct path exactly as discussed. I've had another look, and one ordering case still gets through.

**Contingent cancellation bypasses the in-transit exclusion**:
With `support_contingent_orders` enabled, link two OCO orders A and B across separate submits so A reaches the venue while B's submit is still in transit, then release a cancel-all before B's due time. `process_cancel_all_excluding` filters B from the initial list, but canceling A calls `cancel_order(..., None)`, which runs `cancel_contingent_orders`. B is `Submitted` rather than active-local there, so it receives `OrderCanceled` despite the venue never having received its submit, and the later-arriving submit becomes a no-op on the closed guard. The exclusion needs to hold through contingent cancellation as well, with the same regression coverage.

**One smaller note**:
At `stop()`, a discarded modify or cancel rejection is synthesized even when the order closed while the command was in flight, for example a fill through market data landing before the stop. `(Filled, CancelRejected)` has no FSM transition, so the engine logs an invalid-state-transition error and drops the event. The `SubmitOrderList` arm already filters closed orders; the modify and cancel arms could take the same route.

Once the contingent case is covered, this should fit the sandbox execution path cleanly.

ChrisAB:
@cjdsellers

Both are addressed.

**Contingent cancellation**: `cancel_order_excluding` carries `excluded` into
`cancel_contingent_orders`, which skips a named linked order before the `is_active_local` check that
let a `Submitted` order through. `cancel_order` and `expire_order` delegate with `&[]`, so no other
call site changes. `test_process_cancel_all_excluding_spares_an_excluded_contingent_order` covers the
matching engine and `test_inbound_latency_cancel_all_leaves_in_transit_contingent_unaffected` runs
your 3s/1s timeline in the sandbox. When B's submit arrives the venue rejects it with `Contingent
order O-RECEIVED already closed` from the existing validation in `process_order`, so B still ends
closed, but through a response the venue issued when it saw the order. Say if you want a different
terminal state there.

**Stop rejections**: `needs_rejection` returns false for an order already closed in the cache, so the
modify and cancel arms take the same route as `SubmitOrderList`. It carries the `pending_rejected`
dedup and the no-engine path shares it.
`test_stop_raises_no_rejection_for_an_order_closed_in_flight` covers both arms.

**One path I have not touched**: the same in-transit order can still be canceled by a venue event
rather than a command. A quote that fills A cascades through the `Oco` arm of the fill handler, which
guards only on `is_closed() || is_active_local()`. `update_contingent_order` and `expire_order` have
the same shape; `sync_ouo_leaves` does not, since it skips anything absent from `core`. A per-command
`excluded` cannot reach any of them.

Two things about that. `SimulatedExchange` has both the cancel-all and the fill-cascade case on
`develop`, with no sandbox code involved: `exchange.rs:1917` calls the unexcluded
`process_cancel_all`, and `BacktestExecutionClient` generates `OrderSubmitted` before `send`. I have
throwaway tests reproducing each against the backtest exchange, both ending on `accept_order`'s
`is_closed` guard. A bracket sent as a `SubmitOrderList` takes one latency stamp and processes every
leg in one pass, which is likely why this has not surfaced.

And guarding on `core` instead will not work. An OTO child waiting on its parent is `Submitted`, not
active-local, and absent from `core`, which is what an in-transit submit also looks like, and it has
to stay cancelable. `sync_ouo_leaves` is safe only because OUO siblings are always both live at the
venue. Anything covering the event paths has to tell an order the venue never received from one it is
holding, which neither `core` nor a per-command list does. `process_modify` on an accepted parent
reaches `update_contingent_order`, so that one is a command path `excluded` could carry through if
you want it here.

Do you want the event paths in this PR, or a separate issue covering the backtest as well?

cjdsellers (approved):
@ChrisAB, thanks for sticking with this and for tracing the remaining cases. The latest fixes address my last comments, and I’m happy to proceed with the merge.

I’ll follow up post-merge on some remaining contingent-order paths, ensuring orders still in transit cannot be canceled, modified, or activated early, with regression coverage for sandbox and backtest.

PR #4866, opened 2026-08-26
Await betfair execution lifecycle transitions in tests

folknor:
A follow-up to the betfair test determinism work in #4849, which left `tests/exec_client.rs` for a separate change.

## The execution-client tests wait on a wall-clock deadline instead of on the event

`tests/exec_client.rs` had 64 `wait_until_async` calls, each polling a predicate against a 2 to 5 second budget and asserting on elapsed wall-clock time. `#[tokio::test]` without a `flavor` builds a current-thread runtime, so the test body, the mock server task and the socket tasks are a single schedulable unit; when a parallel build competes for the machine that unit can lose the CPU for seconds and the budget expires with nothing wrong. Widening a budget only moves the threshold.

60 of those waits are converted here. The four that observe `is_connected` are left alone: that predicate combines core connection state, the reconciliation gate and stream readiness, and calls `process_pending_resync` on a true result, so it is not an observation that a watched mirror can represent.

## Reconciliation waits on the generation the gate already publishes

`ReconciliationGate` holds a `watch::Sender<u64>` and advances it on every halt and resume; `wait_for_generation_change` already awaits it on the reconnect path. The 12 waits that polled `is_reconciling` now await `BetfairExecutionClient::wait_for_reconciliation_state`, which subscribes before its first read and then loops, so a transition landing between the read and the await has already advanced the generation and `changed()` returns immediately.

`is_halted` remains the authoritative read; the channel is used only as an edge notification, and the gate's odd and even generation encoding stays private. The method takes a `bool` because the observable state is binary - the distinct odd generations exist to stop a stale recovery reopening a newer halt, and carry no lifecycle meaning a caller could act on.

## The mock server notifies instead of being polled

The remaining 48 waits observe the mock's own counters and request records. `MockState` gains a `watch::Sender<u64>` bumped on each mutation, and predicates read the mock's existing fields directly under their mutexes.

Publishing a snapshot of those fields instead would reintroduce the same class of flake by another route: the fields are independently synchronized, so two concurrent handlers can compose their snapshots and publish them out of order, leaving the channel holding regressed state with no later notification. A monotonic counter cannot regress, and coalesced notifications are harmless because every wake re-evaluates the authoritative state.

## Testing

The assertions are unchanged; only the waiting is different. Each logical phase carries a 15 second bound that reports the expected condition and the state after the deadline, so a regression fails with a diagnosis rather than hanging until the harness intervenes. That bound is a circuit breaker rather than synchronization - above the scheduling gaps that caused the flakes, and below the harness slow-test thresholds so the diagnostic can fire.

Suppressing the mock's notification while leaving its state mutations in place fails a converted wait at the phase bound, and the diagnostic then shows the state satisfying the predicate, which is the missing-notification case rather than a missing transition. Suppressing the gate's `send_replace` while leaving its atomic store in place fails `test_command_before_reconnect_image_keeps_recovery_pending`, which waits on both edges. Both establish that the tests observe transitions through the notifier rather than finding the expected state already present.

This removes scheduler-sensitive polling rather than making the suite faster; `wait_until_async` returned as soon as a poll succeeded, so the change is not a speed one. One representative test moved from 1.03s to 0.85s, which is offered as evidence of no regression.

`tests/data_client.rs` has six waits of the same shape and can now reuse this helper; it is left for a follow-up.

### DIFF
--- a/crates/adapters/betfair/src/execution.rs
+++ b/crates/adapters/betfair/src/execution.rs
@@ -232,6 +232,26 @@ impl BetfairExecutionClient {
         self.reconciliation_gate.is_halted()
     }
 
+    /// Waits for the reconciliation halt state to equal `expected`.
+    ///
+    /// Returns immediately if the gate is already in the expected state; otherwise
+    /// waits for a later transition. A transient expected state that is replaced
+    /// before this task observes it can be missed because transitions are not
+    /// recorded as history. This method has no internal timeout; callers wanting a
+    /// bound should wrap it in [`tokio::time::timeout`].
+    pub async fn wait_for_reconciliation_state(&self, expected: bool) {
+        wait_for_reconciliation_state(&self.reconciliation_gate, expected).await;
+    }
+
+    /// Waits for the semantic connection state to equal `expected`.
+    ///
+    /// Returns immediately if the client is already in the expected state; otherwise
+    /// waits for a later lifecycle transition. This method has no internal timeout;
+    /// callers wanting a bound should wrap it in [`tokio::time::timeout`].
+    pub async fn wait_for_connection_state(&self, expected: bool) {
+        wait_for_connection_state(self, expected).await;
+    }
+
     fn submissions_halted(&self) -> bool {
         !self.core.is_connected()
             || self.reconciliation_gate.is_halted()
@@ -4254,6 +4274,69 @@ async fn wait_for_generation_change(
     while *state_rx.borrow_and_update() == generation {
         if state_rx.changed().await.is_err() {
             return;
+        }
+    }
+}
+
+/// Waits until the gate's halt state equals `expected`.
+///
+/// Subscribes before the first read so a transition landing between the two is
+/// observed rather than missed. The generation channel is used only as an edge
+/// notification; `is_halted` remains the authoritative read.
+async fn wait_for_reconciliation_state(gate: &ReconciliationGate, expected: bool) {
+    let mut state_rx = gate.subscribe();
+
+    while gate.is_halted() != expected {
+        if state_rx.changed().await.is_err() {
+            return;
+        }
+    }
+}
+
+/// Waits until the client's semantic connection state equals `expected`.
+///
+/// `is_connected` reads three terms: the core connected flag, the reconciliation
+/// gate, and the stream's composite order readiness. The reconciliation receiver is
+/// subscribed once, before the first evaluation, so no generation edge is missed
+/// across iterations. The stream term is awaited through a level-triggered wait
+/// that subscribes before reading its own predicate, which closes the same window
+/// for it - an edge-triggered wait cannot, because `select!` subscribes only when
+/// it first polls the future, after this loop has already evaluated `is_connected`.
+///
+/// The stream arm is selected only when the stream term is not already the
+/// satisfied one. Waiting for `true` with the stream already ready means the gate
+/// is what blocks, and a level-triggered wait on an already-satisfied term would
+/// return instantly on every iteration and starve the task that resolves the gate.
+///
+/// The core connected flag has no notification source, so this waits only for the
+/// terms that can move while it runs. That is sufficient because `connect`,
+/// `disconnect`, and `stop` all take `&mut self` while this holds a shared borrow of
+/// the same client, so the core flag cannot change under it.
+async fn wait_for_connection_state(client: &BetfairExecutionClient, expected: bool) {
+    let mut reconciliation_rx = client.reconciliation_gate.subscribe();
+
+    while ExecutionClient::is_connected(client) != expected {
+        let Some(stream_client) = &client.stream_client else {
+            reconciliation_rx
+                .changed()
+                .await
+                .expect("reconciliation sender lives as long as the client");
+            continue;
+        };
+
+        if expected && stream_client.is_order_ready() {
+            reconciliation_rx
+                .changed()
+                .await
+                .expect("reconciliation sender lives as long as the client");
+            continue;
+        }
+
+        tokio::select! {
+            () = stream_client.wait_for_order_ready_state(expected) => {}
+            changed = reconciliation_rx.changed() => {
+                changed.expect("reconciliation sender lives as long as the client");
+            }
         }
     }
 }
--- a/crates/adapters/betfair/src/stream/client.rs
+++ b/crates/adapters/betfair/src/stream/client.rs
@@ -247,6 +247,28 @@ const fn subscription_transport_state(connected: bool, id: u64) -> StreamLifecyc
     }
 }
 
+/// Waits until [`BetfairStreamClient::is_order_ready`] equals `expected`.
+///
+/// Subscribes to both lifecycle components the predicate reads BEFORE evaluating
+/// it, so a transition landing between the caller's own check and this wait is
+/// still delivered. Transport transitions are reflected in the order component,
+/// which is why the socket flag needs no separate subscription.
+async fn wait_for_order_ready_state(client: &BetfairStreamClient, expected: bool) {
+    let mut auth_rx = client.lifecycle.authenticated.changed.subscribe();
+    let mut order_rx = client.lifecycle.order.changed.subscribe();
+
+    while client.is_order_ready() != expected {
+        tokio::select! {
+            changed = auth_rx.changed() => {
+                changed.expect("lifecycle sender lives as long as the borrowed client");
+            }
+            changed = order_rx.changed() => {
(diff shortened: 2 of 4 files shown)


folknor:
@cjdsellers CI fail here is just the benchmark stuff you fixed one commit after I opened this PR :joy:

cjdsellers (commented):
Hey @folknor,

Thanks for the PR. One converted wait now returns before the behavior covered by its assertion has completed. `handle_betting` publishes as soon as it records `cancelOrders`, but `test_cancel_all_orders_sends_request` then checks `rx.try_recv()` and disconnects. That disconnect aborts the pending HTTP task, so the test can pass without consuming the success response and no longer proves that response processing emits no rejection.

Could you add a response-completion barrier before checking the channel? The request-receipt notification is suitable for the payload assertions, but it does not establish completion of the response path.

There is also a smaller consistency gap in the generic mock-state notifier: mutations such as `betting_applied_request_params` do not advance its generation. Could you either publish those changes or narrow and document the fields supported by `wait_for_mock_state`?

folknor:
Thank you.

Both points are fixed.

## The completion barrier

Request receipt is the wrong barrier. The mock records method and params on receipt, so the wait returned before the response existed, and the disconnect then aborted the pending task; the assertion held whether or not the response path had run.

A server-side counter would not close it either, since it proves the response was handed off rather than received and handled by the client. The barrier is therefore client-side. `BetfairExecutionClient` holds a cancel-all completion generation, advanced at the end of the spawned task after the result is handled rather than immediately after the await, so reaching it means error handling ran. `cancel_all_completion_token()` captures the generation before dispatch and `wait()` blocks until a later one: the initial value cannot satisfy it, request receipt does not advance it, and an aborted task never advances it. The test wraps `wait()` in the existing `PHASE_TIMEOUT`.

The token carries no operation identity, so a second concurrent cancel-all would satisfy it; these tests issue one at a time. This also repairs the test's meaning rather than a production bug - the cancel-all task awaits, logs on error and returns, with no path that emits a rejection. The old assertion proved nothing had been emitted by request-receipt time, which is true of any implementation.

`TaskHandles::all_finished()` was the alternative and is vacuous-prone, since "all finished" is initially true. Draining pending work in `disconnect` would have changed production shutdown semantics for one test.

## The notifier

Every handler-side mutation now publishes. Beyond `betting_applied_request_params`, the one-shot status, error and apply-then-status consumptions and the response-sequence `pop_front` were silent. So was `betting_request_count`: it increments before the method is parsed, but its publication sat inside the non-empty-method branch, so a malformed request advanced the count without waking anyone.

`wait_for_mock_state` takes `Fn(&MockState) -> bool` over public fields, so the interface promises any readable field is waitable and an allowlist would contradict it. Extra notifications are harmless because the helper rechecks its predicate before awaiting again.

`test_mock_state_notifier_contract` asserts each branch advances the generation, with a comment requiring new mutations to be added. Its assertions are relative rather than exact counts, which would have failed a contributor adding a legitimate publication. Making the fields private behind notifying accessors would be stronger and is larger than this round.

## The other sites

`test_cancel_all_orders_sends_request` is the only converted wait followed by an immediate `try_recv()` and a disconnect. Of the others, 670, 1121 and 3291 positively await a response-generated event, so a missing response path fails them outright; 564, 722, 780, 1196, 1721, 3357 and 6388 rest their negative assertions on a timed `rx.recv()` loop, which does give the response path a chance to run; and 3089 and 3481 use fixed 200 ms sleeps in place of a completion barrier, the second calling `try_recv()` for a positive `OrderUpdated` and able to fail spuriously.

None of those sleeps is introduced or modified here, since the conversion replaced deadline-based wait helpers. The old helper polled every 100 ms, so removing it lets an adjacent test observe sooner, shortening the margin those sleeps rely on without making them regressions from this change. I am happy to take them alongside the remaining `data_client.rs` conversion, or here if you would rather.

## Testing

The test captures the token before dispatch, keeps the request-receipt wait for the payload assertions, and awaits completion under `PHASE_TIMEOUT` before checking the channel. Against a deliberately broken completion signal, `advance` incrementing but never publishing, it fails with `deadline elapsed waiting for cancel-all completion`. That proves the wait depends on a published completion; it does not break response processing, and no rejection is being caught.

`test_mock_state_notifier_contract` fails if any branch stops publishing, including the malformed-method case.

cjdsellers:
@folknor, thanks for the detailed follow-up. Unfortunately, significant task-ownership work landed on `develop` after this branch was created, and it overlaps all three files changed here. Betfair execution now uses the new `TaskGroup` and `TaskGroupGuard` ownership and shutdown model rather than `TaskHandles` with separate `JoinHandle`s, so this branch needs to be updated and reconciled with that implementation before it can merge.

The merge is textually clean, but the current-base result does not compile: the tests still refer to `BettingResponseGate`, while `develop` now has `MockResponseGate`, and the manual `MockState` initializer is missing `accounts_response_gate`. The generic notifier contract also needs to cover the current account response-gate mutations. While adapting this, the cancel-all completion barrier needs to remain test-owned rather than exposing `cancel_all_completion_token()` and `CancelAllCompletionToken` as public production API solely for this test.

folknor:
Thank you. The branch is rebuilt on current `develop` rather than rebased, because the task-ownership work reshaped the code the old commits touched and every conflict hunk was in code that no longer exists in that form. The conversion is the same idea against the current implementation: 70 deadline waits, up from 64.

## The task-ownership model

`TaskGroup` and `TaskGroupGuard` changed what the tests need rather than how they wait. Connect readiness is still asynchronous, since `subscribe_orders` only queues the subscription and `is_connected` additionally requires `is_order_ready`, so those waits stay. Successful `disconnect` and `stop` now set disconnected and clear reconciliation state before returning, so the waits that followed them are gone rather than converted. Reconciliation transitions still happen in stream handlers and the reconnect task, so every `is_reconciling` wait stays.

`BettingResponseGate` is `MockResponseGate` throughout, and `accounts_response_gate` is in the initializer and in the notifier contract.

## The cancel-all barrier

There is no barrier to make test-owned, and that is the answer rather than an evasion.

`disconnect` looked like one: `TaskGeneration::spawn` selects only on the force token, so `finish_shutdown` drains a pending task for its graceful window rather than cancelling it. But `teardown_partial_connect` calls `BetfairHttpClient::disconnect` before `await_pending_tasks`, and that cancels the token every JSON-RPC operation snapshots, so the in-flight request is cancelled and the task terminates on an error it logs. The drain proves termination, never that a response was handled. A forced abort after the graceful deadline is also recorded as a successful shutdown, so a barrier built on it would degrade silently under load into the failure mode this PR removes.

Nor can any scheduling arrangement recover the information, because the successful cancel-all future publishes nothing: it awaits `send_betting_order`, logs on error, and returns, with no emitter call, no state change, and no follow-up request. So `test_cancel_all_orders_sends_request` now asserts what it can establish - the request reached the venue with a valid customer reference and no `instructions` for the unfiltered form - and no longer claims anything about the response. The ambiguous-response tests keep a sound no-event assertion, because observing the retry proves the client received and classified the first response.

## The notifier

The generation is published after every handler-side mutation, including the request count incremented before the method is parsed, which previously let a malformed request advance the count without waking anyone. Waiters subscribe before their first evaluation and read `MockState` directly, so notifications are hints and the authoritative state is what decides. The contract test pins each branch against a freshly measured baseline.

The handler records the method before the params, publishing after each, so a predicate over the method list can wake between the two and read a payload that is not yet there. `betting_method_count` now counts the params collection, which is the later of the two, so every waiter observes a request whose payload is already readable.

## The connection wait

`is_connected` reads three terms, and the wait subscribes to each that can move. The reconciliation receiver is created once outside the loop; the stream term goes through a crate-private level wait that subscribes to both lifecycle components before reading its own predicate. An edge-triggered wait cannot close that window, because `select!` subscribes only when it first polls the future, after the predicate has been evaluated. The stream arm is excluded from the select when waiting for true with the stream already ready: a level wait on an already-satisfied term returns on every iteration and starves the task that resolves the gate, which deadlocked `test_degraded_order_stream_denies_submit_until_current_heartbeat` while I was building this.

## Testing

Making the mock's publication a no-op fails a converted wait at 15.14s with the state dump showing the awaited request already recorded, which is the missing-notification case rather than a missing transition, and fails the contract test at its first branch. Suppressing a single publication site does not fail a converted wait, and should not: the predicate reads authoritative state and any notification causes a recheck. Publication completeness is the contract test's job, not each waiter's.

cjdsellers (approved):
@folknor, this looks good to merge now. Thanks for rebuilding it against current `develop`.

After merge, I'll follow up separately to move the `wait_for_connection_state` test support out of production code and update the related documentation.

PR #4867, opened 2026-08-26
Consolidate WebSocket initial-connect retry into the client

folknor:
Bybit and both Architect AX WebSocket clients each carry a copy of the same hand-rolled initial-connect retry ladder, because `WebSocketClient::connect` does not retry.

This is a suggested implementation rather than a question: it picks an answer, says which one below, and I am happy to reshape it if you would rather the policy lived elsewhere.

## The copied ladder

`WebSocketClient::connect` bounds its attempt with `connect_timeout_ms` and returns on failure; its `ExponentialBackoff` drives the reconnect loop only. So the three adapters wrap it in a `const MAX_RETRIES: u32 = 5` loop with their own `ExponentialBackoff`, at `bybit/src/websocket/client.rs:439`, `architect_ax/src/websocket/data/client.rs:397` and `.../orders/client.rs:435`.

Each copy has the same three problems. It never reads the client's own cancellation state, so a `close()` arriving during the ladder is observed only after it finishes, up to roughly 33 seconds later. It retries every error, so a malformed URL or a rejected handshake consumes all five attempts. And its parameters are unreachable: the attempt count is a local `const` and the backoff values are literals sitting ten lines below the `reconnect_backoff_factor` the same function sets from configuration.

The raw socket transport already solves this. `SocketConfig::connection_max_retries` drives an initial-connect ladder in `SocketClient::connect` with identical backoff values, so this is an asymmetry between two transports rather than a missing design.

## The policy is a builder input, not a config field

`WebSocketClient::builder()` and `epoch_builder()` gain an optional `InitialConnectRetryPolicy` and an optional `CancellationToken`. With no policy the builder makes exactly one attempt, so every existing caller is unchanged; the three adapters opt in to their current values. `stream_builder` is untouched, since stream mode promises a single bounded connection.

The policy is a builder input rather than a `WebSocketConfig` field because it governs one invocation of `connect` and means nothing once a client exists, which is the same reason `rate_limiter` and `state_sink` are builder inputs. It also avoids a source break: `WebSocketConfig` is constructed by literal in roughly 85 places, and a new field would require every one to change.

The builder does not return a client until the initial dial succeeds, so `WebSocketClient::close()` cannot interrupt the ladder and the token has to reach the builder instead. The adapters hold the token in an `ArcSwap` shared across clones: `connect` installs a fresh one, `close` cancels the current one, and the next `connect` replaces it, so a connect-close-connect sequence is not poisoned by a spent token. The loop selects cancellation against all three blocking phases - the connection rate-limit wait, the dial, and the backoff sleep - since checking only at attempt boundaries would still leave `close()` waiting out a full connection timeout.

## What counts as retryable

`is_retryable_initial_connect_error` retries connection timeouts, the connection-loss variants, and network-shaped `Io` errors, and refuses `InvalidUrl`, `Handshake`, `Tls`, `Protocol`, the size and encoding errors, and `Other`.

The polarity of the `Io` arm is deliberate. Resolver failures commonly surface as `ErrorKind::Other`, so an allowlist of retryable kinds would stop retrying the DNS case these ladders exist for. It therefore names the kinds that are clearly permanent - `InvalidInput`, `InvalidData`, `Unsupported`, `PermissionDenied` - and treats the rest of connection-time `Io` as transient.

Every `Handshake` error is permanent, so an HTTP 429 or 5xx upgrade response no longer retries. Distinguishing those needs the status preserved structurally in `TransportError::Handshake` rather than in its string payload.

The adapters now report `Failed to connect to {url}: {e}`, keeping Bybit's DNS hint. They no longer claim an attempt count, which they cannot know from a `TransportError`: the same error is returned for a permanent failure rejected on the first attempt and for an exhausted ladder.

A structured exhausted-ladder error carrying the real count would fix that and is a reasonable follow-up if you want it.

## Testing

`timed_out_initial_connect_attempt_is_retried` stands up a listener that accepts TCP but never upgrades, so every attempt reaches the connect timeout, and asserts the server saw exactly three connections against a three-attempt policy. It fails at one against three on the previous behaviour, where a timed-out attempt returned instead of retrying. `initial_connect_cancellation_interrupts_in_flight_handshake` cancels after the server has accepted and asserts `Interrupted`; `permanent_initial_connect_error_does_not_wait_for_retry_ladder` proves an invalid URL returns without entering a 30 second backoff. Both fail against the unfixed code by timing out.

The two Architect AX connection-failure tests previously asserted the message named the URL and the attempt count. They now assert only that it names the endpoint, because the attempt count is no longer claimed; the exhaustion property they were standing in for is pinned directly by the network test above, which counts attempts rather than reading a string.

### DIFF
--- a/crates/adapters/architect_ax/src/websocket/data/client.rs
+++ b/crates/adapters/architect_ax/src/websocket/data/client.rs
@@ -17,6 +17,7 @@
 
 use std::{
     fmt::Debug,
+    num::NonZeroU32,
     sync::{
         Arc, Mutex,
         atomic::{AtomicBool, AtomicI64, AtomicU8, Ordering},
@@ -30,14 +31,14 @@ use nautilus_common::live::get_runtime;
 use nautilus_core::{AtomicMap, consts::NAUTILUS_USER_AGENT};
 use nautilus_live::SocketControl;
 use nautilus_network::{
-    backoff::ExponentialBackoff,
     http::USER_AGENT,
     mode::ConnectionMode,
     websocket::{
-        PingHandler, ReconnectHeaders, SubscriptionState, TransportBackend, WebSocketClient,
-        WebSocketConfig, channel_message_handler,
+        InitialConnectRetryPolicy, PingHandler, ReconnectHeaders, SubscriptionState,
+        TransportBackend, WebSocketClient, WebSocketConfig, channel_message_handler,
     },
 };
+use tokio_util::sync::CancellationToken;
 use ustr::Ustr;
 
 use super::{
@@ -136,6 +137,7 @@ pub struct AxMdWebSocketClient {
     cmd_tx: Arc<tokio::sync::RwLock<tokio::sync::mpsc::UnboundedSender<HandlerCommand>>>,
     out_rx: Option<Arc<tokio::sync::mpsc::UnboundedReceiver<AxDataWsMessage>>>,
     signal: Arc<AtomicBool>,
+    cancellation_token: Arc<ArcSwap<CancellationToken>>,
     task_handle: Option<tokio::task::JoinHandle<()>>,
     subscriptions: SubscriptionState,
     request_id_counter: Arc<AtomicI64>,
@@ -168,6 +170,7 @@ impl Clone for AxMdWebSocketClient {
             cmd_tx: Arc::clone(&self.cmd_tx),
             out_rx: None,
             signal: Arc::clone(&self.signal),
+            cancellation_token: Arc::clone(&self.cancellation_token),
             task_handle: None,
             subscriptions: self.subscriptions.clone(),
             subscribe_lock: Arc::clone(&self.subscribe_lock),
@@ -182,6 +185,16 @@ impl Clone for AxMdWebSocketClient {
 }
 
 impl AxMdWebSocketClient {
+    fn initial_connect_retry_policy() -> InitialConnectRetryPolicy {
+        InitialConnectRetryPolicy {
+            max_attempts: NonZeroU32::new(5).expect("initial connect attempts must be non-zero"),
+            delay_initial: Duration::from_millis(500),
+            delay_max: Duration::from_secs(5),
+            backoff_factor: 2.0,
+            jitter_ms: 250,
+        }
+    }
+
     /// Creates a new Ax market data WebSocket client.
     ///
     /// The `auth_token` is a Bearer token obtained from the HTTP `/api/authenticate` endpoint.
@@ -207,6 +220,7 @@ impl AxMdWebSocketClient {
             cmd_tx: Arc::new(tokio::sync::RwLock::new(cmd_tx)),
             out_rx: None,
             signal: Arc::new(AtomicBool::new(false)),
+            cancellation_token: Arc::new(ArcSwap::from_pointee(CancellationToken::new())),
             task_handle: None,
             subscriptions: SubscriptionState::new(AX_TOPIC_DELIMITER),
             request_id_counter: Arc::new(AtomicI64::new(1)),
@@ -243,6 +257,7 @@ impl AxMdWebSocketClient {
             cmd_tx: Arc::new(tokio::sync::RwLock::new(cmd_tx)),
             out_rx: None,
             signal: Arc::new(AtomicBool::new(false)),
+            cancellation_token: Arc::new(ArcSwap::from_pointee(CancellationToken::new())),
             task_handle: None,
             subscriptions: SubscriptionState::new(AX_TOPIC_DELIMITER),
             request_id_counter: Arc::new(AtomicI64::new(1)),
@@ -354,10 +369,10 @@ impl AxMdWebSocketClient {
     /// # Errors
     ///
     pub async fn connect(&mut self) -> AxWsResult<()> {
-        const MAX_RETRIES: u32 = 5;
-        const CONNECTION_TIMEOUT_SECS: u64 = 10;
-
         self.signal.store(false, Ordering::Release);
+        let cancellation_token = CancellationToken::new();
+        self.cancellation_token
+            .store(Arc::new(cancellation_token.clone()));
 
         let (raw_handler, raw_rx) = channel_message_handler();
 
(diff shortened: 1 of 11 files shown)


cjdsellers:
Hi @folknor,

Thank you for the PR. I found one retry-classification issue before this can merge.

`is_retryable_initial_connect_error` treats every `TransportError::Handshake` as permanent. Both WebSocket backends use that variant for HTTP upgrade rejections, including 429 and 5xx, and the proxy path also uses it when the proxy closes before returning its CONNECT response. These three adapters previously retried every initial connection failure up to five attempts; they now return after the first of those transient failures, so startup can fail even when the next attempt would succeed.

Could the transport preserve enough information to distinguish permanent handshake errors from retryable status and connection failures, with a regression test for at least one transient case?

folknor:
Thank you.

Fixed. `TransportError::Handshake` was carrying at least five distinct causes behind one string, so the classifier had nothing to decide on. The transport now preserves the rejection status and the classifier applies policy to it.

## The error shape

Two additive variants, `UpgradeRejected(u16)` and `ProxyConnectRejected(u16)`. `Handshake(String)` is unchanged and still carries malformed HTTP, invalid accept keys, invalid configured header names and values, and malformed proxy responses, all of which remain permanent.

Reshaping `Handshake` itself was the alternative. It would have broken four proxy destructures that test concepts this change does not touch, and an optional status field still could not separate a transient premature close from permanent malformed HTTP when neither carries a status.

`tungstenite::Error::Http` now yields `UpgradeRejected` with the response status rather than a formatted string. A proxy CONNECT refusal yields `ProxyConnectRejected`. A proxy that closes before sending its response becomes `ConnectionClosed`, which already describes that condition and was already retryable.

Sockudo needed more: `sockudo_ws::handshake::parse_response` rejects any non-101 status before returning, so the status never reached the wrapper. The local handshake recovers it from the raw response at that boundary. Recovery is gated on the `HandshakeFailed` variant and a validated status line rather than on the error message, so an upstream rewording cannot change transport semantics, and `InvalidHttp` stays permanent.

## Retry policy

`408 | 425 | 429 | 500..=599` retry on initial connect; every other 4xx is permanent. 408 and 425 both describe a request the server expects to be reattempted. Retrying every 5xx is deliberate rather than an allowlist of the ones that look transient.

`Retry-After` is not honoured. That needs preserved headers, both delta-seconds and HTTP-date parsing, and a server-directed delay reconciled with the configured backoff, so 429 and 503 retry on the configured ladder rather than on server-directed timing. I am happy to take it separately if you want it.

## Two values that were previously decorative

Both are pre-existing. Neither had any effect until the status began selecting retry behavior.

`read_connect_response` extracted the HTTP version and discarded it, so `NOT-HTTP 503 Service Unavailable` would have parsed as a retryable rejection. The version is now checked against HTTP/1.0 and HTTP/1.1.

The same function parsed the status token as a bare `u16`, which accepts tokens HTTP does not and normalizes `0503` to `503`. It now parses as `http::StatusCode`, which requires three digits with a nonzero leader. The accompanying message became "invalid status code", since the rejected input is no longer necessarily non-numeric.

## Testing

`transient_upgrade_rejection_is_retried` covers 429 and 503, `sockudo_too_many_requests_upgrade_rejection_is_retried` covers the sockudo path, and `proxy_close_before_connect_response_is_retried` covers the premature close. Each counts accepted connections against a three-attempt policy rather than asserting on error text, which a single attempt would also satisfy. A 404 control asserts exactly one attempt.

`initial_connect_retryable_status_policy` and `initial_connect_error_classification` pin the full policy including the 499/500 and 599/600 boundaries and the 426 near-miss. `read_connect_response_rejects_malformed_version_with_retryable_status` and `read_connect_response_rejects_malformed_status_token_with_retryable_value` both use 503, because a permanent status would pass whether or not the validation exists.

Each new assertion fails against the previous revision. Neutering the status policy fails the three attempt-count tests and nine classification cases; reverting the proxy premature close to its old construction fails its attempt count and the proxy unit test; accepting a malformed version or reverting the typed status parse each produce `ProxyConnectRejected(503)` where a `Handshake` is expected.

The existing `timed_out_initial_connect_attempt_is_retried` and `permanent_initial_connect_error_does_not_wait_for_retry_ladder` are unchanged and still pass, the latter still returning well inside its 250 ms bound.

A caller receiving a proxy premature close now sees an error for which `is_closed()` returns true, where the old `Handshake` returned false.

cjdsellers:
Thanks for the detailed follow-up. **I found two remaining blockers:**

- `cargo check -p nautilus-network --no-default-features --tests --locked` fails in the new listener code because `AsyncReadExt` and `AsyncWriteExt` are imported only with `transport-sockudo`. The crate documents `default-features = false` as supported, so the tests need to compile in that configuration.
- The initial-retry warning logs `config.url` verbatim. `WebSocketConfig::Debug` deliberately redacts this field because WebSocket URLs can contain credentials, tokens, or signed query data. The shared retry path needs to preserve that redaction.

**One non-blocking follow-up:** If you're able, could the public API docs clarify which failures stop before `max_attempts`, and that the builder cancellation token only applies while establishing the initial connection?

folknor:
Thank you.

Both fixed, and the redaction now has a regression test.

## The test build without default features

`AsyncReadExt` and `AsyncWriteExt` were imported under `transport-sockudo`, but two helpers added last round are ungated and use them. Both traits move out of the gate; `AsyncRead` stays behind it, since only the sockudo response reader names it. `cargo check -p nautilus-network --no-default-features --tests --locked` passes.

## The redacted endpoint

The retry warning substitutes `nautilus_core::string::secret::REDACTED` for the URL, matching what `WebSocketConfig::Debug` does with the same field. The message keeps the attempt counters and the `TransportError`, neither of which reproduces the endpoint.

No other site on the initial-connect or reconnect path formats the URL, the proxy URL, or configured header values into a log line or an error: proxy parse failures report a scheme or a missing host without the input string, and rejections retain only the numeric status.

`test_network_logs_omit_payload_bodies` now covers this. Its capture harness accepted only `Trace` records prefixed `Sending`, `Received`, or `Replaced`, so it would have discarded the warning silently; it now also admits the `Warn` record. The test drives a two-attempt ladder against a listener returning 503, with a marker in the URL query string, and asserts that the captured `WebSocket connection attempt 1/2` message contains `<redacted>`. That assertion is positive so the existing blanket secret assertion cannot pass vacuously. Restoring the interpolation fails the test with the marker visible in the message.

## Docs

`InitialConnectRetryPolicy` now names the retryable set as the classifier has it - the connection-closed variants, I/O errors of every kind except `InvalidInput`, `InvalidData`, `Unsupported`, and `PermissionDenied`, and upgrade or proxy `CONNECT` rejections carrying 408, 425, 429, or 500 through 599 - and states that everything else returns immediately without a backoff delay, however many attempts remain. `max_attempts` is documented as an upper bound rather than a promise. The doc also notes that classification is by error variant rather than by cause, since a TLS failure is permanent as `Tls` but follows the I/O rule where a backend reports it as `Io`.

The builder documents that `cancellation_token` applies only while the initial connection is being established, observed during the connection rate-limit wait, the dial, and each backoff delay, and has no effect once a client exists.

cjdsellers (approved):
@folknor, thanks for all the careful follow-ups here. I've had another look and this now looks good to merge :ok_hand:. I'll follow up separately on keeping retry classification adapter-owned, reusing `RetryManager` where it fits, and adding a regression test for the no-policy single-attempt default.

PR #4873, opened 2026-08-27
Allow external order claim changes while the live node runs

folknor:
Allow external order claim changes while the live node runs

A follow-up to the external-order-claim registration work in #4620.

## Claim registration while the event loop owns the node

`LiveNode::register_external_order_claims` and `deregister_external_order_claims` take `&mut self`, and `run` / `run_with_mode` hold that borrow for the whole run, so as documented neither can be called while the event loop owns the node. A deployed consumer that adds and removes strategies at runtime cannot reach either from a running worker: `start()` releases the borrow but nothing then services the runner's channels, and `LiveNodeHandle` carries only control state and metrics. The consumer's remaining option is halting the account whenever a strategy change would leave claims it knows are wrong.

Two new `SystemCommand` variants, `RegisterExternalOrderClaims` and `DeregisterExternalOrderClaims`, carry the request plus a oneshot responder and ride the runner's existing command channel, dispatched by `process_system_command` inside the running loop (which becomes `&mut self`; every caller already held `&mut`). The synchronous method bodies moved into private helpers that both routes call, so preflight-before-mutation and the engine-then-manager commit order are shared rather than duplicated, and the responder receives the helper's exact `Result` after both commits - a dropped receiver cannot roll back a committed registration.

`LiveNodeHandle` gains async `register_external_order_claims` / `deregister_external_order_claims` backed by a clone of the channel sender, giving worker threads a `Send` route. `new()` and `Default` handles stay control-only, and the synchronous methods remain the idle / started / stopped route, with their rustdoc updated to name the handle route for the running case. Handle failure outcomes are explicit: unattached handle, node not running, channel closed before enqueue, event loop exit before dispatch, and the same preflight errors the synchronous API returns.

`SystemCommand`'s derives reduce from `Debug, Clone, Copy, PartialEq, Eq, Display` to `Debug, Display`: the responder is single-use, the claims payload owns a `Vec<InstrumentId>`, and no production path clones or compares commands - the channel moves them. The tests that relied on the old derives now match the variant and assert its fields.

## Testing

`test_handle_register_external_order_claims_while_event_loop_runs` and its deregistration sibling drive `run()` to the main loop from a task on the same runtime and assert the claim change reached both the execution engine and the reconciliation manager. `test_handle_registration_returns_synchronous_preflight_error` and its deregistration sibling pin that a duplicate claim and mismatched tiers come back through the responder with the same error text the synchronous API produces. `test_shutdown_drain_drops_external_claim_response` pins that a command drained at shutdown resolves the caller with an error rather than leaving it pending. The new node tests run on current-thread runtimes with bounded waits on the handle's own state, no wall-clock sleeps. With the register dispatch arm disarmed (reporting success without registering), the running-path test fails on the manager-tier state assertion.

### DIFF
--- a/crates/common/src/actor/tests.rs
+++ b/crates/common/src/actor/tests.rs
@@ -5485,7 +5485,10 @@ fn test_reconnect_socket_enqueues_typed_command(
     drop(actor);
 
     let SystemCommand::ReconnectSocket(command) =
-        system_rx.try_recv().expect("reconnect command queued");
+        system_rx.try_recv().expect("reconnect command queued")
+    else {
+        panic!("expected reconnect command");
+    };
 
     assert_eq!(command.trader_id, trader_id);
     assert_eq!(command.client_id, ClientId::from("POLYMARKET"));
--- a/crates/common/src/messages/mod.rs
+++ b/crates/common/src/messages/mod.rs
@@ -62,7 +62,9 @@ pub enum DataEvent {
 #[derive(Debug, Display)]
 pub enum SystemCommand {
     ReconnectSocket(system::ReconnectSocket),
+    #[cfg(feature = "live")]
     RegisterExternalOrderClaims(system::RegisterExternalOrderClaims),
+    #[cfg(feature = "live")]
     DeregisterExternalOrderClaims(system::DeregisterExternalOrderClaims),
 }
 
--- a/crates/common/src/messages/mod.rs
+++ b/crates/common/src/messages/mod.rs
@@ -59,9 +59,11 @@ pub enum DataEvent {
 }
 
 /// System command variants routed to a live node.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Display)]
+#[derive(Debug, Display)]
 pub enum SystemCommand {
     ReconnectSocket(system::ReconnectSocket),
+    RegisterExternalOrderClaims(system::RegisterExternalOrderClaims),
+    DeregisterExternalOrderClaims(system::DeregisterExternalOrderClaims),
 }
 
 /// System event variants routed to a live node.
--- a/crates/common/src/messages/system/claims.rs
+++ b/crates/common/src/messages/system/claims.rs
@@ -13,19 +13,19 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
 
-pub mod claims;
-pub mod component;
-pub mod queue;
-pub mod shutdown;
-pub mod socket;
-pub mod trading;
+use nautilus_model::identifiers::{InstrumentId, StrategyId};
 
-// Re-exports
-pub use claims::{DeregisterExternalOrderClaims, RegisterExternalOrderClaims};
-pub use component::ComponentStateChanged;
-pub use queue::{QueueCondition, QueueState, QueueStateChanged};
-pub use shutdown::ShutdownSystem;
-#[cfg(feature = "live")]
-pub(crate) use socket::socket_endpoint;
-pub use socket::{ReconnectSocket, SocketState, SocketStateChange, SocketStateChanged};
-pub use trading::TradingStateChanged;
+/// Command requesting registration of external order claims on a running live node.
+#[derive(Debug)]
+pub struct RegisterExternalOrderClaims {
+    pub strategy_id: StrategyId,
+    pub claims: Vec<InstrumentId>,
+    pub response: tokio::sync::oneshot::Sender<anyhow::Result<()>>,
+}
+
+/// Command requesting deregistration of external order claims on a running live node.
+#[derive(Debug)]
+pub struct DeregisterExternalOrderClaims {
+    pub strategy_id: StrategyId,
+    pub response: tokio::sync::oneshot::Sender<anyhow::Result<()>>,
+}
--- a/crates/common/src/messages/system/mod.rs
+++ b/crates/common/src/messages/system/mod.rs
@@ -13,14 +13,17 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------------------------------
 
-pub mod claims;
 pub mod component;
 pub mod queue;
 pub mod shutdown;
 pub mod socket;
 pub mod trading;
 
+#[cfg(feature = "live")]
+pub mod claims;
+
 // Re-exports
+#[cfg(feature = "live")]
 pub use claims::{DeregisterExternalOrderClaims, RegisterExternalOrderClaims};
 pub use component::ComponentStateChanged;
 pub use queue::{QueueCondition, QueueState, QueueStateChanged};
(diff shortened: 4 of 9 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. I found two blockers before this can merge:

- `nautilus-common` no longer compiles without the `live` feature. The new claims module and `SystemCommand` variants compile unconditionally, but their responder fields use optional `tokio`. I confirmed this with `CARGO_BUILD_JOBS=4 cargo check -p nautilus-common --no-default-features --lib`, which fails with `E0433` at both oneshot fields. The default-feature build needs to remain supported.
- The handle methods can wait indefinitely after `LiveNode::start()`. That route leaves the node in `Running` while deliberately not servicing runner channels, so both methods pass the state check, enqueue successfully, and never receive a response. The API needs to reject this state, with registration and deregistration coverage for the manual-start route.

One smaller API-contract point: once either method enqueues its command, dropping or timing out the future does not cancel the request, and the event loop may still apply the mutation. The public docs should state this so callers do not treat a timeout as proof that no state changed.

folknor:
Thank you.

Both blockers are fixed.

## The default-feature build

`claims.rs` named `tokio::sync::oneshot::Sender` unconditionally while `tokio` is optional behind `live`, and the module declaration, its re-exports and both `SystemCommand` variants were ungated, so the responder fields failed to resolve without the feature. All three sites are now behind `#[cfg(feature = "live")]`, following the `DataEvent::DeFi` variant in the same file. `strum::Display` observes the filtered definition, so the rendered names are unchanged, and every destructuring site already compiles only with `live` enabled: the `actor` test is `#[cfg(feature = "live")]`, and `python` enables `live`.

## The unserviced start route

`LiveNode::start` reaches `finish_startup_trader` and enters `Running` without driving the receivers, so both methods passed the state check, enqueued against a live but unpolled receiver, and awaited a response that could not arrive. `LiveNodeHandle` now carries a private `Arc<AtomicBool>` recording whether a run loop is servicing the runner channels, set in `run_with_mode` immediately before `finish_startup_trader` and cleared by an RAII guard, which covers the startup-abort, stop, runtime-error and future-drop exits uniformly. Both methods reject with a distinct error naming `run`/`run_with_mode` when it is false.

The event-store replay branch reaches `Running` through `finish_startup_replay` and returns before the loop starts, so it is left unserviced as well and those calls report the same condition rather than depending on when the dropped receiver closes the channel.

A `NodeState` variant would have been the other option. It is a `#[pyclass]`, so a new discriminant is a binding change, and the distinction is about one service mechanism rather than lifecycle state.

## The cancellation contract

Both methods now state that once the command is enqueued, dropping or timing out the future does not cancel it, the event loop may still apply the mutation, and a timeout is therefore not evidence that no state changed. Both `# Errors` sections also name the inactive-loop condition.

## Testing

`test_handle_register_external_order_claims_rejects_manual_start` and its deregistration counterpart drive the full `start()` path and assert the error. Both fail against the previous revision by reaching their bound with `Elapsed`, which is the hang you described.

An earlier version of these tests used the event-store replay fixture, whose `start()` returns before `finish_startup_trader` - the function `start()` shares with `run_with_mode`. Setting the flag on that shared call reintroduces the defect, and the replay-fixture tests passed against exactly that regression while the full-path ones failed.

`test_handle_register_external_order_claims_rejects_event_store_replay_start` and `test_handle_register_external_order_claims_rejects_replay_run` cover the two replay routes. The second reaches the branch through `run()`, so it detects a flag set in `run_with_mode` ahead of the replay return: with that injected it fails on the assertion, receiving the channel-closed message instead of the inactive-loop one.

The four existing handle tests are unchanged and still pass. `nautilus-common` builds again without `live`, checked against the same two `E0433` sites you reported.

cjdsellers:
Hey @folknor,

Thanks for the detailed follow-up. I’ve had another look at the latest changes, and the two blockers and cancellation contract are addressed.

The extra machinery is justified here: `run` owns the node, the existing system-command channel is the natural synchronization point, and the servicing flag distinguishes a node that can answer handle requests from one that is only in `Running`. I’m also comfortable treating the `SystemCommand` derive changes as an explicit release-candidate API break.

If you’re able, could you fold two small cleanups into the rebase:

- Have `process_system_command` call the public claim methods directly and remove the `_inner` variants, so there is only one path to maintain.
- Change the replay error to say that handle calls require an active servicing loop and are unavailable after manual `start()` or in replay mode. Recommending `run()` is misleading when replay reached that state through `run()`.

The branch currently conflicts with `develop`, so it needs a rebase, followed by the required checks running and passing. Once that is done, this looks good from my side.

folknor:
Thank you.

Both cleanups are in, folded into the rebase onto current develop.

`process_system_command` now calls the public `register_external_order_claims` / `deregister_external_order_claims` directly and the `_inner` variants are gone, so the dispatcher and the direct API share one body per method.

The error now reads "handle calls require an active event loop servicing the runner channels, and are unavailable after manual start or in event-store replay mode", with the `# Errors` sections and the four test assertions updated to match.

The rebase absorbed nothing beyond the import merge with the fill-report work; the diff shape is otherwise unchanged.

cjdsellers:
Hey @folknor,

Thanks for folding those cleanups into the rebase. I’ve had another look at the latest changes, and this looks good from my side 👌

cjdsellers (approved):

cjdsellers:
@folknor, because this is a slightly larger change - just holding it until after next release (being arranged shortly). Thanks for your patience :pray:

cjdsellers:
Hey @folknor,

Thanks again for the careful implementation and for your patience while I revisited the design after approval. I should have caught this architectural issue before you spent time addressing each review point.

I'm going to close this in favor of the [cache-backed implementation now on `develop`](https://github.com/nautechsystems/nautilus_trader/commit/681607428c6ff7ecedd5c646637964dab87f33b6).

The extra machinery here came from accidental complexity in the Rust port. In v1, `LiveExecutionEngine` inherited from `ExecutionEngine`, so its live reconciliation logic could access the same external order claims map used by the execution engine. Rust could not carry that inheritance pattern across. The port instead separated live reconciliation into `ExecutionManager` and gave it another private claims map alongside the one in `ExecutionEngine`.

That created two owners for one logical mapping. Runtime changes then appeared to require synchronizing both copies. This PR was continuing down that complexifying pathway by adding system messages that carry one-shot senders, acknowledgement handling, runner servicing state, and shutdown and cancellation handling.

The replacement puts the live `external_order_claims` in the shared cache as the canonical state. `ExecutionEngine` and `ExecutionManager` now read the same map, while `external_order_instrument_ids` remains the serializable strategy configuration intent. Strategies update their claims through a direct setter, with no duplicated state or cross-component acknowledgement protocol.

Claim transfers remain explicit: the original strategy releases its claims before another strategy acquires them. This accepts a small handover window rather than adding coordination machinery disproportionate to the risk it closes.

The implementation here was careful and complete. I'm closing it because the underlying architecture has changed.

folknor:
Yes, I was just taking a look at this PR myself after your commit.

There's a small residual here as a result of https://github.com/nautechsystems/nautilus_trader/commit/681607428c6ff7ecedd5c646637964dab87f33b6 that I'll file separately.

PR #4874, opened 2026-08-27
Share the execution event emitter sender across clones

folknor:
Share the execution event emitter sender across clones

## A clone taken before `set_sender` is permanently senderless

`ExecutionEventEmitter` derives `Clone` over `sender: Option<UnboundedSender<ExecutionEvent>>` held by value, so a clone freezes the sender state of the instant it was taken: one taken before the adapter's `start()` installs the sender holds `None` for its lifetime, and `send_order_event` on it warns and drops the event. Every in-tree adapter installs the sender in `start()` and takes its emitter clones afterwards, so the ordering holds by convention rather than by construction - and a host that builds execution clients on one thread and connects them on another takes construction-time clones that stay senderless when the runner later installs the sender. The thread-local `EXEC_EVENT_SENDER` slot cannot heal such a clone from another thread.

The sender now lives in an `Arc<ArcSwapOption<UnboundedSender<ExecutionEvent>>>` shared through the derived `Clone`, so `set_sender` installs - or replaces, exactly as today - the sender for the emitter and all its clones, and `is_initialized` answers for the shared slot. Every send path takes one atomic `load()` snapshot for the whole operation; the added per-send cost is that lock-free load, with no mutex or rwlock. Public signatures are unchanged.

The alternative shape is to enforce the ordering instead - thread the sender through construction or make each adapter clone only after `start()` - but that touches every holder (~150 `emitter.clone()` sites and 68 `set_sender` calls across 18 execution clients) to establish a convention the representation can guarantee in one place. This diff changes only the emitter.

## Testing

`test_clone_before_set_sender_observes_sender` clones before installing, installs through the original, and asserts the clone reports initialized and delivers to the channel; with a freezing `Clone` substituted for the derived one (each clone getting a fresh empty cell), it fails at `is_initialized`. `test_set_sender_replaces_existing_sender` pins the existing overwrite semantics: the displaced channel disconnects and the new one receives. `test_never_initialized_sender_drops_or_errors` pins the warn-and-drop path and `try_send_order_event`'s error for an emitter never given a sender.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6409,6 +6409,7 @@ version = "0.63.0"
 dependencies = [
  "ahash 0.8.12",
  "anyhow",
+ "arc-swap",
  "async-trait",
  "bon",
  "bytes",
--- a/crates/adapters/lighter/fuzz/pornin/Cargo.lock
+++ b/crates/adapters/lighter/fuzz/pornin/Cargo.lock
@@ -636,7 +636,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I've had a close look at this, including the clone-before-initialization path and sender replacement behavior. This looks good to merge.

PR #4875, opened 2026-08-27
Add an account registration wait to the live node handle

folknor:
Add an account registration wait to the live node handle

## No signal exists at the account cache insertion boundary

A live host that registers an execution client cannot await the account row existing in the cache: `Cache::add_account` returns without notifying, and the live route is fire-and-forget end to end - the adapter emits `ExecutionEvent::Account`, the runner forwards it to the portfolio's msgbus endpoint, and the portfolio inserts. A deployed consumer polls the cache every 10 ms under a wall bound to know when its account is usable; an adapter-side latch cannot substitute, because the adapter forwarding the event only queues it.

`LiveNodeHandle` gains `await_account_registered(account_id)`. The node marks an account registered after dispatching each `ExecutionEvent::Account` and confirming the row is in the cache - msgbus dispatch is synchronous on the runner thread, so the mark is at-or-after the insertion, and the cache re-check means a failed portfolio insertion (logged and swallowed today) is not reported as registration. All four node-owned dispatch paths are hooked, including the startup buffering phases and the shutdown drain; the standalone `AsyncRunner` is untouched, as are `Cache`, the portfolio, and every adapter. Accounts restored from storage during cache preparation are seeded, since they never arrive as events.

The carrier is a registered-set plus generation watch shared between the node and its handles, with waits subscribing before the first predicate check so no insertion can fall between check and sleep. A wait on an already-registered account resolves immediately; an unattached or not-running handle errors. The tracker closes only after each terminal path's final event drain, so an account inserted by the shutdown drain releases its waiter with success rather than a stop error; a waiter stranded across a stop-and-restart terminates with a lifecycle error instead of silently surviving; and `dispose()` invalidates the tracker before the kernel clears the cache, so a retained handle cannot resolve against a row disposal removes.

## Testing

`test_account_waiter_released_after_running_loop_caches_account` drives `run()` to the main loop, sends an account event, and asserts the waiter resolves with the row present; with the mark's cache predicate inverted, it fails by timeout. Sibling tests pin immediate resolution for an already-registered account, seeding from a pre-populated cache through `start()`, the not-running and unattached errors, mark-before-close ordering (a drain-inserted account wins over the stop error), termination of a waiter whose stop and restart coalesce before it wakes, and the post-dispose error for a previously registered account. All on current-thread runtimes with bounded waits, no wall-clock sleeps.

### DIFF
--- a/crates/live/src/node/mod.rs
+++ b/crates/live/src/node/mod.rs
@@ -411,6 +411,7 @@ impl LiveNode {
 
         self.kernel.reset_shutdown_flag();
         self.kernel.start_async().await;
+        self.seed_registered_accounts();
 
         if self.kernel.is_event_store_replay() {
             log::info!(
@@ -1051,6 +1052,7 @@ impl LiveNode {
         self.handle.set_starting();
         self.kernel.reset_shutdown_flag();
         self.kernel.start_async().await;
+        self.seed_registered_accounts();
 
         if self.kernel.is_event_store_replay() {
             log::info!(
@@ -1929,7 +1931,6 @@ impl LiveNode {
 
         let cache = self.kernel.cache();
         if !cache.borrow().has_backing() {
-            self.seed_registered_accounts();
             return Ok(());
         }
 
@@ -1940,7 +1941,6 @@ impl LiveNode {
             .is_some_and(|config| config.flush_on_start)
         {
             cache.borrow_mut().flush_db();
-            self.seed_registered_accounts();
             return Ok(());
         }
 
@@ -1953,7 +1953,12 @@ impl LiveNode {
                 .context("Failed to load persistent cache")?;
         }
 
-        self.seed_registered_accounts();
+        // Deliberately no tracker seeding here: `prepare_cache` runs before
+        // `set_starting` bumps the lifecycle generation, so a seed taken now
+        // would tag database-loaded accounts with the PREVIOUS lifecycle and
+        // wrongly release a waiter from it. The unconditional seed after
+        // `kernel.start_async` covers every account loaded here, at the
+        // started generation.
         Ok(())
     }
 
@@ -2149,9 +2154,14 @@ impl LiveNode {
 
     async fn connect_exec_clients(&mut self, deadline: dst::time::Instant) -> anyhow::Result<()> {
         let remaining = deadline.saturating_duration_since(dst::time::Instant::now());
-        dst::time::timeout(remaining, self.kernel.connect_exec_clients())
+        let result = dst::time::timeout(remaining, self.kernel.connect_exec_clients())
             .await
-            .map_err(|_| anyhow::anyhow!("exec-connect timeout"))
+            .map_err(|_| anyhow::anyhow!("exec-connect timeout"));
+        // Seed on the timeout path too: clients connect concurrently, so an
+        // aggregate timeout can leave an already-connected client's account
+        // cached with no later observation point before the tracker closes.
+        self.seed_registered_accounts();
+        result
     }
 
     /// Connects execution clients and checks all engines are connected.
@@ -4085,6 +4095,7 @@ mod tests {
             Arc,
             atomic::{AtomicBool, Ordering},
         },
+        task::{Context, Poll, Waker},
     };
 
     use bytes::Bytes;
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @folknor,

Thank you for the PR. **I found two cases that still break the waiter contract:**

- `seed_registered_accounts()` runs before kernel startup and execution-client connection. Event-store replay and the supported sandbox execution client both insert account rows directly without an `ExecutionEvent::Account`, so the cache can contain the account while the waiter remains parked. In replay mode no later account event can release it.
- A waiter from lifecycle 1 can observe only the final state after stop -> restart -> registration in lifecycle 2. Because the registered predicate is accepted before the generation mismatch, it returns success instead of the required lifecycle error.

Could you make account observation cover those direct insertion paths and keep registrations generation-aware across coalesced updates? The regression coverage should drive restored/direct insertion, stop -> restart -> registration before the old waiter resumes, and the actual startup/final-drain wiring. The `# Errors` section should also mention the absent-account `Idle`/`Stopped` rejection.

folknor:
Thank you.

Both contract breaks are fixed.

## Direct insertions

The tracker was seeded once in `prepare_cache` and marked only from the `ExecutionEvent::Account` dispatch hooks, so replay restoration and the sandbox client's send to the portfolio endpoint were invisible. The tracker is now re-seeded from the cache at the two startup boundaries the node owns: immediately after `kernel.start_async` (after replay restoration, with the new lifecycle generation already active) and inside the shared exec-connect helper both `start()` and `run_with_mode` use - including its timeout path, since clients connect concurrently and an aggregate timeout can leave an already-connected client's account cached with no later observation point. Every seed reads accounts actually present in the cache, keeping the mark truthful against a swallowed portfolio insertion failure.

I considered a cache-level watch and a topic subscription. Neither works: `Cache` has no notification machinery in any of its insertion paths, and the portfolio's endpoint-side `update_account` publishes no account topic, so nothing observable exists downstream of that insertion.

The `prepare_cache` seed is gone rather than kept alongside: it ran before `set_starting` bumps the generation, so a database-loaded account would have been tagged with the previous lifecycle and could wrongly release a waiter from it - the same causal defect at a different observation point.

## Lifecycle generations

`registered` is now a map from account to the lifecycle generation at first registration, immutable on re-mark, cleared only by disposal. The fast path and the wake loop share one rule - registered at or before the waiter's captured generation returns Ok; otherwise stopped or a generation change is the lifecycle error; otherwise keep waiting. A lifecycle-2 registration observed by a coalesced lifecycle-1 waiter now errors, while a registration that landed inside the waiter's own lifecycle still succeeds even when observed after a restart, and an account already registered at call entry returns Ok in any node state. Only an absent account is rejected while `Idle` or `Stopped`, and the `# Errors` section now says so.

## Testing

Six node-level tests exercise the startup, connect, and drain paths: replay restoration through `start()` and through `run()`; direct insertion with a gated client that holds `connect` open until the waiter is provably parked, so the post-connect seed is what wakes it; the aggregate-timeout case above (one client connected, one held past the deadline - `start()` errors, the waiter still releases); stop -> restart -> lifecycle-2 registration returning the lifecycle error to the old waiter; an account event drained and marked before tracker close still releasing its same-generation waiter; and the already-registered fast path after a normal stop. Removing the post-replay seed times out the replay test; making the tag comparison vacuous flips the lifecycle test from the required error to success.

cjdsellers:
@folknor, Thank you for the PR.

I'm going to pause on this one as there are some higher level design considerations that need to be considered and discussed. To allocate bandwidth efficiently on my end, I'm going to prioritize some of your other simpler PRs first, then circle back to this one - thanks for your patience! :pray:

folknor:
@cjdsellers Thank you, understood.

For your planning: beyond the two open RFC issues, I hold roughly twenty verified findings whose fixes each need a design ruling from you - nearly all core machinery questions (reconciliation scope, position accounting, lifecycle semantics) rather than anything specific to my own projects. Happy to send the list by email or Discord, or file them as small PRs at whatever pace suits you.

The one thing that would help me sequence: whether API-shaped changes remain welcome after 2.0.0, or whether some of these should reach you before the release.

cjdsellers:
@folknor, thanks for the heads up. I think at this point it would best direct/coordinate efforts if you send me the findings via email, then I can get a sense of the overall themes of what is being turned up and can advise on where the targeted PRs will work and which order might be best. Much appreciated 🙏

PR #4881, opened 2026-08-29
Refuse emulated order submission from execution algorithms

folknor:
A follow-up to the execution algorithm command routing work in #4848, which fixed the modify half of this class and named the submission half.

## An execution algorithm can submit an emulated order that nothing owns

`ExecutionAlgorithm::submit_order` never reads the order's `emulation_trigger`. It caches the child, publishes `OrderInitialized`, and sends the `SubmitOrder` to the risk engine unconditionally. An algorithm can produce such a child through the crate's own API: `spawn_limit` takes an `emulation_trigger` and passes it to `LimitOrder::new`. `OrderEmulator` never learns the order exists, so nothing holds it local and nothing releases it when the trigger fires.

The fix refuses the submission before the order is cached or its initialization event published, and restores the primary order's deducted quantity on that path so a `reduce_primary` spawn does not strand it. This matches `ExecAlgorithm.submit_order` in the Python implementation, which raises `ValueError` on a live trigger under the stated rationale that emulated orders cannot be sent from execution algorithms.

Refusal rather than emulator routing, though `modify_order` and `cancel_order` in the same trait do route: `OrderEmulator::release` sends a released order carrying an `exec_algorithm_id` to `<algorithm-id>.execute`, and that endpoint dispatches a `SubmitOrder` to `on_order` rather than resuming the interrupted submission. A released child would re-enter the algorithm as fresh primary input and, for `TwapAlgorithm`, be scheduled and spawned again. Submission establishes ownership; modify and cancel act on an order whose owner is already established.

## A spawned MARKET_TO_LIMIT order stored a trigger it can never report

`spawn_market_to_limit` applied its `emulation_trigger` argument with `set_emulation_trigger` after construction, writing to `OrderCore` a value `MarketToLimitOrder::emulation_trigger` always reports as `None`. `MarketToLimitOrder` is initialized with `NO_TRIGGER` in both implementations and cannot be emulated, so the getter is right and the write was contradictory hidden state. The argument is now accepted and unapplied, as it is in the Python `spawn_market_to_limit`, and the rustdoc says so.

## Testing

`test_algorithm_submit_order_refuses_emulated_limit_spawn` spawns a child with a `BidAsk` trigger and `reduce_primary`, then asserts the error names the reason and the client order ID, that neither the risk endpoint nor the emulator endpoint received a command, that the child is absent from the cache with no `OrderInitialized` published for it, that the primary quantity is restored, and that the pending spawn reduction is consumed. Asserting both endpoints pins refusal rather than emulator routing. `test_algorithm_submit_order_routes_unemulated_spawn_to_risk` is the control.

Against the unfixed code the test fails on the error assertion; with that assertion removed it fails on the risk endpoint receiving the command, so the routing claim does not ride on the error.

## Reported rather than fixed

`restore_primary_order_quantity` consumes the `pending_spawn_reductions` entry before it looks up the primary, so a missing primary or a failed cache update discards the bookkeeping without restoring anything. That is pre-existing on the denied and rejected paths, and a repair needs a policy for whether such failures are retried at all.

### DIFF
--- a/crates/trading/src/algorithm/mod.rs
+++ b/crates/trading/src/algorithm/mod.rs
@@ -568,6 +568,10 @@ pub trait ExecutionAlgorithm: DataActor {
     /// by the spawned quantity. If the spawned order is subsequently denied or
     /// rejected (before acceptance), the deducted quantity is automatically
     /// restored to the primary order.
+    ///
+    /// `_emulation_trigger` is accepted for signature parity and is not applied:
+    /// a `MARKET_TO_LIMIT` order is always initialized with `NoTrigger` and
+    /// cannot be emulated.
     #[expect(clippy::too_many_arguments)]
     fn spawn_market_to_limit(
         &mut self,
@@ -577,7 +581,7 @@ pub trait ExecutionAlgorithm: DataActor {
         expire_time: Option<UnixNanos>,
         reduce_only: bool,
         display_qty: Option<Quantity>,
-        emulation_trigger: Option<TriggerType>,
+        _emulation_trigger: Option<TriggerType>,
         tags: Option<Vec<Ustr>>,
         reduce_primary: bool,
     ) -> MarketToLimitOrder
@@ -596,7 +600,7 @@ pub trait ExecutionAlgorithm: DataActor {
                 .track_pending_spawn_reduction(client_order_id, quantity);
         }
 
-        let mut order = MarketToLimitOrder::new(
+        MarketToLimitOrder::new(
             primary.trader_id(),
             primary.strategy_id(),
             primary.instrument_id(),
@@ -619,13 +623,7 @@ pub trait ExecutionAlgorithm: DataActor {
             tags.or_else(|| primary.tags().map(<[Ustr]>::to_vec)),
             UUID4::new(),
             ts_init,
-        );
-
-        if emulation_trigger.is_some() {
-            order.set_emulation_trigger(emulation_trigger);
-        }
-
-        order
+        )
     }
 
     /// Reduces the primary order's quantity by the spawn quantity.
@@ -775,9 +773,11 @@ pub trait ExecutionAlgorithm: DataActor {
 
     /// Submits an order to the execution engine via the risk engine.
     ///
+    /// Execution algorithms cannot submit orders carrying a live emulation trigger.
+    ///
     /// # Errors
     ///
-    /// Returns an error if order submission fails.
+    /// Returns an error if the order carries a live emulation trigger or submission fails.
     fn submit_order(
         &mut self,
         order: OrderAny,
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. I found one Python-facing compatibility issue before this can merge.

The new guard returns through `PyExecutionAlgorithm.submit_order`'s existing `to_pyruntime_err` mapping, so Python execution algorithms receive `RuntimeError` for a live emulation trigger. The legacy `ExecAlgorithm.submit_order` contract, and the parity described in this PR, use `ValueError` for that invalid input. Could you preserve `ValueError` at the Python boundary and add a Python-facing regression test?

While touching this, could you also make the local refusal contract explicit in the public docs and restoration log? This error restores the cached primary, consumes the pending reduction, publishes `OrderUpdated`, and currently logs that the child was denied/rejected even though it was refused before submission.

folknor:
Thank you.

Both points are addressed.

## The Python boundary

The refusal is now a typed `EmulatedOrderSubmissionError` carrying the client order id, and `py_submit_order` maps it to `ValueError` by downcast while every other submission failure keeps `RuntimeError` - the same pattern the strategy wrapper uses for `OrderError`. The Rust guard still runs on the Python path, so the spawn-deducted primary quantity is still restored before the error surfaces; a wrapper-side precheck would have raised without restoring. The error message is unchanged.

## The refusal contract

`restore_primary_order_quantity` takes the outcome explicitly and the log now reads "was refused before submission" on this path, with the denied/rejected wording untouched. The `submit_order` rustdoc states the contract - refusal restores the primary quantity, consumes the pending reduction, and publishes `OrderUpdated`, for spawned orders with a pending primary reduction - and the three spawn methods' restoration sentences include the refusal case.

## Testing

`test_python_submit_order_maps_emulation_refusal_to_value_error` registers a `PyExecutionAlgorithm`, submits an emulated limit order through `py_submit_order`, and asserts `ValueError` with the client order id and "live emulation trigger" in the message; a second assertion pins that an unrelated failure (unregistered algorithm, plain market order) stays `RuntimeError`, covering the fallback arm. With the guard reverted to the untyped error, this test fails at the `ValueError` assertion and the existing Rust test fails at its new downcast assertion; both pass restored.

cjdsellers:
Thanks, @folknor. The typed error, Python `ValueError` mapping, updated refusal wording, and regression coverage address my earlier points. I don't see a blocker in the latest head 👍.

I found two small public-contract gaps that can follow up separately: `spawn_limit` still accepts a live `emulation_trigger`, but the resulting child cannot be submitted; and after a refused submission restores the cached primary, a native Rust caller's existing `&mut primary` value remains reduced and must be discarded or refreshed before reuse. ~~If you're able, could you make both constraints explicit in the execution algorithm docs?~~

folknor:
@cjdsellers Sure thing! I'm a bit confused, do you want further changes here, or all of those in a followup? :joy:

cjdsellers:
@folknor, I could have been clearer. I can action those docs follow-ups on my end :smile:

cjdsellers (approved):

PR #4887, opened 2026-08-30
Bound WMA and HMA periods to their buffer capacity

pucedoteth:
  described an agreed limitation below — **`make format` ran; `make pre-commit` did not complete, see Testing**

## Summary

`WeightedMovingAverage` buffers inputs in an `ArrayDeque` with a fixed `MAX_PERIOD` capacity of 8192 but never validated `period` against it, so a larger period was accepted and then silently degraded. `HullMovingAverage` forwards its `period` straight into a `WeightedMovingAverage` and only checked `period > 0`, inheriting the problem in a worse form. Both now validate against `MAX_PERIOD`.

### WMA: never initializes

`initialized` compares `count()` against the unclamped `period`, so it can never become true once `period` exceeds capacity, while `weighted_average` keeps returning a value built from the last 8192 weights:

```
period            = 8193
inputs buffered   = 8192
initialized()     = false   <- after 8293 updates
value             = 4196.5
```

### HMA: reports ready on a short window

HMA tracks its own count rather than its inner averages', so it does report ready while `ma2` is short of inputs:

```
hma.period        = 10000
hma.count         = 10500
hma.initialized() = true    <- inner WMA(10000) holds at most 8192
hma.value()       = 10281.690433849888
```

That second figure is the configuration `hma`'s own `test_new_with_positive_period_constructs` pinned as `case(10_000)`, so that case asserted successful construction of an indicator that cannot work. It now pins `MAX_PERIOD`, with a separate case covering rejection above it. That is the one existing expectation this changes.

`SimpleMovingAverage`, `LinearRegression`, `CommodityChannelIndex`, `BollingerBands` and `Stochastics` all already reject a period above their capacity, so this brings two outliers into line rather than introducing a new convention.

**Alternative, if you prefer:** raise `MAX_PERIOD`, or move `WeightedMovingAverage` to a heap-allocated buffer, so a 10,000 period genuinely works instead of being rejected. That keeps `case(10_000)` valid but changes allocation behaviour, so it seemed like your call rather than mine. Happy to redo it that way.

## Related issues/PRs

None. I checked open issues and pull requests for existing work on these indicators.

## Type of change

- [x] Bug fix (non-breaking)

## Breaking change details

Constructing `WeightedMovingAverage` or `HullMovingAverage` with `period > 8192` now fails instead of returning a silently broken indicator. Both validate in `new_checked`, and both Python bindings raise `ValueError`.

The moving average factory is an in-tree caller on this path. `MovingAverageFactory::create` calls `HullMovingAverage::new` directly, and the factory-backed indicators forward a user period without a cap of their own, so `RelativeStrengthIndex(10_000, ma_type=MovingAverageType.Hull)` constructed before this change and now reaches `expect(FAILED)`. Those bindings take `py_new(...) -> Self` rather than `PyResult<Self>`, so a release wheel aborts rather than raising. RSI, CMO, MACD, ATR, KC, RVI, DM, KVO, PSL and Pressure all share that route.

This is not a new failure mode for the factory: `SimpleMovingAverage::new` already asserts against its own 1024 cap on exactly that path. Widening the factory to checked constructors is out of scope for this PR and is being tracked separately.

Error messages carry both the offending period and the bound, matching `sma.rs`, `roc.rs` and `swings.rs`: `WeightedMovingAverage: period 8193 exceeds MAX_PERIOD (8192)`. `MAX_PERIOD` is `pub(crate)` and is not exported to Python, so the value has to be in the message for a caller to act on it.

## Documentation

- [x] Doc comments on both constructors updated to list the new bound
- [x] `make py-stubs` not needed. I ran `generate_stubs.py`: `PyResult<Self>` produces the same `__new__` signature for `HullMovingAverage` as `Self` did

## Testing

Rust: three WMA tests (panic above max, `new_checked` errors above max and the message carries both numbers, boundary at `MAX_PERIOD` still reaches `initialized`) and one HMA test (panic above max).

Python: three HMA tests covering construction at the bound, `ValueError` above it, and `ValueError` at zero.

Reverting only the guards and keeping the tests fails them. The boundary test passes either way, so the fix does not over-reject. Reverting only the interpolation in the WMA message fails `new_checked_period_exceeds_max_errors` and nothing else.

Ran locally:

- `cargo test -p nautilus-indicators` — 573 passed, 0 failed (plus 2 doc-tests)
- `cargo clippy -p nautilus-indicators --all-targets --all-features -- -D warnings` — clean
- `cargo +nightly fmt --check` — clean
- `maturin develop --profile nextest`, then `pytest python/tests/unit/indicators/` — 151 passed, which includes the three new HMA cases:

```
tests/unit/indicators/test_hma.py::test_new_at_max_period_constructs PASSED
tests/unit/indicators/test_hma.py::test_new_above_max_period_raises_value_error PASSED
tests/unit/indicators/test_hma.py::test_new_with_zero_period_raises_value_error PASSED
```

**Limitation:** `make pre-commit` did not complete locally. `prek` was still provisioning environments for the 70 configured hooks when it exhausted disk on my machine. I checked the hooks that apply to the changed files by hand instead: no trailing whitespace, no CRLF, no tabs, files end with a newline, and `cargo +nightly fmt --check` is clean.

### DIFF
--- a/crates/indicators/src/average/hma.rs
+++ b/crates/indicators/src/average/hma.rs
@@ -21,7 +21,7 @@ use nautilus_model::{
 };
 
 use crate::{
-    average::wma::WeightedMovingAverage,
+    average::wma::{MAX_PERIOD, WeightedMovingAverage},
     indicator::{Indicator, MovingAverage},
 };
 
@@ -107,12 +107,18 @@ impl HullMovingAverage {
     ///
     /// # Panics
     ///
-    /// Panics if `period` is not a positive integer (> 0).
+    /// Panics if `period` is not a positive integer (> 0), or exceeds `MAX_PERIOD`.
     #[must_use]
     pub fn new(period: usize, price_type: Option<PriceType>) -> Self {
         assert!(
             period > 0,
             "HullMovingAverage: period must be > 0 (received {period})"
+        );
+        // `ma2` below is a `WeightedMovingAverage` over the full `period`, so this
+        // indicator cannot support a period its inner averages cannot buffer.
+        assert!(
+            period <= MAX_PERIOD,
+            "HullMovingAverage: period {period} exceeds MAX_PERIOD ({MAX_PERIOD})"
         );
 
         let half = usize::max(1, period / 2);
@@ -176,7 +182,7 @@ mod tests {
     use rstest::rstest;
 
     use crate::{
-        average::hma::HullMovingAverage,
+        average::{hma::HullMovingAverage, wma::MAX_PERIOD},
         indicator::{Indicator, MovingAverage},
         stubs::*,
         testing::assert_approx_equal,
@@ -279,10 +285,16 @@ mod tests {
     }
 
     #[rstest]
+    #[should_panic(expected = "exceeds MAX_PERIOD")]
+    fn test_new_with_period_above_max_panics() {
+        let _ = HullMovingAverage::new(MAX_PERIOD + 1, None);
+    }
+
+    #[rstest]
     #[case(1)]
     #[case(5)]
     #[case(128)]
-    #[case(10_000)]
+    #[case(MAX_PERIOD)]
     fn test_new_with_positive_period_constructs(#[case] period: usize) {
         let hma = HullMovingAverage::new(period, None);
         assert_eq!(hma.period, period);
--- a/crates/indicators/src/average/hma.rs
+++ b/crates/indicators/src/average/hma.rs
@@ -15,6 +15,7 @@
 
 use std::fmt::Display;
 
+use nautilus_core::correctness::{FAILED, check_predicate_true};
 use nautilus_model::{
     data::{Bar, QuoteTick, TradeTick},
     enums::PriceType,
@@ -110,16 +111,27 @@ impl HullMovingAverage {
     /// Panics if `period` is not a positive integer (> 0), or exceeds `MAX_PERIOD`.
     #[must_use]
     pub fn new(period: usize, price_type: Option<PriceType>) -> Self {
-        assert!(
+        Self::new_checked(period, price_type).expect(FAILED)
+    }
+
+    /// Creates a new [`HullMovingAverage`] instance with the given period.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - `period` is not a positive integer (> 0).
+    /// - `period` exceeds `MAX_PERIOD`.
+    pub fn new_checked(period: usize, price_type: Option<PriceType>) -> anyhow::Result<Self> {
+        check_predicate_true(
             period > 0,
-            "HullMovingAverage: period must be > 0 (received {period})"
-        );
+            &format!("HullMovingAverage: period must be > 0 (received {period})"),
+        )?;
         // `ma2` below is a `WeightedMovingAverage` over the full `period`, so this
         // indicator cannot support a period its inner averages cannot buffer.
-        assert!(
+        check_predicate_true(
             period <= MAX_PERIOD,
-            "HullMovingAverage: period {period} exceeds MAX_PERIOD ({MAX_PERIOD})"
-        );
+            &format!("HullMovingAverage: period {period} exceeds MAX_PERIOD ({MAX_PERIOD})"),
+        )?;
 
         let half = usize::max(1, period / 2);
         let root = usize::max(1, (period as f64).sqrt() as usize);
@@ -130,7 +142,7 @@ impl HullMovingAverage {
         let ma2 = WeightedMovingAverage::new(period, get_weights(period), Some(pt));
         let ma3 = WeightedMovingAverage::new(root, get_weights(root), Some(pt));
 
-        Self {
+        Ok(Self {
             period,
             price_type: pt,
             value: 0.0,
@@ -140,7 +152,7 @@ impl HullMovingAverage {
             ma1,
             ma2,
             ma3,
-        }
+        })
     }
 }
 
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @pucedoteth,

Thanks for the PR. The capacity guard fixes the underlying buffer mismatch, but I found two points to address before merge:

- `HullMovingAverage`'s new upper-bound `assert!` is reached by the Python `__new__` wrapper. Release wheels compile with `panic = "abort"`, so `HullMovingAverage(8193)` terminates the Python process instead of raising `ValueError`. This needs a checked constructor path at the Python boundary, with a Python regression test for the upper bound.
- `MAX_PERIOD` only needs crate visibility for HMA. Making it `pub` adds a downstream Rust API; `pub(crate)` is enough for this use.

pucedoteth:
Thanks, both addressed in f717423.

**1. `panic = "abort"` at the Python boundary**

You're right, and the `assert!` was the wrong tool given how the binding calls it. `py_new` returned `Self` and called `new` directly, so there was nothing between the assertion and the interpreter.

Added `HullMovingAverage::new_checked` returning `anyhow::Result<Self>`, with `new` delegating via `.expect(FAILED)` — the same shape `WeightedMovingAverage` already uses. The binding is now:

```rust
fn py_new(period: usize, price_type: Option<PriceType>) -> PyResult<Self> {
    Self::new_checked(period, price_type).map_err(to_pyvalue_err)
}
```

This also fixes the zero-period case, which had the same problem and predates this PR.

I passed both existing panic messages through unchanged, so `test_new_with_zero_period_panics` and `test_new_with_period_above_max_panics` still pin the same text rather than being rewritten around the new path.

**2. `MAX_PERIOD` visibility**

Narrowed to `pub(crate)`. Checked that hma.rs is the only cross-module reader, and every other indicator (`bias`, `cci`, `roc`, `swings`) declares its limit privately, so `pub` was exporting API for no reason.

**Tests**

- Rust: `new_checked` error path for zero and above-max.
- Python: `ValueError` above the bound, `ValueError` at zero, and successful construction at `MAX_PERIOD`.

**Verification**

`cargo test -p nautilus-indicators` — 573 passed. `cargo clippy -p nautilus-indicators --all-targets` — clean. `cargo check -p nautilus-indicators --features python` — clean.

One thing to flag: I could not run the Python cases locally, as building the extension module needs more disk than this machine has free. They're written against the `pytest.raises(ValueError, match=...)` convention used elsewhere in the suite and the error text is pinned by the Rust side, but CI will be the first real execution — worth a look if it goes red.

cjdsellers:
Hi @pucedoteth,

Thanks for the quick turnaround. The `new_checked` shape is right, and picking up the zero-period case along the way is a good catch.

**One path still reaches the panic**:
`MovingAverageFactory::create` calls `HullMovingAverage::new` directly, and the factory-backed indicators forward a user period without a cap of their own. `RelativeStrengthIndex(10_000, ma_type=MovingAverageType.Hull)` constructed before this change and now hits `expect(FAILED)`, and since `RelativeStrengthIndex.py_new` returns `Self` rather than `PyResult`, a release wheel aborts instead of raising. RSI, CMO, MACD, ATR, KC, RVI, DM, KVO, PSL and Pressure all share that route.

`SimpleMovingAverage` already behaves this way above its own 1024 cap, so it isn't new ground, and I'd rather not pull a factory-wide checked-constructor change into this PR. I'll raise a separate issue for it.

Two things I'd like before merging:

- The description says no in-tree caller is near the bound. The factory is one, and that text carries into the squash commit, so it's worth correcting there along with the breaking-change note, which still describes the `assert!` version.
- `WeightedMovingAverage::new_checked` fails with "`period` must not exceed MAX_PERIOD". `MAX_PERIOD` is `pub(crate)` and isn't exported to Python, so a caller sees a constant they can't look up and no offending value. The HMA message you added in the same change interpolates both, as do the caps in `sma.rs`, `roc.rs` and `swings.rs`.

On the Python cases: stating the limitation up front is the right call, that's what `CONTRIBUTING.md` asks for. It doesn't remove the need for the run though. CI confirms a change you've already validated locally rather than standing in for local compute, and workflows need a maintainer to approve every push from a first-time contributor, so it makes a poor execution loop for both of us. I'd like those three cases to have actually run before this merges. Once the points above are addressed and you have a local run behind them, I'll approve CI and we can take it from there.

pucedoteth:
Thanks, both points were right and both are fixed in 9fccfae plus a description rewrite.

**The factory.** You are correct and my description was wrong. `MovingAverageFactory::create` calls `HullMovingAverage::new` directly, and `RelativeStrengthIndex::py_new` returns `Self`, so `RelativeStrengthIndex(10_000, ma_type=MovingAverageType.Hull)` constructs today and aborts after this change on a release wheel. I had grepped for direct `HullMovingAverage::new` callers and stopped at the indicator modules, so I missed the factory sitting one level up. The description now states this instead of claiming no caller is near the bound, names the ten affected indicators, and notes that `SimpleMovingAverage::new` already asserts on the same path so it is not a new failure mode. Agreed on keeping a factory-wide change out of this PR.

The breaking-change section also still described the `assert!` version, so that is corrected to `new_checked` and `ValueError`.

**The WMA message.** Fixed to match `sma.rs`, `roc.rs` and `swings.rs`:

```
WeightedMovingAverage: period 8193 exceeds MAX_PERIOD (8192)
```

`new_checked_period_exceeds_max_errors` now asserts the message carries both numbers rather than only that the call errors. Reverting just the interpolation fails that one test and nothing else.

**The Python cases.** Point taken, and thanks for spelling out why. Built with `maturin develop --profile nextest` and ran them:

```
tests/unit/indicators/test_hma.py::test_new_at_max_period_constructs PASSED
tests/unit/indicators/test_hma.py::test_new_above_max_period_raises_value_error PASSED
tests/unit/indicators/test_hma.py::test_new_with_zero_period_raises_value_error PASSED
```

The full `python/tests/unit/indicators/` directory is 151 passed. On the Rust side, `cargo test -p nautilus-indicators` is 573 passed and clippy with `-D warnings` is clean.

One note on the stubs, since I could not complete a full drift check: I ran `generate_stubs.py` and `HullMovingAverage.__new__` comes out identical under `PyResult<Self>`, so the `PyResult` change does not move the generated signature. `make pre-commit` still has not completed here, for the same disk reason as before, and the build for the Python run is what filled the disk again.

Developed with assistance from AI.

cjdsellers (approved):
Hi @pucedoteth,

Thanks for following through on the fixes and running the Python tests locally. I've had another look at the latest changes, and this looks good from my side.

PR #4895, opened 2026-08-31
Enable Polymarket data-only resolution subscriptions

### LINKED ISSUE #4878: Polymarket: Publish resolution data for watched markets without positions
# Feature Request: Data-only Polymarket resolution subscriptions

- [x] I’ve searched existing issues and discussions to avoid duplicates.

## Problem statement

Strategies may observe Polymarket markets for paper trading, shadow trading, analytics, or data collection without placing orders or holding positions.

Today, native Polymarket settlement events are effectively tied to position tracking. A data-only consumer can stream market data, but cannot subscribe to a selected instrument and reliably receive its official resolution through the normal data pipeline.

This prevents downstream systems, such as ClickHouse collectors, from recording the final official outcome for watched binary markets.

## Proposed solution

Support the existing native data subscriptions for selected Polymarket instruments:

- `subscribe_instrument_status(instrument_id)`
- `subscribe_instrument_close(instrument_id)`

These subscriptions should establish maintained resolution monitoring for the selected instrument without requiring:

- an execution client;
- credentials or an account;
- an order or position;
- synthetic `PositionEvent`s.

When Polymarket officially resolves the condition, emit the existing native events once:

- `InstrumentStatus(action=Close)`
- `InstrumentClose(close_price=1)` for the winning outcome token;
- `InstrumentClose(close_price=0)` for the losing outcome token.

For a binary market, subscribing to both outcome instruments should therefore produce the official winner/loser pair.

## Example usage

```python
def on_start(self):
    self.subscribe_instrument_status(
        self.up_instrument_id,
        client_id=POLYMARKET_CLIENT_ID,
    )
    self.subscribe_instrument_close(
        self.up_instrument_id,
        client_id=POLYMARKET_CLIENT_ID,
    )

    self.subscribe_instrument_status(
        self.down_instrument_id,
        client_id=POLYMARKET_CLIENT_ID,
    )
    self.subscribe_instrument_close(
        self.down_instrument_id,
        client_id=POLYMARKET_CLIENT_ID,
    )

def on_instrument_status(self, status):
    if status.action == MarketStatusAction.CLOSE:
        self.log.info(f"Market closed: {status.instrument_id}")

def on_instrument_close(self, close):
    self.log.info(
        f"Official outcome: {close.instrument_id}, close={close.close_price}",
    )
```

## Requirements

- Preserve the existing position-owned resolution path.
- Allow position and data subscriptions to independently own resolution monitoring for the same condition.
- Removing one owner must not remove the other.
- Use the maintained Polymarket WebSocket and Gamma/CLOB polling paths; do not add a new scheduler, transport, or event model.
- Respect configured instrument scopes and filters.
- Do not re-enroll terminal or retired markets.
- Ensure duplicate resolution messages or polls do not emit duplicate native closure events.

## Alternatives considered

Users can query resolution later with `PolymarketDataLoader`, but that requires separate scheduling, retries, and deduplication.

Using the final market price is not sufficient because it is not the authoritative settlement result.

## Additional context

This concerns data-only observation of selected Polymarket instruments. It should not change trading, positions, PnL, or existing settlement behavior for open positions.

mystic-io:
# Pull Request

- [ ] A maintainer agreed on the problem and approach in an issue, or this is a small, self-contained fix that does not need prior discussion
- [x] I have read and followed CONTRIBUTING.md and AI_POLICY.md
- [x] I ran all relevant tests locally, or I described an agreed limitation below

## Summary

Allow Polymarket data clients to watch a selected market until it resolves.

A client can now receive native market-close events without an account, order, or position. The adapter uses the existing resolution watchlist and Gamma outcome polling.

## Related issues/PRs

Closes #4878.

## Type of change

- [x] Documentation update

Not applicable.

## Documentation

## Testing

- [x] I added/updated tests to cover new and changed logic

Validated locally:

- `make format`
- `make pre-commit`
- Focused Polymarket resolution tests
- `cargo clippy -p nautilus-polymarket --all-targets -- -D warnings`
- Live data-only probe: emitted `InstrumentStatus(Close)` and `InstrumentClose` prices of `1.00` and `0.00` for a resolved market without an execution client or position.

Known limitation:

- `cargo test -p nautilus-polymarket` has one loopback WebSocket timeout. The same test fails on freshly fetched `develop`, so it is not caused by this change.


cjdsellers:
Hi @mystic-io,

Thanks for the PR. I found several resolution lifecycle cases that need addressing before this is ready:

- An uncached subscription can be accepted and then see `closed=true` during auto-load before the market has a strict outcome. The closed branch creates no resolution watch entry, so that subscriber misses the later official resolution.
- `market_resolved` is venue-wide, but every event is added to `closed_condition_ids`. This grows terminal state for unrelated markets, schedules unnecessary retirement scans, and bypasses the existing WebSocket serialization boundary.
- A missing expiration timestamp is represented by zero. These entries now bypass the configured maximum wait and are polled forever.
- HTTP or manual resolution clears the status and close intents without reconciling the WebSocket or recording terminal state. The token can remain subscribed and later be enrolled again.
- Auto-load checks the active resolution intent separately from inserting the watch entry. An unsubscribe in that window can leave a phantom data owner and continued polling.

There are also three smaller consistency issues: `emitted_condition_ids` should only include conditions that emitted an `InstrumentClose`, the integration guide needs to document independent status and close subscriptions, and the quote-only churn test should count the new ownership sets it seeds.

mystic-io:
@cjdsellers Thanks for the feedback! 🙏🏼 I’ve made the requested changes, rebased on the latest develop, and rerun the checks and tests. Everything looks good and it’s ready for another review.

cjdsellers:
Hi @mystic-io,

Thanks for the update. I've had another look and found four resolution lifecycle cases that still need addressing:

- `apply_condition_resolution` defers the entire condition when any active status or close subscription is missing from the watch entry. If one sibling never loads or is filtered out, settlement is blocked for every data and position owner, while the pending entry suppresses normal market data.
- The position-event updater does not consult `closed_condition_ids`, so a late position event can recreate a terminal watch and allow settlement events to be emitted again.
- The positive `closed=true` auto-load probe stores `Closed(Vec::new())`. It then terminalizes the condition without caching the returned instruments or adding a resolution watch, leaving an uncached data-only subscriber with neither WebSocket nor polling coverage.
- WebSocket dispatch inserts `pending_resolutions` before task admission. If shutdown has closed admission, the apply future is dropped but the pending entry survives reconnect and blocks later processing.

These cases need focused regression coverage before the PR is ready.

There are also three smaller follow-ups:

- Expired resolution-owned instruments are selected for no-op retirement on every poll tick while their WebSocket owner remains active.
- The docs should include status and close subscriptions in runtime auto-loading and explain that resolution terminalizes normal sibling streams.
- `sync_ws_subscription_async` is now test-only production-module surface with branches that production cannot use.

The smaller items can follow separately if that keeps the lifecycle fixes focused.

mystic-io:
hey @cjdsellers thanks again. I’ve pushed fixes for all the lifecycle cases and the smaller follow-ups. I also added tests, rebased on the latest develop, and reran the checks. I left the broader performance work and WebSocket retry redesign for a separate change. It’s ready for another look when you have time. 🙏🏼

cjdsellers:
Hey @mystic-io,

Thanks for the latest update. I've had another look and found two resolution liveness cases that still need addressing:

- When auto-load sees `closed=true` before a strict winner is available, the condition becomes terminal for WebSocket reconciliation even though status or close subscriptions remain active. With polling disabled, or when expiration is future or missing, the client has no maintained path to receive the later official resolution.
- An admitted resolution task can be forcibly cancelled while waiting for `ws_sub_mutex`. Its `pending_resolutions` entry is only cleared after the future completes, so a normal disconnect and reconnect can retain a stale barrier that suppresses market data and blocks later WebSocket resolution messages.

There are also two smaller follow-ups:

- The data capability summary still omits `InstrumentStatus` and `InstrumentClose`.
- The fallback token-ID extraction is repeated three times and should use one shared parser so the symbol contract cannot drift.

The branch also conflicts with current `develop` in `docs/integrations/polymarket.md`. The resolved text should retain the current `InstrumentCloseType.CONTRACT_EXPIRED` label.

mystic-io:
Hi @cjdsellers , I've applied the requested follow ups: unresolved closed markets now retain their resolution WebSocket subscription, cancelled resolution tasks clean up pending state, the capability docs include InstrumentStatus and InstrumentClose, and token-ID parsing is now shared. I also rebased onto the latest develop and verified the relevant tests and checks.

cjdsellers:
Hi @mystic-io,

Thanks for the update. The lifecycle work is holding up: independent position and data ownership, the pending-resolution barrier with its cancellation cleanup, and exactly-once emission across the WebSocket, poll, and manual paths all check out. Three things remain.

**Auto-load no longer detects closure for unusable payloads:**
Swapping the positive closure probe in `auto_load.rs` from `query_positive_closed_condition_ids` to `request_instruments_by_params_with_transient` ties condition-level closure detection to instrument parsability. A market Gamma reports as `closed=true` whose payload cannot be parsed no longer reaches `closed_condition_ids`, so it never terminalizes. Closure detection needs to stay independent of whether the payload yields a usable instrument.

`auto_load_transient_closed_market_uses_positive_closure_probe` was stating that guarantee, and its replacement asserts the opposite outcome. When a test starts failing under new code, treat it as a contract to satisfy rather than an expectation to update. That one matters more than the fix itself.

**The closed auto-load branch discards the resolution it already has:**
The `AutoLoadOutcome::Closed` arm classifies the condition from a Gamma payload that carries the outcome, then leaves the subscriber with only a poll watch. `collect_resolve_watch_selection` pauses that entry on the first tick when expiration is already stale or zero, so no `InstrumentStatus` or `InstrumentClose` is ever emitted. A subscriber reaching this branch still needs both events.

**Resolution intent pins the WebSocket subscription with no release:**
`wants_resolution` keeps the token subscribed regardless of terminal state, and nothing releases it once the watch entry pauses. Separately, `market_resolved` is gated behind `subscribe_new_markets`, which defaults to `false`, so that retained subscription cannot carry a resolution under the shipped default. A resolution owner needs one maintained path that delivers under the default configuration and is released when the condition terminalizes.

**On the approach:**
We're four rounds in, and each round has fixed the specific cases I named while opening up new issues on the next one. I think the missing step is up front. Before the next push, specify the contract this feature guarantees:

- What establishes resolution monitoring for a data-only subscriber, and what tears it down.
- Which delivery paths are maintained, and under which configuration each one actually delivers.
- What releases each path, and what the release is triggered by.
- Which guarantees the pre-existing position-owned, quote, and auto-load closure paths keep unchanged.

Then work through it:

- Walk every affected path against that list rather than only the sites I pointed at.
- Add one regression test per invariant, so the list is enforced rather than remembered.
- When an existing test fails, resolve it against the contract before changing the assertion.

Thank you for sticking with this through the rounds, and for the regression coverage you've added each time. The feature is a genuinely useful addition and it's close now 👍

mystic-io:
Hey @cjdsellers Thanks again for your time and patience, and sorry this has taken so many review rounds. I've pushed the remaining fixes, added tests for these cases, and checked how the fixes work together. the two pre-existing heartbeat test timeouts remain. I really appreciate your feedback and help on getting this right.

cjdsellers:
@mystic-io, thanks for sticking with it, we're nearly there. I found one remaining reconnect issue. A normal disconnect drains the market pool's shards and assignments but retains `ws_open_tokens`. When the client reconnects, `resume_resolution_subscriptions()` therefore treats cached resolution tokens as already subscribed and sends no asset subscription to the rebuilt pool. With `subscribe_new_markets=true` and `resolve_poll_enabled=false`, this leaves the selected instrument without its asset-scoped resolution subscription.

I reproduced this with a focused test that confirms the first wire subscription, disconnects and reconnects the client, then waits for the subscription to be replayed. The second subscription never arrives.

Could you reset that transport state or force retained resolution subscriptions to replay against the rebuilt pool? It would also be worth adding the cached-instrument disconnect/reconnect case as a regression test.

mystic-io:
hey @cjdsellers fixed the stale WebSocket subscription state, added the cached-instrument reconnect regression test.  appreciate your patience and reviews 🙏🏼

cjdsellers (dismissed):
@mystic-io, thanks again for sticking with this through the review rounds and for adding the reconnect regression. I approve this and will merge it.

I found one remaining edge around closed markets whose CLOB token IDs are still hydrating, along with some resolution-state cleanup. I'll handle both as follow-up work rather than extending this PR.

PR #4897, opened 2026-09-01
Stream backtest data lazily across multiple data configs

### LINKED ISSUE #4832: Streaming Multiple Data streams at once
# Feature Request

## Problem statement

`BacktestNode`'s streaming path (`chunk_size` set on `BacktestRunConfig`) only loads data with
bounded, near-constant memory when exactly one `BacktestDataConfig` is supplied. As soon as a run
needs more than one nautilus data type at once (e.g. quotes + trades + order book deltas for the
same instrument — a very ordinary need for a market-making or order-book-aware strategy), the
memory bound disappears silently: `BacktestNode::run_streaming` falls back to fully materializing
every configured data type into one `Vec<Data>` before running any of it, even though `chunk_size`
is still set and still respected for *delivery* to the engine afterward.

Concretely, in `crates/backtest/src/node.rs`:

```rust
fn run_streaming(engine: &mut BacktestEngine, config: &BacktestRunConfig, chunk_size: usize) -> anyhow::Result<()> {
    let data_configs = config.data();

    if data_configs.len() == 1 {
        // Single config: stream directly from catalog iterator without
        // materializing the full dataset, bounded by chunk_size
        let data_config = &data_configs[0];
        let mut catalog = create_catalog(data_config)?;
        let result = dispatch_query(&mut catalog, data_config, config.start(), config.end())?;
        let data = result.map(|item| item.map_err(anyhow::Error::from));
        stream_chunks(engine, config, data.peekable(), chunk_size)?;
    } else {
        // Multiple configs require loading all data to merge-sort across types
        let all_data = load_and_merge_data(config)?;
        stream_chunks(engine, config, all_data.into_iter().map(Ok).peekable(), chunk_size)?;
    }

    Ok(())
}
```

`load_and_merge_data` calls `load_data` once per configured type, and `load_data` itself is:

```rust
fn load_data(config: &BacktestDataConfig, run_start: Option<UnixNanos>, run_end: Option<UnixNanos>) -> anyhow::Result<Vec<Data>> {
    let mut catalog = create_catalog(config)?;
    let result = dispatch_query(&mut catalog, config, run_start, run_end)?;
    Ok(result.collect::<Result<Vec<_>, _>>()?)   // <-- eagerly drains the otherwise-lazy stream
}
```

The single-type path's `dispatch_query(...)` already returns a genuinely lazy iterator —
`ParquetDataCatalog::query::<T>()` (`crates/persistence/src/backend/catalog.rs`) returns a
`QueryResult`, which is a `KMerge<BatchStream, Data, TsInitComparator>`
(`crates/persistence/src/backend/session.rs`) — a pull-based k-way merge across every
file/`BatchStream` it registers for that one type `T`, in `ts_init` order. That's exactly the
mechanism that gives streaming its bounded memory. The limitation is that this lazy merge only
spans *files of the same type* (because `query<T>` is generic over one concrete
`T: DecodeDataFromRecordBatch + CatalogPathPrefix`, and `dispatch_query` picks exactly one `T` per
call via `match config.data_type() { ... }`). There is no equivalent lazy merge across
*different* types, so `load_and_merge_data` has no choice but to fully collect each type's stream
into memory individually before it can concatenate and globally sort them by `ts_init`.

We hit this concretely building a downstream project (real Binance USD-M tick data, Rust core,
not Python bindings): a strategy that wants quotes + trades + the order book (via
`OrderBookDelta`) together has no way to get that combination with bounded memory today. Measured
on our own catalog: a single-type (`QuoteTick` only) streaming run held RSS in the 250MB-850MB
range across a real 1h-46h window-length ladder, 5 instruments, `chunk_size: 100000`. The same
`chunk_size`, run with 5 data types configured instead of 1 (quotes, trades, order book deltas,
mark prices, funding rate updates; 1 instrument, 1h window), sat around ~2.9GB RSS almost
immediately — consistent with the whole window being materialized up front rather than streamed.
For a longer window or more instruments, this reintroduces exactly the unbounded-RAM/OOM risk that
`chunk_size` streaming exists to avoid in the first place, with no warning beyond reading the
source (nothing in the public API surfaces that `data_configs.len() > 1` silently changes the
memory characteristics).

## Proposed solution

Add a lazy merge across *multiple, independently-typed* `QueryResult` streams, so
`BacktestNode::run_streaming`'s `data_configs.len() > 1` branch can stay on the lazy,
bounded-memory path instead of calling `load_and_merge_data`.

Concretely: `QueryResult` already wraps `KMerge<BatchStream, Data, TsInitComparator>` — a k-way
merge over homogeneous streams, ordered by `ts_init`, yielding `Data`. Since every `QueryResult`
already yields the same `Data` enum regardless of which concrete type `T` produced it, one more
level of the same merge combinator would work: something like
`KMerge<QueryResult, Data, TsInitComparator>` (or a small wrapper type, e.g.
`MultiTypeQueryResult(Vec<QueryResult>)`, implementing `Iterator<Item = Result<Data, QueryError>>`
via the same k-way-merge-by-`ts_init` logic `KMerge` already implements one level down). Then:

- `dispatch_query` stays as-is (one `QueryResult` per `NautilusDataType`).
- A new function builds one `QueryResult` per `BacktestDataConfig` (still lazy, still one
  `KMerge` per type) and merges the resulting `Vec<QueryResult>` into a single lazy
  `Iterator<Item = Result<Data, QueryError>>`.
- `run_streaming`'s `else` branch calls that instead of `load_and_merge_data`, and hands the
  result straight to `stream_chunks` exactly like the single-config branch does today.

This keeps `chunk_size` meaning what it currently means for delivery to the engine, and extends
the *existing* bounded-memory guarantee to cover any number of configured data types instead of
exactly one.

## Example usage

```python
from nautilus_trader.backtest.config import BacktestDataConfig, BacktestEngineConfig, BacktestRunConfig
from nautilus_trader.backtest.node import BacktestNode
from nautilus_trader.model.data import QuoteTick, TradeTick, OrderBookDelta

data_configs = [
    BacktestDataConfig(catalog_path=CATALOG_PATH, data_cls=QuoteTick, instrument_ids=[INSTRUMENT_ID]),
    BacktestDataConfig(catalog_path=CATALOG_PATH, data_cls=TradeTick, instrument_ids=[INSTRUMENT_ID]),
    BacktestDataConfig(catalog_path=CATALOG_PATH, data_cls=OrderBookDelta, instrument_ids=[INSTRUMENT_ID]),
]

run_config = BacktestRunConfig(
    venues=[...],
    data=data_configs,
    chunk_size=100_000,  # currently: silently ignored for the memory bound once len(data) > 1
    engine=BacktestEngineConfig(...),
)

node = BacktestNode(configs=[run_config])
node.run()  # today: fully materializes quotes + trades + deltas before running at all
            # proposed: streams all three, merged lazily by ts_init, bounded by chunk_size
```

## Alternatives considered

- **Pre-merge multiple types into one combined catalog "view" outside of `BacktestNode`.** Would
  need its own new persistence-layer concept and duplicate what `QueryResult`'s merge logic
  already does one level down — more surface area than extending the existing combinator.
- **The proposed lazy multi-type merge** is preferred: it reuses the existing `KMerge`/
  `TsInitComparator` machinery at one more level, and makes `chunk_size` mean the same thing
  regardless of how many data types are configured.

## Additional context

- `crates/backtest/src/node.rs`: `run_streaming` (data_configs.len() == 1 check), `load_and_merge_data`, `load_data`, `dispatch_query`.
- `crates/persistence/src/backend/session.rs`: `QueryResult`, wrapping `KMerge<BatchStream, Data, TsInitComparator>`.
- `crates/persistence/src/backend/catalog.rs`: `ParquetDataCatalog::query<T>`, generic over one `T: DecodeDataFromRecordBatch + CatalogPathPrefix`.
- Separately, `NautilusDataType`/`BacktestDataConfig` have no representation for custom/non-native
  data types (e.g. adapter-specific types registered only via `CustomDataTrait`), so those can
  only ever be backtested via a fully materialized `Vec<Data>` regardless of this issue. That's a
  distinct, larger gap (no catalog/streaming path for custom data at all) and probably belongs in
  its own issue rather than folded into this one.

abhijeetvichare76:
## Summary

`run_streaming` only streamed lazily when exactly one `BacktestDataConfig` was present. With two or
more it fell through to `load_and_merge_data`, which collected every record into a single
`Vec<Data>` and sorted by `ts_init` — so `chunk_size` still appeared to apply while the memory bound
it promises was silently gone.

This builds one `QueryResult` per data config and k-way merges them by `ts_init`, feeding
`stream_chunks` exactly as the single-config branch already did. The single-config case becomes k=1
of that same path, so the two can no longer drift, and `load_and_merge_data` is deleted. `load_data`
and `run_oneshot` keep loading eagerly — oneshot needs the whole `Vec<Data>` for
`add_data`/`sort_data`, and `load_data_config` is public API.

The merge holds one peeked item per config and breaks `ts_init` ties by config order, matching the
stable `sort_by_key` the eager path applied. A peeked `Err` is returned immediately so a failed
stream cannot read as exhaustion, the same contract as `QueryResult::next`.

One behavior change: the single-config path now emits the existing `"No data found for config"`
warning it previously skipped. Log-only.

Closes #4832

## On not reusing KMerge

`KMerge<I, T, C>` requires `I: Iterator<Item = std::vec::IntoIter<T>>` — a stream of batches — and
yields a bare `T`, not a `Result`. Reusing it would mean wrapping each item in a one-element `Vec`,
defeating `ElementBatchIter`, plus a `Compare` impl and a second `ErrorSlot` inside
`nautilus-persistence` to satisfy the orphan rule. That is a public API change to another crate to
serve a merge over k = data configs, typically 2-5, so this stays in `node.rs` and leaves
`nautilus-persistence` untouched. Happy to switch if you'd rather it used the shared machinery.

## Testing

Unit, in `node.rs`: ordering across streams including a three-way `ts_init` tie; a merge over two
*unbounded* streams taking four items, which would not terminate if the merge materialized its
input; and a failure surfacing as the next item rather than as exhaustion.

Integration: `test_run_streaming_multiple_data_configs` runs a real catalog with quote and trade
configs at `chunk_size=3` over 20 events and asserts the exact interleaved sequence the strategy
receives, proving ordering holds across chunk boundaries, not just within one.

- `cargo test -p nautilus-backtest --lib --features streaming` — 155 passed
- `cargo test -p nautilus-backtest --test backtest_node --features streaming` — 48 passed
- `cargo clippy --no-deps -p nautilus-backtest --lib --tests --features "python,high-precision,streaming" -- -D warnings` — clean
- `cargo +nightly fmt` — no changes

`arrow`/`ffi` are not features of this crate. `--no-deps` avoids pre-existing `doc_markdown` errors
in `crates/persistence/src/parquet.rs:656`, present on an untouched tree. `canonical_backtest_workloads`
fails locally on Windows but fails identically on clean `develop` with a byte-identical
`result_digest`, so this change is result-neutral there.

### DIFF
--- a/crates/backtest/src/node.rs
+++ b/crates/backtest/src/node.rs
@@ -421,11 +421,14 @@ fn run_streaming(
             .map(|item| item.map_err(anyhow::Error::from))
             .peekable();
 
-        if stream.peek().is_none() {
-            log::warn!("No data found for config: {:?}", data_config.data_type());
-            continue;
+        match stream.peek() {
+            Some(Ok(_)) => streams.push(stream),
+            // Surface a failed query in config order, before opening later ones
+            Some(Err(_)) => {
+                stream.next().transpose()?;
+            }
+            None => log::warn!("No data found for config: {:?}", data_config.data_type()),
         }
-        streams.push(stream);
     }
 
     stream_chunks(
--- a/crates/backtest/src/node.rs
+++ b/crates/backtest/src/node.rs
@@ -407,26 +407,60 @@ fn run_streaming(
 ) -> anyhow::Result<()> {
     let data_configs = config.data();
 
-    if data_configs.len() == 1 {
-        // Single config: stream directly from catalog iterator without
-        // materializing the full dataset, bounded by chunk_size
-        let data_config = &data_configs[0];
-        let mut catalog = create_catalog(data_config)?;
-        let result = dispatch_query(&mut catalog, data_config, config.start(), config.end())?;
-        let data = result.map(|item| item.map_err(anyhow::Error::from));
-        stream_chunks(engine, config, data.peekable(), chunk_size)?;
-    } else {
-        // Multiple configs require loading all data to merge-sort across types
-        let all_data = load_and_merge_data(config)?;
-        stream_chunks(
-            engine,
-            config,
-            all_data.into_iter().map(Ok).peekable(),
-            chunk_size,
-        )?;
+    // Stream directly from the catalog iterators without materializing the full
+    // dataset, so memory stays bounded by chunk_size for any number of configs
+    let mut catalogs = data_configs
+        .iter()
+        .map(create_catalog)
+        .collect::<anyhow::Result<Vec<_>>>()?;
+    let mut streams = Vec::with_capacity(catalogs.len());
+
+    for (catalog, data_config) in catalogs.iter_mut().zip(data_configs) {
+        let result = dispatch_query(catalog, data_config, config.start(), config.end())?;
+        let mut stream = result
+            .map(|item| item.map_err(anyhow::Error::from))
+            .peekable();
+
+        if stream.peek().is_none() {
+            log::warn!("No data found for config: {:?}", data_config.data_type());
+            continue;
+        }
+        streams.push(stream);
     }
 
-    Ok(())
+    stream_chunks(
+        engine,
+        config,
+        merge_streams(streams).peekable(),
+        chunk_size,
+    )
+}
+
+// Merges the data streams of every config in ascending `ts_init` order, taking one
+// item at a time so the merge holds only a single item per config. Ties keep config
+// order, matching the stable sort the eager path applies.
+fn merge_streams<I: Iterator<Item = anyhow::Result<Data>>>(
+    mut streams: Vec<Peekable<I>>,
+) -> impl Iterator<Item = anyhow::Result<Data>> {
+    std::iter::from_fn(move || {
+        let mut next: Option<(usize, UnixNanos)> = None;
+
+        for (i, stream) in streams.iter_mut().enumerate() {
+            match stream.peek() {
+                Some(Ok(data)) => {
+                    let ts_init = data.ts_init();
+                    if next.is_none_or(|(_, ts)| ts_init < ts) {
+                        next = Some((i, ts_init));
+                    }
+                }
+                // A failed stream ends the merge rather than reading as exhaustion
+                Some(Err(_)) => return stream.next(),
+                None => {}
+            }
+        }
+
+        streams[next?.0].next()
+    })
 }
 
 // Feeds data from an iterator to the engine in timestamp-aligned chunks.
@@ -501,21 +535,6 @@ fn take_aligned_chunk<I: Iterator<Item = anyhow::Result<Data>>>(
     }
 
     Ok(chunk)
-}
-
-fn load_and_merge_data(config: &BacktestRunConfig) -> anyhow::Result<Vec<Data>> {
-    let mut all_data = Vec::new();
-
-    for data_config in config.data() {
-        let data = load_data(data_config, config.start(), config.end())?;
-        if data.is_empty() {
-            log::warn!("No data found for config: {:?}", data_config.data_type());
-            continue;
-        }
-        all_data.extend(data);
-    }
-    all_data.sort_by_key(HasTsInit::ts_init);
-    Ok(all_data)
 }
 
 fn create_catalog(config: &BacktestDataConfig) -> anyhow::Result<ParquetDataCatalog> {
@@ -614,7 +633,8 @@ mod tests {
     #[cfg(feature = "python")]
     use nautilus_model::enums::{AccountType, OmsType};
     use nautilus_model::{
-        identifiers::InstrumentId,
+        enums::AggressorSide,
+        identifiers::{InstrumentId, TradeId},
         types::{Price, Quantity},
     };
     #[cfg(feature = "python")]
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @abhijeetvichare76,

Thanks for the PR. I've confirmed that the merge applies to all data types currently exposed through `BacktestDataConfig`; custom data remains outside this path, as noted in #4832.

There's also a new catalog implementation getting close to landing which incorporates this work as part of a broader design. This PR is reasonably focused and addresses the current memory-efficiency issue, so we'll take it here rather than wait for the larger catalog change.

There's one error-ordering edge to address first in `run_streaming`: when `stream.peek()` returns `Some(Err(_))` for an earlier config, the code treats it as non-empty and continues opening later queries. If a later `dispatch_query` then fails, that later error masks the stream error already observed.

Could the peeked error be consumed and returned before opening the next query? That preserves config-order error reporting and the immediate-propagation behavior described in the PR.

abhijeetvichare76:
Thanks @cjdsellers , good catch. Fixed: the peek in `run_streaming` now matches on `Some(Err(_))` and consumes and returns that error before the next `dispatch_query runs`, so the first failing config in order is the one reported. `Some(Ok(_))` pushes the stream and `None` keeps the existing warning.

Could you share what this catalog is that you referred to? That context will help me pick up the next issue so that there's less duplicate work.

cjdsellers (approved):
Hi @abhijeetvichare76,

Thanks for turning that around so quickly. The peek now reports the first failing config in order, which is what I was after, and this looks good to merge.

Worth flagging since it landed after your base: `develop` now renames the `Data` book variants to `BookDelta` and `BookDeltas`. Nothing here touches them, and the merged tree passes the backtest tests and `pre-commit`, so there's no rebase needed on your side.

On the catalog you asked about: it will replace the current Parquet data catalog, and we'll share more details soon.

PR #4898, opened 2026-09-01
Publish Binance Spot dispatch readiness at task admission

folknor:
A follow-up to the task-lifecycle standardization in 4c18691277.

## The dispatch-active flag is published from inside the task it describes

`BinanceSpotExecutionClient::submit_order` and `submit_order_list` reject a command with "Binance Spot user data stream is not active" unless `ws_user_data_active()` holds, and that predicate ands the client's user-data state with a `dispatch_running` flag. Both dispatch flags were stored by the first statement *inside* the spawned dispatch task.

`TaskGroup::spawn` returns once the task is admitted; the body runs when the runtime first polls it. There is no happens-before edge from a successful spawn to that store becoming visible, so a caller can observe `false` after admission has already succeeded.

On the Binance US path that window survives `connect()`. The dispatch task is spawned, `mark_user_data_active()` then runs synchronously, and `connect()` returns without awaiting anything the dispatch task produces - so an order submitted immediately after a successful connect can be rejected for a stream that is connected and subscribed. The non-US path is protected by its own handshake today, because `connect()` waits for the `UserDataSubscribed` notification that the dispatch task itself emits.

Before 4c18691277 the same conjunct read `ws_user_data_handle.as_ref().is_some_and(|handle| !handle.is_finished())`, which was true the moment `spawn` returned. Replacing the handle with an atomic moved the observable transition from admission time to first-poll time.

`ws_order_transport_active()` reads the same predicate, so during the window the submit, cancel, modify, and cancel-all paths silently select HTTP rather than the WS order transport.

## The fix

Both sites now go through `spawn_dispatch_task`, which stores the flag, constructs `DispatchActiveGuard`, and moves the guard into the spawned future. Readiness is therefore published before `spawn` returns, restoring the pre-refactor ordering, and ownership clears the flag on every path: `TaskGroup::spawn` drops a rejected future, the group's wrapper can drop the future without polling it when cancellation wins, and a normal exit or panic drops it as before. Publishing before admission is what makes that last part load-bearing - a flag set before the task exists has to be cleared by a task that may never run.

## Testing

`test_spawn_dispatch_task_clears_active_when_admission_closed` spawns into a closed group and asserts the flag is clear on return, covering the obligation the new ordering introduces.

There is no deterministic regression test for the window itself. `TaskGroup` spawns onto the process-global multi-threaded runtime, so nothing a test can do from outside will hold a successfully admitted future unpolled: saturating workers is scheduler pressure rather than synchronization, and `NAUTILUS_WORKER_THREADS` is read once per process behind a `OnceLock`. A test that merely tends to fail against the old code would be exactly the scheduler-sensitive flake this area has been removing. The evidence the window is real is a full-workspace test run in which `test_binance_us_submit_order_uses_http_transport` failed at its `submit_order` call with this error under parallel load, and passed in 1.84s run alone; it synchronizes on the socket-state event, which says nothing about the dispatch task having been polled.

### DIFF
--- a/crates/adapters/binance/src/spot/execution.rs
+++ b/crates/adapters/binance/src/spot/execution.rs
@@ -47,7 +47,7 @@ use nautilus_core::{
 };
 use nautilus_live::{
     ExecutionClientCore, ExecutionEventEmitter, SocketControlFactory,
-    task::{TaskGroup, TaskGroupGuard, TaskSpawner},
+    task::{TaskGroup, TaskGroupGuard, TaskSpawnError, TaskSpawner},
 };
 use nautilus_model::{
     accounts::AccountAny,
@@ -656,10 +656,7 @@ impl BinanceSpotExecutionClient {
             .spawner()
             .context("Binance Spot session task admission is closed")?;
 
-        if let Err(e) = self.session_tasks.spawn(async move {
-            dispatch_active.store(true, Ordering::Release);
-            let _active = DispatchActiveGuard(dispatch_active);
-
+        if let Err(e) = spawn_dispatch_task(&self.session_tasks, &dispatch_active, async move {
             while let Some(message) = ws_clone.recv().await {
                 if matches!(&message, BinanceSpotWsTradingMessage::Reconnected) {
                     ws_clone.mark_user_data_active();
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR and for pinning down this race. The race is real, but while reviewing it I found the underlying gap in the common `TaskGroup` machinery: the lifecycle standardization had lost per-task identity.

I've added named task references to the shared live machinery in eb42e2bfc6 and migrated Binance Spot and Kraken to use them. This lets adapters observe task lifecycle through the common machinery instead of maintaining local active flags, so I'll close this in favor of the central implementation rather than merge the adapter-specific fix and centralize it afterward. Your diagnosis directly helped expose the gap. Thanks again.

PR #4899, opened 2026-09-01
Stop two task lifecycle tests spinning on a stalled timer

folknor:
A follow-up to the task-lifecycle standardization in 4c18691277.

## Two task tests spin against a timer that cannot advance

`canceled_singular_abort_wait_preserves_abort_classification` and `canceled_finish_preserves_forced_abort_classification` each loop over a `finish_*` future with a zero-duration graceful timeout, racing it against `task::yield_now()` in a `select!`.

The zero-duration `time::timeout` inside the raced future is a timer, and the `yield_now` arm keeps the task runnable, so the remaining worker can be delayed in servicing the timer driver. Neither loop can make progress until it does, because the exit conditions - `slot.abort_requested` and `generation.force.is_cancelled()` - are both set only after that timeout fires. Both tests run on a two-worker runtime whose other worker is parked in a deliberately blocking `recv()`, which leaves nothing else to drive the timer.

Measured on an idle machine, with an iteration counter added to the loops:

- `canceled_singular_abort_wait_preserves_abort_classification`: 15,917,237 iterations, 6.26s in the loop, 6.47s wall. Repeat runs of the same binary took 0.89s, 4.32s and 6.47s.
- `canceled_finish_preserves_forced_abort_classification`: 31,673 iterations, 0.12s wall.

The second spins too. Wall time alone hides it, which is why the iteration count is the measurement that matters here.

## The fix

Both `yield_now` arms become `time::sleep(POLL_INTERVAL)`, a shared 10ms constant. Sleeping makes the task non-runnable, so the runtime parks, services the timer, and the raced timeout fires on the next poll. The loops, the assertions and the blocking task are unchanged.

After: the first test runs 1 iteration, 11.09ms in the loop, 0.28s wall; the second stays at 0.12s.

## Testing

No test was added; these are the tests. Inverting the `abort_requested` condition in `TaskSlot::complete` fails `canceled_singular_abort_wait_preserves_abort_classification` on its `TaskJoinOutcome::Aborted` assertion, so the faster loop still pins the classification it exists to pin.

Both were reached through `cargo test -p nautilus-live`, and the crate is green.

### DIFF
--- a/crates/live/src/task.rs
+++ b/crates/live/src/task.rs
@@ -1330,6 +1330,15 @@ mod tests {
 
     const TEST_TIMEOUT: Duration = Duration::from_secs(1);
 
+    /// Interval a test waits on while letting an in-flight `finish_*` future make progress.
+    ///
+    /// Sleeping rather than yielding is deliberate. A `yield_now` arm keeps the remaining
+    /// worker runnable and can delay it servicing the timer driver, so a zero-duration
+    /// `time::timeout` inside the raced future may not fire promptly and the loop spins
+    /// waiting for it. Sleeping lets the runtime park and service the timer before the
+    /// next poll.
+    const POLL_INTERVAL: Duration = Duration::from_millis(10);
+
     #[rstest]
     fn task_group_guard_closes_all_groups_and_runs_rollback_until_disarmed() {
         let first = Arc::new(TaskGroup::new());
@@ -2114,7 +2123,7 @@ mod tests {
                 tokio::select! {
                     biased;
                     outcome = &mut finish => panic!("finish completed unexpectedly: {outcome:?}"),
-                    () = task::yield_now() => {}
+                    () = time::sleep(POLL_INTERVAL) => {}
                 }
             }
 
@@ -3011,7 +3020,7 @@ mod tests {
                 tokio::select! {
                     biased;
                     outcome = &mut finish => panic!("finish completed unexpectedly: {outcome:?}"),
-                    () = task::yield_now() => {}
+                    () = time::sleep(POLL_INTERVAL) => {}
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR. There's one simulation-build issue to fix: `POLL_INTERVAL` is unconditional, but both uses are excluded under `cfg(all(feature = "simulation", madsim))`. The DST simulation Clippy gate therefore fails under `-D warnings` with `constant POLL_INTERVAL is never used`. Could you give the constant the same `#[cfg(not(all(feature = "simulation", madsim)))]` as those tests?

The new rationale comment also ties both cases to a single remaining test-runtime worker. In `canceled_finish_preserves_forced_abort_classification`, `TaskGroup::spawn` runs the blocking task on the separate global runtime. Could you keep the timer-driver explanation but remove the worker-count-specific claim?

folknor:
Thank you.

Both points are addressed.

## The simulation build

`POLL_INTERVAL` now carries the same `#[cfg(not(all(feature = "simulation", madsim)))]` as its two callers, so the constant and its uses appear and disappear together and the simulation build no longer compiles it with no reader.

## The rationale comment

You are right that the worker-count claim does not hold. `TaskGroup::spawn` reaches `spawn` in `crates/live/src/task.rs`, whose non-simulation arm is `nautilus_common::live::get_runtime().spawn(future)`, so the blocking task runs on the process-global runtime rather than on the test's own. The comment now rests only on the timer driver: a `yield_now` arm leaves the polling task immediately runnable, so the loop can re-poll before the runtime services its timers, and a zero-duration `time::timeout` inside the raced future may then not fire promptly.

cjdsellers (approved):
@folknor, thanks for addressing both points. I've had another look at the latest changes, and this looks good from my side 👌

PR #4902, opened 2026-09-01
Add configurable REST retry count to Kraken execution config

folknor:
`KrakenExecutionClientConfig` exposes every other transport control the raw clients accept - timeout, proxy, rate limit - but not the retry count, which ten other adapter configs carry as `max_retries`.

## The retry count is unreachable from the execution config

`KrakenSpotExecutionClient::new` and `KrakenFuturesExecutionClient::new` construct their raw HTTP clients with `config.timeout_secs`, `config.proxy_url` and `config.max_requests_per_second`, then pass literal `None` for the `max_retries`, `retry_delay_ms` and `retry_delay_max_ms` parameters. The raw constructors resolve those to 3 retries with a 1s-to-10s backoff, so an operator cannot change the retry count without bypassing the adapter and building `KrakenSpotHttpClient` directly - which the Python bindings already allow, since `KrakenSpotHttpClient` and `KrakenFuturesHttpClient` both expose all three parameters.

The consequence reaches order commands through spot cancellation. `cancel_order` and `cancel_all_orders` go through `send_request`, so one cancel command can produce up to four requests against the venue's rate budget, and an operator accounting for order commands outside the adapter cannot observe the extra attempts. `add_order` and `amend_order` use `send_request_once`, and every Futures order command goes through `send_order_request`, so submissions are single-shot.

## The change

One field, `max_retries: u32`, defaulting to 3, forwarded as `Some(config.max_retries)` from both execution constructors. The two delay parameters keep their `None`, and `KrakenDataClientConfig` is untouched. A configuration that omits the field behaves exactly as before; `max_retries = 0` yields a single attempt, since `RetryManager` computes `max_attempts` as `max_retries + 1`.

`max_retries` is client-wide, so setting it to zero also disables retries for the reconciliation and account reads that share the client. A per-operation retry policy would be a larger design question than this opens.

This exposes the retry count alone, where the ten adapters carrying the same control also expose the initial and maximum backoff. I am happy to widen it to the delays, or to add the same field to the data config, if you would rather Kraken match them.

## Testing

`test_execution_config_max_retries_zero_disables_spot_cancel_retries` starts a mock server that answers `/0/private/CancelOrder` with HTTP 429, constructs a `KrakenSpotExecutionClient` from a config carrying `max_retries: 0`, awaits the cancellation to its terminal error, and asserts the server received exactly one request. Awaiting the operation rather than sampling the counter is what makes it meaningful: the retry ladder waits about a second before its second attempt, so an immediate assertion would pass against the unfixed code too. With the constructor restored to `None` the server receives four requests and the test fails.

Three config tests cover the default of 3, an explicit 0 through a serde round trip, and a payload omitting the field resolving to 3.


cjdsellers:
Hey @folknor,

Thanks for the PR. The placement of `max_retries` alongside the other transport controls is the right API shape, and the wiring and default behavior look sound.

If you're able, could you also add `max_retries` and its default to the execution config table in `docs/integrations/kraken.md` as part of this PR? It would keep the public option discoverable. Otherwise, this looks good to merge.

folknor:
Thank you.

`max_retries` and its `3` default are now in the execution client configuration table in `docs/integrations/kraken.md`, placed after `max_requests_per_second` so the table follows the same ordering as the config struct.

cjdsellers (approved):

Issue #4702, opened 2026-08-09
[RFC] Typed fill origin for venue-generated executions


folknor:
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I have no other RFC open.

## Context

Venues distinguish executions they generate themselves - liquidation, ADL, settlement, delivery - from ordinary fills, and each adapter encodes that differently.

Binance futures reads it off the `client_order_id` prefix: `is_liquidation` on `autoclose-`, `is_adl` on `adl_autoclose`, `is_settlement` on `settlement_autoclose-` or `delivery_autoclose-`. Bybit reads `BybitExecType`, with `AdlTrade`, `BustTrade`, `Delivery`, `Settle` and `CorporateAction` behind `is_exchange_generated`. Hyperliquid parses `liquidation.method`, `mark_px` and `liquidated_user` off the user-fills stream and emits them through `log::warn!`; the `FillReport` it builds carries none of it.

A consumer asking "did the venue close this position, and why" reimplements two conventions and cannot see the third at all.

Proposal: a typed fill-origin field on `OrderFilled` and `FillReport` - normal, liquidation, ADL, settlement, delivery - populated by each adapter from what it already parses.

## Considerations

The field is additive with a normal default, so existing consumers are unaffected. Each adapter's detection stays where it is; only its destination changes. Both types are `pyclass`, so this regenerates stubs.

Hyperliquid carries more than an enum holds - mark price and the liquidated account. That either stays in the log or wants a separate carrier; I'd keep this change to the discriminator.

## Alternatives

Leave the per-adapter conventions and let consumers reimplement them. That is the status quo, and it never reaches Hyperliquid's data, which is parsed and dropped.

Adapter-specific custom data events. Workable per venue, but gives a portfolio-level consumer no common shape.

Happy to implement the Rust side and the adapter population if you want this.

PR #4903, opened 2026-09-01
Standardize cancellation reason propagation to OrderCanceled

folknor:
`OrderStatusReport` carries `cancel_reason` and several adapters populate it, but `OrderCanceled` has no field to receive it, so the venue's explanation is discarded at the point the event is constructed.

## The reason is dropped one frame above the consumer

`OrderEvent::reason()` returns `Option<Ustr>`, and `OrderCanceled` answers it with a hardcoded `None`. The information exists: `create_reconciliation_fill_voids` already passes `report.cancel_reason.as_deref().map(Ustr::from)` into `OrderFillVoided::new`, so a report's cancellation reason already reaches an event's reason field one function away. `create_external_terminal_event` and `create_reconciliation_canceled` hold the same `OrderStatusReport` and cannot do it, because the event they build has nowhere to put it.

The same split exists inside adapters. `LighterOrderStatus::as_cancel_reason` maps eleven canceled statuses to `post-only`, `self-trade`, `liquidation`, `invalid-balance` and seven more, and its doc comment names `OrderStatusReport::with_cancel_reason` as the destination; the status-report path preserves the value while the direct WebSocket dispatch path writes `None`, so one venue status produces two different payloads depending on routing. `open_orders_cancel` in the Kraken Futures dispatch already uses `cancel.reason` for its external-report fallback and drops it for tracked orders. BitMEX computes a local `cancel_reason`, uses it for the post-only rejection branch, and discards it in the branch beside it.

## The change

`reason: Option<Ustr>` on `OrderCanceled`, matching `OrderFillVoided`, with `#[serde(default, skip_serializing_if = "Option::is_none")]` so existing payloads deserialize unchanged and nothing new is serialized for callers supplying nothing. `Ustr` is `Copy`, so the event remains `Copy`. The field is the final argument to `OrderCanceled::new` and to the Python constructor, where it defaults to `None`.

Both reconciliation constructors now forward `report.cancel_reason`. Six adapter sites that already hold a reason forward it too: Lighter, BitMEX, Betfair, Deribit, and the Kraken Futures delta and cancel paths, plus Kraken Spot. Every other construction site passes `None` and is unchanged.

Bybit and OKX keep `None` despite having a reason in scope: Bybit's `reject_reason` needs empty and `UNKNOWN` filtering, and OKX's belongs behind the normalization its report parser already performs. Hyperliquid's canceled reports carry no reason today.

This is limited to cancellation. `OrderExpired` has the same hardcoded `None`, but no adapter supplies an expiry reason with defined semantics, and some venues encode expiry as a cancellation subtype; that can follow when there is something to carry.

This is a source-breaking Rust change: `OrderCanceled::new` takes a new final argument and the struct has public fields, so downstream exhaustive construction changes with it. The Python constructor is backward compatible.

Related to #4702 in that both concern venue-originated context discarded in the adapter layer, but this proposes no taxonomy - it fills a channel `OrderEvent` and `OrderStatusReport` already define. I am happy to fold it into a broader provenance direction if you would rather take it that way.

## Testing

`test_external_canceled_order_preserves_cancel_reason` and `test_reconcile_order_report_generates_canceled` cover the two reconciliation constructors, each asserting the reason reaches `reason()`. Reverting either wiring line in isolation fails its own test with `left: None, right: Some("not-enough-liquidity")`, so each path is proven separately rather than jointly.

`parse_lighter_order_event_emits_canceled_for_other_cancel_variants` is parameterized over four statuses and now asserts the expected reason for each, including `None` for the unqualified `Canceled` variant that carries its meaning in the order status itself.

`test_serialization_roundtrip_with_reason` covers a populated reason; `test_deserialization_defaults_missing_reason_to_none` asserts the key is absent from a default event's serialized form and that such a payload still deserializes. The Python tests cover the constructor, the getter, and dictionary round-trips for both a populated reason and the default.


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. I'm happy with this approach. Keeping `reason` optional matches `OrderFillVoided` and lets adapters preserve venue explanations where available. This can stand on its own without waiting for a broader provenance design.

PR #4906, opened 2026-09-02
Document installing the execution event sender at create

folknor:
A follow-up to the emitter sender sharing in #4874.

## The guide ties event routing to the lifecycle thread

`docs/developer_guide/adapters.md` tells an execution client using `ExecutionEventEmitter` to install its sender during `start`. The sender comes from `get_exec_event_sender`, which reads a `thread_local!` slot in `nautilus_common::live::runner`, so the instruction binds correct event routing to whichever thread runs `start`; a host arrangement where that is not the runner's thread leaves the emitter without a sender, and `send_order_event` warns and drops.

`LiveNodeBuilder::build` calls `bind_senders` on the runner before invoking any registered `ExecutionClientFactory::create`, on the same thread, so a factory can resolve the sender with `try_get_exec_event_sender` and install it before returning the client. After #4874 the emitter's clones share that slot, so every clone taken during construction carries the sender. The paragraph now says this, keeps `start`-time installation as sufficient when `LiveNode` drives the lifecycle, and states what a host calling a factory outside the builder must arrange before `create`.

## The sender accessors are not global

The `get_*`, `try_get_*`, and `set_*` doc comments on the four sender families in `crates/common/src/live/runner.rs` describe a "global" sender. The backing store is `thread_local!`, and the module's own doc says so. The comments now say thread-local.

No code changes.

### DIFF
--- a/crates/common/src/live/runner.rs
+++ b/crates/common/src/live/runner.rs
@@ -21,7 +21,7 @@ use std::cell::RefCell;
 
 use crate::messages::{DataEvent, ExecutionEvent, SystemCommand, SystemEvent};
 
-/// Gets the global data event sender.
+/// Gets the thread-local data event sender.
 ///
 /// # Panics
 ///
@@ -37,7 +37,7 @@ pub fn get_data_event_sender() -> tokio::sync::mpsc::UnboundedSender<DataEvent> 
     })
 }
 
-/// Attempts to get the global data event sender without panicking.
+/// Attempts to get the thread-local data event sender without panicking.
 ///
 /// Returns `None` if the sender is not initialized (e.g., in Python/v1 bridge environments
 /// before a runner or adapter bridge has registered a sender).
@@ -46,7 +46,7 @@ pub fn try_get_data_event_sender() -> Option<tokio::sync::mpsc::UnboundedSender<
     DATA_EVENT_SENDER.with(|sender| sender.borrow().as_ref().cloned())
 }
 
-/// Sets the global data event sender.
+/// Sets the thread-local data event sender.
 ///
 /// Can only be called once per thread.
 ///
@@ -61,14 +61,14 @@ pub fn set_data_event_sender(sender: tokio::sync::mpsc::UnboundedSender<DataEven
     });
 }
 
-/// Replaces the global data event sender for the current thread.
+/// Replaces the data event sender for the current thread.
 pub fn replace_data_event_sender(sender: tokio::sync::mpsc::UnboundedSender<DataEvent>) {
     DATA_EVENT_SENDER.with(|s| {
         *s.borrow_mut() = Some(sender);
     });
 }
 
-/// Gets the global system event sender.
+/// Gets the thread-local system event sender.
 ///
 /// # Panics
 ///
@@ -84,7 +84,7 @@ pub fn get_system_event_sender() -> tokio::sync::mpsc::UnboundedSender<SystemEve
     })
 }
 
-/// Attempts to get the global system event sender without panicking.
+/// Attempts to get the thread-local system event sender without panicking.
 ///
 /// Returns `None` if the sender is not initialized (e.g., in test environments).
 #[must_use]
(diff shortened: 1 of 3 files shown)


folknor:
@cjdsellers This one is very simple :joy:

cjdsellers:
Hi @folknor,

Thanks for the PR. The new adapter-guide guidance is correct, but the public `ExecutionEventEmitter` docs still say to install the sender in `start()`: the module diagram, type docs, `new`, and `set_sender` all retain the old instruction. That leaves two conflicting usage contracts, and someone following the rustdoc can still hit the cross-thread sender lookup this PR is meant to prevent, leaving the emitter uninitialized and dropping events.

Can you update those rustdoc references to match the factory-`create` guidance while keeping the `LiveNode` start-time exception clear when you get a chance?

folknor:
Thank you.

All four sites now give the same contract as the guide: the module diagram, the type docs, `new`, and `set_sender`.

`set_sender` carries the reasoning, since it is the one a reader reaches from either direction: `LiveNodeBuilder` binds the runner's senders before it calls any factory, so `try_get_exec_event_sender` resolves inside `create` and every clone taken during construction carries the result. It also keeps the `LiveNode` exception explicit - installing only in the client's `start` is sufficient when `LiveNode` drives the lifecycle, and leaves the emitter uninitialized for a client constructed where those senders are not bound.

cjdsellers:
@folknor, thanks for the follow-up. The remaining issue starts with the last sentence of `set_sender`: constructing a client without bound TLS does not mean the emitter stays uninitialized. A later `set_sender` in `start` updates the shared slot and all existing clones.

Can you revisit the lifecycle guidance from there, including what `try_get_exec_event_sender` returning `None` during `create` should mean? The client still needs a guaranteed sender-installation path before anything can emit.

folknor:
Thank you.

The slot is shared through the derived `Clone`, so a later `set_sender` reaches every clone; that sentence is gone.

`start` is the authoritative install, and the guidance says so now: `LiveNode` rebinds the runner's senders on the calling thread before it starts clients, so a sender resolved earlier can be a superseded runner's, and the `start` install replaces it and reaches the construction-time clones through the shared slot. The four rustdoc sites and the guide state one contract - install in `start` unconditionally, and in `create` when `try_get_exec_event_sender` returns `Some`.

`None` in `create` is then an expected state rather than an error: the calling thread has no bound senders, which is what a factory unit test or a host that binds later looks like.

For a host outside `LiveNode` the guide keeps both options, binding the runner's senders on the client's thread or passing the sender through its own constructor.

cjdsellers:
@folknor, thanks for the follow-up. There is still a conflict in the non-`LiveNode` guidance. It requires `get_exec_event_sender` unconditionally in `start`, then offers passing the sender through the host's own factory or constructor as an alternative to binding the runner's senders to that thread.

Injection goes through `set_sender`, which updates the emitter's shared slot, not the thread-local slot. `get_exec_event_sender` reads only the thread-local slot and panics via `expect` when nothing has bound it, so an injection-only host panics in `start` before the injected sender is used. Every current adapter `start` already does that unconditional lookup, so the injection option cannot work with any existing client.

Can you make that external-host path consistent when you get a chance? It needs either a thread-local binding before the unconditional `start` lookup, or a `start` contract that accepts the already-injected sender.

folknor:
Thank you.

The external-host guidance now requires the thread-local binding: bind the runner's senders on the client's thread before `start`, because the `start` install resolves through `get_exec_event_sender`, which reads only the thread-local slot and panics when nothing has bound it. Constructor injection reaches the emitter's shared slot but not that lookup, so it is no longer offered as a freestanding alternative: it stands alone only for a client whose `start` accepts an already-installed sender, and the emitter-backed execution clients in this repository all perform the unconditional lookup.

cjdsellers (approved):
Hey @folknor,

Thanks for working through the follow-ups. I've had another look, and the lifecycle guidance is consistent now. This looks good to merge 👍

PR #4907, opened 2026-09-02
Expose fallible account state dispatch on the emitter

folknor:
A follow-up to the emitter sender sharing in #4874.

## One fallible send is private

`ExecutionEventEmitter` has three event kinds, each with an infallible `send_*` that warns on failure and a fallible `try_send_*` that takes a caller-built value. `try_send_order_event` and `try_send_execution_report` are public; `try_send_account_state` is private. The public fallible path for account state, `try_emit_account_state`, builds the event itself and stamps `ts_init` from the emitter's `AtomicTime`, so an adapter that stamps account state from its own clock can have fallible dispatch or its own timestamps, not both.

The private helper predates its siblings: it was extracted for `try_emit_account_state` when the order and report sends were still inline blocks with no fallible form, and the public `try_send_order_event` and `try_send_execution_report` arrived later without revisiting it. `send_account_state` is already public and accepts a caller-built state with its own `ts_init`, so making the fallible form public changes failure reporting only.

`try_send_account_state` is now `pub`, documented like its siblings.

## Testing

`test_try_send_account_state_preserves_caller_built_state` in a new `crates/live/tests/emitter.rs` calls the method through `nautilus_live::ExecutionEventEmitter`, so it pins the visibility at compile time: with the method private it fails with E0624. It asserts the uninitialized-sender error on a caller-built `AccountState`, then installs a channel and asserts every field of the received event against the sent state, `ts_init` included; `AccountState::PartialEq` compares identity fields only, so whole-state equality would not have shown the state was delivered unchanged.

### DIFF
--- a/crates/live/src/execution/emitter.rs
+++ b/crates/live/src/execution/emitter.rs
@@ -477,7 +477,12 @@ impl ExecutionEventEmitter {
         }
     }
 
-    fn try_send_account_state(&self, state: AccountState) -> anyhow::Result<()> {
+    /// Emits an account state event and returns any channel error to the caller.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the sender is uninitialized or its receiver is closed.
+    pub fn try_send_account_state(&self, state: AccountState) -> anyhow::Result<()> {
         let sender = self.sender.load();
         let sender = sender
             .as_ref()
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. I've had a look at this, and the change fits the existing emitter API while preserving the current dispatch behavior. The external test pins both public visibility and caller-built state delivery, and this looks good to merge 👍

cjdsellers (approved):

PR #4908, opened 2026-09-02
Apply hedging flip remainders to the flipped position

folknor:
A follow-up to the Hedging position ID conflict rejection introduced with the Binance Futures hedge reconciliation work.

## A partially filled reversal under Hedging drops every fill after the first

Under `OmsType::Hedging` with a virtual position ID, `flip_position` mints a fresh ID for the flipped leg and `open_position` re-indexes the order against it through `Cache::add_position`. The venue keeps stamping the ID it was given, so every later partial fill of that order arrives with the original ID, and `determine_position_id` now sees a cached ID that differs from the fill's. The Hedging arm added for venue-supplied conflicts returns `None`, so the fill is applied to neither the order nor the position: a long 2 reversed by a sell 4 that fills 3 then 1 ends short 1 with the order at `filled_qty` 3, while the venue holds short 2 and a filled order. Netting never reaches the arm because `flip_position` reuses the fill's ID there. The built-in matching engine produces the same sequence on its own: `fill_position_for_order` computes `venue_position_id` once per fill pass and `apply_fills` stamps every fill of the pass with it, so an order flipping a position and filling at two book levels in one pass reaches this in any backtest or sandbox run under Hedging.

Applying the flipping fill writes its ID onto the order, and `add_position` then moves only the cache index, so a flip remainder is a fill whose ID equals the order's own `position_id` while the cache holds a different virtual ID. When that holds and both IDs are virtual, the fill falls through to the existing warn-and-reassign-from-cache path, which lands it on the flipped position. A venue-supplied conflict never satisfies it, because the order holds the ID it was submitted or last filled under, so the rejection written for Binance hedge legs is unchanged.

## Testing

`test_hedging_flip_applies_remainder_to_flipped_position` opens a long 100,000 under a virtual ID, sells 200,000 in a 150,000 flipping partial and a 50,000 remainder, both stamped with the original ID, and asserts the order's `filled_qty` is 200,000, the original position is closed, and the single open position is a distinct ID short 100,000 carrying the remainder's trade ID. Against the unfixed engine the `filled_qty` assertion fails at 150,000. `test_conflicting_venue_position_id_rejects_hedging_fill` and the live manager's `test_reconcile_mass_status_skips_only_position_with_fill_position_id_conflict` are unchanged and still pass.

### DIFF
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -3241,7 +3241,13 @@ impl ExecutionEngine {
             && let Some(fill_position_id) = fill.position_id
             && cached_position_id != fill_position_id
         {
-            if oms_type == OmsType::Hedging {
+            // A flip re-indexes the order to a newly minted virtual position ID while
+            // later fills of that order still carry the ID the order holds.
+            let flip_remainder = order.and_then(OrderAny::position_id) == Some(fill_position_id)
+                && fill_position_id.is_virtual()
+                && cached_position_id.is_virtual();
+
+            if oms_type == OmsType::Hedging && !flip_remainder {
                 log::error!(
                     "Cannot apply hedging fill {} for {}: venue position ID {fill_position_id} conflicts with cached position ID {cached_position_id}",
                     fill.trade_id,
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -3241,27 +3241,23 @@ impl ExecutionEngine {
             && let Some(fill_position_id) = fill.position_id
             && cached_position_id != fill_position_id
         {
-            // A flip re-indexes the order to a newly minted virtual position ID while
-            // later fills of that order still carry the ID the order holds.
-            let flip_remainder = order.and_then(OrderAny::position_id) == Some(fill_position_id)
-                && fill_position_id.is_virtual()
-                && cached_position_id.is_virtual();
+            if oms_type == OmsType::Hedging {
+                if !self.is_flip_remainder(fill, cached_position_id, fill_position_id) {
+                    log::error!(
+                        "Cannot apply hedging fill {} for {}: venue position ID {fill_position_id} conflicts with cached position ID {cached_position_id}",
+                        fill.trade_id,
+                        fill.client_order_id(),
+                    );
 
-            if oms_type == OmsType::Hedging && !flip_remainder {
-                log::error!(
-                    "Cannot apply hedging fill {} for {}: venue position ID {fill_position_id} conflicts with cached position ID {cached_position_id}",
-                    fill.trade_id,
-                    fill.client_order_id(),
+                    return None;
+                }
+            } else {
+                log::warn!(
+                    "Incorrect position ID assigned to fill: \
+                     cached={cached_position_id}, assigned={fill_position_id}; \
+                     re-assigning from cache",
                 );
-
-                return None;
             }
-
-            log::warn!(
-                "Incorrect position ID assigned to fill: \
-                 cached={cached_position_id}, assigned={fill_position_id}; \
-                 re-assigning from cache",
-            );
         }
 
         if let Some(position_id) = cached_position_id {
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hey @folknor,

Thank you for the PR. I extended the sequence to one flipping fill followed by two remainders carrying the original position ID. The first remainder is accepted, but applying it changes `OrderCore.position_id` to the cached flipped ID. The next remainder is then rejected, leaving the order at 175,000 of 200,000 filled.

This exposes a wider lifecycle problem, but I’d like to keep this PR focused. If you’re able, would you be interested in adjusting just the active fill path so a mismatch is accepted only when the recorded split relationship proves that this same order moved from the original virtual position to the cached flipped position? The regression test should include at least two post-flip remainders, retain rejection for unrelated virtual position IDs, and avoid the incorrect-ID warning on the accepted path. I’ll handle restart reconstruction, reconciliation, and fill corrections separately.

folknor:
Thank you, that sequence kills the predicate as filed and the reason is that it consumed its own evidence.

## Why the second remainder was rejected

The filed predicate compared the fill's position ID against the order's own. Applying the first accepted remainder writes the determined ID - the cached flipped one - onto `OrderCore.position_id`, so by the time the second remainder arrives the order no longer holds the ID the fill carries and the mismatch is rejected. One remainder passed and every later one failed, which is why the order stopped at 175,000.

## What it compares now

The split relationship is recorded in the two positions, and the flip writes it once: `flip_position` opens the flipped position from `fill_split2`, so its `opening_order_id` is the reversing order, and `fill_split1` drives the original to zero quantity, where `Position::apply` sets `closing_order_id` to that same order. Neither field moves again.

`is_flip_remainder` accepts a Hedging mismatch only when both halves still name this order: the cached position was opened by it, and the position the fill names is closed with it as the closing order. That is stable across any number of remainders, and it does not consult the order at all.

It also drops the virtuality test as the load-bearing check. `PositionId::is_virtual` is a `P-` prefix test, and the live test corpus spells venue IDs with that prefix throughout, so a conflict between two unrelated `P-` IDs was only ever excluded by the clause that has now gone. The recorded opening and closing orders exclude it directly.

The accepted path no longer reaches the incorrect-position-ID warning: the Hedging branch either declines or falls through silently, and the netting branch keeps the warning unchanged.

## Remainders arriving after the flipped position closes

The predicate does not require the cached position to still be open, so a remainder arriving after the flipped position has itself been closed by another order is accepted too. That fill names a closed position, which `handle_position_update` routes to `Action::Reopen` and the standard Hedging reopen path, so the flipped ID reopens for the remaining quantity. That reopen is the one any Hedging fill naming a closed position already takes.

Rejecting instead drops the fill outright: `determine_position_id` returning `None` makes the `Filled` arm return before `update_cached_order`, so neither the order nor any position records the quantity, which is the 175,000 of 200,000 you saw. Reopening seemed the better of the two, but say if you would rather a remainder past the flipped position's close were refused.

## Testing

`test_hedging_flip_applies_remainders_to_flipped_position` runs one flipping fill of 150,000 against a 200,000 reversal, then two remainders of 25,000 each carrying the original position ID. It asserts the order reaches 200,000 filled, the original position is closed, and the single open position is short 100,000 and carries both remainder trade IDs.

`test_hedging_flip_remainder_reopens_a_closed_flipped_position` covers the case above: the flipped position is closed by a third order, then a 50,000 remainder of the reversal arrives carrying the original ID, and the flipped ID is reopened short 50,000 with the order complete at 200,000.

Restoring the previous predicate fails the first test at the `filled_qty` assertion, so the second remainder is what it pins. `test_conflicting_venue_position_id_rejects_hedging_fill` and the rest of the position-ID suite pass unchanged.

Restart reconstruction, reconciliation and fill corrections are left alone as you asked.

cjdsellers:
@folknor, thanks, the two-position predicate now carries multiple remainders as intended. I found one issue with the added case where another order closes the flipped position before the remainder arrives.

The Hedging reopen behavior predates this PR, but this change newly routes the remainder through it. `Cache::add_position` replaces the closed `Position` under the same ID without first snapshotting that cycle. `Portfolio::realized_pnl` totals current positions and position snapshots, so the closed cycle's realized gain or loss disappears after the ID is reopened. The analyzer keeps its recorded trade, but the cache-derived account PnL does not.

I think preserving the closed cycle before reopening is the right outcome, since rejecting the remainder would return to dropping a real fill. If you're able, could you address that and extend `test_hedging_flip_remainder_reopens_a_closed_flipped_position` with distinct prices and an exact assertion that the earlier realized PnL remains included after the remainder?

folknor:
Thank you.

`reopen_position` now archives the closed cycle for both OMS types before `add_position` replaces it: the Hedging arm takes the same `snapshot_position` path the Netting reopen already took, so the cycle's realized PnL stays in the snapshots `Portfolio::realized_pnl` sums for Hedging beside the current positions' - counted once, since the reopened position starts fresh. `test_hedging_flip_no_snapshot` pins the flip path, where no ID is reused and nothing is replaced, and is unchanged.

## Testing

`test_hedging_flip_remainder_reopens_a_closed_flipped_position` now fills at distinct prices with zero commissions - the flip at 1.00010, the close at 1.00002 - and asserts that after the remainder reopens the flipped ID it holds exactly one snapshot whose realized PnL is 4.00 USD, with the reopened position's own realized PnL at zero. With the snapshot gated back to NETTING only, the test fails at the snapshot-count assertion. The assertion is on the cache frame rather than on `Portfolio::realized_pnl` - the execution test suite has no portfolio harness - and the Hedging summation over snapshots plus current positions is already pinned in the portfolio suite, so this test pins the frame that summation consumes.

cjdsellers (approved):
@folknor, thanks for addressing the PnL preservation issue. This looks good to merge. I'll handle the remaining small follow-ups after merge: update the position docs to describe snapshotting on hedging reopens, and add regression tests confirming unrelated virtual position IDs are still rejected.

Issue #4876, opened 2026-08-27
[RFC] Define Polymarket settlement evidence and fill correction


cjdsellers:
# RFC

## Before you proceed

- [x] I searched existing issues and discussions. This RFC consolidates [Discussion #4722](https://github.com/nautechsystems/nautilus_trader/discussions/4722)
- [ ] I have no other RFC open. I understand that the project considers one RFC from an author at a time

## Context

Polymarket reports trades before on-chain settlement is final. `MATCHED_NOT_BROADCASTED`, `MATCHED`, `MINED`, and `RETRYING` are provisional; `CONFIRMED` and `FAILED` are terminal. The adapter can therefore apply a valid fill that later requires correction.

Polymarket also states that its [user stream does not replay every update missed during a disconnect](https://docs.polymarket.com/trading/realtime-order-updates) and recommends refreshing open orders and recent trades after reconnecting.

This maintainer-owned RFC builds on the design work @seungpyoson developed in [Discussion #4722](https://github.com/nautechsystems/nautilus_trader/discussions/4722), including its corrected two-axis model. It supersedes the earlier state tables in that discussion.

The experimental PRs #4691, #4692, #4693, #4704, #4705, and #4707 are closed. Maker fill ownership from #4662 and the core `OrderFillVoided` path are already in `develop`. #4831 added strict REST report admission, but not the settlement registry proposed here.

**Current `develop` still has four relevant gaps:**

- REST and WebSocket evidence use different validation paths
- The adapter does not represent `MATCHED_NOT_BROADCASTED`
- A WebSocket `FAILED` update can directly void a fill before authoritative REST resolution
- The execution engine does not publish a negative acknowledgment when it rejects an enqueued fill or void

## Maintainer position

The project should accept an adapter-owned, process-local settlement registry with these boundaries:

- Keep settlement state private to the Polymarket adapter. `OrderFilled` and `OrderFillVoided` remain the core boundary
- Represent venue settlement and per-leg core application separately
- Make the registry the sole Polymarket authority for fill creation and correction, including WebSocket dispatch, targeted reports, mass status, and position reconciliation
- Use a fresh, authenticated, account-scoped, targeted REST terminal result as settlement authority
- Treat the first admitted terminal REST result as final. A conflicting later REST result is a hard reconciliation fault
- Allow all four provisional statuses, including `MATCHED_NOT_BROADCASTED`, to apply only for complete trades belonging to the current uninterrupted stream session
[...]

PR #4909, opened 2026-09-02
Publish an event when the execution engine declines a fill

folknor:
A follow-up to the hedging flip remainder fix in #4908. This branch is stacked on that one and includes its commit, so #4908 needs to land first; the diff to review here is the second commit only.

## A fill the execution engine declines to apply leaves no trace a subscriber can see

When `ExecutionEngine::handle_order_event` declines an `OrderFilled`, it returns before `update_cached_order` and before `publish_order_event`: the order is not updated, no event is published, and one log line is the only record. The cached order and position are wrong from then on with nothing a strategy, operator or reconciler can act on. Four decline classes are substantive: a venue-supplied position ID conflicting with the cached one under Hedging (the rejection from the Binance hedge reconciliation work, which #4908 narrows to exclude the engine's own flip), a fill whose instrument differs from the cached position's, an overfill with `allow_overfills` off, and a fill on an order whose state cannot accept one. The overfill case had no log line either: `check_overfill` bailed through `anyhow` and the caller discarded the error.

`process` is bus-driven and returns nothing, so the signal is a message. `OrderFillDeclined` is a standalone type published with `msgbus::publish_any` on `events.order_fill_declined.{strategy_id}`, in the shape `TradingStateChanged` already uses. It is not an `OrderEventAny` variant: it changes no order state, and a variant would enter every order's `apply`, the serialization schemas and the persistence paths for an event none of them should ever see. It carries the fill's identity and exposure fields (`client_order_id`, `venue_order_id`, `trade_id`, `position_id`, `order_side`, `last_qty`, `last_px`), a unit-variant `FillDeclinedReason` a strategy can match on, a `details` string with the diagnostic the log line carries, and `causation_id` set to the declined fill's `event_id`. Duplicate fills stay at warn and publish nothing. One private `decline_fill` logs and publishes at the final decision point; the validators now return typed errors so the caller can classify rather than parse a string.

Rust subscribers only. `PyCallableHandler` forwards `PyMessage` and nothing else, so a Rust-published message cannot reach a Python callable, and this adds no Python wrapper or stub entry, the same position `TradingStateChanged` is in. Bridging that is a separate change to the Python message bus.

## Testing

`test_position_id_conflict_publishes_fill_declined`, `test_position_instrument_mismatch_publishes_fill_declined`, `test_overfill_refusal_publishes_fill_declined` and `test_invalid_order_state_publishes_fill_declined` each subscribe to the topic and assert exactly one message with the expected reason, the fill's trade ID and `causation_id`, and that the order and positions are unchanged. `test_duplicate_fill_does_not_publish_fill_declined` asserts none. With the publish removed from `decline_fill`, all four positive assertions fail on a message count of 0.

### DIFF
--- a/crates/common/src/msgbus/switchboard.rs
+++ b/crates/common/src/msgbus/switchboard.rs
@@ -477,6 +477,10 @@ define_switchboard! {
     get_event_order_topic(strategy_id: StrategyId) -> strategy_id,
     "events.order.{}", strategy_id;
 
+    order_fill_declined_topics: StrategyId,
+    get_order_fill_declined_topic(strategy_id: StrategyId) -> strategy_id,
+    "events.order_fill_declined.{}", strategy_id;
+
     event_position_topics: StrategyId,
     get_event_position_topic(strategy_id: StrategyId) -> strategy_id,
     "events.position.{}", strategy_id;
@@ -691,6 +695,7 @@ define_wrappers! {
     get_snapshot_order_topic(client_order_id: ClientOrderId) -> MStr<Topic>,
     get_snapshot_position_topic(position_id: PositionId) -> MStr<Topic>,
     get_event_order_topic(strategy_id: StrategyId) -> MStr<Topic>,
+    get_order_fill_declined_topic(strategy_id: StrategyId) -> MStr<Topic>,
     get_event_position_topic(strategy_id: StrategyId) -> MStr<Topic>,
 }
 
--- a/crates/execution/src/engine/mod.rs
+++ b/crates/execution/src/engine/mod.rs
@@ -68,12 +68,13 @@ use nautilus_core::{
 use nautilus_model::{
     accounts::Account,
     enums::{
-        AccountType, ContingencyType, OmsType, OrderStatus, OrderType, PositionSide, TimeInForce,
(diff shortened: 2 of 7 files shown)


folknor:
Closing and reopening to re-trigger CI.

cjdsellers:
Hi @folknor,

Thank you for the PR, and for tracing where declined fills disappear. As of 82a03d7bfa the engine publishes a negative acknowledgment on `events.order_fill_declined.{instrument_id}`, which landed with the Polymarket settlement work in #4876, so I'm closing this one in its favor.

We'd rather avoid a new event type here, so that version republishes the declined `OrderFilled` or `OrderFillVoided` unchanged. Every declined fill or correction reaches the subscriber, duplicates included, and an adapter can match it to the event it enqueued. It's keyed by instrument so a venue adapter can subscribe without tracking strategies, and the decline reason stays in the logs rather than on the event.

Your point about the overfill case still stands: `check_overfill` builds its message, but the caller discards the error, so an overfill rejection is still silent in the logs. That could be a focused follow-up if it's still needed.

folknor:
What a strange coincidence! I had literally gotten my nautilus trader repo up to speed less than 10 minutes before you closed this, and was working on rebasing my open PRs.

I'll see about that followup and you can expect more PRs from me later today. If you wish to point me in any specific direction in the email thread, that might be a good idea?

PR #4916, opened 2026-09-02
Preserve cash order locks on unreported balances

### LINKED ISSUE #4911: [Polymarket] CashAccount discards order reservations on every account state refresh, so `locked` is always zero
## Confirmation

Nearest existing issues, none covering this: #2906, #3704 (both closed), #3565 (RFC).

## Expected behavior

For a `CASH` account, `locked` should reflect collateral committed to open orders, with
`free = total - locked`. The portfolio computes this correctly today —
`update_balance_locked` (`portfolio/src/manager.rs:180`) walks `orders_open` into
`balances_locked`, and `recalculate_balance` (`model/src/accounts/base.rs:453`) derives
`locked` and `free` from it.

`is_reported` decides whose figure wins (`model/src/accounts/cash.rs:258`):

```rust
// Only clear locks when the venue reports a fresh balance snapshot
if event.is_reported && !event.balances.is_empty() {
    self.balances_locked.clear();
}
```

Two cases follow:

1. **`true`** — the venue sent a full snapshot including its own reservation figure, so
   clearing the local map is correct. Bybit, Binance spot, Coinbase, Betfair and
   Hyperliquid all supply one.
2. **`false`** — not a full snapshot; the locally computed reservations should survive and
   stay reflected in `locked` / `free`.

Case 2 is the only correct option for an adapter whose venue returns a bare total.
**It does not work.**

## Actual behavior

With `is_reported = false` the `balances_locked` map survives the guard, but
`base_apply` → `update_balances` (`accounts/base.rs:241` → `:185`) then inserts the
incoming balance wholesale:

```rust
pub fn update_balances(&mut self, balances: &[AccountBalance]) {
    for balance in balances {
        self.balances.insert(balance.currency, *balance);   // total, locked AND free
    }
}
```

`locked` is stamped back to the payload's value and nothing re-derives it from the
surviving map until the next order event. So an adapter that cannot supply a reservation
has no way to report a balance without destroying the portfolio's accounting: `true`
clears the map, `false` overwrites the result.

**Observed on Polymarket.** `GET /balance-allowance` returns only the collateral total, so
the adapter hardcodes `Decimal::ZERO` for the reservation
(`adapters/polymarket/src/execution/parse.rs:639`) and emits it as a reported snapshot
(`reports.rs:841`). Across a live capture of 430 `AccountState` lines, `locked` was
`0.000000` on **every one**, while the venue reported 20.00–48.40 pUSD committed to resting
orders over the same window. `total` tracked reality (33 distinct values as fills settled) —
only the reservation was frozen. A strategy reading `balance_free()` therefore sees the full
account total however much is committed, over-commits, and learns of it only from the
venue's rejection.

`adapters/kraken/src/http/spot/client.rs:1857` has the same shape (`Decimal::ZERO` plus
`is_reported = true` at `:1862`; `spot_account_type` defaults to `AccountType::Cash`).

## Steps to reproduce

Deterministic — no market data, credentials, or network. Add to the existing `mod tests` in
`crates/model/src/accounts/cash.rs`, then
`cargo test -p nautilus-model --lib accounts::cash`:

```rust
#[rstest]
fn test_unreported_snapshot_preserves_order_locks(
    mut cash_account: CashAccount,
    audusd_sim: CurrencyPair,
) {
    let instrument = InstrumentAny::CurrencyPair(audusd_sim);
    let usd = Currency::USD();

    // The portfolio computes a reservation for a resting order and stores it.
    cash_account.update_balance_locked(instrument.id(), Money::from("500.00 USD"));
    assert_eq!(
        cash_account.balance_locked(Some(usd)).unwrap(),
        Money::from("500.00 USD"),
    );

    // A refresh carrying only the total, locked = 0, explicitly NOT a full venue
    // snapshot - the only option an adapter has when the venue supplies no reservation.
    let refresh = AccountState::new(
        cash_account.id,
        AccountType::Cash,
        vec![AccountBalance::new(
            Money::from("1000.00 USD"),
            Money::from("0.00 USD"),
            Money::from("1000.00 USD"),
        )],
        vec![],
        false,
        uuid4(),
        1.into(),
        1.into(),
        Some(usd),
    );
    cash_account.apply(refresh).unwrap();

    // EXPECTED: the surviving reservation is still reflected in the balance.
    assert_eq!(
        cash_account.balance_locked(Some(usd)).unwrap(),
        Money::from("500.00 USD"),
        "reservation for the resting order was discarded",
    );
    assert_eq!(
        cash_account.balance_free(Some(usd)).unwrap(),
        Money::from("500.00 USD"),
        "free balance reports collateral that is already committed",
    );
}
```

Live path: run any strategy keeping more than one limit order resting on `POLYMARKET` and
log `AccountState` while orders are open — `locked` stays `0.000000`.

## Code snippets or logs

Test output on `develop` @ `3e4d66a96a`:

```
test accounts::cash::tests::test_unreported_snapshot_preserves_order_locks ... FAILED

thread '...' panicked at crates/model/src/accounts/cash.rs:403:9:
assertion `left == right` failed: reservation for the resting order was discarded
  left: Money(0.00, USD)
 right: Money(500.00, USD)

test result: FAILED. 157 passed; 1 failed; 0 ignored; 0 measured; 3001 filtered out
```

Live Polymarket rejection — the venue reports the reservation inline as
`sum of active orders`, the figure `balances_locked` should be holding:

```
not enough balance / allowance: the balance is not enough ->
  balance:                    66698599      // 66.698599 pUSD
  sum of active orders:       48400000      // 48.400000 committed to resting orders
  sum of matched orders:      21949680      // 21.949680 already spent on fills
  order amount (inc. fees):    2150000      // 5 x 0.43
                                            // 66.698599 - 48.40 - 21.94968 = -3.651081
```

The `AccountState` logged in the same milliseconds:

```
AccountState(account_id=POLYMARKET-001, account_type=CASH, base_currency=pUSD,
  is_reported=true,
  balances=[AccountBalance(total=66.698599 pUSD, locked=0.000000 pUSD, free=66.698599 pUSD)])
```

## Suggested fix

Re-derive `locked` / `free` after `base_apply`, but only where a local reservation actually
survived — otherwise a venue-supplied `locked` is destroyed instead.

```diff
     fn apply(&mut self, event: AccountState) -> anyhow::Result<()> {
         // ... borrowing guard unchanged ...

         // Only clear locks when the venue reports a fresh balance snapshot
         if event.is_reported && !event.balances.is_empty() {
             self.balances_locked.clear();
         }

+        let currencies: Vec<Currency> = event.balances.iter().map(|b| b.currency).collect();
+
         self.base_apply(event);
+
+        // A snapshot that is not a full venue report carries only the total, so re-derive
+        // locked and free from the reservations that survived above. Without this the
+        // incoming zero silently discards collateral committed to open orders.
+        for currency in currencies {
+            // Only re-derive where a local reservation actually survived. A reported
+            // snapshot has just cleared the map, so its own `locked` figure is preserved.
+            if self
+                .balances_locked
+                .keys()
+                .any(|(_, locked_currency)| *locked_currency == currency)
+            {
+                base::recalculate_balance(&mut self.base.balances, &self.balances_locked, currency);
+            }
+        }
+
         Ok(())
     }
```

The `balances_locked` guard is load-bearing: without it
`test_apply_given_new_state_event_updates_correctly` regresses, since that test applies a
snapshot carrying a genuine `locked = 0.5 BTC` against an empty local map and the
unconditional recalculation zeroes it. `recalculate_balance` is already `pub(crate)` in
`accounts::base`, which `cash.rs:92` already calls with `&mut self.base.balances`, so no
visibility change is needed. Stale locks are not a risk — `manager.rs:471` calls
`clear_balance_locked` when an instrument's last open order closes.

Verified on `3e4d66a96a`: `nautilus-model --lib` 3158 passed / 0 failed,
`nautilus-portfolio --lib` 64 passed / 0 failed, new test passes. (`nautilus-common` shows
10–11 `logging::` serial-test failures with and without the patch — pre-existing.)

With that in place the Polymarket adapter can correctly emit `reported = false`:

```diff
-    emitter.emit_account_state(vec![account_balance], vec![], true, ts_event, None);
+    // Polymarket's /balance-allowance returns only the collateral total with no
+    // reservation figure, so this is not a full venue snapshot. Reporting it as one
+    // would clear the reservations the portfolio computed from open orders.
+    emitter.emit_account_state(vec![account_balance], vec![], false, ts_event, None);
```

No adapter currently passes `reported = false` (all 71 emit sites pass `true`), so the core
change is inert until one opts in. Kraken spot (`spot/client.rs:1862`) needs the same
one-word change but should be validated separately — that client chains a `margin_entry`
into `balances`, has a `skip_margin_wallet` branch, and `spot_account_type` is configurable.

Worth covering in regression tests: `false` with surviving locks preserves them; `true`
still clears and reproduces the payload; `false` with an empty map leaves a venue-supplied
`locked` intact.

## Specifications

- OS platform: Pop!_OS 24.04 LTS (Linux 6.18.7-76061807-generic, x86_64)
- Python version: 3.12.3
- `nautilus_trader` version: `2.0.0rc4.dev20260830+18042` (wheel); also confirmed by
  `cargo test` against `develop` @ `3e4d66a96a` (2026-09-02)
- Installed from: package index wheel (pip, `manylinux_2_34_x86_64`), plus a source checkout
  of `develop` for the Rust test
- Adapter/venue: Polymarket (observed), Kraken spot (same shape by inspection); the
  `CashAccount::apply` defect itself is venue-independent

yashwardhan-gautam:
  and, if I used AI,
  this draft
  limitation below

## Summary

`AccountState::is_reported` decides whether a venue snapshot's reservation figure replaces
the portfolio's computed one. `CashAccount::apply` honours it when clearing `balances_locked`,
but `base_apply` → `update_balances` then inserts the incoming `AccountBalance` wholesale,
stamping `locked` back to the payload's value with nothing re-deriving it until the next
order event. An adapter whose venue returns a bare total has no correct option: `true` clears
the map, `false` overwrites the result.

`apply` now re-derives `locked` and `free` after `base_apply`, but only for currencies where
a local reservation survived. The guard is load-bearing — unconditional recalculation
destroys a venue-supplied `locked`, which
`test_apply_given_new_state_event_updates_correctly` catches. A reported snapshot has just
cleared the map, so nothing recalculates and its own figure stands: the path is bit-identical
for all 71 emit sites in the tree, every one of which passes `reported = true`.

Polymarket switches to `reported = false`, now correct and what makes the fix observable:
`GET /balance-allowance` returns only the collateral total, so `parse_balance_allowance`
hardcodes `Decimal::ZERO`. That zero overwrote the portfolio's reservations on every refresh
— including after every finalized trade — pinning `locked` at `0.000000` while orders rested,
so `balance_free()` reported the full total however much was committed.

Kraken spot (`http/spot/client.rs:1862`) shares the shape but is left alone: it chains a
`margin_entry` into `balances` and `spot_account_type` is configurable, so it needs separate
validation.

## Related issues/PRs

Resolves #4911

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

No documentation changes. No PyO3 bindings or wrapped Rust docs were touched, so
`make py-stubs` does not apply.

## Testing

`test_unreported_snapshot_preserves_order_locks` (new, in `accounts::cash::tests`) stores a
`500.00 USD` reservation, applies an `is_reported = false` snapshot carrying `total = 1000.00`
and `locked = 0.00`, then asserts `locked` and `free` are both still `500.00`. It fails before
this change (`left: Money(0.00, USD)`) and passes after.

Unchanged boundary tests: `test_apply_clears_per_instrument_locks` (a reported snapshot still
clears the map) and `test_apply_given_new_state_event_updates_correctly` (a genuine
`locked = 0.5 BTC` against an empty local map survives — the one that fails without the guard).

`cargo test -p nautilus-model --lib` (3158 tests), `cargo test -p nautilus-portfolio --lib`
(64 tests) and `cargo nextest run -p nautilus-polymarket` (1902 tests) all pass.

### DIFF
--- a/crates/adapters/polymarket/src/execution/reports.rs
+++ b/crates/adapters/polymarket/src/execution/reports.rs
@@ -838,7 +838,9 @@ pub(super) async fn fetch_and_emit_account_state(
         "Account state updated: balance={} pUSD",
         account_balance.total
     );
-    emitter.emit_account_state(vec![account_balance], vec![], true, ts_event, None);
+    // Not a full snapshot: `GET /balance-allowance` returns no reservation figure, so
+    // reporting it as one would clear the portfolio's locks from open orders
+    emitter.emit_account_state(vec![account_balance], vec![], false, ts_event, None);
     Ok(())
 }
 
--- a/crates/model/src/accounts/cash.rs
+++ b/crates/model/src/accounts/cash.rs
@@ -260,7 +260,23 @@ impl Account for CashAccount {
             self.balances_locked.clear();
         }
 
+        let currencies: Vec<Currency> = event.balances.iter().map(|b| b.currency).collect();
+
         self.base_apply(event);
+
+        // A non-reported snapshot carries only the total, so re-derive locked and free
+        // from any reservations which survived above
+        for currency in currencies {
+            // A reported snapshot cleared the map, so its own `locked` figure stands
+            if self
+                .balances_locked
+                .keys()
+                .any(|(_, locked_currency)| *locked_currency == currency)
+            {
+                base::recalculate_balance(&mut self.base.balances, &self.balances_locked, currency);
+            }
+        }
+
         Ok(())
     }
 
@@ -363,6 +379,50 @@ mod tests {
         position::Position,
         types::{AccountBalance, Currency, Money, Price, Quantity},
     };
+
+    #[rstest]
+    fn test_unreported_snapshot_preserves_order_locks(
+        mut cash_account: CashAccount,
+        audusd_sim: CurrencyPair,
+    ) {
+        let instrument = InstrumentAny::CurrencyPair(audusd_sim);
+        let usd = Currency::USD();
+
+        // The portfolio reserves collateral for a resting order
+        cash_account.update_balance_locked(instrument.id(), Money::from("500.00 USD"));
+        assert_eq!(
+            cash_account.balance_locked(Some(usd)).unwrap(),
+            Money::from("500.00 USD"),
+        );
+
+        // Total only, with no reservation figure and not flagged as a full snapshot
+        let refresh = AccountState::new(
+            cash_account.id,
+            AccountType::Cash,
+            vec![AccountBalance::new(
+                Money::from("1000.00 USD"),
+                Money::from("0.00 USD"),
+                Money::from("1000.00 USD"),
+            )],
+            vec![],
+            false, // is_reported
+            uuid4(),
+            1.into(),
+            1.into(),
+            Some(usd),
+        );
+        cash_account.apply(refresh).unwrap();
+
+        // The reservation must survive
+        assert_eq!(
+            cash_account.balance_locked(Some(usd)).unwrap(),
+            Money::from("500.00 USD"),
+        );
+        assert_eq!(
+            cash_account.balance_free(Some(usd)).unwrap(),
+            Money::from("500.00 USD"),
+        );
+    }
 
     #[rstest]
     fn test_display(cash_account: CashAccount) {


cjdsellers:
Hi @yashwardhan-gautam,

Thanks for the PR. There are two correctness issues to resolve before this can merge.

Live cash accounts are created with `calculate_account_state=false`, and `Portfolio::update_order` returns before changing per-instrument locks for those accounts. `initialize_orders()` can seed a lock for an order restored during startup, but a later fill or cancel cannot release it. Once Polymarket balance refreshes are always unreported, each refresh preserves that stale lock indefinitely. After a fill, this also subtracts the old reservation from the already-reduced venue total, so the risk engine can reject orders against collateral that is actually free. New orders accepted after startup still do not create a lock, so the normal live-session case remains unfixed.

The regression test seeds `balances_locked` directly and misses this path. It needs to cover a live-style cash account through resting-order creation, a refresh, and a terminal fill or cancel, with exact locked and free balances at each step.

`CashAccount::apply` also records the incoming event before re-deriving the current balances. `last_event()`, persistence, and republished account states therefore retain `locked=0, free=total` while the account object uses derived values. Those representations need to remain consistent.

Finally, `is_reported=false` currently means system-calculated, not a partial venue snapshot. This change relies on the latter meaning, so that contract needs to be represented explicitly or kept within the Polymarket-specific path.

yashwardhan-gautam:
> Hi @yashwardhan-gautam,
> 
> Thanks for the PR. There are two correctness issues to resolve before this can merge.
> 
> Live cash accounts are created with `calculate_account_state=false`, and `Portfolio::update_order` returns before changing per-instrument locks for those accounts. `initialize_orders()` can seed a lock for an order restored during startup, but a later fill or cancel cannot release it. Once Polymarket balance refreshes are always unreported, each refresh preserves that stale lock indefinitely. After a fill, this also subtracts the old reservation from the already-reduced venue total, so the risk engine can reject orders against collateral that is actually free. New orders accepted after startup still do not create a lock, so the normal live-session case remains unfixed.
> 
> The regression test seeds `balances_locked` directly and misses this path. It needs to cover a live-style cash account through resting-order creation, a refresh, and a terminal fill or cancel, with exact locked and free balances at each step.
> 
> `CashAccount::apply` also records the incoming event before re-deriving the current balances. `last_event()`, persistence, and republished account states therefore retain `locked=0, free=total` while the account object uses derived values. Those representations need to remain consistent.
> 
> Finally, `is_reported=false` currently means system-calculated, not a partial venue snapshot. This change relies on the latter meaning, so that contract needs to be represented explicitly or kept within the Polymarket-specific path.

That makes sense, repurposing `is_reported` is not the right way. Making the changes no good. 

what i can think of roughly, is to give cash accounts on venues that report no reservation as same treatment as wallet account: 

- `CashAccount::apply` — derive `locked`/`free` from `balances_locked` and rewrite `event.balances` before `base_apply`, ignoring the incoming split, mirroring `WalletAccount::apply`.
- A per-account marker for "this venue supplies no reservation", so only affected venues change behaviour and `is_reported` keeps its documented meaning.
- `portfolio.rs` — extend the early return so those accounts reach the lock path, alongside the existing `is_wallet` term, and route them the way `update_balance_locked_wallet` is routed.
- A startup rebuild mirroring `initialize_wallet_orders` (`live/src/node/mod.rs`), since the reservations are transient and not persisted.
- Tests covering a live-style cash account through resting-order creation, a refresh, and a terminal fill or cancel, asserting exact `locked` and `free` at each step.

Polymarket then needs no adapter change at all: it keeps `reported = true`, and the hardcoded `Decimal::ZERO` is simply ignored.

That's four touch points across `model`, `portfolio` and `live` and it changes live order-event handling — so I'd rather agree the scope with you on #4911 first, including whether it belongs on `CashAccount` conditionally or as an account-level capability. Happy to close this PR in the meantime if you'd prefer the discussion there.

yashwardhan-gautam:
@cjdsellers  actually I've a better idea, ignore the [above one ](https://github.com/nautechsystems/nautilus_trader/pull/4916#issuecomment-5521862487) 

The polymarket situation is different from other venues, because there's a settling  state in between 

image

what adapter reports 
```
active   = Σ resting BUY orders   (originalSize − sizeMatched) × price
settling = Σ filled BUY spend awaiting settlement   (notional + commission)

AccountState {
    total:  GET /balance-allowance,      // venue's cached total, unchanged
    locked: active + settling,
    free:   total − locked,
}
```

Polymarket has no `hold_trade` equivalent -- no endpoint reports a reservation. But the user websocket already carries everything needed to maintain one adapter-side: order events (`PLACEMENT` / `UPDATE` / `CANCELLATION` with `originalSize`, `sizeMatched`, `price`, `side`) and trade status transitions (`MATCHED -> MINED -> CONFIRMED / FAILED`). Proposal:

- Maintain two sums on `WsDispatchState` (BUY side only; a SELL reserves outcome tokens, not
  collateral): 
  `active` = `sum of (originalSize - sizeMatched) x price over resting orders` 
  and
  `settling` = `sum of notional + commission over filled-but-unconfirmed trades`. 
  A fill converts its reservation from active to settling rather than releasing it.
- Emit `locked = active + settling` with `is_reported = true` unchanged. This reproduces the venue's own reject arithmetic: balance - "sum of active orders" - "sum of matched orders".
- On trade confirmation, call the currently-unused `update_balance_allowance` (clob.rs:562) and refetch; a settling entry drains only after that post-confirmation refresh lands, so any race transient understates free rather than overstating it.
- Seed/resync `active` from `GET /data/orders` on connect and reconnect; reconcile `settling` via `get_trades`.
  

**Questions to settle on before starting to code this approach:** 
1. Is the WS-maintained tracker with REST seed preferred over a simpler per-refresh `GET /data/orders` sum (one extra paginated call after every finalized trade, and it would miss the settling term)?
2. Should the drain wait for `CONFIRMED`, or is `MINED` sufficient?
3. Is one `update_balance_allowance` call per confirmed trade acceptable on rate limits?

cjdsellers:
Hi @yashwardhan-gautam,

Thanks for working the design through rather than forcing it onto `is_reported`. The wallet shape is right, but it's a cross-cutting concern spanning `model`, `portfolio`, `live` and the risk engine, so I'll take that on.

That leaves a self-contained piece in the Polymarket adapter if you'd like it. `GET /data/orders` returns the venue's own open orders with `price`, `original_size`, `size_matched` and `side`, which is the data behind the `sum of active orders` figure in your rejection, and `get_orders` already exists on the CLOB client. `fetch_and_emit_account_state` can fetch those alongside the balance, sum the collateral committed by resting BUY orders, and pass that as `locked` in place of the hardcoded `Decimal::ZERO`. `is_reported` stays `true` and stays truthful, and none of it depends on the core work.

Three constraints to build around:

- Only BUY orders reserve pUSD. A SELL reserves conditional tokens, which we don't carry as a balance.
- Matched but unsettled orders are held against the balance as well, and drop off `/data/orders` once fully filled, so `locked` will sit slightly under the venue's real hold.
- Refreshes fire on finalized trade, reconnect and mass status, not on order accept or cancel, so `locked` is accurate at each refresh and stale between them. Adding a trigger there is a rate-limit decision I'd rather make separately.

One thing to confirm against your account: whether the venue holds remaining or original size on a partially matched order.

Let's close this one and keep #4911 for the core change. The design will build on your diagnosis there 👍

Issue #3867, opened 2026-04-14
Kraken Futures: margin-account balance parse can violate `AccountBalance` `total == locked + free` invariant


Stamppot82:
## Bug report

**Component:** Kraken Futures adapter (`crates/adapters/kraken/src/http/futures/client.rs`)

**Summary**

The margin-account balance parse for Kraken Futures can panic in `AccountBalance::new_checked` with

```
`total` (<X>) - `locked` (<Y>) != `free` (<Z>)
```

when the raw `af` (available funds) value reported by Kraken and the f64 difference `amount - af` round independently at the currency precision (8 decimals for the synthesised crypto currency).

**Expected behaviour**

`AccountBalance` constructed from Kraken margin-account balances should always satisfy the `total == locked + free` invariant enforced by `AccountBalance::new_checked` (see `crates/model/src/types/balance.rs`).

**Actual behaviour**

For certain f64 balance / `af` combinations, `Money::new(amount)`, `Money::new(amount - af)` and `Money::new(af)` round such that `locked + free != total`, tripping the invariant and panicking in `expect(FAILED)` on the `pub fn new` path.

Minimal drift example at currency precision 8:

```rust
let af = 35.0_f64 * 1e-9;          // 0.000_000_035
let amount = 10.0_f64 + af;         // 10.000_000_035

// round(amount * 1e8) = 1_000_000_003
// round(af * 1e8)     = 4
// round((amount - af) * 1e8) = 1_000_000_000
// 4 + 1_000_000_000 = 1_000_000_004 != 1_000_000_003
```

**Root cause**

`parse_margin_account_balances` in `crates/adapters/kraken/src/http/futures/client.rs` passes `Money::new(amount - available, currency)` as `locked` and `Money::new(available, currency)` as `free` independently, rather than deriving one side from the other in fixed-point `Money` arithmetic. The sibling `parse_multi_collateral_balances` already derives `free = total - locked` correctly.

**Proposed fix**

Derive `free` from `total - locked` using fixed-point `Money` subtraction, matching the multi-collateral path:

```rust
let locked = Money::new((amount - available).max(0.0), currency);
let free = total - locked;
balances.push(AccountBalance::new(total, locked, free));
```

This keeps `total == locked + free` by construction at the currency precision, irrespective of f64 drift in the upstream fields.

**Reproduction**

Included as a regression test `test_parse_margin_account_balances_free_is_derived_from_total_minus_locked` in the accompanying PR.

**Environment**

[...]

PR #4922, opened 2026-09-03
Fix Kraken spot wallet balances to report venue-held amounts

### LINKED ISSUE #4921: [Kraken] Spot wallet balances never read BalanceEx `hold_trade`, so `locked` is always zero
# Bug Report

## Confirmation

## Expected behavior

For a Kraken spot account, `AccountState.balances` should report the portion of each wallet the
venue has reserved against resting orders as `locked`, so that `free` reflects genuinely
available funds.

Kraken exposes this per asset via `POST /0/private/BalanceEx`, which returns `balance` (total)
and `hold_trade` ("Total held amount for an asset").

This is what the margin path in the same function already does: it calls `TradeBalance` and
builds the collateral entry with `AccountBalance::from_total_and_free(equity, free_margin)`,
yielding a real `locked`.

## Actual behavior

Every wallet-derived balance reports `locked == 0` and therefore `free == total`, regardless of
how much Kraken is holding.

In `crates/adapters/kraken/src/http/spot/client.rs`,
`KrakenSpotHttpClient::request_account_state_with_metrics` sources wallet balances from
`POST /0/private/Balance`, which returns per-currency totals only, and then hardcodes the locked
component:

```rust
let currency = Currency::new(normalized_code, 8, 0, "0", CurrencyType::Crypto);
AccountBalance::from_total_and_locked(amount, Decimal::ZERO, currency).ok()
```

`AccountBalance::from_total_and_locked` derives `free = total - locked`, so `free` reports funds
the venue has already reserved as available.

`BalanceEx` and `hold_trade` appear nowhere in the repository, so the field the platform needs is
never requested.

Two clarifications on scope:

- This is not limited to `AccountType::Cash`. The quoted line is in the shared wallet loop, so it
  applies in margin mode too — the margin path only replaces the single collateral wallet with the
  synthetic `TradeBalance` entry, and every other wallet still reports `locked == 0`.
- The `total == locked + free` invariant holds numerically (`total = 0 + total`) but is
  semantically false, which is the same class of problem as #3867
  ("Kraken Futures: margin-account balance parse can violate `AccountBalance`
  `total == locked + free` invariant") and the currently open #4911
  ("[Polymarket] CashAccount discards order reservations on every account state refresh, so
  `locked` is always zero").

## Steps to reproduce

1. Connect a Kraken spot account (`AccountType.CASH`) with funds in a fiat or crypto wallet.
2. Place a resting limit order that Kraken reserves collateral against, and leave it working.
3. Request an account state refresh and inspect the resulting `AccountState.balances` entry for
   that currency.
4. Compare against `POST /0/private/BalanceEx` for the same asset.

Observed: nautilus reports `locked = 0` and `free = total`, while `BalanceEx` reports a non-zero
`hold_trade` equal to the reserved amount of the resting order.

The behavior is also visible without venue access, directly in the existing test suite: on
`develop`, `crates/adapters/kraken/tests/http.rs` contains
`test_spot_request_account_state_margin_other_wallets_unlocked`, which asserts
`balance.locked == Decimal::ZERO` and `balance.free == balance.total` for wallet balances. That
assertion passes today because the held amount is never read, so the current test suite encodes
the behavior rather than catching it.

## Code snippets or logs

`crates/adapters/kraken/src/http/spot/client.rs` — the wallet loop in
`request_account_state_with_metrics` (locate by symbol; line numbers move):

```rust
let balances_raw = self.inner.get_balance().await?;   // POST /0/private/Balance -> totals only
...
let balances: Vec<AccountBalance> = balances_raw
    .iter()
    .filter_map(|(currency_code, amount_str)| {
        let amount = Decimal::from_str_exact(amount_str).ok()?;
        ...
        let currency = Currency::new(normalized_code, 8, 0, "0", CurrencyType::Crypto);
        AccountBalance::from_total_and_locked(amount, Decimal::ZERO, currency).ok()
    })
    .chain(margin_entry)
    .collect();
```

Kraken `POST /0/private/BalanceEx` response shape (per
https://docs.kraken.com/api/docs/rest-api/get-extended-balance):

```json
{
  "error": [],
  "result": {
    "ZUSD": { "balance": "25435.21", "hold_trade": "8249.76" },
    "XXBT": { "balance": "1.2435",   "hold_trade": "0.8423" }
  }
}
```

## Specifications

- OS platform: macOS 26.6 (x86_64)
- Python version: 3.14.4
- `nautilus_trader` version: v2.0.0rc5 (`develop` @ `bb721205d99d761e2c1d0120b08265da5d36b467`)

zhaow-de:
  and, if I used AI,
  this draft
  limitation below

## Summary

Kraken spot wallet balances were sourced from `POST /0/private/Balance`, which returns
per-currency totals only, and the locked component was hardcoded to `Decimal::ZERO`. Since
`AccountBalance::from_total_and_locked` derives `free = total - locked`, every wallet-derived
balance reported `free == total`, presenting funds Kraken had already reserved against resting
orders as available. This sources those balances from `POST /0/private/BalanceEx`, which reports
`balance` and `hold_trade` per asset, and passes `hold_trade` as `locked`.

Notes for reviewers:

- This affected cash and margin accounts alike. The hardcoded zero was in the shared wallet loop;
  the margin path only replaces the single collateral wallet with the synthetic `TradeBalance`
  entry, so every other wallet also reported a zero hold.
- Request count per account update is unchanged (one Balance-family call), so rate-limit usage is
  unaffected.
- `get_balance` is retained for the spot wallet position-report path, which wants wallet totals
  rather than available amounts.

## Related issues/PRs

Closes #4921

Related to #4911 ([Polymarket] `CashAccount` discards order reservations, so `locked` is always
zero) — same class of defect in a different adapter. Prior art: #3867 (Kraken Futures
`total == locked + free` invariant).

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

Not a breaking API change. It is a behavior change: `AccountBalance.free` for Kraken spot wallets
now excludes venue-held amounts. Strategies that implicitly relied on `free == total` will see
smaller available balances, which is the corrected value.

## Documentation

  generated output

The PyO3 wrapper doc for `request_account_state_with_metrics` is generated from the underlying
Rust doc comment, so it was regenerated with `python/generate_docstrings.py` and committed rather
than hand-edited. No binding signature changed, so the `.pyi` stubs are unaffected, and a
repo-wide `generate_docstrings.py --dry-run` reports no remaining drift.

## Testing

Added `test_spot_request_account_state_cash_locks_held_amounts`, covering a held wallet, a
partially held wallet, and a wallet with no hold.

Two existing tests asserted the previous behavior and are updated:

- `test_spot_request_account_state_margin_other_wallets_unlocked` asserted
  `locked == 0` / `free == total` for non-collateral wallets. Its intent was that the margin lock
  must not bleed into other wallets, which is preserved and strengthened: it is renamed to
  `..._other_wallets_lock_own_holds` and now asserts each wallet locks its own `hold_trade`.
- `test_spot_request_account_state_margin_with_gbp_asset_tags_currency` asserted the USD wallet was
  unlocked when the margin target is GBP; it now asserts the USD wallet carries its own hold rather
  than the GBP margin lock.

Verified the new tests fail against the previous behavior (`locked` reported `0` where `1500` and
`0.1` were expected) and pass with the fix.

Local validation:

- `cargo test -p nautilus-kraken`: 724 passed, 0 failed.
- `cargo clippy -p nautilus-kraken --all-targets --all-features -- -D warnings`: clean.
- `make format`: clean, no changes produced.
- `make pre-commit`: 63 passed, 4 skipped (no matching files)


cjdsellers:
Hi @zhaow-de,

Thank you for the PR.

There is one remaining `BalanceEx` case to cover: Kraken defines available funds as `balance + credit - credit_used - hold_trade`, with `credit` and `credit_used` present for credit-line accounts. `BalanceExEntry` currently drops those fields, so `request_account_state_with_metrics` derives `free` as `balance - hold_trade` and understates the available balance whenever net credit is nonzero.

Could you include net credit in the account balance calculation and add a regression case for a credit-line response?

zhaow-de:
Thanks for catching that. Pushed a commit that reads `credit` and `credit_used` and includes net credit in the balance.

Because `total` has to equal `locked + free`, I put the net credit into `total`, so `total = balance + credit - credit_used` and `locked = hold_trade`, and `free` then derives to `balance + credit - credit_used - hold_trade`. The side effect is that `total` now sits above the raw wallet balance for credit-line accounts. That looked like the only mapping that gets `free` right without breaking the invariant, but happy to model it differently if you'd rather.

One related thing I ran into that your comment didn't mention: the wallet loop skipped any entry whose balance was zero, so a wallet with no balance but with available credit would have been dropped entirely. I moved that check onto the computed total instead.

Both fields are optional and treated as zero when absent, so accounts without a credit line behave exactly as before and none of the existing test expectations changed. The new test covers a credit-line wallet, a zero-balance wallet with credit, and a wallet with no credit fields at all. I confirmed it fails against the previous code, with the total coming out as 1000 instead of 4000.

cjdsellers (dismissed):
@zhaow-de, thanks for the follow-up. The updated credit balance handling looks correct, including the zero-balance credit case.

This looks good from my side. I'll resolve the conflict from here 👌

cjdsellers (approved):

PR #4930, opened 2026-09-05
Fix Binance Spot cancel report client order ID matching

### LINKED ISSUE #4923: [Binance Spot] Cancel executionReport is keyed on the cancel request's clientOrderId (`c`) instead of the order's (`C`), so in-flight tracking is never cleared
# Bug Report

## Confirmation

Reproduced on the Rust crates at 0.62.0 (`v2.0.0rc3`). The relevant code is unchanged in 0.63.0 (`v2.0.0rc4`) and on `develop` at `e6d5a8528d`; I have not re-run against those.

## Expected behavior

When a Spot order cancelled over the WS API is confirmed by the user-data `executionReport` (`x=CANCELED`), the Spot execution client should recognise the report as belonging to the tracked order and emit `OrderCanceled` for it. That clears the `CancelOrder` entry in `ExecutionManager::inflight_checks` via `observe_order_event`, and no `QueryOrder` is issued for an order that is already confirmed cancelled.

## Actual behavior

On a `CANCELED` executionReport Binance puts the **cancel request's** client id in `c` and the order's id in `C` (`origClientOrderId`). The Spot client cancels by `orderId` without a `newClientOrderId` (`build_cancel_order_params`, `spot/execution.rs:2559`), so `c` is a Binance-generated string.

The adapter only ever reads `c`:

- `decode_sbe.rs:166-175` maps `clientOrderId` → `client_order_id` and `origClientOrderId` → `original_client_order_id`; the JSON struct does the same (`user_data.rs:132-134`, commented "for cancel-replace").
- `original_client_order_id` is never read anywhere in the adapter.
- `decode_client_order_id` passes strings without the broker prefix through unchanged (`common/encoder.rs:219-232`), so the generated id becomes a plausible `ClientOrderId` and nothing is logged.

So in `dispatch_execution_report` (`spot/execution.rs:2622`) the `order_identities` lookup misses, the report goes down `dispatch_untracked_execution_report` (`:3049`), and instead of an `OrderCanceled` event the client emits an `OrderStatusReport` with `client_order_id = Some(<cancel request id>)` and the correct `venue_order_id`.

In `nautilus-live`, both clearing paths key on that id: `observe_order_status_report` (`manager.rs:2826-2843`) clears a non-existent entry, and `LiveNode::process_exec_event` (`node/mod.rs:1919-1937`) finds no such order in the cache. `ExecutionEngine::reconcile_order_status_report` (`execution/src/engine/mod.rs:1104-1120`) recovers the real order through the `venue_order_id` fallback and synthesises `OrderCanceled { reconciliation: true }`, so order state ends up correct and the strategy notices nothing. The `CancelOrder` in-flight entry is simply orphaned: `check_inflight_orders` polls it four times (`QueryOrder`) and clears it on the fifth retry via the "already resolved" branch.

`NEW` reports carry the order's own id in `c`, so submits take the tracked path and are cleared by `OrderAccepted` as expected. The WS `order.cancel` response itself is bookkeeping only ("OrderCanceled event is synthesized from UDS executionReport", `spot/execution.rs:2090-2099`), so nothing else clears the entry. The HTTP fallback emits `OrderCanceled` directly from the REST response and is not affected.

Related: those four queries all fail because `decode_order` (`spot/http/parse.rs:549-590`) reads `time` straight after `iceberg_qty` without the `cursor.advance(16)` that `decode_orders` (`:673`) and `decode_new_order_full` (`:435`) have, so `time` lands on the null sentinel. With that fixed, the first query response would clear the entry after one poll. I can file it separately if you prefer.

## Steps to reproduce

1. Rust `LiveNode`, Binance Spot on `BinanceEnvironment::Demo`, Ed25519 key, `use_ws_trading` at its default (`true`), `LiveExecEngineConfig` in-flight settings at defaults.
2. Submit a post-only limit bid far from mid so it rests. Every 5s, `Strategy::cancel_order` it and submit a replacement after `OrderCanceled`.
3. Watch the `ExecutionEngine` log: no `Query` while the order rests; after each `OrderCanceled`, `Query <client_order_id>` four times at ~6s spacing for that order.

2m30s run: 27 orders, 84 `QueryOrder`, 4 per cancelled order in steady state.

To see the mis-keyed report directly, set `LiveExecEngineConfig.debug = true`: the received `OrderCanceled` shows `reconciliation: true`. Or subscribe to `reconciliation_raw_order_status_report_topic()` and compare `client_order_id` with the order's.

## Code snippets or logs

One order, timestamps trimmed:

```
06:29:06.807053 [INFO] nautilus_trading::strategy: [CMD]--> CancelOrder
06:29:06.807163 [INFO] nautilus_execution::engine: Cancel O-20260903-062857-001-001-3 294869605
06:29:06.885866 [INFO] nautilus_trading::strategy: <--[EVT] OrderCanceled(instrument_id=BTCU.BINANCE, client_order_id=O-20260903-062857-001-001-3, venue_order_id=294869605, ...)
06:29:13.408121 [INFO] nautilus_execution::engine: Query O-20260903-062857-001-001-3
06:29:13.491757 [WARN] nautilus_binance::spot::execution: Failed to query order status: invalid negative Binance Spot SBE order time timestamp: -9223372036854775808
06:29:19.605819 [INFO] nautilus_execution::engine: Query O-20260903-062857-001-001-3
06:29:25.707370 [INFO] nautilus_execution::engine: Query O-20260903-062857-001-001-3
06:29:31.907676 [INFO] nautilus_execution::engine: Query O-20260903-062857-001-001-3
```

The existing dispatch tests feed the order's own id as `c` on `CANCELED` reports (`spot/execution.rs:3745-3830`), which is why this doesn't show up in CI.

Suggested fix: in `dispatch_execution_report` and `parse_spot_exec_report_to_order_status`, use `original_client_order_id` when it is non-empty and fall back to `client_order_id`. Binance sets `C` on `CANCELED`, `EXPIRED` and `REPLACED` and leaves it empty otherwise. This also covers cancels not initiated by Nautilus, which setting `newClientOrderId` on the request would not.

## Specifications

- OS platform: macOS 26.6.2, arm64
- Python version: n/a (pure Rust `LiveNode`)
- `nautilus_trader` version: Rust crates 0.62.0 (`v2.0.0rc3`)
- Installed from: crates.io, pinned `=0.62.0`, Rust 1.97.1
- Adapter/venue: Binance Spot, Demo environment, WS trading (SBE), Ed25519

abhijeetvichare76:
## A Spot cancel report is matched on the cancel request's ID

On a `CANCELED` `executionReport` Binance puts the cancel request's client ID in `c` and the
order's own ID in `C`. The Spot client cancels by `orderId` with no `newClientOrderId`, so `c` is
a Binance-generated string, and `dispatch_execution_report` only ever read `c`. The lookup in
`order_identities` missed, the report went down the untracked path, and the emitted
`OrderStatusReport` carried the cancel request's ID. The engine still recovered the order through
the `venue_order_id` fallback, so order state ended up right, but nothing cleared the in-flight
`CancelOrder` entry and `check_inflight_orders` polled it with four `QueryOrder`s per cancel.

`BinanceSpotExecutionReport::order_client_order_id` returns `C` when non-empty and `c` otherwise,
and the dispatch lookup and the two parsers in `websocket/trading/parse.rs` use it. `C` is empty
on every other execution type, so `NEW`, `TRADE` and `REJECTED` reports are unchanged, and cancels
not initiated by Nautilus resolve the same way, which setting `newClientOrderId` on our own cancel
requests would not give us.

Preferring `C` on its own regresses cancel-replace. The cancel half of a `cancelReplace` produces
a `CANCELED` report with the same shape, so it would now reach the tracked path, emit an
authoritative `OrderCanceled` for the order being modified and drop its identity before the
replacement `NEW` arrives. The HTTP modify fallback relies on that report staying untracked so the
engine's superseded-cancel check can discard it. Each cancel-replace now sends a unique
`cancelNewClientOrderId` (`CR-` plus a UUID, 35 characters) that the execution client records in
`WsDispatchState` before the request goes out, on both the WS and HTTP paths.
`dispatch_execution_report` skips a `CANCELED` report only when its `c` is an id this client
issued, so a standalone cancel from another client cannot be mistaken for one. The replacement
`NEW` then drives `OrderUpdated` through the already-accepted branch that exists for it. The id
is a parameter through `build_cancel_replace_params`, `BinanceSpotHttpClient::modify_order` and
`BinanceRawSpotHttpClient::cancel_replace_order`.

Withholding that report has one failure mode: Binance can cancel the original order and then reject
the replacement (`-2021`), and `OrderModifyRejected` restores the order to accepted and clears its
in-flight entry, so with periodic reconciliation off nothing would ever cancel it locally. Each
cancel-replace outcome is therefore tracked by its cancel ID: pending on send, canceled once the
report arrives, rejected once the venue answers. If the report came first, the rejection handler
on both the WS and HTTP paths emits `OrderModifyRejected` and then `OrderCanceled` carrying the
original venue order ID and the report's timestamp. If the rejection came first, the later report
is no longer withheld and dispatches as a standalone cancel. Ambiguous codes (`-1006`, `-1007`)
leave the outcome pending, since the replacement may have gone through and its `NEW` still drives
`OrderUpdated`.
The withheld report is kept as its parsed `OrderStatusReport`, so an order without a dispatch
identity (recovered after restart or claimed through reconciliation) is closed too: the replay emits
`OrderCanceled` when an identity exists and otherwise sends the status report for reconciliation to
apply, as an untracked cancel report does today.

Closes #4923

## Testing

`test_dispatch_execution_report_canceled_resolves_order_by_orig_client_order_id` feeds a
`CANCELED` report with a generated `c` and the tracked order's ID in `C`, and asserts
`OrderCanceled` for that order with its identity cleared. On `develop` it fails with an
`OrderStatusReport` instead. `test_dispatch_execution_report_cancel_replace_skips_cancel_then_updates_on_new`
registers an issued cancel id, feeds the cancel report carrying it and asserts nothing is emitted and the identity survives, then feeds
the replacement `NEW` and asserts `OrderUpdated` with the new `venue_order_id`. `test_build_cancel_replace_params_sends_cancel_new_client_order_id` builds the params from a
`ModifyOrder` and asserts the field is set and serializes as `cancelNewClientOrderId`, which is the
form the WS handler sends. `test_cancel_replace_order_sends_cancel_new_client_order_id` in the
HTTP integration suite drives `BinanceRawSpotHttpClient::cancel_replace_order` against the mock
server and asserts the query carried `cancelOrderId`, `cancelNewClientOrderId` and
`newClientOrderId`. `test_cancel_replace_rejected_after_confirmed_cancel_emits_canceled` runs both arrival orders
against the same fixture and asserts `OrderModifyRejected` then `OrderCanceled` with the old venue
order ID and the report's `ts_event`, then empty identity and pending maps.
`test_cancel_replace_ambiguous_rejection_keeps_withholding_cancel` covers `-1006` and `-1007`
followed by the replacement `NEW`.
`test_http_modify_failure_replays_withheld_cancel_without_identity` withholds a cancel report with
an empty dispatch state and drives `handle_http_modify_failure`, the function the HTTP modify task
calls on failure, asserting `OrderModifyRejected` then a canceled status report with the order's
client ID, original venue order ID and event time; the exec test server has no user-data push, so
the report cannot be landed between request and rejection end to end without new harness plumbing.
The parse test
covers `C` set, `C` empty as the JSON feed sends it, and `C` absent. The SBE decoder already maps
an empty `origClientOrderId` to `None`, so it is untouched.

`cargo test -p nautilus-binance` passes (1111 lib, 170 spot integration), as do
`cargo clippy -p nautilus-binance --no-deps --all-targets -- -D warnings` and the
`check-formatting-rs` and `fmt` pre-commit hooks on the changed files.

One adjacent observation, not addressed here: the `decode_order` cursor issue also noted in #4923 is
a separate change.

### DIFF
--- a/crates/adapters/binance/src/common/dispatch.rs
+++ b/crates/adapters/binance/src/common/dispatch.rs
@@ -88,6 +88,7 @@ pub struct WsDispatchState {
     replacements: DashMap<ClientOrderId, PendingReplacement>,
     emitted_accepted: Mutex<FifoCache<ClientOrderId, 10_000>>,
     filled_orders: Mutex<FifoCache<ClientOrderId, 10_000>>,
+    cancel_replace_cancel_ids: Mutex<FifoCache<String, 10_000>>,
 }
 
 impl Default for WsDispatchState {
@@ -100,6 +101,7 @@ impl Default for WsDispatchState {
             replacements: DashMap::new(),
             emitted_accepted: Mutex::new(FifoCache::new()),
             filled_orders: Mutex::new(FifoCache::new()),
+            cancel_replace_cancel_ids: Mutex::new(FifoCache::new()),
         }
     }
 }
@@ -121,6 +123,17 @@ impl WsDispatchState {
     /// Marks an order as having received a fill.
     pub fn insert_filled(&self, cid: ClientOrderId) {
         self.filled_orders.lock().add(cid);
+    }
+
+    pub fn is_cancel_replace_cancel_id(&self, id: &str) -> bool {
+        self.cancel_replace_cancel_ids
+            .lock()
+            .contains(&id.to_string())
+    }
+
+    /// Records the `cancelNewClientOrderId` sent with a cancel-replace request.
+    pub fn insert_cancel_replace_cancel_id(&self, id: String) {
+        self.cancel_replace_cancel_ids.lock().add(id);
     }
 
     pub fn insert_algo_order_id(&self, cid: ClientOrderId, venue_order_id: VenueOrderId) {
--- a/crates/adapters/binance/src/common/dispatch.rs
+++ b/crates/adapters/binance/src/common/dispatch.rs
@@ -29,6 +29,7 @@ use nautilus_model::{
     enums::{OrderSide, OrderType},
     events::{OrderAccepted, OrderCanceled, OrderEventAny},
     identifiers::{AccountId, ClientOrderId, InstrumentId, PositionId, StrategyId, VenueOrderId},
+    reports::OrderStatusReport,
     types::{Price, Quantity},
 };
 use parking_lot::Mutex;
@@ -61,10 +62,10 @@ pub struct PendingRequest {
 enum CancelReplaceOutcome {
     Pending,
     /// The cancel report arrived; withheld while the replacement is pending or succeeded.
-    Canceled {
-        venue_order_id: VenueOrderId,
-        ts_event: UnixNanos,
-    },
+    ///
+    /// Parsed at that point so it can be replayed even for an order without a
+    /// dispatch identity, such as one recovered after restart.
+    Canceled(Box<OrderStatusReport>),
     /// The venue rejected the request before any cancel report arrived.
     Rejected,
 }
@@ -151,18 +152,20 @@ impl WsDispatchState {
             .insert(cancel_id, CancelReplaceOutcome::Pending);
     }
 
+    /// Returns `true` when `cancel_id` belongs to a cancel-replace this client issued.
+    pub fn has_cancel_replace(&self, cancel_id: &str) -> bool {
+        self.cancel_replace_outcomes
+            .lock()
+            .contains_key(&cancel_id.to_string())
+    }
+
     /// Records the cancel half's `CANCELED` report for a cancel-replace request.
     ///
     /// Returns `true` when the report must be withheld because the replacement is
     /// pending or succeeded, and `false` when it should dispatch as a standalone
     /// cancel because the venue already rejected the replacement or the ID is not
     /// one this client issued.
-    pub fn on_cancel_replace_canceled(
-        &self,
-        cancel_id: &str,
-        venue_order_id: VenueOrderId,
-        ts_event: UnixNanos,
-    ) -> bool {
+    pub fn on_cancel_replace_canceled(&self, cancel_id: &str, report: OrderStatusReport) -> bool {
         let mut outcomes = self.cancel_replace_outcomes.lock();
         let Some(outcome) = outcomes.get_mut(&cancel_id.to_string()) else {
             return false;
@@ -170,22 +173,19 @@ impl WsDispatchState {
 
         match outcome {
             CancelReplaceOutcome::Pending => {
-                *outcome = CancelReplaceOutcome::Canceled {
-                    venue_order_id,
-                    ts_event,
-                };
+                *outcome = CancelReplaceOutcome::Canceled(Box::new(report));
                 true
             }
-            CancelReplaceOutcome::Canceled { .. } => true,
+            CancelReplaceOutcome::Canceled(_) => true,
             CancelReplaceOutcome::Rejected => false,
         }
     }
 
     /// Records a rejected cancel-replace request.
     ///
-    /// Returns the confirmed cancellation when its report already arrived, so the
-    /// caller can emit `OrderCanceled` for the original order.
-    pub fn on_cancel_replace_rejected(&self, cancel_id: &str) -> Option<(VenueOrderId, UnixNanos)> {
+    /// Returns the withheld cancel report when it already arrived, so the caller
+    /// can emit the confirmed cancellation for the original order.
+    pub fn on_cancel_replace_rejected(&self, cancel_id: &str) -> Option<OrderStatusReport> {
         let mut outcomes = self.cancel_replace_outcomes.lock();
         let outcome = outcomes.get_mut(&cancel_id.to_string())?;
 
@@ -194,10 +194,7 @@ impl WsDispatchState {
                 *outcome = CancelReplaceOutcome::Rejected;
                 None
             }
-            CancelReplaceOutcome::Canceled {
-                venue_order_id,
-                ts_event,
-            } => Some((*venue_order_id, *ts_event)),
+            CancelReplaceOutcome::Canceled(report) => Some((**report).clone()),
             CancelReplaceOutcome::Rejected => None,
         }
     }
--- a/crates/adapters/binance/src/common/dispatch.rs
+++ b/crates/adapters/binance/src/common/dispatch.rs
@@ -22,7 +22,7 @@
 //! - Untracked orders fall back to execution reports for reconciliation.
 
 use dashmap::DashMap;
-use nautilus_common::cache::fifo::FifoCache;
+use nautilus_common::cache::fifo::{FifoCache, FifoCacheMap};
 use nautilus_core::{UUID4, UnixNanos};
 use nautilus_live::ExecutionEventEmitter;
 use nautilus_model::{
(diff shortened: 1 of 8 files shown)


cjdsellers:
Hi @abhijeetvichare76,

Thanks for the PR. The `C` fallback fixes the standalone cancel path, but I found two issues to address before this is safe to merge:

- `dispatch_execution_report` currently treats every `CANCELED` report whose `c` starts with `CR-` as a cancel-replace leg. Another Spot client can use that prefix for a normal cancel, which would cause us to drop the report before resolving the tracked order from `C`. The marker needs to be correlated with the actual cancel-replace operation rather than matched by prefix alone.
- The new dispatch test creates the marker directly, so it does not prove that either production cancel-replace path sends and serializes `cancelNewClientOrderId`. Could you add coverage for both `build_cancel_replace_params` and `BinanceRawSpotHttpClient::cancel_replace_order`?

abhijeetvichare76:
Thanks @cjdsellers, both fair.

The prefix match is gone. Each cancel-replace now generates a unique `cancelNewClientOrderId` (`CR-` plus a UUID) that the execution client records in `WsDispatchState` before sending, on both the WS and HTTP paths, and `dispatch_execution_report` skips a `CANCELED` report only when its `c` is an id this client issued. A standalone cancel from another client with a `CR-` id falls through to the `C` lookup as before. The id is passed down as a parameter through `build_cancel_replace_params`, `BinanceSpotHttpClient::modify_order` and `BinanceRawSpotHttpClient::cancel_replace_order`, so nothing below the execution client invents it.

Coverage added for both production paths: a unit test for `build_cancel_replace_params` asserting the field and its serialized `cancelNewClientOrderId` key, and an integration test driving `BinanceRawSpotHttpClient::cancel_replace_order` against the mock server and asserting the query carried it. The dispatch test now registers the id rather than fabricating one.

cjdsellers:
@abhijeetvichare76 Thanks for addressing the correlation and request coverage.

The partial-failure case needs handling before merge. When Binance cancels the original order but rejects the replacement, the new filter drops the confirmed cancellation. `OrderModifyRejected` then restores the previous order status and clears in-flight tracking, while periodic open-order reconciliation is disabled by default. Previously, that cancel report reached reconciliation through the venue order ID.

Could you preserve the confirmed cancellation when replacement fails and add regression coverage for both arrival orders of the cancel report and rejection? The standalone cancel fix and successful replacement path should remain covered.

cjdsellers:
Hi @abhijeetvichare76,

Thanks for adding the arrival-order coverage. One cancellation case still needs handling before merge: an order recovered after restart or claimed through reconciliation can be in the cache without an entry in `order_identities`.

If its cancel report arrives before a definitive HTTP replacement rejection, the new filter withholds it, but `emit_canceled_after_rejected_replace` then returns because the identity is missing. `OrderModifyRejected` clears in-flight tracking, leaving the cached order open despite the confirmed cancellation.

Could you preserve the cancellation for these orders and add coverage through the HTTP execution path with no dispatch identity?

abhijeetvichare76:
Thanks @cjdsellers . 
The withheld cancel report is now stored as its parsed `OrderStatusReport` rather than just the venue order ID and timestamp.

cjdsellers (approved):
@abhijeetvichare76, thanks for following through on the cancellation fixes and regression coverage. I've had another look at the latest changes, and this looks good to merge. I'll handle some minor logging and documentation follow-ups after merge.

PR #4934, opened 2026-09-05
Guard SortinoRatio against single-observation samples

### LINKED ISSUE #4915: Sortino ratio returns a degenerate value from a single observation where Sharpe returns NaN
﻿## Confirmation

- [x] I've reviewed or skimmed the source code to confirm the behavior is not by design.
- [x] I've tested this issue using a recent pre-release or development wheel and can still reproduce it.

## Expected behavior

`SortinoRatio` should decline to produce a ratio when the sample is too small to estimate dispersion, consistent with the rest of the crate.

`PortfolioStatistic::calculate_std` in `crates/analysis/src/statistic.rs` already establishes that rule:

```rust
fn calculate_std(&self, returns: &Returns) -> f64 {
    let n = returns.len() as f64;
    if n < 2.0 {
        return f64::NAN;
    }
    ...
}
```

`SharpeRatio`, `ReturnsVolatility`, `RiskReturnRatio`, `ReturnsSkewness` and `ReturnsKurtosis` all route through it and return `NaN` on a one-observation sample.

## Actual behavior

`SortinoRatio` computes downside deviation inline and has no minimum-observation guard. Its only validity check is `check_valid_returns`, which is an emptiness test.

`crates/analysis/src/statistics/sortino_ratio.rs`:

```rust
let returns = self.downsample_to_daily_bins(raw_returns);
let total_n = returns.len() as f64;
let mean = returns.values().sum::<f64>() / total_n;

let downside = (returns
    .values()
    .filter(|&&x| x < 0.0)
    .map(|x| x.powi(2))
    .sum::<f64>()
    / total_n)
    .sqrt();

if downside < f64::EPSILON {
    return Some(f64::NAN);
}

let annualized_ratio = (mean / downside) * (self.period as f64).sqrt();
```

Three problems on a single-observation sample.

**1. It answers where the equivalent statistic declines.** Same input: Sharpe returns `NaN`, Sortino returns `-15.8745`.

**2. The value is degenerate.** With one negative return, `mean / downside` is `r / |r| = -1` exactly, so the result is always `-sqrt(252) = -15.8745` regardless of magnitude. A -0.1% day and a -50% day report the same number, so the output carries no information about the input.

**3. Definedness depends on the sign of the single observation.** One positive return gives `downside = 0`, trips the epsilon guard and returns `NaN`. One negative return returns `-15.8745`.

| input | Sharpe | Sortino |
|---|---:|---:|
| `[-0.02]` | NaN | **-15.8745** |
| `[0.02]` | NaN | NaN |

## Steps to reproduce

Self-contained, generates its own data, no market data files needed:

```python
from nautilus_trader.analysis import SharpeRatio, SortinoRatio

T = 1_600_000_000_000_000_000  # any timestamp, one bin
one_negative_day = {T: -0.02}

print("Sharpe :", SharpeRatio().calculate_from_returns(one_negative_day))
print("Sortino:", SortinoRatio().calculate_from_returns(one_negative_day))
print("one positive day:", SortinoRatio().calculate_from_returns({T: 0.02}))

# magnitude is irrelevant to the result
for r in (-0.001, -0.02, -0.5):
    print(f"  {r} -> {SortinoRatio().calculate_from_returns({T: r})}")
```

Actual output:

```
Sharpe : nan
Sortino: -15.874507866387544
one positive day: nan
  -0.001 -> -15.874507866387544
  -0.02 -> -15.874507866387544
  -0.5 -> -15.874507866387544
```

## Code snippets or logs

The existing tests in `sortino_ratio.rs` cover empty returns, all-positive returns (zero downside), a five-element valid case and the name string. There is no small-sample case, which is why this is uncovered.

## Suggested direction

Give `SortinoRatio` the same minimum-observation behaviour as `calculate_std`: return `NaN` for `n < 2`, measured after `downsample_to_daily_bins` since that is the sample the statistic actually uses.

Possibly worth lifting the guard into a shared helper so the crate has one rule rather than per-file rules. `check_valid_returns` currently only tests emptiness, so every statistic that relies on it inherits an n >= 1 contract regardless of how many observations its estimator actually needs.

`OmegaRatio`, `ValueAtRisk` and `ExpectedShortfall` have the same missing guard. For the historical-quantile statistics returning the single observation is more defensible, so I have not raised those as defects here, but flagging in case you want them treated consistently.

Happy to open a PR with the fix and tests if you agree with the approach.

## Specifications

- OS platform: Windows 11
- Python version: 3.12.10
- Installed from: PyPI pre-release wheel
- Adapter/venue: n/a

raunak2007:
- [ ] I ran `make format`, then ran `make pre-commit` locally and confirmed it passed, or I described an agreed limitation below

## Summary

`SortinoRatio::calculate_from_returns` computed a downside deviation from a single daily observation and returned a finite but meaningless value (about -15.87 for one -2% return), while `SharpeRatio` returns NaN through the `n < 2` rule in `calculate_std`. This adds the same guard after daily downsampling, so a single-observation sample yields NaN regardless of sign and the two statistics agree on when dispersion is undefined.

## Related issues/PRs

Closes #4915

## Type of change

- [x] Bug fix (non-breaking)

## Testing

Added `test_single_observation_returns_nan` (rstest cases -0.02 and 0.02). The negative case fails on `develop` and passes with the guard; `cargo test -p nautilus-analysis sortino` is green. Formatting: ran `cargo +nightly fmt` on the crate; `make pre-commit` is not yet run locally (pinned tool install pending on this machine), and I will run it and update this description. The issue notes that OmegaRatio, ValueAtRisk and ExpectedShortfall lack the same guard; kept this PR to Sortino and happy to follow up with a shared minimum-observation check if wanted.

### DIFF
--- a/crates/analysis/src/statistics/sortino_ratio.rs
+++ b/crates/analysis/src/statistics/sortino_ratio.rs
@@ -79,6 +79,12 @@ impl PortfolioStatistic for SortinoRatio {
         }
 
         let returns = self.downsample_to_daily_bins(raw_returns);
+
+        // Match `calculate_std`: a single observation cannot estimate dispersion
+        if returns.len() < 2 {
+            return Some(f64::NAN);
+        }
+
         let total_n = returns.len() as f64;
         let mean = returns.values().sum::<f64>() / total_n;
 
@@ -142,6 +148,17 @@ mod tests {
     fn test_zero_downside_deviation() {
         let ratio = SortinoRatio::new(None);
         let returns = create_returns(&[0.02, 0.03, 0.01]);
+        let result = ratio.calculate_from_returns(&returns);
+        assert!(result.is_some());
+        assert!(result.unwrap().is_nan());
+    }
+
+    #[rstest]
+    #[case(-0.02)]
+    #[case(0.02)]
+    fn test_single_observation_returns_nan(#[case] value: f64) {
+        let ratio = SortinoRatio::new(None);
+        let returns = create_returns(&[value]);
         let result = ratio.calculate_from_returns(&returns);
         assert!(result.is_some());
         assert!(result.unwrap().is_nan());


cjdsellers (approved):
Hi @raunak2007,

Thanks for the PR. This looks good from my side. The guard is applied after daily downsampling, and the focused Sortino tests also pass against the latest `develop`. Keeping this scoped to `SortinoRatio` is the right call here.

raunak2007:
Thanks @cjdsellers. One question so I don't open something unwanted: `OmegaRatio`, `ValueAtRisk` and `ExpectedShortfall` also downsample without going through `calculate_std` (#4915 flagged them). The reporter's view was that returning the single observation is more defensible for the quantile-based ones. If you'd rather have them aligned with Sharpe/Sortino I'm happy to open that; otherwise I'll leave them as they are.

PR #4935, opened 2026-09-05
Upsert general table entries in the Postgres cache

### LINKED ISSUE #4918: PostgreSQL cache logs duplicate key errors during position reconciliation
**Version:** nautilus_trader v2.0.0rc5 (develop branch)
**File:** `crates/infrastructure/src/sql/queries.rs:70`
**Also involves:** `crates/common/src/cache/mod.rs:4980`
**Severity:** Medium — logged as ERROR on every strategy restart; does not crash but fills logs with noise

---

## Description

When a live strategy with PostgreSQL cache restarts, the engine reconciles its state by
re-adding positions to the cache. The `Cache::add_position_inner()` method writes the
position's OMS type to the `general` table via `DatabaseQueries::add()`. This method uses a
plain `INSERT INTO general (id, value) VALUES ($1, $2)` statement. The `general` table has a
`PRIMARY KEY` on `id`, so when the same position key already exists from a previous session,
the INSERT fails with:

```
duplicate key value violates unique constraint "general_pkey"
```

The error is logged at ERROR level but does not crash the engine — `drain_buffer()` catches
the error and continues. However, the error reappears on every restart and pollutes logs.

## Reproduction

1. Start a live strategy with `PostgresCacheConfig`.
2. Open a position (creates a `general` row with key
   `position_oms:<position_id>`).
3. Restart the strategy.
4. During reconciliation, `add_position_inner()` calls `database.add(key, value)` for the
   same position.
5. The INSERT fails — the key already exists from step 2.

**Relevant log line:**

```
[ERROR] DEMO-001.nautilus_infrastructure::sql::cache:
  Error on query: Failed to insert into general table:
  error returned from database: duplicate key value violates unique constraint "general_pkey"
  at line 673
```

## Root Cause

**`queries.rs:70`** — `DatabaseQueries::add()` uses a plain INSERT:

```rust
pub async fn add(pool: &PgPool, key: String, value: Vec<u8>) -> anyhow::Result<()> {
    sqlx::query("INSERT INTO general (id, value) VALUES ($1, $2)")
        .bind(key)
        .bind(value)
        .execute(pool)
        .await
        .map(|_| ())
        .map_err(|e| anyhow::anyhow!("Failed to insert into general table: {e}"))
}
```

**`cache/mod.rs:4980`** — `add_position_inner()` calls this unconditionally:

```rust
if let Some(database) = &mut self.database {
    database.add_position(position)?;
    database.add(key, value)?;  // <-- INSERT, not UPSERT
}
```

The in-memory `self.general.insert(key, value)` on line 4969 is an `IndexMap::insert`
(upsert by design), but the database `add()` is a bare INSERT (fails on duplicate).

## Proposed Fix

Change `DatabaseQueries::add()` in `queries.rs:70` from INSERT to UPSERT:

```rust
// Before (line 70):
sqlx::query("INSERT INTO general (id, value) VALUES ($1, $2)")

// After:
sqlx::query("INSERT INTO general (id, value) VALUES ($1, $2) \
             ON CONFLICT (id) DO UPDATE SET value = EXCLUDED.value")
```

This matches the in-memory semantics (`insert` = upsert) and prevents the duplicate key
error on restart.

Alternatively, `add_position_inner()` could check whether the key already exists in the
database before calling `add()`, but the UPSERT approach is simpler and more robust.

## Impact

- **Log pollution:** ERROR-level messages on every strategy restart.
- **No data loss:** The in-memory cache is updated correctly; only the database write fails.
- **No crash:** `drain_buffer()` catches the error and continues.
- **Affected data:** Position OMS type (`position_oms:<position_id>`) and position snapshots
  (when implemented — see the TODO on line 4972).

## Environment

- **OS:** Windows 10/11 (AMD64)
- **Python:** 3.12+
- **nautilus_trader:** v2.0.0rc5 (develop branch, built with maturin)
- **Database:** PostgreSQL 14+
- **Venue:** Bybit (demo account)
- **Table schema:** `general(id TEXT PRIMARY KEY, value BYTEA)`

raunak2007:
- [ ] I ran `make format`, then ran `make pre-commit` locally and confirmed it passed, or I described an agreed limitation below

## Summary

`DatabaseQueries::add` used a bare INSERT for the `general` table, so re-adding an existing key (for example the entry written from `add_position_inner` when a position is re-added) hit the primary key, and the error was logged in `drain_buffer` rather than surfacing to the caller. The query now uses `ON CONFLICT (id) DO UPDATE SET value = EXCLUDED.value`, matching the insert semantics of the in-memory `Cache::add` and the upsert pattern the other tables in `queries.rs` already use. `DO UPDATE` rather than `DO NOTHING` because a re-add with a changed value must win, as it does in memory.

## Related issues/PRs

Closes #4918

## Type of change

- [x] Bug fix (non-breaking)

## Testing

Added `test_add_general_object_twice_updates_value`, which asserts on the loaded value rather than the return of `add()`, since the return is `Ok` before and after the change (the failure is asynchronous). The Postgres integration tests are Linux-gated; I ran them on macOS with the gate temporarily lifted against the compose Postgres from `make init-services`: the new test fails on `develop` with a 5s timeout and passes with the change, and both `general` tests pass on a truncated database. Formatting: ran `cargo +nightly fmt` on the crate; `make pre-commit` is not yet run locally (pinned tool install pending on this machine), and I will run it and update this description.

### DIFF
--- a/crates/infrastructure/src/sql/queries.rs
+++ b/crates/infrastructure/src/sql/queries.rs
@@ -67,13 +67,16 @@ impl DatabaseQueries {
     ///
     /// Returns an error if the INSERT operation fails.
     pub async fn add(pool: &PgPool, key: String, value: Vec<u8>) -> anyhow::Result<()> {
-        sqlx::query("INSERT INTO general (id, value) VALUES ($1, $2)")
-            .bind(key)
-            .bind(value)
-            .execute(pool)
-            .await
-            .map(|_| ())
-            .map_err(|e| anyhow::anyhow!("Failed to insert into general table: {e}"))
+        sqlx::query(
+            "INSERT INTO general (id, value) VALUES ($1, $2) \
+             ON CONFLICT (id) DO UPDATE SET value = EXCLUDED.value",
+        )
+        .bind(key)
+        .bind(value)
+        .execute(pool)
+        .await
+        .map(|_| ())
+        .map_err(|e| anyhow::anyhow!("Failed to insert into general table: {e}"))
     }
 
     /// Loads all entries from the `general` table via the provided `pool`.
--- a/crates/infrastructure/tests/integration/test_cache_database_postgres.rs
+++ b/crates/infrastructure/tests/integration/test_cache_database_postgres.rs
@@ -100,6 +100,32 @@ mod serial_tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    async fn test_add_general_object_twice_updates_value() {
+        let mut pg_cache = get_test_pg_cache_database().await.unwrap();
+
+        let key = String::from("test_upsert_id");
+        let first = Bytes::from("first_value");
+        let second = Bytes::from("second_value");
+
+        pg_cache.add(key.clone(), first.clone()).unwrap();
+        wait_until(
+            || pg_cache.load().unwrap().get(&key) == Some(&first),
+            Duration::from_secs(5),
+        );
+
+        // Re-adding an existing key must update the stored value, matching the
+        // in-memory cache's insert semantics, rather than fail on the primary key
+        pg_cache.add(key.clone(), second.clone()).unwrap();
+        wait_until(
+            || pg_cache.load().unwrap().get(&key) == Some(&second),
+            Duration::from_secs(5),
+        );
+
+        pg_cache.flush().unwrap();
+        pg_cache.close().unwrap();
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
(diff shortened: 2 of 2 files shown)


cjdsellers:
Hi @raunak2007,

Thanks for the PR. The change and regression test look good :ok_hand:.

Could you update the `DatabaseQueries::add` doc comment in this PR to say that it inserts or replaces a raw key-value entry? The method now overwrites an existing row, so the documentation needs to state that behavior.

PR #4940, opened 2026-09-06
Derive Polymarket locked balances from open orders

yashwardhan-gautam:

## Summary

Polymarket's `GET /balance-allowance` reports only the collateral total, and `parse_balance_allowance` hardcoded `locked` to `Decimal::ZERO`, so every account state reported `free == total` however much collateral was committed to resting orders (#4911). Account state refreshes now fetch the venue's open orders alongside the balance and report
the remaining notional of resting BUY orders as `locked`, with `free = total - locked`. `is_reported` stays `true` and none of this touches core.

Behavior decisions, per the scope agreed on #4916:

- Only BUY orders reserve pUSD; a SELL reserves conditional tokens, which are not carried as a balance.
- Collateral spent on matched-but-unsettled fills has left the open-order set but is not yet reflected in the cached total, so `locked` sits under the venue's real hold while fills settle. This is accepted and documented rather than tracked.
- Refresh cadence is unchanged (finalized trade, reconnect, mass status); `locked` is accurate at each refresh and stale between them.
- If the open-orders request fails, the refresh logs a warning and reports `locked = 0` — the previous behavior — so a flaky orders endpoint cannot block balance updates.
- The sum uses remaining size, `(original_size - size_matched) x price`: the venue holds matched-but-unsettled spend separately, so holding original size as well would double-count the matched portion. I will confirm this against a live partially matched order and follow up on the thread.

The integration guide documents the three behavior notes under a new "Locked balances"
section.

## Related issues/PRs

Related to #4911 (kept open for the core reservation work)
Scope agreed in #4916 (closed)
Related to #4922 (same defect class in the Kraken spot adapter)

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

## Documentation

No PyO3 bindings or wrapped Rust docs were touched, so `make py-stubs` does not apply.

## Testing

New unit tests in `execution::parse`:

- `test_locked_from_open_orders_sums_remaining_buy_notional` — BUY orders summed at remaining size (including a partially matched order); SELL orders contribute nothing.
- `test_locked_from_open_orders_empty_and_overmatched` — empty set yields zero, and a venue-side rounding overshoot of `size_matched` clamps the remainder at zero instead of going negative.
- `test_parse_balance_allowance_with_locked` — four cases covering a partial reservation, none, fully committed, and `locked > total` clamped into `[0, total]`.
- The existing `test_parse_balance_allowance` cases are updated to the new signature and still pin `free == total` when nothing is resting.

One `#[ignore]`d test, `test_locked_from_open_orders_misses_unsettled_matched_spend`,
documents the accepted unsettled-spend limitation with the figures from the #4911 capture
(`balance 66.698599`, `active 48.400000`, `matched 21.949680`): run with `--ignored` it
fails by design, showing `free 18.298599` against a real venue availability of
`-3.651081`. It pins the boundary of this change rather than asserting behavior this
change provides.

`cargo nextest run -p nautilus-polymarket` passes (1929 tests, 2 skipped).

### DIFF
--- a/crates/adapters/polymarket/src/execution/lifecycle.rs
+++ b/crates/adapters/polymarket/src/execution/lifecycle.rs
@@ -229,11 +229,19 @@ impl PolymarketExecutionClient {
     }
 
     pub(super) async fn refresh_account_state(&self) -> anyhow::Result<()> {
+        let user_address = self
+            .secrets
+            .funder
+            .clone()
+            .unwrap_or_else(|| self.secrets.address.clone());
+        let api_key = SecretString::from(self.secrets.credential.api_key_str().to_string());
         fetch_and_emit_account_state(
             &self.http_client,
             &self.emitter,
             self.clock,
             self.config.signature_type,
+            &user_address,
+            api_key.expose_secret(),
         )
         .await
     }
@@ -344,10 +352,17 @@ impl PolymarketExecutionClient {
                             let http = http_client.clone();
                             let emit = emitter.clone();
                             let session_spawner = session_spawner.clone();
+                            let refresh_user_address = user_address.clone();
+                            let refresh_api_key = user_api_key.clone();
 
                             let future = async move {
                                 match fetch_and_emit_account_state(
-                                    &http, &emit, clock, signature_type,
+                                    &http,
+                                    &emit,
+                                    clock,
+                                    signature_type,
+                                    &refresh_user_address,
+                                    refresh_api_key.expose_secret(),
                                 )
                                 .await
                                 {
@@ -375,9 +390,18 @@ impl PolymarketExecutionClient {
 
                         let http = http_client.clone();
                         let emit = emitter.clone();
+                        let refresh_user_address = user_address.clone();
+                        let refresh_api_key = user_api_key.clone();
                         let future = async move {
-                            match fetch_and_emit_account_state(&http, &emit, clock, signature_type)
-                                .await
+                            match fetch_and_emit_account_state(
+                                &http,
+                                &emit,
+                                clock,
+                                signature_type,
+                                &refresh_user_address,
+                                refresh_api_key.expose_secret(),
+                            )
+                            .await
                             {
                                 Ok(()) => {
                                     log::info!("Account state refreshed after WebSocket reconnect");
--- a/crates/adapters/polymarket/src/execution/parse.rs
+++ b/crates/adapters/polymarket/src/execution/parse.rs
@@ -38,7 +38,7 @@ use crate::{
             PolymarketEventType, PolymarketLiquiditySide, PolymarketOrderSide,
             PolymarketOrderStatus,
         },
-        models::PolymarketMakerOrder,
+        models::{PolymarketMakerOrder, is_owned_by_account},
         parse::parse_decimal_exact,
     },
     http::models::{ClobBookLevel, PolymarketOpenOrder, PolymarketTradeReport},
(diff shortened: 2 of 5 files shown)


yashwardhan-gautam:
@cjdsellers 
One data point on the under-hold while fills settle, before I finalise: I replayed the 15 distinct collateral rejects from the #4911 capture against `free = balance − active`, and 14 still pass the local check and fail at the venue. The unsettled matched spend was 33–44% of the balance at those instants.

There's a fix in the same pull shape: get_trades already returns status, size, price and fee_rate_bps, so the same refresh could add the non-terminal BUY spend to locked — stateless, self-correcting, one extra REST call, and happy to gate it behind a config flag defaulting off.

Worth including, or keep it out of scope for this issue?

cjdsellers:
Hi @yashwardhan-gautam,

Thanks for the PR and for replaying the rejects. That shows the unsettled-spend gap accounts for most of the captured rejections; my earlier description of it as a slight under-hold understated the impact.

I'd keep this PR focused on open-order reservations and take unsettled spend as a separate adapter follow-up. The REST approach is worth investigating, but it needs to establish when the cached balance includes each fill, correctly select the account's maker and taker BUY legs, and bound pagination without losing older pending trades. A default-off flag doesn't settle those accounting questions.

There is one correctness issue to address here: the new sum includes every BUY row returned by `get_orders`, while reconciliation filters those same rows by account ownership. The existing tests cover foreign rows, which would now inflate this account's locked balance and could cause valid orders to be denied. The balance calculation needs the same ownership rule.

The ignored test also asserts a condition that remains impossible even with full unsettled-spend accounting, because free is clamped to zero. A passing test of the documented limitation would be more useful. The refresh documentation should list connect and `QueryAccount` in place of mass status, and the refresh path needs exact emitted-balance coverage for success and the orders-request failure fallback.

If you're able, the live partially matched order check would also close the remaining venue assumption.

yashwardhan-gautam:
> Hi @yashwardhan-gautam,
> 
> Thanks for the PR and for replaying the rejects. That shows the unsettled-spend gap accounts for most of the captured rejections; my earlier description of it as a slight under-hold understated the impact.
> 
> I'd keep this PR focused on open-order reservations and take unsettled spend as a separate adapter follow-up. The REST approach is worth investigating, but it needs to establish when the cached balance includes each fill, correctly select the account's maker and taker BUY legs, and bound pagination without losing older pending trades. A default-off flag doesn't settle those accounting questions.
> 
> There is one correctness issue to address here: the new sum includes every BUY row returned by `get_orders`, while reconciliation filters those same rows by account ownership. The existing tests cover foreign rows, which would now inflate this account's locked balance and could cause valid orders to be denied. The balance calculation needs the same ownership rule.
> 
> The ignored test also asserts a condition that remains impossible even with full unsettled-spend accounting, because free is clamped to zero. A passing test of the documented limitation would be more useful. The refresh documentation should list connect and `QueryAccount` in place of mass status, and the refresh path needs exact emitted-balance coverage for success and the orders-request failure fallback.
> 
> If you're able, the live partially matched order check would also close the remaining venue assumption.

all addressed, rebased on `develop`

cjdsellers (dismissed):
Hey @yashwardhan-gautam,

Thanks for addressing the feedback. I've had another look at the latest changes, and this looks good to merge.

If you're able, could you share the live partially matched order result to confirm the remaining-size assumption? That would close the remaining validation follow-up.

yashwardhan-gautam:
> Hey @yashwardhan-gautam,
> 
> Thanks for addressing the feedback. I've had another look at the latest changes, and this looks good to merge.
> 
> If you're able, could you share the live partially matched order result to confirm the remaining-size assumption? That would close the remaining validation follow-up.

Thanks for the second look.

**Partially matched order check** — answered from the #4911 capture. Replaying its submits,
fills and cancels reconstructs the open BUY set at each of the 15 rejects; 7 match the
payload's `sum of active orders` exactly, which validates the method. At `01:52:41.582` a
partially matched BUY was resting:

```
resting BUY O-…-115:   qty 5.000000, filled 1.402595 @ 0.23  →  remaining 3.597405

venue active hold (payload)              24.025700
− 16 other resting BUYs, none filled     23.200000   (size × price, unambiguous)
= venue's hold on this one order          0.825700

if the venue holds remaining size:  3.597405 × 0.23 = 0.827403   Δ +0.0017
if the venue holds original size:   5.000000 × 0.23 = 1.150000   Δ +0.3243
```

The venue holds **remaining** size, so the sum here is correct. The 0.0017 is the venue
truncating remaining size to 0.01 shares

**Open-orders request frequency** — agreed. The capture shows 126 fills in 530 s (≈14/min,
peak 76/min), each triggering a refresh and now a `GET /data/orders`. Two ways to take it
off the hot path, either doable in this PR:

1. Post-trade refreshes derive `active` from `cache.orders_open()` (zero HTTP; the cache
   already reflects the fill and holds only our orders); keep `/data/orders` for connect,
   reconnect and `QueryAccount`.
2. Debounce: reuse the last open-order sum if fetched within N ms. Simpler; bursts leave
   `locked` briefly stale in the over-locking direction.

One heads-up, not for this PR: once `free` is realistic, the cash-account SELL check
(`cum_notional_sell > free`) becomes live for non-reducing SELLs, and for instruments with no
base currency it compares token notional against pUSD — the same assumption as
`base_calculate_balance_locked`. I'll file that separately.

cjdsellers:
Hey @yashwardhan-gautam,

Thanks for the remaining-size replay and the request-rate figures. I've implemented the Polymarket fix in 3cc6c9d909a3034b259433606272f77e35898b53 and am closing this PR as superseded. Your work on #4940 is credited in the release notes.

Your original solution fetched open orders alongside each balance refresh. The frequency you measured reinforces why I revisited that approach. I've taken the local-calculation direction you suggested and applied it to all balance refreshes: the adapter maintains its own reservation map using processed order events, including reconciliation updates. We need to be able to rely on locally maintained order data for this calculation, so refreshes now account for accepted open BUY remaining notional without additional open-order requests.

Your cash-account SELL-check warning remains relevant and needs a separate follow-up. This change does not address that check, matched-but-unsettled spending, or the broader shared-accounting issue in #4911.

I've validated this live. If you're able, could you also test it with your strategy and compare the reported locked/free balances while orders are resting and partially filled? That would be useful confirmation against the behavior you originally captured.

yashwardhan-gautam:
> Hey @yashwardhan-gautam,
> 
> Thanks for the remaining-size replay and the request-rate figures. I've implemented the Polymarket fix in [3cc6c9d](https://github.com/nautechsystems/nautilus_trader/commit/3cc6c9d909a3034b259433606272f77e35898b53) and am closing this PR as superseded. Your work on #4940 is credited in the release notes.
> 
> Your original solution fetched open orders alongside each balance refresh. The frequency you measured reinforces why I revisited that approach. I've taken the local-calculation direction you suggested and applied it to all balance refreshes: the adapter maintains its own reservation map using processed order events, including reconciliation updates. We need to be able to rely on locally maintained order data for this calculation, so refreshes now account for accepted open BUY remaining notional without additional open-order requests.
> 
> Your cash-account SELL-check warning remains relevant and needs a separate follow-up. This change does not address that check, matched-but-unsettled spending, or the broader shared-accounting issue in #4911.
> 
> I've validated this live. If you're able, could you also test it with your strategy and compare the reported locked/free balances while orders are resting and partially filled? That would be useful confirmation against the behavior you originally captured.

will do, thanks

PR #4942, opened 2026-09-08
Preserve queue positions across book snapshots

cuishuang:
# Pull Request

- [ ] A maintainer agreed on the problem and approach in an issue, or this is a small,
  and, if I used AI,
  this draft
  limitation below

## Summary

Book snapshots and depth10 replacements reset tracked queue positions to zero. This makes existing
limit orders immediately fill-eligible after a snapshot, even when visible liquidity remains ahead.

This change rebases queue positions after the replacement using the previous estimate capped by the
new visible quantity. L3/MBO queues additionally retain previously tracked orders that are still
present in the replacement snapshot.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Documentation

## Testing

### DIFF
--- a/crates/execution/src/matching_engine/mod.rs
+++ b/crates/execution/src/matching_engine/mod.rs
@@ -129,6 +129,7 @@ pub struct OrderMatchingEngine {
     queue_pending: IndexMap<ClientOrderId, PriceRaw>,
     queue_ahead_orders: IndexMap<ClientOrderId, IndexMap<OrderId, QuantityRaw>>,
     queue_ahead_total: IndexMap<ClientOrderId, (PriceRaw, QuantityRaw)>,
+    queue_snapshot_in_progress: bool,
     queue_ids_by_price: IndexMap<PriceRaw, IndexSet<ClientOrderId>>,
     queue_excess: IndexMap<ClientOrderId, QuantityRaw>,
     queue_id_scratch: Vec<ClientOrderId>,
@@ -224,6 +225,7 @@ impl OrderMatchingEngine {
             queue_pending: IndexMap::new(),
             queue_ahead_orders: IndexMap::new(),
             queue_ahead_total: IndexMap::new(),
+            queue_snapshot_in_progress: false,
             queue_ids_by_price: IndexMap::new(),
             queue_excess: IndexMap::new(),
             queue_id_scratch: Vec::new(),
@@ -291,6 +293,7 @@ impl OrderMatchingEngine {
         self.queue_pending.clear();
         self.queue_ahead_orders.clear();
         self.queue_ahead_total.clear();
+        self.queue_snapshot_in_progress = false;
         self.queue_ids_by_price.clear();
         self.queue_excess.clear();
         self.queue_id_scratch.clear();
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @cuishuang,

Thanks for the PR. I found two cases that need addressing before merge:

- The single-delta path misses snapshot completion when only the opening clear carries `F_SNAPSHOT` and the final add carries `F_LAST`. The new condition requires both flags on the same record, so the old quantity ahead remains even when the replacement removes the level, blocking fills. The Tardis loader is one existing producer of this sequence.
- L3 rebasing can discard an order that remains ahead. Starting with A=5 and B=5, a replacement showing A=10 and B=5 assigns the entire capped quantity to A and drops B from tracking. Deleting A then leaves zero ahead while B is still present, allowing premature fills.

Could you cover both cases with regression tests? The new tests currently exercise only L2 and put both flags on the final add.

It would also be useful to document the snapshot and depth10 replacement rule under “Book changes” in the trade-execution guide.

cuishuang:
> Hi @cuishuang,
> 
> Thanks for the PR. I found two cases that need addressing before merge:
> 
> * The single-delta path misses snapshot completion when only the opening clear carries `F_SNAPSHOT` and the final add carries `F_LAST`. The new condition requires both flags on the same record, so the old quantity ahead remains even when the replacement removes the level, blocking fills. The Tardis loader is one existing producer of this sequence.
> * L3 rebasing can discard an order that remains ahead. Starting with A=5 and B=5, a replacement showing A=10 and B=5 assigns the entire capped quantity to A and drops B from tracking. Deleting A then leaves zero ahead while B is still present, allowing premature fills.
> 
> Could you cover both cases with regression tests? The new tests currently exercise only L2 and put both flags on the final add.
> 
> It would also be useful to document the snapshot and depth10 replacement rule under “Book changes” in the trade-execution guide.

Thanks for the detailed feedback. Both cases are now addressed.

- The single-delta path now tracks an in-progress snapshot, so a snapshot opened by a `F_SNAPSHOT` clear is rebased when a later delta carries only `F_LAST`.

- L3 rebasing now preserves each previously tracked order independently using its order ID and `min(previous_size, snapshot_size)`. This keeps B tracked when A grow  from 5 to 10.

cuishuang:
It seems that the CI failure is unrelated to this change

cjdsellers:
@cuishuang, thanks for addressing the two earlier cases. The CI failures are deterministic consequences of this change: both cases of `test_l3_queue_batch_snapshot_clears_prior_depth` still expect the snapshot to clear the queue ahead. With the new behavior, 8 units remain ahead, so the first 2-unit trade produces no fill. Only the later 3-unit trade fills after the book order is deleted, which matches the CI result.

The test needs to assert the intended preservation behavior before this can merge, including the order remaining open with 3 units filled. The current expectations of two fills and a fully filled order describe the old behavior.

cuishuang:
> @cuishuang, thanks for addressing the two earlier cases. The CI failures are deterministic consequences of this change: both cases of `test_l3_queue_batch_snapshot_clears_prior_depth` still expect the snapshot to clear the queue ahead. With the new behavior, 8 units remain ahead, so the first 2-unit trade produces no fill. Only the later 3-unit trade fills after the book order is deleted, which matches the CI result.
> 
> The test needs to assert the intended preservation behavior before this can merge, including the order remaining open with 3 units filled. The current expectations of two fills and a fully filled order describe the old behavior.

@cjdsellers Thanks. I updated the regression test expectations to reflect the intended queue-preservation behavior. Both cases now assert that the 2-unit trade after the snapshot does not fill, the later 3-unit trade fills after the book order is deleted, and the simulated order remains PartiallyFilled with 3 units filled and 2 units remaining.

Please review it again.

cjdsellers (approved):
@cuishuang, thanks for updating the regression tests. I've had another look at the latest changes, and this looks good from my side :+1:.

I'll follow up and document the L3 snapshot rule in the trade-execution guide: only surviving tracked order IDs remain ahead, each capped at its previous quantity.

cjdsellers (approved):

PR #4954, opened 2026-09-08
Fix risk engine account resolution for broker-routed instruments

### LINKED ISSUE #4946: Risk engine silently skips all account-scoped pre-trade checks for broker-routed instruments
# Bug Report

## Confirmation

Tested on `2.0.0rc4` (the latest published RC). The relevant code is unchanged on `develop` HEAD at the time of writing, so this is not already fixed.

## Expected behavior

Account-scoped pre-trade risk checks (`max_notional_per_order`, free balance, cumulative notional) apply to orders for broker-routed instruments, where the account is registered under the broker venue and the instrument carries a routing venue or exchange MIC. With Interactive Brokers this means an account under venue `IB` and instruments under venue `SMART`.

## Actual behavior

Every account-scoped check is silently skipped for such instruments. Observed live against an IB paper account (instruments `AAPL=STK.SMART`, account `IB-DU...`):

* `LiveRiskEngineConfig(max_notional_per_order={"AAPL=STK.SMART": "10"})` did not deny a roughly $307 order.
* A roughly $153M notional order (500,000 AAPL) passed local risk untouched.

The only diagnostic is a `log::debug!`, so at the default `INFO` level an operator sees nothing at all. The configured limits appear active while enforcing nothing.

Note that one check in that branch does still fire: `MARKET_PRICE_UNAVAILABLE` is raised for `MARKET` / `MarketToLimit` orders with no price. So the failure is specifically of the account-scoped checks, not of the whole method.

## Steps to reproduce

1. Configure an IB execution client so the account registers under venue `IB`, and load instruments with `SymbologyMethod.RAW` so they carry venue `SMART`.
2. Set `LiveRiskEngineConfig(max_notional_per_order={"AAPL=STK.SMART": "10"})`.
3. Submit a BUY 1 AAPL limit order, whose notional is far above that cap.
4. The order is not denied. It is submitted to the venue.

## Code snippets or logs

`crates/risk/src/engine/mod.rs`, in `check_orders_risk_for_account` (`develop` HEAD, line 1213; `v2.0.0rc4`, line 1206):

```rust
// Get account for risk checks: use explicit account_id if provided, otherwise venue lookup
let resolved_account = {

    if let Some(account_id) = account_id {
        cache.account(&account_id).map(|account| account.clone_without_events())
    } else {
        cache
            .account_for_venue(&instrument.id().venue)   // venue SMART -> miss
            .map(|account| account.clone_without_events())
    }
};

let Some(mut account) = resolved_account else {
    log::debug!(
        "Cannot find account for venue {} (account_id={account_id:?})",
        instrument.id().venue
    );

    for (&order, price) in orders.iter().zip(&market_prices) {
        if matches!(order, OrderAny::Market(_) | OrderAny::MarketToLimit(_)) && price.is_none() {
            self.deny_no_market_price(instrument.id(), order);
            return false;
        }
    }

    return true;   // fail-open: all account-scoped checks skipped
};
```

## Relationship to #4450 / #4451

This is the same venue-identity mismatch as #4450, which was fixed for the portfolio in #4451 by centralizing resolution in `resolve_account_for_instrument`. That helper currently lives in `crates/portfolio/src/portfolio.rs` and the risk engine was not updated to use it.

The impact differs in kind, which is why this seems worth a separate report. #4450 failed loudly, with an ERROR per fill and visibly broken analytics. Here a safety control fails open and silently.

## Suggested fix

1. Resolve the account using the same fallback logic as #4451, ideally by sharing `resolve_account_for_instrument` rather than duplicating it.
2. Reconsider the unresolved branch. `return true` is fail-open on a safety control. At minimum this should log at `warn!` so it is visible at default log levels. Denying, or making the behavior explicit via config, would be safer still.

I am happy to prepare a PR mirroring #4451, whose additions to `crates/portfolio/tests/portfolio.rs` are a ready template for the test.

## Specifications

- Python version: 3.13.15
- Installed from: PyPI wheel
- Adapter/venue: Interactive Brokers (TWS paper, API v178)

Artur-Sulej:
**TL;DR:** The risk engine looked up the account by the instrument's venue, so for broker-routed instruments (account under `IB`, instrument under `SMART`) the lookup missed and every account-scoped pre-trade check was silently skipped while appearing configured. This adds an ordered fallback for resolving the account, hoists the shared venue and position resolution onto `Cache` so the portfolio and risk engine use one implementation, and raises the unresolved-branch log from DEBUG to WARN so a fail-open is visible. Fixes #4946.

## Summary

`check_orders_risk_for_account` resolved the account with `cache.account_for_venue(&instrument.id().venue)`. For broker-routed instruments the account is registered under the broker venue while the instrument carries the routing venue or exchange MIC, so the lookup missed and the method returned `true` with only a `log::debug!`, skipping `max_notional_per_order`, free-balance and cumulative-notional checks entirely.

Resolution now falls back, in order, to the routing execution client's account, the account owning an existing position, and the sole registered account. The fail-open branch is unchanged in behaviour but now logs at WARN, since risk limits silently ceasing to apply was not observable at the default log level.

The second commit hoists the portfolio's private `resolve_account_for_instrument` onto `Cache` as `account_for_instrument`, so the venue and position resolution has one implementation instead of two. The portfolio's resolution order is unchanged, and its call sites now delegate to the shared method.

The position fallback alone is not sufficient here, which is why this is not a straight reuse of #4451: a pre-trade order usually has no `account_id` and no position yet, so the first order of a session, exactly when the limits should apply, still missed.

## Related issues/PRs

Closes #4946.

Related to #4450 and #4451, which fixed the same venue-identity mismatch for the portfolio. That fix did not extend to the risk engine, and the failure mode differs: #4450 failed loudly with an ERROR per fill, whereas this fails open and silently on a safety control.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

Three tests in `crates/risk/tests/risk_engine.rs`:

- `test_max_notional_denies_order_for_broker_routed_instrument` is the regression test. It builds `AAPL.SMART` with an account under `IB-DU123456` and submits an order with no explicit `client_id`, matching default routing. Verified to fail on the unpatched engine (no events emitted, the order passes risk untouched) and pass with the fix.
- `test_account_scoped_checks_still_skipped_when_no_account_registered` pins the fail-open behaviour when nothing is resolvable, so this is not a silent tightening.
- `test_account_sole_returns_none_when_multiple_accounts_registered` covers the new `Cache::account_sole` boundary at zero, one and two accounts.

Two tests in `crates/common/src/cache/tests.rs` cover the new `Cache::account_for_instrument` directly: the explicit-`account_id` and venue-miss paths, and resolution via the position-owning account when the instrument's own venue has no account.

`cargo test -p nautilus-risk -p nautilus-portfolio` passes (698 and 229), and `cargo test -p nautilus-common --lib cache::` passes (442). `cargo clippy` is clean on the three touched crates and `cargo +nightly fmt` reports no changes.

Two notes for reviewers:

The sole-account fallback is a heuristic, and it is the step that actually fixes the reported case under default routing. It only runs after the venue, client and position lookups have all missed, and it cannot cause a spurious denial: every downstream balance check is `Option`-gated on the account's currency, so an unrelated account yields the same skipped checks as today rather than a rejection. Happy to drop it if you would rather the fix stay limited to precise resolution.

The refactor commit touches the portfolio, so it is worth reviewing separately. It is intended to be behaviour-preserving there: `Cache::account_for_instrument` has the same resolution order as the helper it replaces, and the portfolio suite passes unchanged (229 tests). The two venue-scoped fallbacks elsewhere in `portfolio.rs` were deliberately left alone, since they resolve by venue rather than by instrument.

### DIFF
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -8338,6 +8338,37 @@ impl Cache {
             .map(|account_cell| account_cell.borrow().clone())
     }
 
+    /// Returns a borrow of the account for `instrument_id` (if found).
+    ///
+    /// A plain venue lookup is not sufficient for broker-routed instruments: the account is
+    /// registered under the broker venue (e.g. `IB`) while the instrument carries the routing
+    /// venue or exchange MIC (e.g. `SMART`, `IBIS`). Resolution order:
+    ///
+    /// 1. The explicit `account_id`, when provided.
+    /// 2. Exact venue match on the instrument's venue.
+    /// 3. The account owning an existing position for the instrument.
+    #[must_use]
+    pub fn account_for_instrument(
+        &self,
+        instrument_id: &InstrumentId,
+        account_id: Option<&AccountId>,
+    ) -> Option<AccountRef<'_>> {
+        match account_id {
+            Some(account_id) => self.account(account_id),
+            None => self
+                .account_for_venue(&instrument_id.venue)
+                .or_else(|| self.account_owning_position(instrument_id)),
+        }
+    }
+
+    /// Returns a borrow of the account owning an existing position for `instrument_id`.
+    fn account_owning_position(&self, instrument_id: &InstrumentId) -> Option<AccountRef<'_>> {
+        self.positions(None, Some(instrument_id), None, None, None)
+            .into_iter()
+            .next()
+            .and_then(|position| self.account(&position.account_id))
+    }
+
     /// Returns a borrow of the sole account in the cache, when exactly one is registered.
     ///
     /// Useful as a last-resort resolution for broker-routed instruments, where the account is
--- a/crates/common/src/cache/mod.rs
+++ b/crates/common/src/cache/mod.rs
@@ -8338,6 +8338,23 @@ impl Cache {
             .map(|account_cell| account_cell.borrow().clone())
     }
 
+    /// Returns a borrow of the sole account in the cache, when exactly one is registered.
+    ///
+    /// Useful as a last-resort resolution for broker-routed instruments, where the account is
+    /// registered under the broker venue while the instrument carries a routing venue or
+    /// exchange MIC, so a venue lookup misses. Returns `None` when zero or multiple accounts
+    /// are registered, since the choice would then be ambiguous.
+    #[must_use]
+    pub fn account_sole(&self) -> Option<AccountRef<'_>> {
+        if self.accounts.len() != 1 {
+            return None;
+        }
+        self.accounts
+            .values()
+            .next()
+            .map(|account_cell| AccountRef::new(account_cell.borrow()))
+    }
+
     /// Returns a reference to the account ID for the `venue` (if found).
     #[must_use]
     pub fn account_id(&self, venue: &Venue) -> Option<&AccountId> {
--- a/crates/common/src/cache/tests.rs
+++ b/crates/common/src/cache/tests.rs
@@ -48,6 +48,7 @@ use nautilus_model::{
         AccountState, OrderAccepted, OrderCancelRejected, OrderCanceled, OrderEmulated,
         OrderEventAny, OrderFilled, OrderPendingCancel, OrderRejected, OrderReleased,
         OrderSnapshot, OrderSubmitted, OrderUpdated,
+        account::stubs::cash_account_state_million_usd,
         order::spec::{
             OrderCanceledSpec, OrderEmulatedSpec, OrderFilledSpec, OrderReleasedSpec,
             OrderUpdatedSpec,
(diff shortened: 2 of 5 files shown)


cjdsellers:
Hi @Artur-Sulej,

Thanks for the PR. Account selection needs to match the account that executes the order before this can merge. The existing IB factory supports client `IB-TEST` with account `IB-001`, so treating the client name as a venue can still bypass the checks. Multiple accounts sharing an issuer can also select the wrong balance. The sole-account fallback has a similar problem: an unrelated USD account can reject a USD order, so currency gating does not make that heuristic safe.

There is also a panic path: `MY:CLIENT` is a valid `ClientId`, but `Venue::new` rejects it with `defi` enabled. If you're able to revise this, regression coverage should exercise these cases and assert the specific denial reason.

cjdsellers:
Hi @Artur-Sulej,

Checking in on this one. Are you intending to revise the account resolution along the lines of my earlier feedback, or would it help to talk through the approach first? If time is short at the moment, no pressure, it would just be good to know whether to keep this one active.

cjdsellers:

PR #4959, opened 2026-09-09
Refactor Parquet catalogs and Arrow encoding

faysou:
### Summary

Parquet catalogs and streaming writers use shared persistence interfaces, typed query batches, and named factories. Catalog files store prices, timestamps, enums, JSON, and variable-depth books in open Arrow representations. The `nautilus catalog migrate-parquet` command converts supported existing catalogs into a separate destination while preserving the source.

Parquet remains the default and only built-in catalog backend. Storage uses native `object_store`, and runtime consumers adapt the catalog interfaces to the existing backtest and data-engine paths.

#### Shared catalog and writer interfaces

`CatalogReader` owns queries, sessions, identifiers, and coverage. `CatalogWriter` owns data, instrument, record, and empty-coverage writes; `DataCatalog` combines both. Query objects name filters explicitly, while unsupported optional capabilities return a distinguishable error. Catalog and writer registries reject duplicate factory names and support external Rust implementations.

Parquet implementation code moves under `backend/parquet`. Shared storage, Arrow conversion, query sessions, and writer lifecycle code have their own modules. Compatibility modules retain access through earlier persistence paths where provided.

```mermaid
flowchart TD
    Config[Backtest and live configuration] --> Factories[Catalog and writer factories]
    Factories --> Reader[CatalogReader]
    Factories --> Writer[Streaming writer]
    Reader --> Batches[DataBatch queries]
    Batches --> Adapt[Batch-to-row conversion]
    Adapt --> Runtime[Existing replay and data paths]
    Writer --> Parquet[Parquet catalog]
    Parquet --> Reader
```

The catalog boundary changes while existing runtime consumers retain their execution model.

Source: [catalog contracts](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/catalog/traits.rs), [factory registration](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/backend/mod.rs), and [Parquet modules](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/backend/parquet/mod.rs).

#### Typed batches and query sessions

`DataBatch` represents one family of rows, including instruments and custom data. `BatchView<T>` can retain a shared vector and a range, and `DataRef` provides borrowed row access. `for_each_data_type!` maps concrete types, selectors, batch variants, and storage prefixes in one place. Data, record, and concrete instrument selectors also drive Python queries and serialization.

Parquet core-data sessions merge ordered typed pages and feed `TypedDataBatchSession`.
Typed page decoders stop at the first query or decoding error, so the asynchronous producer never
polls a failed stream again. The chunker extends a batch through rows sharing the boundary timestamp, so its size is a target rather than a hard row or byte bound. Instrument and custom-data fallbacks collect their results before chunking. Parquet rejects historical snapshot selection through `as_of`.

The backtest node converts returned batches through `to_data_vec_for_compat` before the existing replay iterator consumes them. This PR does not claim end-to-end borrowed replay or a new historical execution state machine.

```mermaid
flowchart LR
    Files[Parquet files] --> Pages[Ordered typed pages]
    Pages --> Session[Timestamp-aligned session]
    Session --> Batch[DataBatch]
    Batch --> Convert[to_data_vec_for_compat]
    Convert --> Replay[Existing backtest iterator]
```

Typed storage batches convert into rows at the existing replay boundary.

Source: [batch types](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/model/src/data/mod.rs), [session chunking](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/catalog/session.rs), and [backtest loading](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/backtest/src/node.rs).

`write_data_enum` and `CatalogWriter::write_data_batch` share delta-batch flattening. They preserve
per-instrument row order and boundary flags, and report unsupported batch types as errors.

#### Catalog workers, coverage, and storage

`CatalogWorker` serializes access to a catalog and owns sessions behind opaque IDs. Its bounded command queue applies backpressure, completed sessions close automatically, and `flush` reports accumulated asynchronous write errors. Query forks isolate per-query state while sharing backend storage resources.

`ClosedInterval` replaces the interval-tree dependency for inclusive coverage arithmetic. Parquet continues to derive catalog coverage from files and their ranges, including zero-length files that record a known-empty interval. Identifier discovery, metadata queries, and instrument-family queries use the shared catalog surface.

`StorageBackend` normalizes local and remote locations, retains native object-store options, and supplies the object-store handle used by DataFusion and Parquet. Shared runtime entry points support synchronous persistence operations without each caller constructing its own runtime bridge.

```mermaid
sequenceDiagram
    participant Caller
    participant Worker as CatalogWorker
    participant Session as Parquet session
    Caller->>Worker: open query
    Worker->>Session: create typed session
    Worker-->>Caller: session ID
    loop pull while data remains
        Caller->>Worker: pull session ID
        Worker->>Session: next batch
        Session-->>Worker: batch or completion
        Worker-->>Caller: result
    end
    Caller->>Worker: close session ID
```

The worker retains session ownership and returns data or errors to its caller.

Source: [catalog worker](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/catalog/worker.rs), [coverage](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/backend/parquet/catalog/coverage.rs), and [native storage](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/common/storage.rs).

#### Open Arrow representation

Catalog encoders replace build-dependent binary price and quantity columns with `Decimal128(38, 16)`. Encoding and decoding check range and exactness; undefined values use Arrow nulls, and invalid sentinels fail encoding. Instrument precision metadata remains available for reconstruction. Nanosecond instants use `Timestamp(Nanosecond, Some("UTC"))` with UTC semantics, preserving timestamps that differ within one microsecond. Durations remain integers.

Runtime catalog queries use the final Arrow schema directly. Only explicit migration normalizes
supported older formats; ordinary queries provide no backward- or mixed-format compatibility.
The depth fixture used by runtime catalog tests is already stored in the final Arrow format.
Arrow readers and Nautilus queries retain exact nanoseconds. DuckDB's `TIMESTAMPTZ` query type has
microsecond precision, so direct DuckDB views can truncate sub-microsecond values.

| Value                   | Stored representation                | Relevant behavior                                        |
| ----------------------- | ------------------------------------ | -------------------------------------------------------- |
| Prices and quantities   | `Decimal128(38, 16)`                 | Exact scale and range checks; nulls for undefined values |
| Instant timestamps      | `Timestamp(Nanosecond, Some("UTC"))` | UTC epoch nanoseconds with an explicit UTC annotation    |
| Market-data enums       | Dictionary-encoded names             | Readable without a Nautilus enum lookup                  |
| JSON fields             | UTF-8 with `arrow.json` metadata     | Parquet JSON logical annotation                          |
| Money in custom schemas | Amount and currency struct           | Decimal amount and dictionary currency                   |
| Depth sides             | Lists of level structs               | Price, size, count, and order ID per level               |

Borrowed encoders accept owned values or references without first building a temporary reference vector. Shared column readers centralize accepted Arrow physical types and retain decode errors. Native schema validation rejects opaque binary representations, with explicitly supported custom byte fields as an exception. Files do not store a cluster key.

```mermaid
flowchart LR
    Prices[Prices and quantities] --> Decimal[Decimal128 columns]
    Time[UTC instants] --> Nanos[Nanosecond timestamps]
    Enums[Model enums] --> Names[Dictionary names]
    JSON[JSON fields] --> Annotation[arrow.json]
    Decimal --> Parquet[Open Parquet files]
    Nanos --> Parquet
    Names --> Parquet
    Annotation --> Parquet
```

External readers see standard Arrow values instead of Nautilus-specific fixed-width binary integers.
JSON-backed record decoders match columns by name, so reordered columns retain their values.
Encoders reject duplicate field specifications, and decoders reject duplicate specifications or column names.

Source: [Arrow codecs](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/serialization/src/arrow/mod.rs), [JSON field definitions](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/serialization/src/arrow/json.rs), and [instrument encoders](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/serialization/src/arrow/instrument/mod.rs).

#### Variable-depth Arrow books

`OrderBookDepth` stores variable-length sides with inline capacity for ten levels. Arrow and Cap'n Proto retain the supplied depth; raw Arrow also stores supplied order IDs. `OrderBookDepth10` remains a compatibility name, and affected adapter, cache, actor, backtest, and Python call sites accommodate the model's owned variable-length storage.

The legacy C FFI and SBE formats still require exactly ten levels per side. Their conversion or encoding checks reject incompatible depths. The display encoder returns nested `bids` and `asks` lists containing every level, including price, size, count, and exact integer order ID. Empty sides remain empty lists, and rows with different level counts share the same Arrow list schema. Prices and sizes use `Float64`; raw Arrow remains the choice for exact decimal values.

Old develop depth files omit order IDs. Migration preserves the zero IDs returned by the old reader; it cannot recover information absent from those files. Subsequent writes preserve IDs supplied by the model.

```mermaid
flowchart TD
    Depth[OrderBookDepth] --> Arrow[Arrow level lists]
    Depth --> Capnp[Cap'n Proto lists]
    Depth --> Legacy{Exactly ten levels?}
    Legacy -->|Yes| Fixed[Legacy FFI and SBE]
    Legacy -->|No| Error[Conversion or encoding error]
    Depth --> Display[Nested display lists]
    Arrow --> All[All levels and supplied order IDs]
```

The model and raw storage support variable depth while legacy wire formats retain their fixed-depth contract.

Source: [depth model](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/model/src/data/depth.rs), [FFI checks](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/model/src/ffi/data/depth.rs), [SBE checks](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/serialization/src/sbe/market/book.rs), and [display encoder](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/serialization/src/arrow/display/depth.rs).

The legacy C constructor preserves all ten supplied slots, including padding, counts, and metadata.
General Rust-to-FFI conversion still requires exactly ten levels per side.

#### Custom-data generation and registration

The former persistence proc-macro crate becomes `nautilus-macros`, with separate model and Arrow expansion layers. `nautilus_model::custom_data` generates model traits, JSON conversion, and optional core Python methods. `nautilus_serialization::arrow_custom_data` adds schemas, batch codecs, and optional PyArrow methods. Model-only custom types therefore do not need Arrow dependencies.

Arrow mappings cover prices, quantities, decimals, Money, timestamps, and explicitly marked native enums. Serde fields use JSON, and custom byte fields remain an explicit option. Rust callers register custom Arrow codecs before catalog or streaming use; Python `customdataclass` supplies its corresponding schema and conversion path. Adapter custom types adopt the split macros, and Databento registers its Arrow-enabled custom types at the Python module boundary. `customdataclass` resolves postponed field annotations before Arrow schema generation and JSON field conversion.
Type-level custom-data subscriptions include identifier-scoped payloads. They retain one subscription
at the data client, and disposal releases the original route and parameters once.

```mermaid
flowchart TD
    Type[Custom data struct] --> Model[custom_data]
    Type --> Arrow[arrow_custom_data]
    Model --> Traits[Model traits and JSON]
    Arrow --> Codecs[Arrow schema and codecs]
    Codecs --> Register[Custom-type registration]
    Register --> Catalog[Catalog and streaming writes]
    Traits --> Python[Optional Python methods]
    Arrow --> Python
```

Model behavior and Arrow persistence compose through separate macros and explicit registration.

Source: [macro definitions](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/macros/src/lib.rs), [Python custom data](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/python/nautilus_trader/model/custom.py), and [Databento module registration](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/adapters/databento/src/python/mod.rs).

#### Streaming configuration and Parquet promotion

Backtest and live runtimes share persistence-owned catalog and streaming configuration. `StreamingConfig` remains a single optional configuration under the existing `streaming` feature. Its writer backend selects stage-only Feather or the promoting Parquet writer; Feather remains the default writer. Typed filters select data, records, concrete instrument families, and record identifiers. Explicitly empty filter lists fail validation.

The shared staged writer owns Feather staging, rotation, promotion scheduling, and finalization. The Parquet backend converts sealed Feather files into catalog files, records content-derived promotion identities, and writes run manifests. Completed promotion identities avoid repeating an already-recorded source conversion; these object-store operations do not form a multi-file catalog transaction.

| Parquet writer option         | Default | Effect                                                       |
| ----------------------------- | ------- | ------------------------------------------------------------ |
| `promote_on_close`            | `true`  | Promote staged data when the writer closes                   |
| `delete_feather_after_commit` | `false` | Retain committed source files unless deletion is enabled     |
| `use_ts_event_for_ts_init`    | `false` | Keep initialization timestamps unless remapping is requested |
| `parquet_commit_interval_ms`  | Unset   | Unset or zero disables interval-based promotion              |

```mermaid
flowchart TD
    Events[Filtered bus events] --> Feather[Feather staging]
    Feather --> Sealed[Sealed files and run manifest]
    Sealed --> Promote[Explicit, timed, or close promotion]
    Promote --> Files[Parquet catalog files]
    Files --> Identity[Record promotion identity]
    Identity --> Delete{Delete source enabled?}
    Delete -->|Yes| Remove[Remove committed Feather file]
    Delete -->|No| Keep[Retain Feather file]
```

Flushing Feather staging does not itself promise catalog visibility. A positive commit interval uses a
wall-clock timer in live runs; backtests check the supplied test clock during writes and flushes.
Close-time promotion defaults to enabled, while source deletion remains opt-in. See the
[Parquet streaming guide](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/docs/how_to/stream_parquet_catalog.md) for the
complete configuration example and recovery behavior.

Source: [streaming configuration](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/config.rs), [writer factories](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/writer/factory.rs), [Parquet writer](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/backend/parquet/writer.rs), and [kernel wiring](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/system/src/kernel.rs).

#### Catalog layout and migration

Parquet retains type-and-identifier directories. Custom data uses `data/custom/<type>/<identifier>/`, and instruments use concrete instrument-type folders. The migration planner recognizes supported earlier directory layouts and schema fingerprints, normalizes legacy columns, and resolves destination identifiers before writing. Empty coverage files survive migration, and source objects are checked against the plan before conversion.

`nautilus catalog migrate-parquet SOURCE TARGET` writes a separate catalog. It rejects overlapping locations, nonempty destinations, and unsupported schemas before conversion. Destination files use create-only writes. A dry run reports the plan without creating destination files; a completed or partially populated destination cannot be reused as a resumable target. The source remains unchanged even if conversion fails. Feather streams are outside this command's scope.

```bash
nautilus catalog migrate-parquet /data/catalog-old /data/catalog-new --dry-run
nautilus catalog migrate-parquet /data/catalog-old /data/catalog-new
```

Remote locations accept repeated `--source-option key=value` and `--target-option key=value` arguments. From a source checkout, the command is available through `cargo run -p nautilus-cli -- catalog migrate-parquet ...`.

```mermaid
flowchart TD
    Source[Existing Parquet catalog] --> Plan[Inventory and schema preflight]
    Plan --> Mode{Dry run?}
    Mode -->|Yes| Report[Report without destination writes]
    Mode -->|No| Empty[Require empty separate destination]
    Empty --> Normalize[Normalize legacy rows and layout]
    Normalize --> Target[Create open-format Parquet files]
    Source -.-> Preserve[Source bytes preserved]
```

Migration rewrites a separate destination and provides no in-place overwrite or resume contract.

Source: [CLI options](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/cli/src/opt.rs), [migration planner](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/backend/migration.rs), and [Parquet migration executor](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/backend/parquet/migration.rs).

Migration resolves remote object keys through the catalog resolver, retaining encoded identifiers
without adding the catalog prefix twice. Empty coverage markers use canonical type directories while
preserving instrument and custom-data layouts. Filtered remote instrument queries retain object-store
registration across repeated queries.

#### Python catalog and DataFrame APIs

PyO3 exposes the Parquet catalog through shared configuration, typed selectors, data and record queries, concrete instrument queries, and writer dispatch. Arrow IPC byte methods remain available; `query_catalog` uses Arrow C streams to avoid serializing the result to Python bytes before conversion. Output can be PyArrow, pandas, Polars, or a DuckDB relation, with pandas as the default and optional output libraries imported on demand.

`display=True` requires current-format Arrow input and returns readable display columns, including floating-point prices. Migrate legacy catalogs first; display conversion does not normalize old schemas. `display=False` preserves raw data and record schemas; instrument queries reject that option. Empty results retain the selected schema, and incompatible batches fail rather than silently losing fields. Identity groups sort lexically while preserving order within a group, so callers needing global chronological order must sort explicitly. Parquet requires `as_of=None`.

```mermaid
flowchart LR
    Query[query_catalog] --> Selector[Data, record, or instrument selector]
    Selector --> Stream[Arrow C stream]
    Stream --> Arrow[PyArrow table]
    Arrow --> Pandas[pandas]
    Arrow --> Polars[Polars]
    Arrow --> DuckDB[DuckDB relation]
```

All DataFrame outputs share the typed Arrow query result and its ordering contract.

Source: [DataFrame conversion](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/python/nautilus_trader/persistence/catalog_to_df.py), [Parquet Python bindings](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/python/backend/parquet/catalog.rs), and [Python writer dispatch](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/src/python/backend/writer.rs).

#### Integration and build changes

Adapter changes update custom-data macros, Arrow encoders, and owned depth handling. Backtest and live configuration use the shared factories while retaining the existing `streaming` feature and replay algorithms. Generated Python stubs and schema mirrors follow the exposed types and defaults.

The workspace replaces `nautilus-persistence-macros` with `nautilus-macros`, enables Parquet canonical extension annotations, and removes unused Arrow-row and interval-tree dependencies. Makefile crate lists and changed-file checks follow the macro crate move. A WebSocket test resumes real time around TCP handshakes so virtual time cannot advance directly into a connection timeout.

```mermaid
flowchart LR
    Model[Model and macro changes] --> Adapters[Adapter call sites]
    Model --> Bindings[Python bindings and stubs]
    Catalog[Shared catalog interfaces] --> Backtest[Backtest loading]
    Catalog --> Live[Live catalog configuration]
    Macros[Macro crate move] --> Build[Workspace and check scripts]
```

Integration changes connect the catalog and model contracts to existing consumers.

Source: [workspace dependencies](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/Cargo.toml), [backtest configuration](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/backtest/src/config.rs), [live configuration](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/live/src/node/config.rs), and [WebSocket regression](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/network/src/websocket/client.rs).

### Related issues and PRs

None specified.

### Type of change

- [x] New feature.
- [x] Improvement.
- [x] Breaking change.
- [x] Documentation update.
- [x] Maintenance.

### Breaking change details

| Surface                 | Change                                                           | Required adaptation                                                 |
| ----------------------- | ---------------------------------------------------------------- | ------------------------------------------------------------------- |
| Stored catalog data     | Open Arrow schema and normalized directory layout                | Migrate supported existing files into an empty separate destination |
| Depth model             | Variable-length `OrderBookDepth`; compatibility alias retained   | Handle owned variable-length sides and respect fixed-format limits  |
| Custom data             | Separate model and Arrow macros                                  | Apply and register Arrow support for types that need persistence    |
| Catalog integration     | Typed selectors, query objects, and factory-backed configuration | Update direct Rust integrations to the shared contracts             |
| Streaming configuration | Shared persistence-owned type with writer selection and filters  | Keep a single optional config and select the intended writer        |
| Display output          | Floating-point prices and sizes in nested depth lists            | Read bids and asks as lists; use raw output for exact decimals      |

Old depth-file order IDs cannot be reconstructed. Parquet has no snapshot or time-travel guarantee, and migration does not add one. The catalog and writer defaults remain Parquet and Feather, respectively.

### Documentation

- [x] Add the [Parquet migration guide](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/docs/how_to/migrate_parquet_catalog.md), including dry-run and object-store examples.
- [x] Update the [depth-model documentation](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/docs/concepts/data/order_book_depth10.md) and [data overview](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/docs/concepts/data/index.md).
- [x] Document the [develop-written migration fixtures](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/test_data/nautilus/catalog_develop/README.md).

### Release notes

- [x] Add entries to the existing unreleased section of [RELEASES.md](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/RELEASES.md) for migration, factories, variable depth, the Arrow format, and split custom-data macros.

### Testing

- [x] Add or update Rust and Python tests for catalog queries, streaming, migration, custom data, and Arrow representation.

The migration regression converts seven develop-written files containing eight rows, checks exact decoded values and physical Arrow schemas, and compares the original source bytes after conversion. Additional cases cover dry runs, empty coverage files, overlapping paths, nonempty destinations, and unsupported fixed-width byte columns. The CLI test exercises command validation and conversion through the binary entry point.

Catalog and Python tests cover typed queries, identifier discovery, metadata, display/raw conversion, custom registration, and streaming run manifests. Depth tests cover variable-length model and Arrow behavior, Cap'n Proto round trips, and fixed-depth compatibility checks.

```mermaid
flowchart TD
    Fixture[Develop-written catalog fixtures] --> Migration[Migration and CLI tests]
    Types[Typed data and custom types] --> Catalog[Catalog and Arrow tests]
    Catalog --> Python[Python query and DataFrame tests]
    Writer[Feather and Parquet writers] --> Lifecycle[Promotion and run-state tests]
    Depth[Variable-depth books] --> Formats[Arrow, Cap'n Proto, FFI, and SBE checks]
    Migration --> Assert[Exact values, schemas, and preserved source]
```

Tests check the stored representation and lifecycle contracts as well as decoded values.

Source: [migration regression tests](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/tests/test_parquet_migration.rs), [CLI regression](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/cli/tests/catalog_migration.rs), [Parquet catalog tests](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/crates/persistence/tests/test_parquet_catalog.rs), and [Python DataFrame tests](https://github.com/faysou/nautilus_trader/blob/e800a31e81552aaf30afbcbc5bb94ed26044ca5d/python/tests/persistence/test_parquet_catalog_to_df.py).

Current local validation:

- All seven query-error and JSON/Arrow regressions pass in the core suite; they fail before the fixes.
- Python persistence tests pass with encoding warnings treated as errors; manifest reads specify UTF-8.
- Workspace and fuzz lockfiles pass locked metadata checks; jiff feature checks and supply-chain audits pass.
- Script tests pass with Homebrew Python and a physical temporary-directory path on macOS.
- Review regressions: 14 passed, covering fixed-slot FFI construction, delta-batch writes, remote paths,
  repeated filtered instrument queries, and normalized empty coverage markers.
- Core Rust: 15,973 passed, 1 failed, and 16 skipped. The unchanged HTTP cancellation test
  `canceled_request_releases_partial_response` fails with `ConnectionReset` on macOS, including in isolation.
- Full Python suite and separate live-node run: passed.
- UTC schemas and custom-data round trips: covered by the full Python 3.14 suite.
- Generated API outputs and the branch-specific Parquet configuration example: checked.

Pre-commit, Markdown, and Mermaid checks pass. The macOS debug linker emits an unwind-table-size
warning. The catalog regressions and Python tests pass; the HTTP test failure remains noted above. These results cover `catalog-pr` itself.
DuckLake integration and the broader data-engine and backtest changes remain later PR scope. The UTF-8 fix passes locally with the encoding policy used in CI. After rebasing, both Cargo lockfiles are
regenerated to retain the locked macro dependency version. Replacement CI is pending. Live venue checks are not run.


cjdsellers:
Hey @faysou,

Thanks for all the work on the schema and catalog changes. I’m happy with the variable-depth model. I’ve pushed the adapter test updates and streaming feature guards, and the full Rust and Python suites pass locally. I’ll handle any remaining adapter-specific follow-up.

I found four core issues that need addressing before merge. These cases aren’t covered by the passing suites.

**P1: Padded depth input aborts through the legacy FFI**

`orderbook_depth10_new` accepts ten-slot arrays, but the model constructor removes empty levels. Converting back to the fixed-size FFI then panics and aborts the process. I reproduced this with padded input.

- Suggested fix: Construct `OrderBookDepth10Ffi` directly from the validated input arrays, preserving padding and metadata. Keep the exactly-ten-level requirement on the general Rust-to-FFI conversion.
- Regression coverage: Exercise the exported constructor with populated, partially padded, and fully empty arrays, checking every slot and metadata field.

**P2: `write_data_enum` panics on `Data::BookDeltas`**

Grouping produces a `DataBatch::BookDeltas` variant that the shared writer doesn’t handle, then reaches the `expect` asserting exhaustive dispatch.

- Suggested fix: Move the existing delta-flattening logic from `CatalogWriter::write_data_batch` into the shared batch writer so both entry points use it. Return an error for unsupported variants instead of panicking.
- Regression coverage: Write multiple delta batches through `write_data_enum` and read them back, checking every delta, its ordering, and boundary flags.

**P2: Remote migration adds the catalog prefix twice**

For `s3://bucket/catalog`, the migration plan already contains keys such as `catalog/data/...`. The `ParquetCatalogSource` path resolver prepends `catalog/` again, so migration fails after successful preflight.

- Suggested fix: Override the trait’s `to_object_path_parsed` method for `ParquetDataCatalog` and delegate to the catalog’s existing resolver. It already handles prefixed keys and preserves percent-encoding.
- Regression coverage: Migrate from a non-empty remote prefix, including an encoded instrument path, and verify the destination records. Retain local and bucket-root coverage.

**P2: Filtered remote instrument queries lose the registered store**

The instrument query registers the remote object store, then clears the session by replacing its DataFusion context. The query subsequently runs without that registration.

- Suggested fix: Clear the session before registering the remote store, matching the ordering in the other query paths.
- Regression coverage: Query a remote instrument catalog with a SQL predicate, check the exact matching records, and repeat on the same catalog instance.

**Two smaller follow-ups**

- Empty coverage markers retain legacy directory aliases during migration. Write them under the normalized destination type directory so the destination catalog can discover them.
- The depth documentation still says Rust and Python constructors require exactly ten levels. Update that statement to distinguish variable-depth constructors from the fixed-depth FFI and SBE formats.

faysou:
Thanks @cjdsellers. Addressed all six points in 35d9da836b and rebased onto cbb8bce905, preserving your adapter-test and streaming-guard updates.

- The exported FFI constructor preserves all ten supplied slots, including padding, counts, and metadata. General Rust-to-FFI conversion still requires exactly ten levels.
- Both catalog write entry points share delta-batch flattening, preserve per-instrument ordering and boundary flags, and return errors for unsupported batches.
- Remote migration uses the catalog's existing path resolver. Filtered instrument queries clear the session before registering the object store.
- Empty coverage markers use normalized directories, and the depth documentation distinguishes variable-length constructors from fixed-depth FFI/SBE formats.

All 14 focused regression cases pass, including bucket-root and prefixed remote paths, encoded identifiers, and repeated filtered queries. Assertions compare every FFI slot field and complete serialized records rather than ID-only equality. The full Python suite, separate live-node tests, and pre-commit checks pass.

The full core Rust run passed 15,973 tests with 16 skipped. Its one failure is the unchanged `canceled_request_releases_partial_response` HTTP test, which receives `ConnectionReset` on macOS and also fails in isolation. I left network code outside this PR and recorded that failure in the summary.

cjdsellers (approved):
@faysou, thanks for addressing the review points and adding the regression coverage. I've finished another pass and pushed some further fixes around streaming error propagation, schema preservation during Feather conversion, query table collisions, and custom-data inheritance. The schema changes are a clear improvement, and I appreciate the migration CLI and documentation.

I've kept legacy conversion behind the explicit CLI migration. Runtime normalization was inconsistent: some query paths converted old schemas while others rejected them, so the same catalog could work or fail depending on how it was queried. Requiring migration gives runtime queries a single schema contract and makes the conversion explicit.

This looks good to merge and thank you again for all of this work :pray:

Issue #4948, opened 2026-09-08
[Interactive Brokers] Unrepresentable fractional trade sizes log at WARN once per tick (99.7% of session log)


Artur-Sulej:
# Bug Report

## Confirmation

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I've tested this issue using a recent pre-release or development wheel (`2.0.0rcN`, `dev` develop, or `a` nightly) and can still reproduce it.

Tested on `2.0.0rc4`. The relevant code is identical on `develop` HEAD, at the same line numbers.

## Expected behavior

Dropping an IB trade print whose size cannot be represented at the instrument's size precision is expected and intended behavior. It was introduced deliberately in #4022, which fixed #3951, and that PR logs the drop at **debug** level in the Python market data path:

> "Adds a shared quantity conversion helper for IB market data paths, so invalid sizes are debug-logged and skipped before constructing Nautilus `QuoteTick` or `TradeTick` objects."

The v2 Rust adapter should treat the same condition the same way. This is a common, expected condition on US equities: FINRA-reported odd lots and fractional share fills appear continuously in the consolidated tape.

## Actual behavior

The v2 Rust adapter logs every dropped print at **WARN**, once per tick.

In a single regular trading session on five US equities (AAPL, NVDA, TSLA, AMD, SNDK), this produced:

```
1,603,188 WARN lines out of 1,607,952 total  (99.7% of the log)
284 MB session log
```

Every line has the same shape:

```
2026-09-04T13:29:33.838997000Z [WARN] ORB-001.nautilus_interactive_brokers::data::core::streams:
Failed to parse trade tick: quantity size=0.63462 cannot be represented with precision=0
```

The wording is also misleading. "Failed to parse trade tick" reads like a decode failure, when the condition is a deliberate, sanctioned drop of a print that carries no actionable size for a whole-share instrument.

At this volume the log is effectively unusable for diagnosing anything else, and the disk cost is non-trivial for an unattended session.

## Steps to reproduce

1. Subscribe to tick-by-tick `AllLast` trades for any liquid US equity with `size_precision=0` (AAPL, NVDA, TSLA).
2. Hold the subscription through a regular trading session.
3. Every FINRA odd lot or fractional share print produces one WARN line.

## Code snippets or logs

[...]

PR #4961, opened 2026-09-09
Make the IB tick-by-tick trade type selectable

### LINKED ISSUE #4960: [Interactive Brokers] Make the tick-by-tick trade type selectable (Last vs AllLast)
# Feature Request

## Problem statement

The IB data client always subscribes to tick-by-tick trades as `AllLast`, with no way to request `Last`. In `crates/adapters/interactive_brokers/src/data/core_streams.rs`:

```rust
let mut subscription = client
    .tick_by_tick(&contract, 0)
    .all_last()
    .await
    .context("Failed to create tick-by-tick trade subscription")?;
```

`AllLast` includes trade types beyond regular trades, such as combos, derivatively-priced and average-price prints, which are not part of the price-forming tape. `Last` excludes them. `ibapi 3.3.0` exposes both on the same builder (`last()` at `market_data/realtime/builder/tick_by_tick.rs:190`, `all_last()` at `:221`), so only the choice is missing.

### Why this matters: recorded data cannot reproduce a live session

Replaying a recorded IB session through the same strategy code reproduced every data-derived decision exactly: opening range 229.82/232.58, entry trigger 232.70, size 17, entry fill 17 @ 232.70, protective stop at 229.82, exit reason `protective_stop_hit`.

One thing diverged, the stop's fill price:

```
live   NVDA stop filled @ 229.69     realised -51.17
replay NVDA stop filled @ 224.93     realised -132.09
```

A 13-cent slip became a 4.9-point slip and the loss more than doubled. The recorded tape contains prints far below the stop trigger which were not executable market prices, and the simulator fills a triggered market stop at the next tick it sees.

The direction is what makes it hard to work with: backtests are pessimistic on stop-outs by a margin which varies with the instrument and the day, and which cannot be recovered from the recorded data. For any strategy exiting on stops, backtest P&L is not comparable to live P&L.

Filtering downstream by value, for example discarding prints far from their neighbours, is not a sound substitute: a filter cannot distinguish a non-market print from a genuine violent move, which is precisely the event a protective stop exists for. Being able to subscribe to `Last` moves the decision to the venue, which knows which trades are regular.

### Possible secondary benefit

Odd-lot and irregular prints arrive through `AllLast`. Subscribing to `Last` would likely reduce the volume of unrepresentable sizes reported in #4948 at the source, rather than only lowering how loudly they are logged.

## Proposed solution

A configuration field on `InteractiveBrokersDataClientConfig` selecting the trade tick type, defaulting to the current behaviour so nothing changes for existing users.

This follows the existing convention in the same config, which already lets the caller choose which IB mechanism supplies quotes:

```rust
/// Whether to use batch quotes (reqMktData) by default instead of tick-by-tick.
#[builder(default = true)]
pub batch_quotes: bool,
```

## Example usage

```python
InteractiveBrokersDataClientConfig(
    host="127.0.0.1",
    port=7497,
    client_id=1,
    market_data_type=MarketDataType.REALTIME,
    # Regular trades only, so recorded tapes contain price-forming prints.
    all_last_trades=False,
)
```

A `bool` matches `batch_quotes` for consistency; a small enum (`Last` / `AllLast`) would read better at the call site. Either is fine by us.

## Alternatives considered

**Preserve the trade conditions on `TradeTick` instead.** The labels already arrive: `ibapi`'s `Trade` carries `trade_attribute.unreported` and `special_conditions`, and `parse_trade_tick` drops both. Carrying them through, as `OrderBookDelta`, `OrderBookDepth10` and `OrderBookDeltas` already do with `flags: u8`, would be the more general fix, would let consumers filter by label rather than by feed choice, and would help other adapters too (Databento's record flags are dropped for the same reason).

That is a core data type change touching Arrow schemas, persistence and bindings, so we are raising this narrower option first. Happy to open the broader one as an RFC if you would prefer that direction instead, or in addition.

**Value-based filtering downstream.** Covered above; unsound for stop-driven strategies.

**Recording through a separate IB client which keeps the conditions.** Means maintaining a second data path purely for recording, and gives up the guarantee that live and backtest consume the same code path.

## Additional context

Happy to open a PR for this if the approach is acceptable, including whichever of `bool` or enum you prefer.

Related to #2194 (fill simulation), though the cause here is upstream of the fill model: the information needed to model the fill is discarded before the matching engine sees it.

- Python version: 3.13.15
- Adapter/venue: Interactive Brokers (TWS paper, API v178), US equities

Artur-Sulej:
**TL;DR:** The IB data client hardcoded tick-by-tick trades to `AllLast`, which includes prints that are not part of the price-forming tape and cannot be identified downstream. This adds an `all_last_trades` config field so `Last` can be selected instead, defaulting to the current behaviour. Fixes #4960.

## Summary

`handle_trade_subscription` called `.all_last()` unconditionally. `AllLast` includes trade types which are not price-forming, such as combos, derivatively-priced and average-price prints, and those cannot be filtered downstream because the trade conditions are dropped when a `TradeTick` is constructed. The practical consequence is that a recorded tape cannot reproduce the live session it came from.

Replaying one recorded IB session through the same strategy code reproduced every data-derived decision exactly: opening range 229.82/232.58, entry trigger 232.70, size 17, entry fill 17 @ 232.70, protective stop at 229.82, exit reason `protective_stop_hit`. Only the stop's fill price diverged, live 229.69 against replay 224.93, turning a realised -51.17 into -132.09. The tape contained prints far below the trigger which were never executable prices.

`InteractiveBrokersDataClientConfig` now carries `all_last_trades`, following the existing `batch_quotes` convention in the same struct, which already lets the caller choose which IB mechanism supplies quotes. `ibapi` exposes `last()` alongside `all_last()` on the same builder, so this only surfaces a choice which already existed.

The default is `true`, so behaviour is unchanged unless a user opts in. The data client option table in `docs/integrations/interactive_brokers.md` gains a row for the new field.

## Related issues/PRs

Closes #4960.

Related to #4948: odd-lot and irregular prints arrive through `AllLast`, so selecting `Last` may reduce the volume of unrepresentable sizes at the source rather than only lowering how loudly they are logged. The two changes are independent.

Related to #2194 (fill simulation), though the cause here is upstream of the fill model.

## Type of change

- [ ] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

Four tests in `crates/adapters/interactive_brokers/src/config.rs`:

- the default is `AllLast`, pinning that existing users are unaffected;
- the builder can select `Last`;
- the field survives a serde round trip;
- a config document written before this field exists still deserializes, and keeps the previous behaviour.

`cargo test -p nautilus-interactive-brokers` passes (422 tests). `cargo clippy --all-targets` is clean and `cargo +nightly fmt` reports no changes.

`make py-stubs` runs and reproduces the committed stubs with no diff. My earlier failure to build with `--features python` was a local environment problem (PyO3 picked up the system Python 3.9 because the checkout had no `python/.venv`), not a build issue.

The subscription branch itself needs a live IB connection, so it is covered by the config tests plus the `AllLast`/`Last` marker now included in the subscription's debug log, rather than by a unit test.

### One judgement call

I placed `all_last_trades` next to `batch_quotes` in the Python signature rather than appending it, since that matches the struct field order and reads better. That does shift the positions of `instrument_provider` and `dockerized_gateway` for anyone passing them positionally. Happy to move it to the end if you would rather not change positional order at all.

### DIFF
--- a/crates/adapters/interactive_brokers/src/config.rs
+++ b/crates/adapters/interactive_brokers/src/config.rs
@@ -116,6 +116,14 @@ pub struct InteractiveBrokersDataClientConfig {
     /// Whether to use batch quotes (reqMktData) by default instead of tick-by-tick.
     #[builder(default = true)]
     pub batch_quotes: bool,
+    /// Whether to subscribe tick-by-tick trades as `AllLast` rather than `Last`.
+    ///
+    /// `AllLast` includes trade types which are not part of the price-forming tape, such as
+    /// combos, derivatively-priced and average-price prints. `Last` returns regular trades
+    /// only, which is usually what should be recorded for replay. Defaults to `true` to
+    /// preserve existing behaviour.
+    #[builder(default = true)]
+    pub all_last_trades: bool,
     /// Instrument provider configuration.
     #[builder(default)]
     pub instrument_provider: InteractiveBrokersInstrumentProviderConfig,
@@ -402,5 +410,42 @@ mod tests {
         assert!(formatted.contains("password: Some(<redacted>)"));
         assert!(!formatted.contains("test-user"));
         assert!(!formatted.contains("test-password"));
+    }
+
+    #[rstest]
+    fn data_client_config_defaults_to_all_last_trades() {
+        // Defaults must preserve the previous behaviour for existing users.
+        let config = InteractiveBrokersDataClientConfig::default();
+        assert!(config.all_last_trades);
+    }
+
+    #[rstest]
+    fn data_client_config_can_select_last_trades() {
+        let config = InteractiveBrokersDataClientConfig::builder()
+            .all_last_trades(false)
+            .build();
+        assert!(!config.all_last_trades);
+    }
+
+    #[rstest]
+    fn data_client_config_all_last_trades_round_trips_through_serde() {
+        let config = InteractiveBrokersDataClientConfig::builder()
+            .all_last_trades(false)
+            .build();
+
+        let json = serde_json::to_string(&config).unwrap();
+        let decoded: InteractiveBrokersDataClientConfig = serde_json::from_str(&json).unwrap();
+
+        assert!(!decoded.all_last_trades);
+    }
+
+    #[rstest]
+    fn data_client_config_omitted_all_last_trades_deserializes_to_default() {
+        // `#[serde(default)]` on the struct means an existing config file without the new
+        // field must still deserialize, and keep the previous behaviour.
+        let decoded: InteractiveBrokersDataClientConfig =
+            serde_json::from_str(r#"{"host":"127.0.0.1","port":7497}"#).unwrap();
+
+        assert!(decoded.all_last_trades);
     }
 }
--- a/crates/adapters/interactive_brokers/src/config.rs
+++ b/crates/adapters/interactive_brokers/src/config.rs
@@ -121,7 +121,7 @@ pub struct InteractiveBrokersDataClientConfig {
     /// `AllLast` includes trade types which are not part of the price-forming tape, such as
     /// combos, derivatively-priced and average-price prints. `Last` returns regular trades
     /// only, which is usually what should be recorded for replay. Defaults to `true` to
-    /// preserve existing behaviour.
+    /// preserve existing behavior.
     #[builder(default = true)]
     pub all_last_trades: bool,
     /// Instrument provider configuration.
@@ -414,7 +414,7 @@ mod tests {
 
     #[rstest]
     fn data_client_config_defaults_to_all_last_trades() {
-        // Defaults must preserve the previous behaviour for existing users.
+        // Defaults must preserve the previous behavior for existing users.
         let config = InteractiveBrokersDataClientConfig::default();
         assert!(config.all_last_trades);
     }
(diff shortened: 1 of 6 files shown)


cjdsellers:
Hi @Artur-Sulej,

Thanks for the PR and the detailed reports. I appreciate the time you've put into investigating these problems and contributing fixes.

Could you slow down on opening further issues and PRs for now, and focus on addressing feedback on the existing work? Maintainer bandwidth is limited, and several issues and PRs arriving together take time to assess properly. With `develop` moving quickly, multiple open PRs are also likely to become stale before I can finish reviewing them, adding rebase and validation work for you.

There are a few contribution requirements to address:
- The [contributing guide](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md#start-with-an-issue) asks contributors to wait for agreement on substantial changes. This PR followed #4960 before that discussion took place; the configuration and API choices need agreement first.
- The [local validation requirements](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md#run-local-checks-and-follow-repository-rules) require generated stubs to come from `make py-stubs`, rather than hand edits. Thanks for explaining the build limitation, but an exception needs to be agreed before requesting review. Asking maintainers to regenerate the files leaves part of the required validation with them.
- The account-selection feedback on #4954 still needs addressing. Readiness checkboxes and safety claims need to reflect the submitted change and its validation.

The [AI Policy](https://github.com/nautechsystems/nautilus_trader/blob/develop/AI_POLICY.md#maintainer-effort) also makes clear that AI-assisted work must not shift design, debugging, or validation onto maintainers. These quality expectations apply regardless of which tools you use.

For now, could we work through the existing contributions one at a time and agree on scope before you start another substantial change? That would make it easier to give each contribution the attention it needs.

cjdsellers:

Issue #3951, opened 2026-04-28
[Interactive Brokers] Fractional tick sizes can stall market data dispatch


francofst:
# Bug Report

**NautilusTrader version:** 1.225.0 (also reproducible in `develop` HEAD — `client/market_data.py` line 823 still calls `instrument.make_qty(size)` without sanitization)
**Component:** `nautilus_trader/adapters/interactive_brokers/client/market_data.py::process_tick_by_tick_all_last`
**Severity:** High in production — silent dispatch hang under load (no crash, just stalls)

## Summary

`InteractiveBrokersClientMarketDataMixin.process_tick_by_tick_all_last` (line ~787, calling `instrument.make_qty(size)` at line ~823) does not sanitize `size < instrument.size_increment`. Most US equities use `size_precision=0`, so any incoming tick with `size < 1.0` raises `ValueError` from `make_qty`.

This actually happens often: **FINRA-reported dark-pool odd-lot trades** and **fractional-share fills from brokers like Robinhood/Fidelity** appear in the consolidated tape with sub-share sizes. The exception propagates from the executor task, the IB socket buffer fills, and tick dispatch silently stalls — the bridge looks alive (no Python crash) but stops receiving market data.

In our production deployment we filter these ticks via monkey-patch and counted **>10,000 fractional ticks rejected per trading session** across SPY/AAPL/MSFT subscriptions. Without the patch, the bridge would freeze within minutes during RTH.

## Reproduction

1. Subscribe to tick-by-tick `AllLast` for any high-volume US stock (SPY/AAPL/MSFT) with `size_precision=0`.
2. Hold the subscription through a busy RTH window (typically a few minutes is enough).
3. The first FINRA dark-pool odd-lot fill (e.g. `size=0.5`) reaches `process_tick_by_tick_all_last` and raises:

```
ValueError: invalid size precision...
  at make_qty(size)
  at adapters/interactive_brokers/client/market_data.py:823 process_tick_by_tick_all_last
```

4. The exception is not caught inside the dispatch coroutine. Subsequent ticks pile up in the asyncio queue, executor saturates, no more market data is processed.

## Evidence from production

JUPITER instruments the patched call site with a counter. From a single trading session today:

```
[PATCH_5] activated count=1     (process_tick_by_tick_all_last_fractional)
[PATCH_5] activated count=5
[PATCH_5] activated count=20
[PATCH_5] activated count=100
[PATCH_5] activated count=500
[...]

PR #4022, opened 2026-05-06
Drop unrepresentable IB market data tick sizes

faysou:
# Pull Request

**NautilusTrader prioritizes correctness and reliability, please follow existing patterns for validation and testing.**

- [ ] I have reviewed the `CONTRIBUTING.md` and followed the established practices

## Summary

- Fixes Interactive Brokers market data dispatch stalling when IB emits fractional tick sizes that cannot be represented by the cached instrument size precision.
- Adds a shared quantity conversion helper for IB market data paths, so invalid sizes are debug-logged and skipped before constructing Nautilus `QuoteTick` or `TradeTick` objects.
- Applies the guard to live tick-by-tick bid/ask data, live tick-by-tick all-last trades, historical bid/ask ticks, and historical trade ticks.
- Adds regression coverage for AAPL-style equity instruments with `size_precision=0` receiving `Decimal("0.5")` sizes.

### Design notes

- The fix follows the Databento adapter pattern of resolving or validating feed precision before constructing Nautilus model values.
- IB differs from Databento because the affected sizes arrive as decimal quantities for instruments that only support whole-share quantities. Databento feed sizes are decoded as integral quantities, so this guard is specific to the IB adapter.
- Dropping unrepresentable market data ticks is preferable to rounding fractional prints into actionable whole-share sizes.
- Missing cached instruments now log an error before returning from the affected conversion paths.

### Files changed

- `nautilus_trader/adapters/interactive_brokers/client/market_data.py`
- `tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py`

### Verification

- `uv run --no-sync pytest tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py`.
- `uv run --active --no-sync ruff format nautilus_trader/adapters/interactive_brokers/client/market_data.py tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py`.
- `uv run --active --no-sync ruff check nautilus_trader/adapters/interactive_brokers/client/market_data.py tests/integration_tests/adapters/interactive_brokers/client/test_client_market_data.py`.
- `git diff --check`.

### Risk

- Low to medium. The change only affects IB market data conversion and skips ticks that already could not be represented without raising.
- Consumers may see fewer IB trade or quote ticks when the feed contains fractional sizes for whole-share instruments, but this avoids dispatch task failures and preserves valid ticks.

## Related Issues/PRs

https://github.com/nautechsystems/nautilus_trader/issues/3951

## Type of change

[...]

PR #4962, opened 2026-09-09
Log unrepresentable IB tick sizes at debug instead of warn

### LINKED ISSUE #4948: [Interactive Brokers] Unrepresentable fractional trade sizes log at WARN once per tick (99.7% of session log)

Artur-Sulej:
**TL;DR:** IB reports FINRA odd-lot and fractional-share prints with sub-increment sizes, and dropping them is deliberate, but the v2 adapter logs each drop at WARN once per tick. In one five-symbol regular session that was 1,603,188 of 1,607,952 log lines, 99.7% of the file. This adds a typed error so the expected drop can be told apart from a real decode failure, and logs it at debug. Fixes #4948.

## Summary

Dropping a print whose size cannot be represented at the instrument's size precision is intended behaviour, established in #4022 for #3951. That PR chose **debug** for the equivalent Python path: "invalid sizes are debug-logged and skipped". The v2 Rust adapter never adopted that level and logs the same condition at WARN, once per tick.

This is a high-frequency condition, not a rare one. FINRA-reported odd lots and fractional-share fills arrive continuously in the US consolidated tape for instruments with `size_precision=0`. Measured on `2.0.0rc4` over one regular session across AAPL, NVDA, TSLA, AMD and SNDK:

```
1,603,188 WARN lines out of 1,607,952 total   (99.7% of the log)
284 MB session log
```

Every line reads `Failed to parse trade tick: quantity size=0.63462 cannot be represented with precision=0`, which also reads like a decode failure rather than a sanctioned drop.

The problem is that one log site collapses two different things: an expected, high-frequency drop and a genuine parse error. Lowering the level indiscriminately would hide real failures. So `checked_quantity` now returns a typed `UnrepresentableSize` error, and the call sites match on it: the expected drop logs at debug with a clearer message, and anything else keeps WARN.

The four affected paths are live trades, live quotes, historical trades and historical quotes, matching the surface #4022 covered on the Python side. The duplicate size check in `data/cache.rs` is lowered to debug for the same reason, while its "invalid size" branch keeps WARN.

No ticks change: the same prints are dropped as before, only the reporting differs.

## Related issues/PRs

Closes #4948.

Follows #3951 and #4022, which established that dropping these prints is preferred over rounding them into actionable whole-unit sizes. This PR aligns the v2 Rust adapter with the log level chosen there.

Related to #4960 and its PR #4961, which makes the IB trade tick type selectable. Many of these sub-increment prints arrive through `AllLast`, so selecting `Last` would reduce their volume at the source, while this PR fixes how the remainder are reported. The two are independent and touch different lines; I verified the branches merge cleanly in either order.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Testing

The two existing `rejects_unrepresentable_size` tests asserted only `is_err()`, so they passed either way. They now assert the classification, and the trade one asserts the error carries the reported size and precision. Verified they fail when the typed error is reverted to a plain `bail!`.

Two tests were added:

- `test_invalid_size_is_not_flagged_unrepresentable` submits a negative size and asserts it is **not** classified as an expected drop, so genuine failures keep WARN. This is the test that stops the fix from becoming a blanket silencing.
- `test_fractional_size_parses_when_precision_allows_it` shows the same 0.5 size parses at `size_precision=1`, so the drop is precision-dependent rather than a rejection of fractional sizes.

`cargo test -p nautilus-interactive-brokers` passes (420 tests). `cargo clippy --all-targets` is clean and `cargo +nightly fmt` reports no changes.

One note for reviewers: an alternative would be a periodic aggregate ("dropped N unrepresentable prints for X in the last minute"), which would keep some signal at INFO. I kept this PR to the level change since that matches the existing decision in #4022, but I am happy to add the aggregate if you would prefer the volume visible rather than at debug.

### DIFF
--- a/crates/adapters/interactive_brokers/src/data/cache.rs
+++ b/crates/adapters/interactive_brokers/src/data/cache.rs
@@ -36,7 +36,9 @@ fn checked_quantity(size: f64, precision: u8) -> Option<Quantity> {
 
     let tolerance = 10_f64.powi(-i32::from(precision)) * 1e-9;
     if (quantity.as_f64() - size).abs() > tolerance {
-        tracing::warn!(
+        // Expected on venues which report sub-increment sizes (FINRA odd lots, fractional
+        // fills), so this is not a decode failure; see `parse::UnrepresentableSize`.
+        tracing::debug!(
             "Ignoring unrepresentable IB quote size {} with precision {}",
             size,
             precision
--- a/crates/adapters/interactive_brokers/src/data/core.rs
+++ b/crates/adapters/interactive_brokers/src/data/core.rs
@@ -1994,6 +1994,9 @@ impl DataClient for InteractiveBrokersDataClient {
                                 ts_init,
                             ) {
                                 Ok(quote_tick) => batch_quotes.push(quote_tick),
+                                Err(e) if super::parse::is_unrepresentable_size(&e) => {
+                                    tracing::debug!("Dropping quote for {}: {}", instrument_id, e);
+                                }
                                 Err(e) => {
                                     tracing::warn!("Failed to parse quote tick: {:?}", e);
                                 }
@@ -2185,6 +2188,13 @@ impl DataClient for InteractiveBrokersDataClient {
                                 trade_id,
                             ) {
                                 Ok(trade_tick) => batch_trades.push(trade_tick),
+                                Err(e) if super::parse::is_unrepresentable_size(&e) => {
+                                    tracing::debug!(
+                                        "Dropping trade print for {}: {}",
+                                        instrument_id,
+                                        e
+                                    );
+                                }
                                 Err(e) => {
                                     tracing::warn!("Failed to parse trade tick: {:?}", e);
                                 }
--- a/crates/adapters/interactive_brokers/src/data/core_streams.rs
+++ b/crates/adapters/interactive_brokers/src/data/core_streams.rs
@@ -1050,6 +1050,13 @@ pub(super) async fn handle_tick_by_tick_quote_subscription(
                                         return Ok(());
                                     }
                                 }
+                                Err(e) if crate::data::parse::is_unrepresentable_size(&e) => {
+                                    tracing::debug!(
+                                        "Dropping quote for {}: {}",
+                                        instrument_id,
+                                        e
+                                    );
+                                }
                                 Err(e) => tracing::warn!("Failed to parse quote tick: {:?}", e),
                             }
                         }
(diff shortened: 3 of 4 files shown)


cjdsellers:

PR #4965, opened 2026-09-10
Fix Binance Spot SBE order decoder field alignment

abhijeetvichare76:
## Spot SBE order decoders read fields at the wrong offsets

The hand-written Spot SBE decoders in `crates/adapters/binance/src/spot/http/parse.rs` disagree with the generated codecs. `decode_order` skipped neither `trailingDelta`/`trailingTime` nor the `strategyId`..`workingFloor` run, so `icebergQty`, `time`, `updateTime`, `isWorking`, `workingTime`, `origQuoteOrderQty` and `selfTradePreventionMode` were read from the wrong offsets. `decode_new_order_full` read `selfTradePreventionMode` one byte late, from the low byte of `tradeGroupId`, and then made a phantom `commissionExponent` read; `decode_cancel_order` read it 46 bytes early, from the low byte of `stopPrice`. Nothing failed loudly because the trailing skip is driven by the wire block length, so the variable-length strings still landed correctly. `decode_orders` was already aligned.

The single-order path feeds `OrderStatusReport` through `parse_order_status_report_sbe`, which takes `ts_accepted` and `ts_event` from `time` and `updateTime`, so `query_order` reported wrong timestamps. The mainnet captures under `test_data/spot/user_data_sbe/mainnet` decode to `ExpireMaker` through the generated decoders where the hand decoders returned null.

The fix folds `decode_order` and `decode_orders` onto one shared block walk with a minimum block-length guard, corrects the two skips in the other decoders, and drops the phantom read. The integration mock single-order response had mirrored the wrong layout and is corrected to match the batch response builder.

The previous unit tests hand-assembled bytes at the decoder's own offsets, which is why they passed. They are replaced with fixtures built from the generated encoders, and a handler test asserts the hand decoders agree with the generated decoders on the mainnet captures.

Noted, not changed here: `decode_new_order_full` has no short-block guard, unlike `decode_cancel_order`, so an undersized block falls through to `read_trailing_expiry_reason` returning `None`.

## Testing

Fixtures cover each template at its pre-3:4, 3:4 and future block lengths, short blocks, groups with variable-length strings, empty prevented-match groups and null optionals. Reverting any of the three decoder fixes on its own fails four or five of the new tests; the old `decode_orders` tests did not catch the shared-walk change because they were zero-padded.

- `cargo test -p nautilus-binance --lib`: 1112 passed
- `cargo test -p nautilus-binance --test integration -- spot::`: 185 passed
- `cargo clippy -p nautilus-binance --no-deps --all-targets -- -D warnings`: clean
- `prek run --files` on the three changed files: `check formatting (Rust)` and `cargo fmt` pass

### DIFF
--- a/crates/adapters/binance/src/spot/http/parse.rs
+++ b/crates/adapters/binance/src/spot/http/parse.rs
@@ -359,14 +359,18 @@ pub fn decode_klines(buf: &[u8]) -> Result<BinanceKlines, SbeDecodeError> {
 
 /// Bytes consumed from the fixed block before the end-of-block skip in each decoder.
 /// These represent explicit reads and advances up to the last field we extract.
-const NEW_ORDER_FULL_FIELDS_END: usize = 135;
-const CANCEL_ORDER_FIELDS_END: usize = 63;
+const NEW_ORDER_FULL_FIELDS_END: usize = 117;
+const CANCEL_ORDER_FIELDS_END: usize = 109;
 const CANCEL_ORDER_LIST_FIELDS_END: usize = 21;
 const CANCEL_ORDER_LIST_ORDER_FIELDS_END: usize = 8;
 const CANCEL_ORDER_LIST_REPORT_FIELDS_END: usize = 107;
 const CANCEL_ORDER_LIST_REPORT_V0_BLOCK_LENGTH: u16 = 124;
 const CANCEL_ORDER_LIST_REPORT_V1_BLOCK_LENGTH: u16 = 135;
-const ORDER_FIELDS_END: usize = 104;
+const ORDER_FIELDS_END: usize = 134;
+
+/// Minimum fixed-block length of an order in `orderResponse` and the
+/// `ordersResponse` group (schema 3:3 baseline).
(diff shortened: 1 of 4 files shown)


cjdsellers:
Hi @abhijeetvichare76,

Thanks for the PR. This looks good to merge.

If you get a chance, could you update the remaining new-order and cancel-order mocks in `tests/integration/spot/http.rs` and `exec_client.rs` to match the schema offsets too? Their zero-valued STP fields hide the stale layouts.

abhijeetvichare76:
Thanks, done!

cjdsellers (approved):
Hey @abhijeetvichare76,

Thanks for aligning the mocks with the schema offsets. I've had another look at the latest changes, and this looks good to merge 👍.

Rebuilding the tests on the generated encoders was a great call too, the fixtures can no longer silently mirror the decoder's own offsets, and the mainnet capture cross-check against the generated decoders locks that in.

On the `decode_new_order_full` short-block guard you noted: happy to leave that as a follow-up, and if you get a chance it would be nice to have it consistent with `decode_cancel_order` and `decode_order_block`. Low risk either way since the venue controls the block length.

PR #4967, opened 2026-09-10
Split reversal fills into closing and opening portions

folknor:
A single fill that reverses a position through zero currently contaminates the new episode's close accounting in `Position`.

## Reversal fills carry the old episode's close state into the new one

When an opposite-side fill exceeds the open quantity, `handle_buy_order_fill` and `handle_sell_order_fill` reset only `avg_px_open`. The closing arm feeds the full `last_qty` into `calculate_avg_px_close_px`, so the close average absorbs the opening portion of the fill, and `avg_px_close`, `realized_return`, `buy_qty`, and `sell_qty` all carry across the reversal - the next episode's close averages and returns are weighted by the previous episode's volume. Realized PnL was already correct: `calculate_pnl_raw` clamps to the open quantity, so it has always used only the closing portion. The fix completes that closing-portion pattern: the fill is split with exact `Quantity` arithmetic at the pre-fill open quantity, only the closing portion reaches the close-average and PnL calculations, and the reversal establishes a fresh episode - `avg_px_open` at the fill price, `avg_px_close` cleared, `realized_return` zeroed, and the quantity accumulators re-seeded to the opening residual. `realized_pnl` stays cumulative and commissions are charged once. This departs from the legacy Python implementation, which carries the same contamination.

- Applying the reversal as two synthetic fragments via `OrderFilled::split_for_position_flip`, as `flip_position` does across two objects: inside one `Position` the fragments would duplicate events, trade IDs, and commissions.
- Routing the zero crossing through `reset_cycle`, the path an exact close already takes: it resets everything including `realized_pnl`, and a single object has nowhere to archive the closed episode, so cumulative PnL from cache- and replay-derived positions would be lost.
- Clamping in `calculate_avg_px_close_px` the way `calculate_pnl_raw` already does: fixes the reversal fill's own weighting but leaves `avg_px_close`, `realized_return`, and the accumulators stale, so the second close after a reversal still misweights.
- Restructuring `Position` into per-episode records: a public-surface redesign out of proportion to the defect.

## Testing

`test_position_reversal_starts_new_close_episode` covers both directions: open, partial close, oversized reversal fill, then two differently priced closes. The two-close sequence is the load-bearing part - `calculate_avg_px_close_px` returns `last_px` when `avg_px_close` is `None`, so a single close looks correct even unfixed, and only the second close exposes stale accumulators. Assertions pin the episode reset, exact close averages and returns, and cumulative PnL with commissions counted once. `test_fill_void_replays_partially_voided_reversal` covers a reversal fill later partially voided. `test_fill_void_replays_split_fragments_in_one_corrected_cycle` is updated to the episode-scoped accumulator values; `test_position_true_reversal_uses_fill_price` is unchanged, demonstrating PnL preservation. The new assertions fail against the unfixed implementation.

### DIFF
--- a/crates/model/src/position.rs
+++ b/crates/model/src/position.rs
@@ -101,7 +101,9 @@ pub struct Position {
     pub realized_pnl: Option<Money>,
     #[serde(with = "nautilus_core::serialization::sorted_hashset")]
     pub trade_ids: AHashSet<TradeId>,
+    /// Total buy quantity in the current position episode.
     pub buy_qty: Quantity,
+    /// Total sell quantity in the current position episode.
     pub sell_qty: Quantity,
     pub commissions: IndexMap<Currency, Money>,
 }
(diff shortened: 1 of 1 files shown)


cjdsellers (approved):
Hey @folknor,

Thanks for the PR. This looks good to merge. I'll follow up with the position docs to clarify that fill-driven reversals reset close averages and returns while realized PnL stays cumulative.

PR #4968, opened 2026-09-10
Restore spawn reductions on canceled and expired children

folknor:
A spawned exec-algorithm child that reaches the venue and terminates unfilled permanently loses its quantity from the primary order.

## Canceled and expired children commit the pending spawn reduction

When an exec algorithm spawns a child with `reduce_primary`, the deducted quantity is tracked and restored by `restore_primary_order_quantity` if the child is denied or rejected, capped at the child's `leaves_qty`. The `Accepted`, `Canceled`, and `Expired` arms of `handle_order_event` instead committed the reduction with a bare `take_pending_spawn_reduction`, so a canceled or expired child restored nothing and the algorithm under-executed by the child's unfilled amount.

The reduction entry now lives until the child's first terminal outcome: acceptance retains it, `Canceled` and `Expired` restore through the existing helper - whose `leaves_qty` cap restores exactly the unfilled portion after a partial fill and nothing after a full fill - and a fill that leaves zero quantity consumes the entry. `Denied` and `Rejected` are unchanged, restoration cannot double-apply because the take is destructive, and `reduce_primary=false` spawns track no entry and are unaffected.

Two boundaries are preserved rather than handled: a late fill arriving after a cancellation restoration is not compensated back into the primary, and a `FillVoided` that reopens a fully filled child cannot restore, since by then neither the original reduction nor `reduce_primary` eligibility is recoverable. Both would need reduction history retained beyond the child's closure.

- Fixing only the `Canceled`/`Expired` arms while `Accepted` keeps committing the entry: incomplete by construction, since an accepted child's entry is gone before its terminal event arrives.
- Decrementing the map entry on each partial fill: `min(reduction, leaves_qty)` at terminal time already yields the unfilled portion, and per-fill updates open duplicate-fill and `FillVoided` bookkeeping questions.
- Committing at acceptance and recreating the entry at terminal time: after consumption a reduced child is indistinguishable from a `reduce_primary=false` one and the original reduction amount is gone, so re-derivation could add quantity never deducted.
- A non-removing read on the reduction map: the destructive take is what makes restoration idempotent, so a peek would add surface and lose that property.

## Testing

New cases pin accepted-then-canceled and accepted-then-expired zero-fill children restoring the full reduction, partial-fill cancellation and expiry restoring exactly `leaves_qty`, full-fill consumption with no restoration, and a denied or rejected child followed by a directly dispatched terminal event restoring only once. `test_spawned_order_accepted_prevents_restoration` is renamed to `test_spawned_order_accepted_then_canceled_restores_reduction` with its post-accept assertion intact. The new assertions fail against the unfixed implementation - six failing cases, covering both the canceled and expired paths.

### DIFF
--- a/crates/trading/src/algorithm/core.rs
+++ b/crates/trading/src/algorithm/core.rs
@@ -49,6 +49,18 @@ pub struct StrategyEventHandlers {
     pub position_handler: TypedHandler<PositionEvent>,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct SpawnReduction {
+    /// The quantity deducted from the primary at spawn time.
+    pub deducted_qty: Quantity,
+    /// Whether the spawned quantity was quote-denominated at deduction time.
+    pub spawn_was_quote_quantity: bool,
+    /// The gross quantity released by restoration - including any portion
+    /// applied against primary late-fill debt rather than returned to the
+    /// primary - serving as the child's remaining late-fill accounting budget.
+    pub restored_qty: Quantity,
+}
+
 /// The core component of an [`ExecutionAlgorithm`](super::ExecutionAlgorithm).
 ///
 /// This struct manages the internal state for execution algorithms including
@@ -70,8 +82,13 @@ pub struct ExecutionAlgorithmCore {
     exec_spawn_ids: AHashMap<ClientOrderId, u32>,
     /// Tracks strategies that have been subscribed to for events.
     subscribed_strategies: AHashSet<StrategyId>,
-    /// Tracks pending spawn reductions for quantity restoration on denial/rejection.
-    pending_spawn_reductions: AHashMap<ClientOrderId, Quantity>,
+    /// Tracks spawn reductions through restoration and possible late-fill netting.
+    spawn_reductions: AHashMap<ClientOrderId, SpawnReduction>,
+    /// Tracks uncompensated late-fill quantity per primary order, discharged
+    /// against later spawn restorations.
+    spawn_fill_debts: AHashMap<ClientOrderId, Quantity>,
+    /// Tracks primary orders whose submission was handed off to the risk engine.
+    handed_off_primaries: AHashSet<ClientOrderId>,
     /// Maps primary order client IDs to the command params supplied at submission.
     submit_params: AHashMap<ClientOrderId, Params>,
     /// The portfolio shared by the trader.
@@ -119,10 +136,9 @@ impl Debug for ExecutionAlgorithmCore {
             .field("exec_algorithm_id", &self.exec_algorithm_id)
             .field("exec_spawn_ids", &self.exec_spawn_ids.len())
             .field("subscribed_strategies", &self.subscribed_strategies.len())
-            .field(
-                "pending_spawn_reductions",
-                &self.pending_spawn_reductions.len(),
-            )
+            .field("spawn_reductions", &self.spawn_reductions.len())
+            .field("spawn_fill_debts", &self.spawn_fill_debts.len())
+            .field("handed_off_primaries", &self.handed_off_primaries.len())
             .field("submit_params", &self.submit_params.len())
             .field(
                 "strategy_event_handlers",
@@ -156,7 +172,9 @@ impl ExecutionAlgorithmCore {
             exec_algorithm_id,
             exec_spawn_ids: AHashMap::new(),
             subscribed_strategies: AHashSet::new(),
-            pending_spawn_reductions: AHashMap::new(),
+            spawn_reductions: AHashMap::new(),
+            spawn_fill_debts: AHashMap::new(),
+            handed_off_primaries: AHashSet::new(),
             submit_params: AHashMap::new(),
             portfolio: None,
             strategy_event_handlers: IndexMap::new(),
@@ -233,7 +251,7 @@ impl ExecutionAlgorithmCore {
         std::mem::take(&mut self.strategy_event_handlers)
     }
 
-    /// Clears all spawn tracking state.
+    /// Clears spawn ID tracking state.
     pub fn clear_spawn_ids(&mut self) {
         self.exec_spawn_ids.clear();
     }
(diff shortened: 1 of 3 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR. I found two lifecycle cases that need coverage before merge:

- A child can cancel or expire after the remaining primary reaches the venue. Restoration then increases only the cached primary quantity. For example, a primary submitted for 0.5 becomes 1.0 locally, so its actual 0.5 fill leaves it incorrectly marked `PartiallyFilled`.
- The documented late-fill exclusion introduces an over-execution path. A 1.0 TWAP can spawn 0.5, restore that quantity after cancellation, receive the child's late 0.5 fill before the final timer, and still submit the restored primary for 1.0. Total execution becomes 1.5. The previous behavior totals 1.0 for that sequence.

If you're able to revise this, restoration needs to preserve the submitted primary's venue quantity and account for late child fills before restored quantity is reused, with regression coverage for both sequences.

The spawn-method docstrings and execution algorithms guide also need updating: the guide still says acceptance commits the reduction.

folknor:
Thank you for the review - both cases are covered, with regression tests pinning your exact sequences, and fixing the second uncovered two adjacent accounting defects that are also addressed.

## Restoration preserves a submitted primary

`restore_primary_order_quantity` now restores only while the primary is still local (`is_active_local`, the same boundary `modify_order_in_place` enforces); otherwise the pending reduction is dropped and the quantity is untouched, which keeps totals correct since the deduction is already reflected in the submitted quantity. The guard covers `Denied` and `Rejected` as well - the local-only mutation was identical on those paths. `test_spawned_order_canceled_after_primary_submission_does_not_restore_reduction` pins your first sequence: the submitted primary stays at 0.5 and its full fill closes it.

## Late child fills re-deduct restored quantity

Each reduced spawn now carries a lifecycle record (`SpawnReduction { deducted_qty, restored_qty }`). Restoration keeps the record with the gross released amount, and a subsequent fill on that child re-deducts `min(last_qty, restored_qty)` from the primary before the algorithm's fill callback runs, so a restored primary shrinks back before it can be resubmitted. `test_late_spawn_fill_rededucts_restored_primary_before_final_submission` pins your second sequence: the final submission totals 1.0.

Two further interleavings are handled. Restored quantity already reused by a later spawn caps the re-deduction at the primary's remaining quantity and books the excess as debt, which any later restoration for that primary discharges before returning quantity (`test_late_spawn_fill_after_restored_quantity_reused_caps_at_primary_quantity`). And a child whose restoration went to debt keeps its gross budget, so its own late fill recreates the debt instead of escaping the accounting (`test_late_fill_on_child_that_fully_discharged_debt_recreates_debt`, plus the partial-discharge case).

One window remains open: a late fill arriving after the restored primary has itself been submitted cannot be compensated locally, since the submitted quantity is no longer touchable - the same boundary as the first fix. Records for restored children are retained until consumed; cleanup when the primary leaves local control would be a follow-up if you want it.

The spawn docstrings and the execution algorithms guide now describe the terminal-outcome lifecycle, the submitted-primary boundary, and the late-fill netting. All new assertions fail against the previous revision: the two headline tests fail at 1.0 against 0.5, the reuse test hits the unchecked subtraction, and the discharge test fails on the missing record.

cjdsellers:
Thanks @folknor, I can handle rebasing to solve the merge conflicts.

cjdsellers:
@folknor, thanks for the changes and regression tests. I've had another look, and two cases still need coverage before merge:

- `is_active_local()` can remain true after the adapter captures the primary's submission quantity, before its queued `OrderSubmitted` is applied. A child cancellation queued first can still restore the cached primary from 0.5 to 1.0 while the venue receives 0.5. Both restoration and late-fill netting need to respect the submission handoff.
- A quote-sized child can be converted to base quantity before cancellation. Restoration then compares the quote deduction with base-denominated leaves. For example, a 50-quote spawn at price 10 becomes 5 base units; canceling it unfilled restores only 5 quote units. Restoration and late-fill accounting need to preserve the primary's quantity units.

If you're able to revise these, regression coverage for both paths would help close them out.

folknor:
Thank you for the review - both cases are covered, with regression tests pinning your sequences.

## Restoration and netting respect the submission handoff

`is_active_local()` is a status test, and the status only moves when the queued `OrderSubmitted` applies - which is exactly the window you found. The algorithm now marks a primary as handed off synchronously in `submit_order`, immediately before the command is dispatched, and both restoration and late-fill re-deduction require the primary to be unmarked as well as locally active.

A blocked event drops the child's record and discards the primary's late-fill debt, since no permitted restoration can discharge it after handoff. The marker never clears within the order lifecycle.

`test_spawned_order_canceled_during_primary_submission_handoff_does_not_restore` drives your sequence through the real `submit_order` - one dispatched command, the cached primary still `Initialized`, the cancellation handled before any `OrderSubmitted` - and asserts the quantity stays at the venue-received 0.5 with no `OrderUpdated` published; a companion test pins the same window for a late fill after restoration.

## Accounting stays in the primary's quantity units

Each reduction record now captures the spawn's denomination at deduction time. When a venue has converted the child to base units, restoration and late-fill charges are computed proportionally in primary units - `floor(deducted * child / total)` in exact integer arithmetic, widening to a 256-bit intermediate product where the raw multiplication would overflow - rounding down only, so restoration can never exceed the deducted budget. Unit-matched children keep the existing `min` arithmetic unchanged.

`test_converted_quote_spawn_canceled_unfilled_restores_full_quote_quantity` pins your example: a 50-quote spawn converted to 5 base restores the full 50 on cancellation. Companion tests cover the partial fill (2 of 5 base restores 30 quote), the late-fill charge, per-event downward rounding on a non-exact ratio, and repeated fractional fills.

All headline assertions fail against the previous revision: the handoff tests see the primary restored to 1.0 with the marker disarmed, and the conversion tests restore 5 instead of 50 with the unit handling bypassed. The docstrings and the execution algorithms guide now describe handoff as the boundary and the proportional unit conversion.

cjdsellers (approved):
@folknor, thanks for working through the revisions and adding the regression coverage. This looks good to merge once the remaining CI finishes.

I'll follow up on cumulative rounding for converted late fills, reversing tracked debits on `FillVoided`, and cleaning up obsolete lifecycle records. Those can be handled separately.

PR #4969, opened 2026-09-10
Refresh Hyperliquid instruments at runtime

riven314:
# Pull Request

  and, if I used AI,
  this draft
  limitation below

## Summary

A Hyperliquid market listed after process start could not be traded without a restart:
`HyperliquidDataClientConfig.update_instruments_interval_mins` was parsed but never consumed, so the
data client loaded instruments once during `connect()`, and the execution client bootstrapped its
asset-index map once and never refreshed it.

The data client now runs a periodic refresh built on the same shape as the OKX data client
(`0` disables it), diffs the fetched universe against its cache, and publishes only new or
materially changed definitions. The execution client picks those up through the existing
`ExecutionClient::on_instrument` hook. Because Hyperliquid signs orders with a numeric asset index
that `InstrumentAny` did not carry, `create_instrument_from_def` now writes the index onto the
instrument's `info` map, and `HyperliquidHttpClient::cache_instrument` registers it from there;
WebSocket submission signs through the same HTTP client, so one registration serves both paths. An
instrument that arrives without the key logs a warning and leaves the cached index untouched rather
than guessing one.

Each refresh pass also rebuilds the `allDexsAssetCtxs` instrument-id mapping, which is positional
over each perp dex universe: a listing or delisting shifts it even for coins whose own definition is
unchanged, so a stale mapping would misattribute incoming `ctxs` entries.

## Related issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

None. `update_instruments_interval_mins` keeps its existing default of 60 and now takes effect;
set it to `0` to keep the previous load-once behavior.

## Documentation

`docs/integrations/hyperliquid.md` documented both limitations explicitly, so the "Instrument
loading" section and the data client config table are corrected. No PyO3 binding signatures
changed; `crates/adapters/hyperliquid/src/python/http.rs` carries only the regenerated
`cache_instrument` docstring, which mirrors the Rust doc comment.

## Testing

Added to the existing test modules:

- `http/parse.rs`: every market type (main-dex perp, HIP-3 perp, spot, outcome) carries its asset
  index on `info`.
- `http/client.rs`: `cache_instrument` registers the index from `info`, and an instrument without
  the key leaves the map untouched.
- `execution.rs`: `on_instrument` makes the asset index resolvable through the execution client's
  HTTP client.
- `data.rs`: the diff helper reports a newly listed market and a changed tick size, and ignores a
  later `ts_init` on an otherwise identical definition; the refresh task is spawned at a positive
  interval and skipped at `0`.

Validated locally with `make format`, `make pre-commit`, and `make pre-flight`. The pre-flight Rust
suite ran 30,926 tests; the only failures were four pre-existing ones in `nautilus-network` and
`nautilus-polymarket` (WebSocket/HTTP timing tests). Neither crate depends on `nautilus-hyperliquid`
and neither is touched by this branch. Every commit compiles on its own.

### DIFF
--- a/crates/adapters/hyperliquid/src/common/consts.rs
+++ b/crates/adapters/hyperliquid/src/common/consts.rs
@@ -62,6 +62,12 @@ pub const HYPERLIQUID_BUILDER_APPROVAL_DOCS_URL: &str =
 /// Hyperliquid signing chain ID (0x66eee = 421614 decimal).
 pub const HYPERLIQUID_CHAIN_ID: u64 = 421614;
 
+/// Key carrying the venue asset index on an instrument's `info` map.
+///
+/// Order signing needs the numeric asset index, and `InstrumentAny` is the only
+/// shape published on the message bus, so the index travels with the instrument.
+pub const ASSET_INDEX_INFO_KEY: &str = "asset_index";
+
 // Error message substrings for detecting specific rejection reasons
 pub const HYPERLIQUID_POST_ONLY_WOULD_MATCH: &str =
     "Post only order would have immediately matched";
--- a/crates/adapters/hyperliquid/src/config.rs
+++ b/crates/adapters/hyperliquid/src/config.rs
@@ -88,6 +88,8 @@ pub struct HyperliquidDataClientConfig {
     #[builder(default = 3)]
     pub stale_stream_max_targeted_resubscribes: u32,
     /// Interval for refreshing instruments in minutes.
+    ///
+    /// Set to 0 to disable the periodic refresh.
     #[builder(default = 60)]
     pub update_instruments_interval_mins: u64,
     /// WebSocket transport backend (`Sockudo` by default; `Tungstenite` when
--- a/crates/adapters/hyperliquid/src/data.rs
+++ b/crates/adapters/hyperliquid/src/data.rs
@@ -419,7 +419,7 @@ impl HyperliquidDataClient {
         let client_id = self.client_id;
 
         self.session_tasks.spawn(async move {
-            log::debug!("Hyperliquid instrument refresh started, interval={interval:?}");
+            log::info!("Hyperliquid instrument refresh started, interval={interval:?}");
 
             loop {
                 tokio::select! {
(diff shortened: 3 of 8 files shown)


cjdsellers (commented):
Hi @riven314,

Thanks for the PR. This is a useful improvement, and the approach fits the existing OKX pattern. Carrying the asset index in instrument metadata and consuming updates through `on_instrument` makes sense here.

There are two cases that need addressing before this is ready:

- `RequestInstrument` and `RequestInstruments` populate the cache used by the periodic diff without publishing instrument updates. If either request discovers a new market first, subsequent refreshes treat it as unchanged and execution never receives its asset index. Even requesting an existing instrument can trigger this because that path caches the whole universe. Instruments discovered through either path need to reach execution.
- A partial context-mapping refresh can discard working HIP-3 mappings. When `allPerpMetas` fails but the standard `meta` fallback succeeds, the result replaces the entire mapping with the standard dex only. HIP-3 contexts are then skipped until a successful refresh. A partial result needs to preserve working mappings for dexes it could not refresh.

Could you add regression coverage for both sequences to keep these behaviors protected?

There are also two smaller things to address here: the WebSocket count-mismatch guidance still says the mapping is built only at bootstrap and recommends reconnecting, and the new missing-index warning also fires for synthetic vault instruments used for balances, where no signing index is expected.

riven314:
Hi @cjdsellers 

Thanks for the review, all points are addressed.
- **Request paths:** RequestInstrument and RequestInstruments now go through the same reconcile step as the refresh. Newly discovered markets are published, with their asset index, before the response.
- **Partial mapping refresh:** the allDexsAssetCtxs mapping now merges by dex in both the WebSocket client and the handler, so a meta fallback keeps the existing HIP-3 entries.
- **Smaller points:** I corrected the count-mismatch guidance and skipped the missing-index warning for vault tokens.

Since requests and the refresh can now overlap, fetch-and-apply passes run one at a time under a per-client lock. Without that, an older fetch that finished later could republish a stale definition.

Regression tests cover both sequences you described, plus the concurrent case. Each one fails without its fix.

cjdsellers:
Thanks for the follow-up @riven314, I've had another look at the latest changes and everything from my review is addressed. Instruments discovered through `RequestInstrument` and `RequestInstruments` now publish before the response, the per-dex context mappings survive a partial refresh, and the docs and vault-token warning are corrected.

The regression coverage for each of those sequences is good to have, and serializing the fetch-and-apply passes so an older fetch can't republish over a newer one was a solid addition beyond what I'd raised.

This looks good to merge 👍

cjdsellers (approved):

PR #4974, opened 2026-09-11
Restore GTD expiry management after a rejected cancel

folknor:
A cancel request that the venue rejects currently leaves an open GTD order with no managed expiry.

## A rejected cancel strips the order of its expiry timer

With `manage_gtd_expiry` enabled, `Strategy::cancel_order` cancels the GTD expiry timer when it dispatches the cancel command. If the venue rejects the cancel, `OrderCancelRejected` restores the order's previous status, so the order is open again - but nothing re-arms the timer, and the order can rest past its expire time indefinitely on venues that do not enforce GTD natively.

The fix extends the restoration pattern the reopened `OrderFillVoided` path already uses: `handle_order_event` re-arms via `set_gtd_expiry` on `CancelRejected` when the cached order is open and no timer exists. When the expire time has already passed, `set_gtd_expiry`'s existing branch dispatches a fresh cancel instead of installing a timer, so an expired order is re-attempted rather than left unmanaged; each further attempt requires a new rejection event, and the pending-cancel guard in `cancel_order` suppresses duplicate dispatch while a cancel is outstanding.

The restoration runs before the user callback, so the already-expired retry can return the cached order to `PendingCancel` before `on_order_cancel_rejected` observes it. A stopped strategy does not restore; `reactivate_gtd_timers` covers open GTD orders on the next start.

- Deferring the timer drop in `cancel_order` until `OrderCanceled` confirms: changes event-ordering semantics for every cancel, managed or not, and leaves a window where a firing timer races the in-flight cancel.
- Restoring outside the running-state gate so stopped strategies also re-arm: a stopped strategy would schedule timers and dispatch cancels nobody manages, and the restart path already re-derives timers from the cache.
- Damping the reject-restore-cancel repetition for expired orders: adds retry policy and state for a cycle that is already bounded by the venue's own rejection events.

## Testing

`test_handle_cancel_rejected_restores_future_gtd_timer` pins the restored timer entry; `test_cancel_rejected_restores_timer_after_cancel_order_removes_it` runs the full sequence - dispatch removes the timer, the rejection applied to the cache restores the order status, the event restores the timer. `test_restored_gtd_timer_event_cancels_order` advances the test clock to expiry and routes the emitted time event, asserting the restored timer is consumed and exactly one cancel command is emitted. `test_handle_cancel_rejected_retries_already_expired_gtd_order` pins the widened case: no timer, one new cancel command, cached order pending cancel again. `test_handle_cancel_rejected_preserves_existing_gtd_timer` and its already-expired variant pin that an existing timer is never reset - the variant fails if the existing-timer guard is removed. Five negative cases cover missing, closed, and non-GTD orders, management disabled, and a stopped strategy. All tests use the paused test clock; no wall-clock waits. The new positive assertions fail against the unfixed implementation.

### DIFF
--- a/crates/trading/src/strategy/mod.rs
+++ b/crates/trading/src/strategy/mod.rs
@@ -1357,6 +1357,22 @@ pub trait Strategy: DataActor {
             return;
         }
 
+        if matches!(&event, OrderEventAny::CancelRejected(_)) {
+            let order = StrategyNative::strategy_core_mut(self)
+                .cache_ref()
+                .order(&client_order_id)
+                .map(|order| order.clone());
+            if let Some(order) = order
+                && (order.is_open() || order.is_inflight())
+                && !self.has_gtd_expiry_timer(&client_order_id)
+                && let Err(e) = self.set_gtd_expiry(&order)
+            {
+                log::error!(
+                    "Failed to restore GTD expiry for cancel-rejected order {client_order_id}: {e}"
+                );
+            }
+        }
+
         if matches!(&event, OrderEventAny::FillVoided(event) if event.is_reopened) {
             let order = StrategyNative::strategy_core_mut(self)
                 .cache_ref()
@@ -2483,8 +2499,8 @@ mod tests {
         events::{
             OrderAccepted, OrderCanceled, OrderFilled, OrderRejected, PositionAdjusted,
             order::spec::{
-                OrderAcceptedSpec, OrderCanceledSpec, OrderEmulatedSpec, OrderExpiredSpec,
-                OrderFillVoidedSpec, OrderFilledSpec, OrderRejectedSpec,
+                OrderAcceptedSpec, OrderCancelRejectedSpec, OrderCanceledSpec, OrderEmulatedSpec,
+                OrderExpiredSpec, OrderFillVoidedSpec, OrderFilledSpec, OrderRejectedSpec,
             },
         },
         identifiers::{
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hi @folknor,

Thanks for the PR. The restore path mirrors the reopened fill-void pattern cleanly, and the test coverage pins the behavior well.

One reachable race keeps the same defect alive for orders that haven't been accepted yet. `set_gtd_expiry` arms the timer at submit time, and `cancel_order` drops it at dispatch for any GTD order, including one still in `Submitted`. If the venue rejects that cancel, `cancel_rejected` restores `Submitted` as the previous status, so the new `order.is_open()` guard in `handle_order_event` skips the restore and the timer is gone. When the venue then accepts the order (the FSM explicitly allows `PendingCancel` -> `Accepted` for failed cancel requests), nothing re-arms on `OrderAccepted`, and with `use_gtd=False` the venue won't enforce the expiry either. The accepted order rests past its expire time with no managed cancel, the same outcome this fixes for already-accepted orders.

The restore needs to cover the inflight case as well: include `Submitted` in the restore condition, or re-arm on `Accepted` when no timer exists, with a test for the cancel-rejected-while-submitted sequence. Everything else here looks good to merge once that's covered.

folknor:
Thank you for the review - the inflight case is covered.

The restore condition is now `order.is_open() || order.is_inflight()`, which adds exactly the `Submitted` case while keeping emulated orders excluded (`is_inflight` carries the same emulation-trigger exclusion as `is_open`). I took this shape rather than an `Accepted`-side re-arm because acceptance can arrive while a deliberate cancel is still in flight (`PendingCancel -> Accepted`), and re-arming there would let the expiry callback dispatch a second cancel - the situation dropping the timer at dispatch exists to prevent.

`test_cancel_rejected_while_submitted_restores_timer_through_acceptance` pins your sequence: submit with timer, cancel drops it, the rejection restores `Submitted` and re-arms, and the subsequent acceptance leaves the single timer in place with no further cancel dispatched. An already-expired variant asserts the restore path dispatches an immediate cancel instead of arming, the seeded-timer variant asserts no new clock alert is created when a timer already exists, and `Emulated` joins the ineligible cases. The two positive tests fail against the previous revision.

The opposite delivery order remains uncovered: when the acceptance overtakes the rejection, the FSM has no `Accepted -> CancelRejected` transition and the engine does not publish an event it could not apply, so the strategy never observes the rejection. Covering that needs the rejection preserved across the FSM or published despite failing to apply.

cjdsellers (approved):
@folknor, thanks for covering the inflight case. I've had another look, and this looks good to merge. I'll address the pre-existing acceptance-before-rejection race in a separate FSM fix and update the managed-GTD docs to cover restoration after a cancel rejection and the immediate retry when expiry has already passed.

Issue #4952, opened 2026-09-08
BacktestNode.add_strategy_from_config before build() fails with "No engine for run config"


Artur-Sulej:
# Bug Report

## Confirmation

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I've tested this issue using a recent pre-release or development wheel (`2.0.0rcN`, `dev` develop, or `a` nightly) and can still reproduce it.

Tested on `2.0.0rc4`.

## Expected behavior

Registering a strategy on a `BacktestNode` either works regardless of call ordering, or fails with an error that names the required ordering.

## Actual behavior

`add_strategy_from_config` requires `build()` to have been called first. If it has not, the call fails with a message that points at the run config id rather than at the missing `build()`:

```
RuntimeError: No engine for run config 'repro-07'
```

That message reads like the id is wrong, a typo or a mismatch against `BacktestRunConfig.id`, which is where I looked first. The id is in fact correct. Engines are constructed lazily in `build()`, so before that call there is nothing to attach the strategy to.

## Steps to reproduce

```python
from nautilus_trader.backtest import BacktestNode, BacktestRunConfig, BacktestVenueConfig
from nautilus_trader.model import AccountType, BookType, OmsType
from nautilus_trader.trading import ImportableStrategyConfig

run = BacktestRunConfig(
    venues=[BacktestVenueConfig(
        name="SIM", oms_type=OmsType.NETTING, account_type=AccountType.CASH,
        book_type=BookType.L1_MBP, starting_balances=["1_000_000 USD"],
    )],
    data=[],
    id="repro-07",
)
cfg = ImportableStrategyConfig(
    strategy_path="my_module:MyStrategy",
    config_path="nautilus_trader.trading:StrategyConfig",
    config={},
)

node = BacktestNode([run])
node.add_strategy_from_config("repro-07", cfg)   # RuntimeError: No engine for run config 'repro-07'

# Works only in this order:
node = BacktestNode([run])
node.build()
node.add_strategy_from_config("repro-07", cfg)
```

## Code snippets or logs

```
--- add_strategy_from_config BEFORE build() ---
  RuntimeError: No engine for run config 'repro-07'

--- add_strategy_from_config AFTER build() ---
  (proceeds to strategy import)
```

[...]

PR #4977, opened 2026-09-12
Clarify backtest data type and missing engine errors

### LINKED ISSUE #4951: BacktestDataConfig rejects TradeTick.fully_qualified_name() and the error does not list valid values
# Bug Report

## Confirmation

Tested on `2.0.0rc4`.

## Expected behavior

Either `BacktestDataConfig(data_type=...)` accepts the identifier the Python API itself advertises for a data type, `TradeTick.fully_qualified_name()`, or the resulting error states which values are accepted.

## Actual behavior

`fully_qualified_name()` is rejected, and the error does not say what a valid value looks like. Only the bare class name works.

## Steps to reproduce

```python
from nautilus_trader.backtest import BacktestDataConfig
from nautilus_trader.model import TradeTick, InstrumentId

fqn = TradeTick.fully_qualified_name()
print(fqn)   # 'nautilus_trader.model:TradeTick'

iid = [InstrumentId.from_str("AAA.SIM")]
for dt in (fqn, "TradeTick", "trades", "trade_tick"):
    try:
        BacktestDataConfig(data_type=dt, catalog_path="/tmp/x", instrument_ids=iid)
        print(f"ACCEPTED {dt!r}")
    except Exception as e:
        print(f"REJECTED {dt!r}: {type(e).__name__}: {e}")
```

## Code snippets or logs

```
'nautilus_trader.model:TradeTick'
REJECTED 'nautilus_trader.model:TradeTick': ValueError: Invalid `NautilusDataType`: 'nautilus_trader.model:TradeTick'
ACCEPTED 'TradeTick'
REJECTED 'trades': ValueError: Invalid `NautilusDataType`: 'trades'
REJECTED 'trade_tick': ValueError: Invalid `NautilusDataType`: 'trade_tick'
```

The accepted set is the `NautilusDataType` enum in `crates/persistence/src/python/backend/session.rs`: `OrderBookDelta`, `OrderBookDepth10`, `QuoteTick`, `TradeTick`, `Bar`, `MarkPriceUpdate`, `OptionGreeks`, `InstrumentStatus`.

## Why this is confusing

Three plausible spellings are in play and only one works:

* `fully_qualified_name()` is a public method on the data types and the natural thing to reach for when a config asks for a data type as a string. It is also close to the v1 `data_cls` convention, so it is what a migrating user will try first.
* `"trades"` is the on-disk directory name used by the catalog layout and by the catalog query APIs.
* `"TradeTick"` is the one the config wants.

Another user hit the same confusion independently and documented the workaround as a code comment in the reproducer in #4933:

> `NOTE: data_type takes the NautilusDataType spelling "Bar"; the catalog query APIs take the on-disk directory name "bars".`

## Suggested fix

The minimal fix is to make the error actionable by listing the accepted values, for example:

```
Invalid `NautilusDataType`: 'nautilus_trader.model:TradeTick'
(expected one of: OrderBookDelta, OrderBookDepth10, QuoteTick, TradeTick, Bar,
 MarkPriceUpdate, OptionGreeks, InstrumentStatus)
```

Optionally, also accept a fully qualified name by matching on the trailing segment after `:`. That is a design call for maintainers; the error message improvement stands on its own either way.

## Specifications

- Python version: 3.13.15
- Installed from: PyPI wheel
- Adapter/venue: n/a

abhijeetvichare76:
## Two backtest configuration errors that do not say what went wrong

Closes #4951 and #4952, both reported against `2.0.0rc4`.

`BacktestDataConfig(data_type=...)` accepts only the bare `NautilusDataType` variant name. The error for anything else was `Invalid `NautilusDataType`: '<value>'` with no hint of that, so `TradeTick.fully_qualified_name()` and the catalog directory name `trades` both failed without saying what would have worked. Parsing in `crates/backtest/src/config.rs` now iterates the enum with strum `EnumIter`, as the rest of the workspace does, and the error names the accepted values. I have not made the parser accept a fully qualified name; the reporter left that as a maintainer call, and it is a one-line change on top of this if wanted.

`BacktestNode` builds its engines lazily in `build()`. Calling `add_strategy_from_config` or a post-run inspection method before that failed with `No engine for run config '<id>'`, which reads as an ID mismatch. The lookup in `crates/backtest/src/python/node.rs` now goes through one helper that says which case applies: an unknown ID, listing the known ones; or a known ID with no engine, which tells the caller to call `build()` and that a build which already ran may have failed (see the log) or been disposed. Those last states all leave the engine map empty, so the message covers them together rather than adding state to the node. The built-in strategy path uses the same helper instead of its own copy of the message.

## Testing

- `test_nautilus_data_type_from_str_round_trips_every_variant` and `test_nautilus_data_type_from_str_error_lists_expected_values` (fully qualified name and catalog directory cases) in the config unit tests.
- Python: `test_data_config_invalid_data_type` asserts the full message for three invalid inputs; `test_node_missing_engine_explains_build_requirement` covers not built, suppressed build failure and raised build failure for `get_engine_cache` and `add_strategy_from_config`; `test_node_inspection_after_dispose_explains_missing_engine` covers `dispose()`; the unknown-ID test also checks the known-ID list. `pytest tests/unit/backtest/test_backtest_node.py tests/unit/backtest/test_backtest_configs.py` against a local debug build: 77 passed.
- `cargo test -p nautilus-backtest --lib config::`: 49 passed; the engine tests fail under plain `cargo test` from the process-global logger shutdown regardless of this change and pass in isolation.
- `cargo clippy -p nautilus-backtest --no-deps --all-targets --all-features -- -D warnings`: clean
- `prek run --files` on the changed files: pass apart from the make-dependent collection hook


cjdsellers (changes requested):
Hi @abhijeetvichare76,

Thanks for the PR. There’s one remaining case in `missing_engine_err`: when `build()` fails with `raise_exception=False`, the engine collection stays empty, so the next call says “call build() first” even though it already ran. `BacktestNode` currently allows only one run config, which also makes the “engine failed to build” branch unreachable.

Could you make the message account for failed builds and add a Python regression for that case? The existing `test_build_respects_raise_exception` covers the underlying Rust state.

abhijeetvichare76:
Both cases now share one message: `No engine for run config '<id>': call build() before accessing the engine; if build() already ran, it may have failed (check the log) or the engine was disposed`. The engine map is empty in each of those states, so the node cannot tell them apart without new state, and I did not want to add a flag to `BacktestNode` just for a message. The unreachable "failed to build" branch is gone with it.

`test_node_missing_engine_explains_build_requirement` covers not built, a suppressed build failure (`raise_exception=False`) and a raised one, for both `get_engine_cache` and `add_strategy_from_config`; `test_node_inspection_after_dispose_explains_missing_engine` covers `dispose()`. The unknown-ID test now also checks the known-ID list.

Separately, the hand-written `ALL` array is replaced by strum `EnumIter`, which the workspace already uses for this, so the accepted-values list cannot drift from the enum. Rebased onto develop.

cjdsellers:
Thanks for addressing the failed-build case and adding the regressions, @abhijeetvichare76. I've had another look at the latest changes, and this looks good to merge 👌

cjdsellers (approved):

PR #4982, opened 2026-09-12
Gate report task completions during shutdown grace

folknor:
A shutdown-lifecycle gap in `LiveNode`'s continuous reconciliation, on the completion side of the report tasks whose start side already yields to shutdown.

## Report-task completions during shutdown grace

`initiate_shutdown` stops the trader, sets the grace deadline, and cancels none of the three in-flight report tasks; the maintenance path then refuses to start a new open-order or position check under `ShuttingDown`, but the completion arms in the run loop gate only on task presence. A task completing inside the grace window therefore reconciled unconditionally - `reconcile_open_order_reports`, `reconcile_targeted_order_reports`, and `handle_position_report_result` mutated the cache and published inferred events after the trader had stopped - and the open-order arm could start a fresh targeted-report task inside the window, bypassing the guarded start path. The start-side guards and the ungated completions arrived in the same commit (f8c103795512).

Each `Completed` arm now consults the loop iteration's shutdown snapshot: during grace the result is discarded and the cancellation cleanup runs (planned-ID removal for targeted results, deferred-instrument flush for all), and no follow-up task starts. A completion winning an iteration that began `Running` was decided before that iteration could initiate shutdown and is processed exactly as before. `TimedOut` arms and the post-loop task cancellation are unchanged: timeout cleanup is lifecycle housekeeping and remains correct during shutdown, and queued residual events keep draining as they always did.

## Testing

Seven integration tests in the existing blocking-report-client idiom, all on paused Tokio time with notify-released responses and no wall-clock waits: discards for all four completion variants (open-order, targeted, position, position-fill), positive controls for pre-shutdown open-order and fill reconciliation, timeout cleanup during shutdown, and final-deadline cleanup. Each driver waits on a response-returned acknowledgment from the fixture, so a future change that stops polling completions during grace fails the wait rather than passing vacuously.

The four discard assertions fail against the unfixed implementation: the open-order and targeted orders become `Canceled` instead of staying `Accepted`, the position arm issues its fill-report request, and the fill arm books `PartiallyFilled`. Removing the discard's planned-ID cleanup fails the targeted test on a post-run `check_inflight_orders` probe that a leaked targeted-query marker suppresses.


cjdsellers:
Hey @folknor,

Thanks for the PR. I agree that shutdown should prevent new reconciliation checks, including follow-up queries, but I'd prefer to keep applying completed results where no further query is needed. The grace period allows residual events to arrive, and stopping the trader doesn't mean the cache should stop receiving valid fills or order status updates.

Would you be able to narrow this to preventing follow-up queries, while preserving completed-result processing and the existing timeout and deadline cleanup? Position reports that need another fill query can remain unapplied during shutdown. That keeps the existing behavior where possible without discarding results that are already safe to process.

folknor:
Thank you - narrowed as requested. Completed results now apply during the grace window and only follow-up queries are suppressed; the position arm stays deferred, since its only action is to start a fill query. Dropping the open-order arm's targeted-query plan also removes its `targeted_order_queries` markers, which are registered when the plan is built and would otherwise suppress the inflight-check ladder for those orders indefinitely.

During grace no later check corrects a wrong inference: `reconcile_open_order_reports` synthesizes inferred fills when a report's `filled_qty` disagrees with the local order, and a snapshot fetched before shutdown, applied while stop-time cancels and residual stream fills are still landing, can persist that inference as the session's final state - a discarded result would have been re-derived from venue truth at the next start. The recent-local-activity threshold covers part of this window, not a fill still in flight on the stream.

`ShuttingDown` now selects three behaviors at once: the start side refuses new checks, completions apply without querying, and timeout housekeeping runs. If this surface grows, a distinct draining state between running and stopped may name it better.

cjdsellers (approved):
Thanks for narrowing this, @folknor. I've had another look and the code looks good to merge 👌

Let's keep the inference concern separate from this fix; changing which completed reports are safe to apply needs its own reconciliation review.

**Small note:** the PR description no longer matches this changeset. Might be worth updating, even after merge.

PR #4986, opened 2026-09-12
Add pyarrow to the test dependency group

clementbvt69:
python/tests/unit/persistence/test_persistence.py imports pyarrow
unconditionally (line 26, used once at line 773 for
pa.ipc.open_stream(...).read_all()), but pyarrow is declared neither in
python/pyproject.toml nor in python/uv.lock. On a clean checkout,
`make sync && make pytest` aborts at collection with
"ModuleNotFoundError: No module named 'pyarrow'", so the entire Python
suite reports no counts.

Add pyarrow>=21.0.0 to [dependency-groups] test (it is test-only; the
other reference, tests/unit/test_macos_extension.py:79, already uses
pytest.importorskip), and add pyarrow to [tool.uv] no-build-package so
scripts/check-no-build-packages.sh stays in sync with the lock.

make pytest now exits 0: 5124 passed / 13 skipped, plus 38 passed for
tests/unit/test_live_node.py.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01F13tWCSvJ8PFuzC1oi5qkz

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -112,6 +112,7 @@ no-build-package = [
   "plotly",
   "pluggy",
   "psutil",
+  "pyarrow",
   "pygments",
   "pytest",
   "pytest-asyncio",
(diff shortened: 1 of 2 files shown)


cjdsellers:
Hi @clementbvt69,

Thanks for the PR. I've pushed aeed6b6cf166dca7d7cafaa63cc5f84aec72ef09 to `develop`, which removes the `pyarrow` import from that test instead of adding it as a dependency, so I'll close this.

We haven't decided between pyarrow and polars for Python-side Arrow support yet.

PR #4988, opened 2026-09-13
Retire a strategy whose component the registry has lost

folknor:
`Trader::remove_strategy` is the only public door to strategy retirement - `retire_strategy`, `remove_strategy_subscriptions`, `release_component` and `dispose_registered_component` are all private - and it fails in one state.

## The retirement path errors when the component registry has no entry

`retire_strategy` calls `dispose_registered_component`, which opens with `component_state(&id)?`. That returns `Err` when the registry has no entry for the id, so the `?` propagates and the function returns before clearing the external order claims, removing the msgbus order and position handlers and the strategy control endpoint, releasing the component clock, deregistering either registry, or dropping the id from `strategy_ids`. The id then cannot be retired or redeployed for the life of the process. `retire_actor` and `retire_exec_algorithm` open with the same call and have the same property.

The fix tests registry presence with `get_component` and skips only the disposal transition when the component is absent, leaving the caller's own cleanup to run. Presence is tested directly rather than inferred from a `component_state` failure, because that error channel does not distinguish an absent component from one that is already mutably borrowed, and a borrowed component is still live.

This extends the same progression as 60208b0c11, which added the `Disposed` short-circuit so a second disposal transition could not strand the trader's bookkeeping, and e3625c583d and 069566dafb, which did the same for `Faulted`.

## Execution algorithm retirement has a second obstruction at the same point

`retire_exec_algorithm` strands at the next statement instead. It runs the stored cleanup closure, which resolves the algorithm through `try_get_actor_unchecked` and returns `Execution algorithm {id} not found while cleaning subscriptions` when the actor entry is also gone, so retirement still fails one line later. The actor-dependent cleanup is now skipped when the actor is absent, while the execute endpoint, clock, registry entries, tracked id and the restore and cleanup maps are removed as before. Cleanup errors that are not about absence still propagate.

## What this does not do

The retirement releases what the trader installed. Subscriptions the component installed on its own behalf are not released by this path, because `Component::release_subscriptions` needs the object. A surviving handler of that kind resolves the id, so it can dispatch into a later component registered under the same id. Absence in the component registry also does not prove the object has been dropped, since an actor entry or an external handle may still hold it.

No contract-respecting runtime path in this repository produces the state. `Trader::add_strategy` registers one `Rc` into both registries through `register_component_actor`, and `release_component` deregisters both together, so nothing removes a component entry alone. It is reachable through public API: `deregister_component` removes the component entry on its own, and `add_strategy_id_with_subscriptions` and its actor and exec-algorithm siblings document the prerequisite that the object is already in both registries while enforcing only the already-tracked and order-id-tag checks, so a caller that violates it installs handlers and tracks an id for an object in neither registry.

## Testing

`test_remove_strategy_retires_trader_state_when_component_is_missing` establishes the registrations, removes the component entry, and asserts the retirement clears the claims, the handlers, the control endpoint, the clock, both registries, the tracked id and the callback and stop-fn maps, that a strategy sharing the same event topics is untouched, and that the id can be registered again. It then publishes an order event to the redeployed strategy and asserts exactly one delivery, which fails if the retired strategy's handler survived.

`test_remove_strategy_preserves_bookkeeping_when_component_is_borrowed` drives a re-entrant disposal so the component is present and mutably borrowed, and asserts the call still errors, leaves the strategy `Faulted`, and preserves every registration.

`test_remove_actor_retires_trader_state_when_component_is_missing` and `test_clear_exec_algorithm_retires_trader_state_when_component_is_missing` cover the other two retirement paths, the second parameterized over whether the actor entry survives alongside the component entry.

All four missing-component assertions fail against the previous revision with `Component '...' not found in global registry`. Reverting the execution algorithm's actor guard so cleanup always runs fails the actor-absent case on the closure's own error; disabling that cleanup entirely fails the actor-present case on a stale delivery to the redeployed algorithm. Replacing the presence test with a match that treats any `component_state` error as absence fails the borrowed case, which is what keeps a live borrowed component from being retired.

### DIFF
--- a/crates/system/src/trader.rs
+++ b/crates/system/src/trader.rs
@@ -28,12 +28,12 @@ use nautilus_common::python::wrappers::release_python_wrapper;
 use nautilus_common::{
     actor::{
         DataActor, DataActorNative,
-        registry::{deregister_actor, try_get_actor_unchecked},
+        registry::{actor_exists, deregister_actor, try_get_actor_unchecked},
     },
     cache::Cache,
     clock::Clock,
     component::{
-        Component, component_state, deregister_component, dispose_component,
+        Component, component_state, deregister_component, dispose_component, get_component,
         register_component_actor, release_component_subscriptions, reset_component,
         start_component, stop_component,
     },
@@ -1415,10 +1415,15 @@ impl Trader {
     ///
     /// # Errors
     ///
-    /// Returns an error if the strategy is not registered, the cache is already borrowed, or
-    /// disposal fails. A cache borrow failure preserves the strategy registration and its external
-    /// order claims. A failed disposal keeps the strategy registered and tracked, and leaves it
-    /// `Faulted`; see [`Component::dispose`]. Calling this again retires the strategy.
+    /// Returns an error if the strategy is not tracked by this trader, the cache is already
+    /// borrowed, or disposal fails. A cache borrow failure preserves the strategy registration and
+    /// its external order claims. A failed disposal keeps the strategy registered and tracked, and
+    /// leaves it `Faulted`; see [`Component::dispose`]. Calling this again retires the strategy.
+    ///
+    /// A strategy the trader still tracks whose component the global registry no longer holds is
+    /// retired rather than rejected: the component is unreachable through the component API, and
+    /// the registrations this trader installed are still owed their cleanup. Subscriptions the
+    /// component installed on its own behalf are not released by this path, which needs the object.
     pub fn remove_strategy(&mut self, strategy_id: &StrategyId) -> anyhow::Result<()> {
         if !self.strategy_ids.contains(strategy_id) {
             anyhow::bail!("Cannot remove strategy, {strategy_id} not found");
@@ -1478,7 +1483,13 @@ impl Trader {
     /// Disposes an execution algorithm, then releases everything its registration created.
     fn retire_exec_algorithm(&mut self, exec_algorithm_id: ExecAlgorithmId) -> anyhow::Result<()> {
         Self::dispose_registered_component(exec_algorithm_id.inner())?;
-        self.cleanup_exec_algorithm_subscriptions(exec_algorithm_id)?;
+        if actor_exists(&exec_algorithm_id.inner()) {
+            self.cleanup_exec_algorithm_subscriptions(exec_algorithm_id)?;
+        } else {
+            log::debug!(
+                "Execution algorithm {exec_algorithm_id} not found, skipping subscription cleanup"
+            );
+        }
 
         let endpoint: Ustr = format!("{exec_algorithm_id}.execute").into();
         msgbus::deregister_any(endpoint.into());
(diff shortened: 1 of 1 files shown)


cjdsellers:
Hey @folknor,

Thanks for the PR. Recovering the trader's bookkeeping seems reasonable, but I'd like to settle the recovery contract before reviewing this further. When would you expect this state to occur in practice, given that the supported registration and removal paths keep both registries together?

Should removal report success and allow the same ID to be reused when component-owned subscriptions can still deliver events to the replacement? The missing registry entry also doesn't establish that the original object has stopped.

I think revisiting the registry design may be the better starting point, with a tighter structure and an explicit contract for registration, ownership, and retirement. That would be maintainer-owned work; I'm raising it here to establish which inconsistent states recovery should support before extending the removal behavior.

folknor:
Thank you - settling the contract first sounds right, and revisiting the registry design as maintainer-owned work supersedes this shape; I am happy to hold the PR, rework it against whatever contract you settle on, or close it in favor of that work.

On when the state occurs: not through the supported paths - `Trader::add_strategy` and `release_component` keep both registries together, as the body says. It arises from an embedding application driving the public registry API directly: `deregister_component` removes the component entry alone, and `add_strategy_id_with_subscriptions` and its actor and exec-algorithm siblings document the both-registries prerequisite while enforcing only the already-tracked and order-id-tag checks. When an actor outside the trader's own cleanup damages the registry this way, the id can be neither retired nor reused for the life of the process, and the only exit is a restart.

The case this PR argues is completing the removal path for a state the public API can construct, not one the runtime produces.

On reporting success and allowing reuse: the trader-installed registrations are fully removed, but subscriptions the lost object installed on its own behalf are not, and such a handler can dispatch into a replacement under the same id - the exec-algorithm redeployment test observes exactly that delivery. And absence in the component registry does not establish the object stopped; an actor entry or an external handle may still hold it. Plain `Ok(())` with a debug log leaves the caller to accept that residual implicitly, so if recovery of this state is supported at all, a distinguished outcome (or refusing reuse until the process can prove the object is gone) may be the better contract.

cjdsellers:
Thanks for the detailed follow-up, it's sharpened the contract question.

It's also surfaced a philosophical difference worth calling out. This PR extends in-process dynamic teardown, retiring and redeploying components while the node runs, which is where the hard questions about id reuse and surviving handlers come from. The model I want to keep optimizing for is crash-only and fail-fast: a node comes up with its configuration, and larger config or componentry changes bring the node down rather than mutate a live process. That's simpler to reason about, and it's the direction the design docs already describe: [crash-only design](https://github.com/nautechsystems/nautilus_trader/blob/develop/docs/concepts/architecture.md#crash-only-design) and the fail-fast and recovery sections of the [design principles](https://github.com/nautechsystems/nautilus_trader/blob/develop/docs/developer_guide/design_principles.md). Incoherent registry state should be prevented, or fail loudly at the point it's introduced, rather than recovered from.

So for now we'll prioritize the simpler approach: the registry work I mentioned earlier, with a tighter structure and an explicit contract for registration, ownership, and retirement. Your analysis maps the failure surface well, and the points about enforcing the documented prerequisites and failing fast at the point of damage feed straight into that. We may revisit a dynamic teardown capability later, and this thread will be a useful reference when we do.

I'll close the PR for now rather than leave it parked. Thanks again for the careful investigation and tests, this made the trade-off much clearer.

PR #4991, opened 2026-09-14
Reconcile Hyperliquid execution state on reconnect

riven314:
# Pull Request

  and, if I used AI,
- [ ] This change is complete, locally validated, and ready for review, or a maintainer requested
  this draft
  limitation below

## Summary

Hyperliquid does not replay a snapshot when `orderUpdates` and `userEvents` resubscribe. An accept,
cancel, or fill received while the execution WebSocket is disconnected can therefore be absent
from the engine indefinitely when periodic open-order checks are disabled.

The execution client now waits until both execution subscriptions are acknowledged, then fetches
`userFills` and `historicalOrders` over REST. It keeps rows from the existing five-minute recovery
window only when their CLOID maps to an order still tracked by this client. Filtering happens
before instrument conversion, so unrelated account history cannot fail the sweep because an
instrument is unavailable locally.

Selected fills are replayed oldest first and before order statuses through the same per-report path
as live WebSocket messages. This preserves the CLOID mapping until fills have been processed.
Existing deterministic trade-ID deduplication suppresses duplicate `OrderFilled` events when
recovery overlaps the resumed stream.
Failed sweeps retain the existing limit of four attempts with exponential backoff, and reconnect
bursts retain the existing 20-second minimum interval.

REST fill parsing now keeps the venue CLOID out of the domain `ClientOrderId` field. Reconnect
reconciliation resolves the raw CLOID through the WebSocket cache only after confirming that its
Nautilus order context is still tracked. There are no new query types, configuration options, or
Python binding changes.

### Open review items

The following maintainer review concerns remain open and are not claimed as resolved by this
narrow update:

- Duplicate or stale fills must not trigger bracket side effects.
- A cancellation racing REST recovery must not suppress an earlier unseen fill.
- Replayed fills from an older venue leg must not consume a pending modify intent.
- Older accepted reports must not restore a superseded venue binding.

The readiness checkbox remains unchecked until these behaviors are resolved or explicitly scoped
out with the maintainer.

## Related issues/PRs

None.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

None.

## Documentation

`docs/integrations/hyperliquid.md` documents the subscription cutover, fixed recovery window,
replay ordering, filtering, and retry behavior. No PyO3 bindings or wrapped Rust docs changed, so
stub regeneration does not apply.

## Testing

Added or strengthened coverage for:

- Exact real-wire CLOID fixture values and isolation of venue CLOIDs from domain client IDs.
- Filtering untracked, unavailable-instrument history before conversion.
- Resolution of tracked CLOIDs and five-minute-window filtering.
- Oldest-first fill replay with all fills ahead of order statuses.
- Waiting for both execution subscription acknowledgements for the same account.
- Bounded retry behavior using the existing `userFills` and `historicalOrders` requests.
- Existing dispatch deduplication for a fill replayed after being seen live.

Validated on the final branch with:

- `make format`
- `make pre-commit` (all hooks passed)
- `cargo test -p nautilus-hyperliquid --quiet` (1,160 passed; 8 ignored)
- `cargo clippy -p nautilus-hyperliquid --all-targets -- -D warnings`


cjdsellers (commented):
Hi @riven314,

Thanks for the PR. I support reconnect recovery and the reuse of the existing report path, but replay still has several correctness gaps:

- A child fill of 0.6 against quantity 1.0 is correctly deduplicated on replay, yet the surrounding bracket code counts it again when deciding whether to cancel the sibling.
- A cancellation received while REST recovery runs can remove the context and CLOID mapping needed to recover an earlier missed fill. If the fill is already selected, the terminal marker suppresses it instead.
- Replaying a processed fill from an older venue leg during a pending modify can promote that old leg and consume the modify intent before deduplication.
- An older REST Accepted report can also restore a superseded venue binding after the live stream has already applied its replacement.

Before merging, replay needs to preserve unseen fills after cancellation and leave bracket actions, modify intents, and newer venue bindings unchanged when processing duplicates or superseded reports. Regression coverage needs to exercise the complete report-processing path and overlap between REST recovery and live events.

riven314:
Hi @cjdsellers 

Thanks for the careful review, all four points were valid. They had the same root cause: replayed reports could change live state and trigger bracket actions before we knew whether they were new.

**Changes**
- Duplicate fills cause no side effects. Dispatch now reports whether a fill was actually applied. Sibling cancels and resizes, child activation, modify handling and venue-binding promotion only follow an applied fill.
- Cancels during recovery don't lose missed fills. On reconnect we take a snapshot of each tracked order's ID, context and filled quantity, and the REST sweep uses it. Missed fills are still applied after a live cancel, and trade-id dedup still applies.
- Old venue legs can't win. Each order remembers its current venue order ID plus the ones it replaced. A report for a replaced or older leg is ignored, which covers both the old-leg fill and the stale REST Accepted.

Tests go through the full report processing path, including bracket actions, and cover:
- a replayed child fill leaving the sibling alone
- a live cancel during the REST sweep, where two missed fills are still applied in order through the recovery batch
- an old-leg fill during a modify, both already seen and new
- an older Accepted arriving after its replacement

I also fixed a related edge case: when a parent fill is held until the replacement ACCEPTED arrives, its bracket children still activate.

cjdsellers:
@riven314, thanks for the follow-up and regression coverage. I've had another look, and I think the reconciliation boundary needs consideration before adding more recovery bookkeeping.

**Reconciliation design**

The engine already reconciles order snapshots with fills against cached order state. Can reconnect recovery use that path, with the adapter handling subscription readiness, fetching reports, and the Hyperliquid-specific state that must remain local?

Bracket handling and pending modifies need coordination, so forwarding reports alone may not be sufficient. It's worth establishing why historical reports need to pass through live dispatch.

**Remaining correctness gaps**

- **Engine order identity:** An unseen intermediate venue leg can regress the engine's current venue order ID when its fill is replayed. For A -> B -> C, where B was never observed, the adapter preserves C but the engine treats the recovered fill's B as its current ID. The regression test checks the adapter binding without applying the event to the engine order.
- **Bracket protection:** A live parent cancellation can remove staged bracket children before REST recovery emits an earlier missed fill. The fill is recovered, but the children are no longer available to protect the resulting position.
- **Replacement cancellation:** If a replacement is accepted and then canceled during the outage, recovery drops its cancellation because its ID differs from the cached binding. The old leg's cancellation is also suppressed by the pending modify, leaving the order open locally.

**Managing complexity**

These are requirements to preserve order identity, terminal state, and bracket protection. They don't require reconstructing every intermediate leg. I'd prefer a smaller, clearly bounded recovery design over adding a separate exception for each race.

**Failed-sweep throttling**

Failed sweeps don't advance the timestamp used for the 20-second interval, allowing queued reconnects to start another sweep immediately. The shared limiter mitigates overall usage, but recovery can still consume capacity needed by other requests. This is secondary to the correctness gaps above.

**Before merging**

The three correctness cases need regression coverage through the affected engine and bracket paths. The existing reconciliation machinery is worth evaluating as the basis for that work.

cjdsellers:
Hey @riven314, just checking in to see if you're still planning to work on this? Happy to discuss any of the feedback. The aim is to keep complexity manageable by reusing existing state tracking where possible, so the Hyperliquid adapter stays maintainable over time.

riven314:
Hi @cjdsellers I will work on the feedback sometimes next week. Running low on tokens this week

PR #4999, opened 2026-09-15
Make netting reopen cost independent of replay history

folknor:
A follow-up to the NETTING replay-cost work in e00ef01863 and 908c571cae, extending it to the sites that still scale with archived-cycle history.

## Duplicate fill detection scans the full replay log

`Position::is_duplicate_replay_fill` runs a linear scan of `replay_events` on every recorded fill, and `ExecutionEngine::is_duplicate_closed_fill` scans it again before every NETTING reopen. With `carry_replay_events_on_reopen` enabled the log spans every archived cycle, so per-fill cost grows with the number of retained fills, and a backtest's per-bar cost rises with accumulated cycles (measured downstream at 3.6 to 6.6 us/bar between 5,000 and 50,000 bars).

The fix derives a replay identity index on `Position` - trade-id counts, `(trade_id, causation_id)` fill keys, and voided trade ids - excluded from serialization and maintained incrementally by every mutation method. Each lookup reproduces the existing predicate exactly and falls back to the previous linear scan whenever the index is stale (for example immediately after deserialization), so observable behavior is unchanged in every state. The engine's closed-fill check now goes through the same index via `Position::has_replay_trade_id`.

The index fields are private, so `Position` can no longer be constructed by struct literal or functional record update outside the model crate; the in-tree test sites were converted. Code mutating `replay_events` or `fill_voids` directly must call `rebuild_replay_index`, and the field docs say so.

## The archive snapshot copies the history it discards

`Cache::build_position_snapshot` cloned the complete position, replay history included, then cleared the copied vectors - while its rustdoc claimed the copy excludes them. `Position::clone_for_snapshot` now constructs the copy without cloning the replay or fill-void vectors; the snapshot's serialized form is unchanged.

## A reopen copies the retained history several times

`Action::Reopen` cloned the cached position before the carry flag was consulted, `open_position` cloned the prior again to carry replay state, spliced it with two further vector copies, and `add_position` cloned the spliced result into the cache.

The reopen now selects the action by id, checks the duplicate against the canonical cached position, and installs the replacement through `Cache::replace_position`, which swaps the stored value under the reused id and moves the prior replay vectors into it, leaving no per-reopen copy of the history in either carry mode.

Indexing reads the installed replacement, preserving the existing external-strategy ownership behavior. The position state snapshot still publishes and persists the complete position, and the `PositionOpened` event is built from state captured before publication, since snapshot subscribers run synchronously and may mutate the cache.

The two new cache mutators are classified for event-store recovery alongside `add_position`: replaying the captured `OrderFilled` reproduces the replacement.

## Testing

`test_replay_index_matches_linear_duplicate_predicate` compares the indexed predicate against a copy of the previous linear predicate across duplicate trades, causal fragments with both outcomes, decisive fill-void cases, purged history, adjustments, and a serialize/deserialize round trip, asserting the index is synchronized at each indexed checkpoint; removing the index maintenance fails it.

`carry_replay_reopen_transfers_history_allocations` pins the ownership transfer by vector pointer identity across an engine reopen and fails when the transfer is replaced by a copy. 

`position_opened_uses_state_captured_before_snapshot_subscriber` mutates the cached position from a snapshot subscriber and fails under the previous fetch-after-publish ordering.

Engine tests pin that a carry=false replacement holds only the new cycle and that carry=true preserves cross-cycle correction, duplicate rejection, and realized PnL.

A criterion benchmark (`position_replay`) drives reopens through the public `ExecutionEngine::process` at 8 to 4096 archived cycles, with unique and historical-duplicate candidates and an isolated archive-snapshot case.


cjdsellers (dismissed):
Hey @folknor,

Thank you for the PR. This looks good from my side.

~~When you get a chance, this needs a rebase onto current `develop`.~~

One minor follow-up I'll handle separately: `apply_pending_orderless_flip_opened` in the event-store replay still splices the replay vectors by hand; folding it onto `replace_position_without_order` with carry would match the new field docs. It's correct as-is through the fallback.

folknor:
@cjdsellers Thank you for rebasing. I'm bedridden at the moment, and will be for a few more days so don't expect timely replies :sneezing_face:

cjdsellers:
> @cjdsellers Thank you for rebasing. I'm bedridden at the moment, and will be for a few more days so don't expect timely replies :sneezing_face: 

Understood, hope you feel better soon! 🙏

cjdsellers:
@folknor, I had a deeper look at this while you're out, and profiled the reopen path with a criterion benchmark from 8 to 4096 archived cycles. I've pushed the result to your branch: rebased onto current `develop`, with one commit on top. The rebase rewrote your commit, so you'll want to reset rather than pull if you pick it up locally.

The short version is that the replay identity index wasn't paying for itself. It only helped duplicate rejection, and it made the reopen itself slower, because the three sets plus the sync fields added 216 bytes to every `Position` and the cache holds one of those per archived snapshot frame. The reopen was paying an index tax on state it never read.

Almost all of the win turns out to come from the copy elimination, which I kept intact:

| at 4096 cycles    | `develop` | with index | after   |
| ----------------- | --------- | ---------- | ------- |
| reopen, carry on  | 1857 µs   | 226 µs     | 160 µs  |
| reopen, carry off | 167 µs    | 212 µs     | 162 µs  |
| archive snapshot  | 135 µs    | 2.6 µs     | 5.0 µs  |

Carry-off is the default, and it had regressed against `develop`; it no longer does. The one real giveback is rejecting a duplicate whose trade ID sits at the far end of a long log, now around 30 µs against 8 µs with the index, still well under the 183 µs on `develop`. If that lookup depth turns out to matter in practice, the index is worth revisiting on its own terms, sized so it doesn't inflate the struct.

`clone_for_snapshot`, the ownership transfer, and `replace_position` are all unchanged, and `make cargo-test-extras` and `make pre-commit` are green on the rebased tree.

I'll take it from here, so nothing needed on your side.

cjdsellers (approved):

cjdsellers (approved):

PR #5002, opened 2026-09-16
Fix historical BinanceBar custom data delivery

### LINKED ISSUE #4998: Binance historical BinanceBar responses use Vec<BinanceBar> and fail Python on_historical_data
## Environment

- `nautilus_trader==2.0.0rc5`
- macOS arm64
- CPython 3.13.14
- Official wheel, no patches
- Also reproduced with the official `2.0.0rc4` wheel

No exchange account is required for the reproduction.

## Reproduction

A small integration test starts the official Binance DataClient against loopback HTTP/WebSocket fixtures.

The test requests historical `BinanceBar` custom data using the canonical BinanceBar descriptor, including the BarType, instrument metadata, and identifier.

The client sends the expected public Kline request and successfully parses the response into Binance extended bars.

However, the Python historical callback is never invoked. The Native layer logs:

```text
Failed to convert historical data to Python: unsupported type
```

The failure occurs after the Binance response has been parsed and before dispatch to the Python `on_historical_data` callback.

The corresponding live path works: a standard bars subscription produces BinanceBar custom data using the canonical BinanceBar DataType, and the extended fields such as quote volume can reach Python consumers.

## Source boundary

In:

```text
crates/adapters/binance/src/futures/data.rs
```

the historical BinanceBar branch calls `request_binance_bars(...)` and passes the resulting `Vec<BinanceBar>` directly into `CustomDataResponse`.

In:

```text
crates/trading/src/python/strategy.rs
crates/common/src/python/actor.rs
```

historical custom-data conversion accepts only:

```text
CustomData
Vec<CustomData>
```

and otherwise returns:

```text
Failed to convert historical data to Python: unsupported type
```

The generic custom-response pipeline in:

```text
crates/data/src/engine/mod.rs::custom_response_data
```

likewise consumes `CustomData` / `Vec<CustomData>` containers.

This leaves the Binance historical response shape inconsistent with the generic custom-data delivery contract.

## Expected behavior

Historical BinanceBar requests should reach Python Actor and Strategy consumers through the normal `on_historical_data` custom-data path while preserving:

- the canonical BinanceBar DataType;
- all BinanceBar fields, including quote volume, trade count, taker-buy base volume, and taker-buy quote volume;
- request correlation and requested time bounds.

The historical payload shape should be consistent with the CustomData container used by the live producer and the generic custom-response pipeline.

## Suggested minimal correction

One possible adapter-side correction is to convert each returned BinanceBar into a `CustomData` before constructing the response, using the canonical BinanceBar descriptor.

Conceptually:

```text
Vec<BinanceBar>
    ↓
Vec<CustomData>
    ↓
CustomDataResponse
```

For example, the response could construct the canonical DataType with `binance_bar_data_type(bar_type)` and wrap each bar with `CustomData::new(Arc::new(bar), data_type.clone())`.

This is only a proposed source correction, not a compiled or certified patch.

## Suggested regression coverage

A regression test should exercise the actual public historical delivery path:

```text
Binance DataClient
→ RequestCustomData(BinanceBar)
→ historical Kline response
→ CustomDataResponse
→ Python DataActor / Strategy
→ on_historical_data
```

It should verify that:

- the callback receives a list of `CustomData`;
- each payload contains a BinanceBar;
- the canonical DataType is preserved;
- extended fields such as quote volume and trade count are preserved;
- an empty successful response reaches the callback as an empty batch;
- the generic response/time-range pipeline continues to accept the response shape.

A Catalog roundtrip for the resulting BinanceBar custom data would also be useful coverage, although it is not required to reproduce this delivery failure.

## Impact

The standard historical `Bar` path does not preserve all Binance Kline fields needed by consumers of `BinanceBar`.

Without a working historical BinanceBar CustomData delivery path, Python consumers must either lose those extended fields or reimplement exchange acquisition outside the adapter.

The underlying Binance client already performs the acquisition and parsing correctly; the observed failure is at the historical custom-data delivery boundary.

abhijeetvichare76:
## Summary

Closes #4998, reported and diagnosed by @pengjiu.

A `request_data` for `BinanceBar` custom data was fetched and parsed correctly on both Spot and Futures, but the response never reached `on_historical_data`; the node logged `Failed to convert historical data to Python: unsupported type`. Both handlers put the raw `Vec<BinanceBar>` into `CustomDataResponse`, while every consumer downcasts to `CustomData` or `Vec<CustomData>`. These were the only two producers in the repo emitting a bare vector.

The two sites now share a small helper in `common/bar.rs` that builds the descriptor the live BinanceBar stream attaches (`binance_bar_data_type`, with `bar_type` and `instrument_id` metadata) once and wraps each bar with it, so historical and live items carry the same `DataType` and keep quote volume, trade count and the taker-buy volumes. Deribit and Hyperliquid attach the request's `DataType` to each item instead; matching the live stream seemed the better contract for a type that has one. The response-level `data_type`, correlation ID and time bounds are unchanged, and an empty venue response becomes an empty batch.

## Noted, not changed

- A fetch or parse failure after the request is spawned is still logged without emitting a response, on every Binance request path; not covered by a test here.
- No Python-level test: the container conversion is covered by `test_python_dispatch_historical_custom_data_preserves_payload_shape`, and the `BinanceBar` pyclass conversion is the one the live path uses.

## Testing

- The existing historical BinanceBar integration tests (Futures USD-M and COIN-M, Spot one-second SBE) now assert a `Vec<CustomData>` batch, the per-item `DataType` with `bar_type` and `instrument_id`, every extended field, the correlation ID and the start and end bounds.
- New `test_request_historical_binance_bars_empty_response_emits_empty_batch` for Spot and Futures: an empty kline response (JSON `[]`, SBE group count 0 through the production decoder) reaches the requester as an empty batch with the request's `data_type` and correlation ID.
- With the production change reverted at either site, that product's tests fail at the payload downcast; with it, `cargo test -p nautilus-binance --test integration -- binance_bars` passes 11.
- Full `nautilus-binance` integration suite single-threaded: 471 passed, one pre-existing exec client log-capture test failing on a second global logger install in one process (not under nextest).
- `cargo clippy -p nautilus-binance --no-deps --all-targets --all-features -- -D warnings` clean; `prek run --files` on the changed files passes apart from the Windows-only clippy and machete hooks.


cjdsellers:
Hey @abhijeetvichare76,

Thanks for the PR, and thank you @pengjiu for the report and diagnosis in #4998. I've gone over the change and its consumers, and it lines up with the delivery contract: attaching the live stream's descriptor to each historical bar is the right call for a type that has one, so historical and live items now carry the same `DataType`. The expanded assertions and the empty-response coverage through the production SBE decoder look good.

Leaving the spawned fetch or parse failure as log-only is fine here since it matches the other Binance request paths, and the existing Python dispatch coverage is enough. This looks good to merge 👍

cjdsellers (approved):

PR #5003, opened 2026-09-16
Keep working orders when a reconciliation lifecycle is replaced

### LINKED ISSUE #5000: Startup reconciliation drops a working order when an instrument's fills do not reconstruct the venue position
# Bug Report

## Confirmation

- [x] I've searched existing issues and discussions to avoid duplicates. Closest are #4736 (same function, warm-restart path, cached position present) and PR #4693 (accounting for dropped *position* reports); neither covers order reports dropped on a cold start.
- [x] I've reviewed the source code to confirm the behavior is not by design — see below, where a sibling branch states the opposite rule in a comment and follows it.
- [x] I've tested this issue using a recent pre-release or development wheel. Reproduced on `develop` @ `67eb71a5a9` with the failing test below. All line numbers in this report are that revision.

## Expected behavior

`process_mass_status_for_reconciliation` should keep an order that is still working at the venue. Such an order has no fills, so it moves the position by zero and cannot be the source of a discrepancy between the simulated position and the venue's.

Its `FilterToCurrentLifecycle` branch states exactly that rule:

```rust
// crates/execution/src/reconciliation/positions.rs:174
// Keep only orders that have fills or are still working
let orders_with_fills: ahash::AHashSet<VenueOrderId> = fill_map.keys().copied().collect();
order_map.retain(|id, order| {
    orders_with_fills.contains(id)
        || !matches!(
            order.order_status,
            OrderStatus::Denied
                | OrderStatus::Rejected
                | OrderStatus::Canceled
                | OrderStatus::Expired
                | OrderStatus::Filled
        )
});
```

## Actual behavior

`ReplaceCurrentLifecycle`, the sibling branch, clears the map instead:

```rust
// crates/execution/src/reconciliation/positions.rs:157
// Replace with only synthetic
order_map.clear();
fill_map.clear();
order_map.insert(first_venue_order_id, order);
fill_map.insert(first_venue_order_id, vec![fill]);
```

One condition chooses between them — whether the simulated position already matches the venue:

```rust
// positions.rs:275  (returns FilterToCurrentLifecycle at :276)
// Current lifecycle matches - filter out old lifecycles

// positions.rs:282  (returns ReplaceCurrentLifecycle at :292)
// Current lifecycle doesn't match - replace with synthetic fill
```

Whether an order is still working at the venue is independent of whether the historical fills add up, so the two branches should treat it the same way. Of the four outcomes, three keep it (`NoAdjustment`, `AddSyntheticOpening`, `FilterToCurrentLifecycle`) and only `ReplaceCurrentLifecycle` does not.

The caller propagates the loss, replacing every order report for the instrument with the returned set:

```rust
// crates/live/src/execution/manager.rs:4716
final_orders.retain(|_, order| order.instrument_id != instrument_id);
```

`adjust_mass_status_fills` (`manager.rs:4611`) reaches this when the mass status has no lookback window (`:4635`), the instrument is in scope (`:4642`), it has a position report, it is not hedge mode, the instrument is cached, and the cache holds no retained open position for it (`:4662`). That last condition holds on every start when no cache backing is configured, so for such a deployment this is every cold restart with a position, not an edge case.

### Consequence

Startup reconciliation reports `open=0` for an instrument that has orders working at the venue, so a strategy that cancels its resting orders on start has nothing to cancel and the orders stay live and unclaimed.

The periodic open-order check does not recover them. `reconcile_open_order_reports` resolves each report through `client_order_id` or `venue_order_id` and `continue`s when neither is in the cache (`manager.rs:2201-2209`); it never calls `handle_external_order`, so an order the cache has not seen is never adopted.

## Steps to reproduce

Added to `crates/execution/src/reconciliation/tests.rs`. One parameterised test, two cases that differ only in the position report: the historical fills are identical, so the only thing that changes is whether they reconstruct the venue position. **It is written to fail** — `case_2` is the defect.

The first two assertions identify which branch was taken, so the test does not require you to take that on trust: the surviving fill keys separate all four outcomes, and only the replacing branch contributes an order report of its own.

```rust
#[rstest]
// The branch is chosen by the quantity comparison alone. The current lifecycle (the two fills
// after the zero-crossing) simulates to 0.10 @ 4050, so `Quantity::from("0.10")` matches and the
// fills are filtered to that lifecycle, while `Quantity::from("0.05")` mismatches and the
// lifecycle is replaced by a synthetic report. `check_position_match` returns on the quantity
// line, so neither average price reaches the comparison; they are carried over from the fixture
// in `test_adjust_fills_multiple_zero_crossings_mismatch` so that both tests describe the same
// venue state, and in the mismatching case the average price prices the synthetic report.
#[case::lifecycle_matches_venue(
    Quantity::from("0.10"),
    dec!(4050.00),
    vec![VenueOrderId::from("ORDER4"), VenueOrderId::from("ORDER5")],
    vec![]
)]
#[case::lifecycle_mismatches_venue(
    Quantity::from("0.05"),
    dec!(4142.04),
    vec![VenueOrderId::from("ORDER4")],
    vec![VenueOrderId::from("ORDER4")]
)]
fn test_process_mass_status_keeps_working_order_for_either_lifecycle_outcome(
    instrument: InstrumentAny,
    #[case] venue_qty: Quantity,
    #[case] venue_avg_px: Decimal,
    #[case] expected_fill_keys: Vec<VenueOrderId>,
    #[case] expected_synthetic_order_keys: Vec<VenueOrderId>,
) {
    let account_id = AccountId::from("TEST-001");
    let instrument_id = instrument.id();
    let venue_order_id1 = VenueOrderId::from("ORDER1");
    let venue_order_id2 = VenueOrderId::from("ORDER2");
    let venue_order_id4 = VenueOrderId::from("ORDER4");
    let venue_order_id5 = VenueOrderId::from("ORDER5");
    let working_venue_order_id = VenueOrderId::from("ORDER9");
    let make_fill =
        |venue_order_id: VenueOrderId, trade_id: &str, side: OrderSide, px: &str, ts_event: u64| {
            FillReport::new(
                account_id,
                instrument_id,
                venue_order_id,
                TradeId::from(trade_id),
                side,
                Quantity::from("0.05"),
                Price::from(px),
                Money::from("0.00 USD"),
                LiquiditySide::Taker,
                None,
                None,
                UnixNanos::from(ts_event),
                UnixNanos::from(ts_event),
                None,
            )
        };

    let mut mass_status = ExecutionMassStatus::new(
        ClientId::from("TEST"),
        account_id,
        Venue::from("SIM"),
        UnixNanos::default(),
        Some(UUID4::new()),
    );
    // Lifecycle 1 closes out at ts 2000, lifecycle 2 simulates to 0.10 @ 4050.
    mass_status.add_fill_reports(vec![
        make_fill(venue_order_id1, "TRADE1", OrderSide::Buy, "4000.00", 1_000),
        make_fill(venue_order_id2, "TRADE2", OrderSide::Sell, "4050.00", 2_000),
        make_fill(venue_order_id4, "TRADE4", OrderSide::Buy, "4000.00", 3_000),
        make_fill(venue_order_id5, "TRADE5", OrderSide::Buy, "4100.00", 4_000),
    ]);
    // A working order with no fills at all.
    mass_status.add_order_reports(vec![OrderStatusReport::new(
        account_id,
        instrument_id,
        Some(ClientOrderId::from("O-WORKING")),
        working_venue_order_id,
        OrderSide::Buy.into(),
        OrderType::Limit,
        TimeInForce::Gtc,
        OrderStatus::Accepted,
        Quantity::from("0.05"),
        Quantity::from("0.00"),
        UnixNanos::from(4_500),
        UnixNanos::from(4_500),
        UnixNanos::from(4_500),
        None,
    )]);
    mass_status.add_position_reports(vec![PositionStatusReport::new(
        account_id,
        instrument_id,
        PositionSide::Long,
        venue_qty,
        UnixNanos::from(5_000),
        UnixNanos::from(5_000),
        None,
        None,
        Some(venue_avg_px),
    )]);

    let result = process_mass_status_for_reconciliation(&mass_status, &instrument, None).unwrap();

    // The surviving fill keys identify the branch: filtering to the current lifecycle keeps the
    // two fills after the zero-crossing, replacing that lifecycle leaves one synthetic fill keyed
    // on its first fill, and leaving the fills unadjusted would keep all four.
    let fill_keys: Vec<VenueOrderId> = result.fills.keys().copied().collect();
    assert_eq!(
        fill_keys, expected_fill_keys,
        "unexpected adjustment branch"
    );

    // Only the replacing branch adds an order report of its own, keyed on that same first fill.
    let synthetic_order_keys: Vec<VenueOrderId> = result
        .orders
        .keys()
        .copied()
        .filter(|id| *id != working_venue_order_id)
        .collect();
    assert_eq!(
        synthetic_order_keys, expected_synthetic_order_keys,
        "unexpected synthetic order reports"
    );

    // An unfilled working order moves the position by zero, so it cannot be the source of a
    // position discrepancy and its fate must not depend on whether the historical fills
    // reconstruct the venue position.
    assert!(
        result.orders.contains_key(&working_venue_order_id),
        "working order was dropped, orders: {:?}",
        result.orders.keys().collect::<Vec<_>>()
    );
}
```

```bash
cargo test -p nautilus-execution --lib \
  reconciliation::tests::test_process_mass_status_keeps_working_order_for_either_lifecycle_outcome
```

```
running 2 tests
test ...::case_1_lifecycle_matches_venue ... ok
test ...::case_2_lifecycle_mismatches_venue ... FAILED

---- ...::case_2_lifecycle_mismatches_venue stdout ----
panicked at crates/execution/src/reconciliation/tests.rs:1218:5:
working order was dropped, orders: ["ORDER4"]

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 916 filtered out
```

`case_2` reaches the third assertion, so the two branch assertions above it already held. Swapping the two cases' expected key sets makes both fail and prints the real ones — `["ORDER4", "ORDER5"]` for the filtering case and `["ORDER4"]` for the replacing case — so those assertions are load-bearing rather than vacuous.

Replacing `order_map.clear()` with a retain over still-working orders makes both cases pass.

## Code snippets or logs

From the restart that prompted this, on a Bybit linear (testnet) node with `generate_missing_orders: true` (default), no `reconciliation_lookback_mins`, and no cache backing. The strategy held an open position and 14 working limit orders on the same instrument; the process was `SIGKILL`ed and started again.

```
Received 3173 order(s), 2557 fill(s), 1 position(s)
Adjusting fills for 1 instrument(s) with position reports
After adjustment: 2430 order(s), 2289 fill group(s)
```

`3173 − 744 + 1 = 2430`: 744 order reports for the instrument removed, one synthetic `MARKET` order inserted. Of those 744, 14 were the working orders and 730 were closed history. Polling the venue's open orders across the restart — three times after the kill and nine more spanning the reconciliation — returned all 14, in status `New`, every time.

Nothing was lost on the way in: reconstructing the mass status from the venue's own endpoints gives 3159 closed from `/v5/order/history` plus the 14 working from `/v5/order/realtime`, which is the 3173 the engine reports receiving. The removal is unrecorded — `skipped` and `filtered` both stay 0.

## Suggested direction

`ReplaceCurrentLifecycle` replaces a fill history shown not to explain the venue position. That scepticism fits derived history, but an order still working at the venue is present state reported in the same payload, so the minimal repair looks like retaining still-working orders instead of clearing the whole map, leaving `fill_map.clear()` as it is.

Worth a maintainer's judgment rather than a patch from me: `FilterToCurrentLifecycle`'s predicate also retains orders that *have* fills, which is safe there because that branch filters fills rather than clearing them. Here the fills are replaced wholesale, so a `PartiallyFilled` order whose fills have just been discarded may be better dropped than kept.

Two things that may help in scoping it:

- No existing test pins the clearing. Replacing `order_map.clear()` with such a retain leaves `nautilus-execution --lib` at 918 passed, and `nautilus-live`'s suites at the same counts as a clean tree.
- The two entry points already disagree. With `generate_missing_orders: false`, `ReplaceCurrentLifecycle` falls through to the no-op arm at `positions.rs:190`, so every order report survives — including the stale ones the synthetic path exists to replace.

## Workarounds

Both cost something, which is why this is a report rather than a config change on our side.

- `reconciliation_lookback_mins` — any window makes `adjust_mass_status_fills` return at `manager.rs:4635`. But the Bybit adapter does not send the window to the venue; `generate_order_status_reports` passes `None` for the time bounds and applies the window locally as `reports.retain(|r| r.ts_last >= start)`. Long-resting orders are dropped at that filter instead.
- `generate_missing_orders: false` — clears the `if generate_synthetic_reports` guard at `positions.rs:141` so nothing is deleted, but it also disables the synthetic fill that makes the opening position agree with the venue, which is the condition that selected this branch.

## Specifications

- OS platform: Linux (x86_64)
- Python version: n/a — reproduced through the Rust crates
- `nautilus_trader` version: `develop` @ `67eb71a5a9` (`v2.0.0rc5`, crates `0.64.0`)
- Adapter/venue: Bybit linear, testnet — the three sites above are venue-agnostic core code

abhijeetvichare76:
## Summary

Closes #5000.

When an instrument's fills do not reconstruct the venue position and the current lifecycle is replaced with a synthetic fill (`ReplaceCurrentLifecycle` in `crates/execution/src/reconciliation/positions.rs`), the order map for that instrument was cleared before the synthetic report went in. Orders still working at the venue with no fills went with it, so the node reported `open=0` and never adopted them; the periodic open-order check only reconciles orders the cache already knows. The sibling `FilterToCurrentLifecycle` branch keeps working orders, and the choice between the two depends only on whether the simulated lifecycle matches the venue position, which says nothing about whether an unfilled order is live.

The branch now retains reports with zero filled quantity that are open or in flight (the `is_open() || is_inflight()` idiom from the `OrderStatus::is_open` docs, so a resting order reported as `Submitted` survives), then clears the fills and inserts the synthetic report as before. Accepted and Triggered reports are adopted as external orders; in-flight statuses hit the existing unhandled-status arm and are only cached. Orders with a filled quantity are dropped with the history their fills belong to: with the fills gone, the external order events would carry an inferred fill for that quantity on top of the synthetic fill and double count the position. A partially filled working order therefore still does not survive this path; carrying it safely is a design call I have left open.

Diagnosis and the first regression test are from @Martingale42's report.

## Noted, not changed

- With `generate_missing_orders=false` the non-generating arm keeps every report for the instrument, closed and partially filled included; what that mode should filter is a separate decision.
- The `FilterToCurrentLifecycle` retain lists terminal statuses by hand and omits `Voided`. One-word follow-up if wanted.
- A retained report that contradicts a cached `PartiallyFilled` order (venue says Accepted, zero filled) now takes the normal snapshot path and can void the cached fills, where before it was discarded. That is the documented snapshot rule, reachable only with no open cached position for the instrument.

## Testing

- `crates/execution/src/reconciliation/tests.rs`: the reporter's two-branch test, a 12-case status and filled-quantity matrix on the replacing branch, and the synthetic report overwriting the first-fill order's own report; the existing mismatched-lifecycle test now shares the fixture.
- `crates/live/tests/integration/manager.rs`: the mass status runs through `ExecutionManager`; the unfilled order comes out Accepted under `EXTERNAL`, the partially filled one does not, and the only fill event is the synthetic one equal to the venue quantity.
- With the fix reverted, the reporter's mismatching case, the five keep cases and the live test fail.
- `cargo test -p nautilus-execution --lib`: 938 passed. `nautilus-live` `manager::` integration tests single-threaded: 341 passed. Clippy with `-D warnings`, `cargo +nightly fmt --check` and `prek run --files` on the changed files pass apart from the make-dependent collection hook.


cjdsellers (approved):
Hey @abhijeetvichare76,

Thank you for the PR. I've been through the change and the downstream reconciliation paths, and it all holds together: the retained orders can't produce an inferred fill on top of the synthetic one, and the snapshot-path consequence you noted is real but consistent with the sibling branches, which never discarded those reports.

On the items you left open:

- Carrying a partially filled working order safely needs its own design pass, so keeping it out of this path makes sense for now.
- What `generate_missing_orders=false` should filter is a separate decision, agreed.
- The `Voided` addition to the `FilterToCurrentLifecycle` retain list would be a welcome one-word follow-up if you're able, bringing it in line with `is_closed()`.

This looks good to merge. Thanks also to @Martingale42 for the report and diagnosis behind this.

PR #4890, opened 2026-08-30
Fix Kraken Spot instrument fees from TradeVolume

matvt-cell:
**NautilusTrader can execute live trades involving real capital. Pull requests are held to a very
high standard for correctness, reliability, testing, clarity, and maintainability.**

> External contributions must not modify files under `.github/workflows` or `.github/actions`;
> workflow changes are maintainer-only.

- [x] A maintainer agreed on the problem and approach in an issue, or this is a small,
  self-contained fix that does not need prior discussion
- [x] I have read and followed
  https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md
  and, if I used AI,
  https://github.com/nautechsystems/nautilus_trader/blob/develop/AI_POLICY.md
- [x] I understand and can explain every submitted change and all information in this PR description
- [x] This change is complete, locally validated, and ready for review, or a maintainer requested
  this draft
- [x] I ran `make format`, then ran `make pre-commit` locally and confirmed it passed, or I
  described an agreed limitation below
- [x] I ran all relevant tests locally, no tests apply to this change, or I described an agreed
  limitation below
- [x] I have not modified `RELEASES.md` (maintainers keep it current to avoid merge conflicts)

## Summary

Resolve Kraken Spot instrument fees from the authenticated private `TradeVolume`
endpoint when credentials are available, so instruments carry the connected
account's actual maker and taker fees instead of relying on the stale public
`AssetPairs` base tier.

This applies to both regular Spot instruments and tokenized/xStocks instruments.
Tokenized TradeVolume requests use the `equity_pair` asset class. When credentials
are unavailable, the existing `AssetPairs` base-tier fallback is preserved and
documented as potentially stale.

## Related issues/PRs

Fixes #4789
Supersedes #4826

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Improvement (non-breaking)
- [ ] Breaking change (impacts existing behavior)
- [x] Documentation update
- [ ] Maintenance / chore

## Breaking change details (if applicable)

Not applicable.

## Documentation

- [x] Documentation changes follow the style guide (`docs/developer_guide/docs.md`)
- [ ] For PyO3 binding or wrapped Rust doc changes, I ran `make py-stubs` and committed the generated output

## Testing

**New or changed logic must be covered by tests.** Select all that apply:

- [ ] Affected code paths are already covered by the test suite
- [x] I added/updated tests to cover new or changed logic
[...]

Issue #3742, opened 2026-03-21
[Bybit] Demo trading exec client fails with error 10001 on connect


jindrichsirucek:
# Bug Report

| Term                          | Definition |
|-------------------------------|------------|
| **Bug**                       | Behavior that contradicts the platform’s documented or intended design as per code, docs, or specs. (i.e., the implementation is incorrect.) |
| **Expectation&nbsp;mismatch** | Behavior that follows the platform’s documented or intended design but differs from what you expected. (i.e., the design/spec might be the problem.) |
| **Enhancement request**       | A request for new functionality or behavior that is not implied by existing design. (i.e., *“It would be great if the platform could…”*) |

**Note:**

- Submitting this issue automatically applies the `bug` label.
- `bug`-labeled issues are triaged with higher priority because they require corrective implementation work.
- **Expectation mismatches** and design-level concerns should be opened as [Discussions](https://github.com/nautechsystems/nautilus_trader/discussions), or RFCs instead, where they can be validated and discussed to consensus before any work is scheduled.
- The absence of a feature is typically not an expectation mismatch, and should be filed as an enhancement request.

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I’ve re-read the relevant sections of the documentation.
- [x] I’ve searched existing issues and discussions to avoid duplicates.
- [x] I’ve reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I’ve tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

### Expected Behavior

When setting `demo=True` in `BybitExecClientConfig` alongside `base_url_http="https://api-demo.bybit.com"`, the `BybitExecutionClient._connect` routine should successfully authenticate and query the `/v5/account/info` metadata needed to establish a live connection to the Bybit V5 Demo Trading environment (Mainnet Demo).

### Actual Behavior

The connection immediately fails with `ValueError(Bybit error 10001: Request Parameter Error)` during the initial metadata handshake within `_connect`.

The Rust HTTP client (`BybitHttpClient`) attempts to fetch account details immediately upon authentication. In the Bybit V5 Demo Trading environment, this request throws `10001`, which suggests the adapter is either:
1. Sending a `POST` request (which signs the body) instead of a `GET` request (which signs the query string).
2. Using unsupported parameters (like strict UTA/margin flags) that the `api-demo` subdomain rejects during the initial sync.
[...]

PR #5005, opened 2026-09-16
Fall back to public Kraken fees when TradeVolume fails

### LINKED ISSUE #5004: [Kraken] A failed TradeVolume request aborts spot instrument loading, so the execution client never connects
# Bug Report

## Confirmation

## Expected behavior

A failure to resolve **account** fee rates should not prevent Kraken spot instruments from loading,
and should never prevent the execution client from connecting.

The adapter already carries a public fee fallback. `resolve_fee_rates`
(`crates/adapters/kraken/src/common/parse.rs`) returns the public base-tier rates from the
`AssetPairs` definition whenever no account rate is supplied, which is how unauthenticated clients
get fees today. The commit that introduced account fees describes this explicitly as
"Preserve public fee fallback for unauthenticated clients".

When the authenticated `TradeVolume` request fails, instruments should fall back to those same
public rates rather than failing to load.

## Actual behavior

The error aborts the entire instrument listing, and because the listing runs on the execution
client's connect path, the execution client never connects.

In `crates/adapters/kraken/src/http/spot/client.rs`, `request_instruments` propagates the fee error
(locate by symbol; line numbers move):

```rust
let asset_pairs = self.inner.get_asset_pairs(pairs.clone(), None).await?;
let fee_rates = self.request_fee_rates(&asset_pairs, None).await?;   // aborts the whole listing
```

`request_fee_rates` skips out only when there are no credentials at all:

```rust
if self.inner.credential().is_none() || pairs.is_empty() {
    return Ok(AHashMap::new());
}
```

so an authenticated client has no path to the public rates once the call errors.

`request_instruments` is called from `KrakenSpotExecutionClient::connect`
(`crates/adapters/kraken/src/execution/spot.rs`) via
`.context("Failed to load Kraken spot instruments")?`. The consequence is far larger than missing
fees: the execution client does not connect, so there is no account in the `Cache` and no
reconciliation. A live trading node comes up disarmed and silently without account state.

That asymmetry is the heart of this report. "Correct fees or no instruments" is a defensible policy
for a pricing consumer and an unsafe one for an execution consumer, and the adapter cannot tell
which it is serving. Fees are instrument metadata with an existing documented public fallback; the
execution client is not there to price fees.

Note there are **two** fee requests on this path and both are fatal. The second requests tokenized
(xStocks) fees with `aclass=equity_pair`. Inside that block, failing to *fetch* the tokenized pairs
is already tolerated (`Err(e) => log::warn!("Failed to fetch tokenized asset pairs: {e}")`), while
failing to get their *fees* aborts everything — an inconsistency worth resolving in the same place.

## Steps to reproduce

The defect itself is deterministic and needs no venue access. Only the *trigger* is intermittent.

**Deterministic reproduction, no credentials required.** Make `/0/private/TradeVolume` return a
Kraken API error and connect a spot execution client. Against `develop` @ `88cb0c7640`,
`KrakenSpotExecutionClient::connect()` returns `Err`:

```
execution client must connect when the account fee request is denied:
Failed to load Kraken spot instruments
```

which is the same top-level context seen in production. Any `TradeVolume` failure reproduces it;
the response body we used was `{"error":["EGeneral:Permission denied"],"result":{}}`.

**How it arises in production.** The denial is intermittent, which is what makes the hard abort
dangerous rather than merely inconvenient:

1. Configure a Kraken spot execution client (`KrakenExecutionClientConfig`) with valid credentials
   holding `Funds permissions - Query`.
2. Repeat `POST /0/private/TradeVolume` with the full spot pair list.
3. Occasionally the call returns `EGeneral:Permission denied` while identical neighbouring calls
   with the same key and the same body succeed.

When that lands on a node's `connect()`, the node loses its execution client entirely:

```
[ERROR] nautilus_execution::engine: Failed to connect execution client:
Failed to load Kraken spot instruments: API error: EGeneral:Permission denied
```

once, with no retry and no recovery.

We probed this against a live account (about 30 read-only `TradeVolume` calls, 27 of them with a
valid nonce, no orders) specifically to find a deterministic cause, and found none. Each of these
was ruled out:

- **Payload size.** The full public `AssetPairs` pair list succeeds, and so does that same list plus
  one additional pair — the latter immediately after a denial on the former.
- **The tokenized/equity-class request.** `aclass=equity_pair` succeeds outright on this account.
- **JSON encoding.** Single-pair JSON calls succeed; form versus JSON makes no difference.
- **Repetition.** Eight consecutive identical full-list calls all succeeded.

Observed frequency, from a small sample and offered as nothing more: one denial in ten identical
well-formed requests, plus the failure that prompted this report, on one account on one day. That
supports "intermittent" and nothing quantitative beyond it.

We are therefore reporting this as a **rare transient venue response**, not a property of the
request. To be explicit: we cannot reproduce the denial on demand, and we are not asking anyone to
try. The deterministic reproduction above is the one that can be run.

This also touches the retry classification: `kraken_http_should_retry` treats an `ApiError` as
retryable only when the message contains `"Rate limit"`, so a transient `EGeneral:Permission denied`
is a single attempt. That classification may be incomplete, but the fallback is the more robust
remedy because it holds regardless of how retryability is resolved.

There is no way to avoid the call from configuration: `KrakenExecutionClientConfig` has no
fee-resolution switch, and neither client config can narrow the instrument listing, so the adapter
always loads the venue's full universe.

## Code snippets or logs

The public fallback that is never reached on error
(`crates/adapters/kraken/src/common/parse.rs`):

```rust
fn resolve_fee_rates(
    definition: &AssetPairInfo,
    account_fee_rates: Option<(Decimal, Decimal)>,
) -> (Option<Decimal>, Option<Decimal>) {
    account_fee_rates.map_or_else(
        || {
            (
                definition.fees_maker.first().map(|(_, fee)| *fee / dec!(100)),
                definition.fees.first().map(|(_, fee)| *fee / dec!(100)),
            )
        },
        |(maker, taker)| (Some(maker), Some(taker)),
    )
}
```

Passing an empty fee map therefore yields fully-formed instruments carrying public base-tier fees.
Catching the error inside `request_fee_rates` and returning an empty map gives
correct-fees-if-available, public-fees-if-not.

## Related

- #4789 — the motivating issue for the account-fee change (stale `AssetPairs` ladder), closed by the
  commit that introduced this path. Different defect: it reports wrong fee *values* on instruments
  that load successfully.
- #4890 — the contributor PR for that change. Its review already raised this class of concern:
  "`get_trade_volume` bypasses the normal retry and cancellation path... A retryable network or 5xx
  failure can now abort instrument initialization." The landed change addressed **retry**; a
  non-retryable error such as this one still propagates and aborts the listing.
- #4289 (merged) — "Fix Binance futures leverage initialization aborting execution client connect",
  which made the auxiliary call best-effort "so a leverage-init issue never prevents the execution
  client from connecting". Same remedy, already accepted in this project.
- Commit `a21304e9f6`, "Fix Bybit demo exec client fee rate error on connect" (resolving #3742), is
  the closest precedent of all: "Fall back to default fees on `BybitError` in `fetch_fee_map`".
  That fallback is still in place today — `BybitHttpClient::fetch_fee_map`
  (`crates/adapters/bybit/src/http/client.rs`) matches `Err(BybitHttpError::BybitError { .. })`,
  logs a warning and returns an empty fee map so default rates apply. That is structurally the
  change proposed here, for the same reason, on the same connect path, in this repository.

We note #4729 and #4617 were closed on the principle that a failed venue query must not be silently
swallowed, and we think this case is distinguishable: position and order reports are authoritative
state, whereas fees are instrument metadata for which this adapter already ships a public fallback
and already uses it for unauthenticated clients. The proposal is to log a warning, not to hide the
failure.

## Specifications

- Reproduced against: `develop` @ `88cb0c7640`

Deployment details for the affected node are omitted deliberately. They reached us second-hand and
we could not verify them first-hand, so we would rather leave the field empty than state it wrongly.
The deterministic reproduction above is against `develop` and needs no venue access.

zhaow-de:
  and, if I used AI,
  this draft
  limitation below

## Summary

A failed `TradeVolume` request aborted the entire Kraken spot instrument listing. Because
`request_instruments` runs inside `KrakenSpotExecutionClient::connect`, that cost the execution
client its connection, and with it the account state and reconciliation — over fee metadata for
which the adapter already ships a public fallback. This catches the error inside `request_fee_rates`
and returns an empty fee map, so `resolve_fee_rates` supplies the public base-tier rates from
`AssetPairs` exactly as it already does for unauthenticated clients.

The failure we observed was a rare transient `EGeneral:Permission denied` from `TradeVolume`; probing
a live account ruled out payload size, the tokenized `aclass=equity_pair` request, and JSON encoding
as causes. Since the denial is not a property of the request, there is no request shape to correct —
the fallback is the fix.

Notes for reviewers:

- The error is caught at the `get_trade_volume` call **inside** `request_fee_rates`, deliberately not
  at the call sites in `request_instruments`. Catching at the call sites would also swallow the
  `ParseError` raised when a successful response omits a fee key — behavior that is intentional and
  protected by `test_spot_domain_request_instruments_rejects_missing_account_fee`. That test is
  unchanged and still passes; the weaker fix was considered and rejected for that reason.
- Both fee call sites benefit, including the tokenized (`aclass=equity_pair`) request. That request
  sat in a block where failing to *fetch* tokenized pairs was already tolerated with a warning while
  failing to fetch their *fees* was fatal; the two are now consistent.
- Retry behavior is untouched. `kraken_http_should_retry` still classifies a non-"Rate limit"
  `ApiError` as non-retryable, so this changes what happens *after* retries are exhausted or
  declined, not whether they run.

## Related issues/PRs

Closes #5004

This mirrors `BybitHttpClient::fetch_fee_map`, which already handles a venue error on the same
connect path by logging a warning and returning an empty fee map so default rates apply (commit
`a21304e9f6`, "Fix Bybit demo exec client fee rate error on connect", resolving #3742). It also
follows #4289, which made Binance futures leverage initialization best-effort so it never prevents
the execution client connecting. Related to #4890, whose review raised the retryable-failure form of
this concern; the landed change addressed retry, while a non-retryable error still aborted the
listing.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

Not a breaking API change. Behaviorally, an account whose `TradeVolume` request fails now receives
instruments carrying public base-tier fees instead of no instruments at all. A warning is logged, so
the failure is reported rather than hidden.

## Documentation

  generated output

The `request_instruments` doc comment changed, and its PyO3 wrapper doc is generated from it, so it
was regenerated with `python/generate_docstrings.py` and committed rather than hand-edited. No
binding signature changed, so the `.pyi` stubs are unaffected, and a repo-wide
`generate_docstrings.py --dry-run` reports no remaining drift.

## Testing

Added `test_spot_domain_request_instruments_falls_back_to_public_fees_on_error`, which makes the
`TradeVolume` mock return `{"error":["EGeneral:Permission denied"],"result":{}}` and asserts that
instruments still load carrying the public rates from the fixtures — `0.0025`/`0.004` for `XBTUSDT`
and `-0.0002`/`0.001` for the tokenized `AAPLxUSD`, which are distinct from the account rates the
mock would otherwise return. Asserting the tokenized instrument covers the second call site.

Also added `test_spot_execution_client_connects_when_trade_volume_denied`, which drives the fault
through `KrakenSpotExecutionClient::connect()` and asserts the client connects. That covers the
reported harm directly rather than by implication: against the previous behavior it fails with
`Failed to load Kraken spot instruments`, the same top-level context seen in production.

No existing test needed changing. Both new tests fail against the previous behavior and pass with
the fix, and they are the only failures.

Local validation: `cargo test -p nautilus-kraken` 773 passed, 0 failed.


cjdsellers:
Hi @zhaow-de,

Thanks for the PR. The fallback looks sound, including preserving the missing-fee-key error.

Could you also update the “Spot instrument fees” section in `docs/integrations/kraken.md`? It still says a failed `TradeVolume` request prevents connection. It needs to describe the warning and public-rate fallback, while retaining the distinction for a successful response that omits a requested fee key.

zhaow-de:
Thanks! I've updated the "Spot instrument fees" section in `docs/integrations/kraken.md`.

It now says that if the `TradeVolume` request fails, the adapter logs a warning and uses the public base-tier rates from `AssetPairs`, so the Spot data or execution client still connects. It keeps the other case too: if the request succeeds but the response leaves out the fee for a requested pair, the client still can't connect.

I also searched the rest of the repo for the old wording, and this was the only place left.

cjdsellers (commented):
@zhaow-de, thanks for following up on the issues. Catching the failure inside `request_fee_rates` so the missing-fee-key error still fails loudly was the right call.

This looks good to merge. Two small things you could take or leave:

- `SpotTradeVolumeResponse.fees` is a required field, so the `{"error": [...], "result": {}}` mock bodies fail deserialization and the new tests exercise `ParseError` rather than the production `ApiError`. Would you drop `"result": {}` from both mocks so they pin the incident's error shape? The fallback path is identical either way; it just keeps the tests honest about what they cover.
- "so the Spot data or execution client still connects" holds for a transient `TradeVolume` failure, but a key broadly lacking Query Funds would still fail `connect` later at the balance fetch. If you're able, would you scope that sentence to transient or isolated `TradeVolume` failures?

cjdsellers (dismissed):

zhaow-de:
Thanks, both good catches. I've taken both.

**Mock bodies.** You're right that `"result": {}` failed deserialization, because `fees` is required while `fees_maker` has `#[serde(default)]`. The deserialize step also runs before the error-array check, so the tests never reached the `ApiError` branch at all. I dropped `"result": {}` from both mocks. To confirm the shape actually changed, I re-ran the tests against the old code: the execution client test now fails with `API error: EGeneral:Permission denied`, which is the error from the incident.

**Docs sentence.** Also right. `connect` requests account state after loading instruments, so a key without Query Funds would fail there instead. The sentence now says a transient or isolated `TradeVolume` failure does not stop the client connecting, and adds that a key missing `Funds permissions - Query` altogether still fails later at the account state request.

All 773 Kraken tests pass, and no other code changed.

cjdsellers (approved):

PR #5019, opened 2026-09-18
Fix Bybit non-terminating cursor pagination

Martingale42:
  this draft
  limitation below

## Summary

Bybit can serve a `/v5/order/history` page whose `nextPageCursor` addresses that page rather than
the next one; requesting it returns the same rows and the same cursor indefinitely. The empty
cursor the pagination loops wait for never arrives, so startup reconciliation does not finish and
the node never starts. Seven of the eight cursor loops in the Bybit HTTP client had no defence
against this, and the eighth caught a stalled cursor but not a cycle. All eight now share a
`CursorWalk` that refuses a cursor which has stopped advancing or which the walk has already
followed.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

## Detail

Observed on a live unified account: from page 12 of `/v5/order/history`
(`settleCoin=USDT`, no `symbol`), `nextPageCursor` addressed the page it arrived with.

    page 11: rows=16  cursor=b63425c7...:1789221904182,902f5495...:1789221005855
    page 12: rows= 3  cursor=45311eca...:1789221005855,902f5495...:1789221005855
    page 13: rows= 3  cursor=45311eca...:1789221005855,902f5495...:1789221005855   (identical to page 12)

Both halves of the page-12 cursor address a row of page 12. Five replays of the same request
returned the same three order IDs and the same cursor, so the venue is deterministic here rather
than racing. A node whose reconciliation timeout was raised from 30 s to 300 s to 3599 s consumed
each of them in turn without making progress.

The budget exit does not help. `generate_order_status_reports` passes `None` for the limit at both
of its call sites (`execution.rs:1094`, `:1109`), and `generate_fill_reports` does the same
(`:1149`, `:1163`), so `if remaining == 0` is unreachable on the live reconciliation path. Two of
the loops compute `remaining` from a deduplicated count (`client.rs:4296`, `:4384`), so a repeated
page adds nothing to it and a limit would not bound them even if one were passed. The reconciliation
instrument filter cannot narrow the query either: it is applied to reports already fetched
(`manager.rs:3106`, `:3118`).

### Why one fix rather than a class of them

This is not a new idea in the codebase, and the closest precedent is in this very file.
`paginate_instruments` already ended a walk whose cursor stopped advancing, by comparing against
the previous one (`client.rs:3352`).

Three other adapters guard theirs as well: `nautilus-architect-ax` with a `seen_cursors` set
(`http/client.rs:1700`, `:1794`), `nautilus-lighter` with one in each of its three reconciliation
walks (`execution.rs:4829`, `:5346`, `:5594`), and `nautilus-polymarket` with a dedicated
paginator. `nautilus-dydx` paginates by a strictly decreasing height and needs no guard.

`nautilus-polymarket` also separates the two cases this PR adopts
(`http/pagination.rs:160`, `:167`, `:332-338`):

- `StalledCursor` - the cursor did not advance from the one just used. This is what we observed.
- `RepeatedCursor` - a cursor seen earlier in the walk reappears, so the pages have cycled.

`CursorWalk::advance` reports both, with the same wording, so a reader who knows one adapter reads
the other without translating.

Worth noting: the `nextPageCursor` in this repository's own
`crates/adapters/bybit/test_data/http_get_executions.json` is `132766%3A2%2C132766%3A2`, which
URL-decodes to `132766:2,132766:2` - both halves already address the same row. The execution-list
test below uses that value verbatim.

Line numbers for `crates/adapters/bybit/src/http/client.rs` in this description are `develop`
line numbers; the Files changed tab shows them shifted, since this PR edits that file.

### Scope

The eight cursor-paginated loops covered by this PR now share `CursorWalk`. Seven return the error:
three in `request_order_status_reports`, one in `request_fill_reports`, two in
`request_position_status_reports`, and one in `request_instrument_statuses`. Returning a
truncated result there would be indistinguishable from a complete one, and reconciliation would
act on it.

`paginate_instruments` uses the same detection but logs and keeps the pages it has read, because
the cost to its caller is an order of magnitude lower. A short instrument list announces itself:
the risk engine denies the next order on a missing instrument with `InstrumentNotFound` and names
it (`crates/risk/src/engine/mod.rs:651`). A short reconciliation history announces nothing. That
walk also feeds a `?` in both the execution and data clients
(`adapters/bybit/src/execution.rs:768`, `adapters/bybit/src/data.rs:796`), so returning an error
would stop the node from starting over a gap it could otherwise report and continue past.

Net effect on startup: nothing that started a node before this change fails to start after it. A
stalled cursor already ended that walk, so it gains only the warning it never had - previously a
normal finish and an abandoned walk both took the same `break`, and the caller could not tell
them apart. A cursor cycle of two or more pages, which the previous-cursor comparison could not
see, now ends instead of repeating forever. After this PR, these eight walks stop on stalled or cycling cursors.
The three account-management walks (`fetch_all_sub_members_paged`,
`fetch_all_escrow_sub_members`, and `fetch_all_sub_api_keys`) remain outside this change;
they have no production callers.

## Documentation

## Testing

Four integration tests in `crates/adapters/bybit/tests/integration/pagination.rs` drive the order
and fill walks against a mock that reproduces the observed page sequence: a page that advances
normally, then a page whose cursor addresses itself and is served unchanged afterwards. The
position-status walk takes the same helper but has no integration test of its own. A fifth test
covers `paginate_instruments` with cursors that cycle `a -> b -> a`, which the previous-cursor
comparison does not catch, and asserts that the instruments read before the cycle are still
returned.

Each test asserts the cursor sequence first, so the output shows which path ran, then that the
call returned. Rows use the field sets from `test_data/http_get_orders_history_with_duplicate.json`
and `test_data/http_get_executions.json` so the mock deserializes through the same models as a
live response.

Two of those tests pass a limit deliberately, because their loops size the budget from a
deduplicated count and a repeating page never advances it.

Five unit tests cover `CursorWalk` directly, including a four-page cycle that a comparison against
the previous cursor alone would not catch.

On `develop`, `nautilus-bybit --test integration` goes 229 passed / 0 failed, to 229 passed /
5 failed with the tests and without the production change, to 234 passed / 0 failed with both.
`nautilus-bybit --lib` goes 719 to 724, all passing. Without the production change the tests
report, for example:

    request_order_status_reports never returned; /v5/order/history was requested 250 times,
    248 of them with the cursor the venue repeated

    request_instruments never returned; instruments-info was requested 599 times,
    cycling between cursor-a and cursor-b

`make format`, `make pre-commit` (66 hooks) and `cargo clippy -p nautilus-bybit --all-targets` are
clean.


cjdsellers:
Hi @Martingale42,

Thank you for the PR. The fix looks right to me: the stalled and cycling cursor cases are both covered, the error-vs-keep split between the reconciliation walks and `paginate_instruments` is the correct call, and the new tests pin the observed venue behavior. This looks good from my side.

It looks like the `pre-commit` check failed. The current failure is `check formatting (Rust)`: missing blank line above `if` in `crates/adapters/bybit/src/http/client.rs` at lines 4426, 4508, 4591, 4866, and 4919. The contributing guidelines need this passing before merge.

This should be easy to fix. Run the following from the repository root, address anything it reports, and push the changes to the same PR:

```bash
make pre-commit
```

On Windows, or if `make` isn't available, run:

```bash
prek run --all-files
```

You can also install the hooks so they run before future commits:

```bash
prek install
```

One thing you might consider when you get a chance: `CursorWalk` covers eight of the eleven cursor walks in `client.rs`, since `fetch_all_sub_members_paged`, `fetch_all_escrow_sub_members`, and `fetch_all_sub_api_keys` still loop without a guard. Those have no production caller, so would you be open to either extending the helper to them or narrowing the "no walk can spin without end" claim to the converted walks? It would keep the stated invariant accurate.

cjdsellers (approved):

Issue #3991, opened 2026-05-04
Custom position_id silently accepted under NETTING OMS


pusteckiy:
## Confirmation

**Before opening a bug report, please confirm:**

- [X] I've re-read the relevant sections of the documentation.
- [X] I've searched existing issues and discussions to avoid duplicates.
- [X] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [X] I've tested this issue using a recent *development* wheel (`dev` develop or `a` nightly) and can still reproduce it.

---

### Expected Behavior

Only positions explicitly opened via `submit_order` with a custom `position_id` should be stored in Redis. No additional positions should be created from for same fills. 

### Actual Behavior

On the third run, the execution engine created additional duplicate positions in Redis using auto-generated IDs (derived from the strategy/instrument name), alongside the expected GRID-tagged positions:

**Expected keys:**
```
trader-67cf6d6f-e8d2-4a3c-aa8d-49a104e521c3:positions:BCHUSDT-LINEAR.BYBIT-GRID
trader-67cf6d6f-e8d2-4a3c-aa8d-49a104e521c3:positions:LTCUSDT-LINEAR.BYBIT-GRID
```

**Unexpected keys that appeared on 3rd run:**
```
trader-67cf6d6f-e8d2-4a3c-aa8d-49a104e521c3:positions:BCHUSDT-LINEAR.BYBIT-marketmaker-BCHUSDT-LINEAR
trader-67cf6d6f-e8d2-4a3c-aa8d-49a104e521c3:positions:LTCUSDT-LINEAR.BYBIT-marketmaker-LTCUSDT-LINEAR
```

Additionally, the execution engine repeatedly detected a position discrepancy (`cached_qty=0.00, venue_qty=0.05`) and failed to resolve it after 3 reconciliation attempts, ultimately giving up.

### Steps to Reproduce the Problem

1. Set up a grid/market-making strategy that submits orders with an explicit `position_id`:
   ```python
   self.submit_order(order, position_id=PositionId(f"{self.instrument_id.value}-GRID"))
   ```
2. Run the strategy — confirm positions are correctly stored in Redis with the `-GRID` suffix.
3. Stop and restart the strategy for a **third** run.
4. Observe that additional positions with auto-generated IDs appear in Redis alongside the expected GRID positions.
5. Observe position discrepancy warnings in the logs that are never resolved.

### Code Snippets or Logs

**Order submission:**
```python
[...]

PR #5020, opened 2026-09-18
Preserve ExecutionEngine handoffs in later submit commands

### LINKED ISSUE #4944: Rust ExecutionEngine can deny an already handed-off order when a later list fails validation
## Expected behavior

Rejecting a later command must not mark an order from an earlier, successfully handed-off submission as `Denied`. Fresh orders in the invalid command should still receive the appropriate native refusal, and the earlier submission must retain a usable lifecycle.

The custom position ID used below is deliberately invalid for NETTING. Rejecting it is correct; this report concerns the effect of that rejection on the earlier submission.

## Actual behavior

The Rust `ExecutionEngine` can apply `OrderDenied` to an order already handed to an execution client when a later invalid list includes that order before its `OrderSubmitted` event reaches the cache.

I reproduced the denial at NT commit `f4a6e629c20d9c77f76e819d1766aadf6b9d1d18` through the native engine and command queue in a downstream Rust integration harness with a stub `ExecutionClient`. The original client invocation occurred exactly once; the invalid list never reached the client; both orders ended in `Denied`. No venue request was executed.

## Steps to reproduce

1. Register an execution client with effective NETTING OMS and cache a matching instrument. The client records submission calls and returns successfully while withholding submission events, modeling the interval before an asynchronous adapter emits `OrderSubmitted`.
2. Submit a valid single order A through the execution engine, with no custom position ID.
3. Assert that the client received A exactly once and that cached A remains `Initialized`.
4. Build a new list containing A and a fresh order B for the same instrument and strategy, with matching initialization events. Set its position ID to a noncanonical value such as `invalid-netting-position`.
5. Execute the list through the native command queue before delivering A's first submission event.
6. Assert that the list did not invoke the client and that B received the NETTING denial. These assertions pass.
7. Assert that A remains `Initialized`, preserving the first submission. This fails: A is `Denied`.

## Code snippets or logs

Executed test: `native_prevalidation_must_not_deny_an_already_forwarded_original`.

```text
assertion `left == right` failed: a later invalid command must not close the original adapter-owned submission
  left: Denied
 right: Initialized
```

The same native mechanism remains in the source inspected at current `develop`, commit `55dd0a65ae2b48866d42480816d580c3a80abe53`:

- [`validate_submission`](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/execution/src/engine/mod.rs#L2001-L2045) permits cached `Initialized`/`Released` orders without distinguishing a prior successful client handoff.
- [`handle_submit_order_list`](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/execution/src/engine/mod.rs#L2380-L2391) applies the position-ID denial to every resolved list member before invoking the client.
- [`deny_order`](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/execution/src/engine/mod.rs#L4500-L4527) updates the canonical cache and publishes the event immediately.

The Polymarket Rust adapter provides a relevant asynchronous interval: its [closed-order check](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/adapters/polymarket/src/execution/orders.rs#L513-L522) occurs before dispatching to submission, while [`OrderSubmitted` is emitted inside the spawned limit-order task](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/adapters/polymarket/src/execution/orders.rs#L129-L140). This adapter scheduling path was inspected, not executed in the reproduction.

There is a potential downstream lifecycle consequence: the [native transition table](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/model/src/orders/mod.rs#L215-L298) has no `Denied` to `Submitted`, `Accepted`, or `Filled` transition, and the [engine drops events whose transition fails](https://github.com/nautechsystems/nautilus_trader/blob/55dd0a65ae2b48866d42480816d580c3a80abe53/crates/execution/src/engine/mod.rs#L3523-L3556). Later submission/fill processing and financial reservation release were not executed in this reproduction; those must not be treated as observed venue outcomes.

## Verification requested

- Preserve the earlier submission when the later invalid command arrives before its first submission event and after it.
- Refuse fresh members without another client invocation for the original.
- Apply later events from the original submission correctly.
- Cover this in native execution tests. Application-side cache restoration, ignored terminal events, or a second submission ledger should not be required.

## Specifications and confirmation

- OS: macOS, arm64.
- Native Rust source build; no Python wheel was used.
- Executed NT pin: `f4a6e629c20d9c77f76e819d1766aadf6b9d1d18`.
- Source also inspected at `069566dafb959b09808a8f1bf4c19733f5999b5d` and current `develop` `55dd0a65ae2b48866d42480816d580c3a80abe53`; the reproduction has not been rerun at either newer pin.
- Reviewed the documented NETTING position-ID rule and native order transitions; searched relevant issues, open PRs, and discussions. No matching report was found in those searches.

## Relations

Related to #3991, which established the NETTING validation used to trigger this case. Its position-ID rule remains correct; this report concerns preserving an earlier client handoff when another command fails that rule.

s1amese2003:

  self-contained fix that does not need prior discussion (the maintainer confirmed the engine path
  on #4944 and asked for a native regression test covering preservation of A, refusal of B, and
  processing of A's later events)
  and, if I used AI,
  this draft
  limitation below

## Summary

`ExecutionEngine::handle_submit_order_list` denied every resolved list member when a later
`SubmitOrderList` failed validation (for example an invalid custom position ID under NETTING),
including an order that an earlier `SubmitOrder` had already handed to the execution client and
that was still `Initialized` because the adapter's `OrderSubmitted` had not arrived yet. That order
became `Denied`, and since the state machine has no `Denied -> Submitted/Accepted/Filled`
transitions, its real venue events were then dropped. A valid later list naming the same order had
the opposite problem: it routed the order to the client a second time. The same held for a command
published to an external client.

The engine now records each successful client handoff (registered client call returned `Ok`, or
the command was published to an external client) until that order's first status transition is
applied, and treats a handed-off order as ineligible for any later submit command, through the
same eligibility gate that already covers `Submitted` and terminal orders. A repeated `SubmitOrder`
for such an order is skipped with a warning and never reaches the client. A `SubmitOrderList`
naming it is denied without reaching the client: only the list's fresh members receive
`OrderDenied` (`ORDER_LIST_DENIED: <list id>`), while the handed-off order keeps its lifecycle and
its later `OrderSubmitted`, `OrderAccepted`, and `OrderFilled` events apply normally.

## Related issues/PRs

Fixes #4944
Related to #3991 (the NETTING position-ID rule itself is unchanged)

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

Not applicable. The only behavior change is for a submit command that names an order the engine
has already handed to a client and that has not yet received its first status event. Before, that
command either denied the order (when the command failed validation) or routed/published it a
second time (when it passed). The normal `Strategy` APIs already reject re-submission of a cached
order before dispatch, so this only affects commands that reach the engine another way.

## Design notes

- Handoff record. After a successful `client.submit_order(...)`, nothing in the cache distinguishes
  the order from one that has not been routed yet: its status stays `Initialized` (or `Released`)
  until the adapter emits `OrderSubmitted`. The cache's execution-client claim is not a handoff
  marker either, because `Strategy::submit_order` / `submit_order_list` already claim the client at
  `add_order` time when the user passes a `client_id`, and the existing test
  `test_submit_order_denied_with_custom_position_id_under_netting` relies on denying a claimed
  `Initialized` order. The engine therefore keeps a private
  `handed_off_orders: RefCell<AHashSet<ClientOrderId>>` (the `execute` path takes `&self`, matching
  the existing `command_count: Cell<u64>`).
- Lifecycle of the record. Inserted in `handle_submit_order` / `handle_submit_order_list` only
  after the client call returns `Ok` (a failed call still denies with `SUBMIT_FAILED` as before),
  and in `execute_command` after a `SubmitOrder` / `SubmitOrderList` is published to an external
  client. Removed in `update_cached_order` as soon as an applied event leaves the order in a
  non-submittable status (`OrderUpdated` on an `Initialized` order keeps the record); this is the
  single funnel for `handle_event` and the reconciliation paths. Cleared on `reset()`. A stale
  entry can never change behavior because status-based staleness takes precedence, so removal is
  only hygiene.
- Where the rule is enforced. In `validate_submission`, which runs before routing, before the
  external-client publish, and before the no-client denial, so none of the denial sites inside
  `handle_submit_order` / `handle_submit_order_list` (venue mismatch, position-ID checks, claim
  failure, `SUBMIT_FAILED`) can ever see a handed-off order. A new
  `SubmissionValidationResult::HandedOff` variant makes a duplicate `SubmitOrder` a logged skip.
  `deny_submission` uses the same predicate, so a denied list denies only its eligible members and
  logs each handed-off member it preserved. This mirrors how the existing gate already preserves
  an order that is `Submitted` when a later list re-submits it, so the fresh member gets the same
  `ORDER_LIST_DENIED` denial before and after the first submission event.
- External-client recording (second commit). The external-client branch publishes the command
  without caching its orders, so the handoff check in `validate_submission` no longer depends on
  the order being cached (cached orders are still checked for status staleness first). Without
  this, a later list naming an externally published order was republished and its fresh member
  was never denied, the same bug class in a different branch.
- `ORDER_LIST_DENIED` versus `INVALID_POSITION_ID`. The fresh member of a later invalid list now
  receives `ORDER_LIST_DENIED: <list id>` (the list is denied because it names an order that is
  not eligible for submission), not the NETTING `INVALID_POSITION_ID` reason the buggy path
  produced. The eligibility gate runs before position-ID validation, consistent with how the
  existing stale-member gate already behaves when the handed-off order is `Submitted`, and it
  applies whether or not the list would otherwise pass validation. This differs from the assertion
  in the issue's harness, so it is called out again under Notes for reviewers.
- Process-local, not persisted. The record lives in the engine only: it is cleared by `reset()`
  and is absent after a restart, and it is released only by the order's first applied status
  transition. After a restart, a still-`Initialized` order that was handed off before the restart
  is treated like any other unrouted `Initialized` order until reconciliation or the client's
  events resolve it (the pre-existing behavior for that state). If an adapter returns `Ok` from
  `submit_order` but never emits any order event, the engine now refuses a re-submission it
  previously forwarded. No current Rust caller (strategy, execution algorithm, emulator, order
  manager, risk throttler, live reconciliation) re-submits such an order, so this is not a
  regression, but it is a deliberate semantic change: after handoff, the client owns the order.
- Docs. `docs/concepts/execution/policies.md` gains one paragraph under "Command outcomes"
  describing submit eligibility (the existing gate was previously undocumented).

## Documentation

- [ ] For PyO3 binding or wrapped Rust doc changes, I ran `make py-stubs` and committed the
  generated output (not applicable: no binding or wrapped doc changed)

## Testing

New `#[rstest]` tests in `crates/execution/tests/integration/exec_engine.rs`, using the existing
`StubExecutionClient` (records submit calls, emits no events):

- `test_submit_order_list_preserves_order_already_handed_to_client`
  (`invalid_list_before_submitted`, `invalid_list_after_submitted`, `valid_list_before_submitted`):
  submits A through the engine, asserts the client saw A exactly once and A is `Initialized`,
  optionally applies `OrderSubmitted`, then executes a list of A + fresh B (with the invalid
  NETTING position ID `invalid-netting-position`, or none). Asserts the client was not invoked
  again, A keeps its status and event count, B is `Denied` with `ORDER_LIST_DENIED: L-LATER`, and
  that A's later `OrderSubmitted`, `OrderAccepted`, and `OrderFilled` reach `Filled` with the
  NETTING position opened. The two `before_submitted` cases fail on `develop`; the
  `after_submitted` case was already handled by the existing gate.
- `test_submit_order_skips_duplicate_before_submitted_event`: a repeated `SubmitOrder` for a
  handed-off order does not reach the client again and does not disturb the later `OrderSubmitted`
  (fails on `develop`).
- `test_submit_order_list_records_handoff_for_every_member`: a valid list (A, B) is handed off,
  then a duplicate `SubmitOrder` for A and a later list naming B plus a fresh C arrive before any
  `OrderSubmitted`. Asserts the client saw A and B exactly once, A and B stay `Initialized` with
  one event each, C is `Denied` with `ORDER_LIST_DENIED: L-LATER`, and both `OrderSubmitted`
  events then apply (covers the list-handoff record site).
- `test_external_client_handoff_preserves_order_in_later_submit_commands`: A is published to an
  external client, then a duplicate `SubmitOrder` and a later list naming A plus a fresh B arrive
  before `OrderSubmitted`. Asserts only the first command was published, A stays `Initialized`
  with one event, B is `Denied` with `ORDER_LIST_DENIED: L-EXTERNAL`, and A's `OrderSubmitted`
  then applies (fails without the second commit).

Validation:

- `make format`: no changes (Linux and Windows).
- `make pre-commit` on Linux (all 71 configured hooks, no exclusions, on a clean clone of this
  branch with the extension built): 66 passed, 0 failed, 4 skipped (no matching files), working
  tree unmodified by hooks.
- `make pre-commit` on a Windows host: same result apart from two pre-existing Windows-only
  issues unrelated to this change and reproduced on unmodified `develop`: the `actionlint`
  self-test compares paths with forward slashes, and `nautilus-network`'s test target calls the
  Unix-only `socket2::Socket::tcp_keepalive_time` (`crates/network/src/net.rs`), which fails the
  workspace clippy pass on Windows. This PR touches neither.
- `cargo nextest run -p nautilus-execution --lib --tests`: 1615 passed, 0 failed (rebased onto
  current `develop`; run on Windows and on Linux).
- `scripts/clippy-changed.sh` (the clippy hook, scoped to `nautilus-execution` with the hook's
  feature set) and the pinned nightly `cargo fmt --check`: clean.
- `nautilus-risk` and `nautilus-backtest` suites: pass (the follow-up commit only touches the
  engine's external-client branch, `validate_submission`, and tests).

## Notes for reviewers

- Denial reason for the fresh member: `ORDER_LIST_DENIED` rather than the NETTING
  `INVALID_POSITION_ID` reason the buggy path produced (see Design notes). Consistent with the
  existing gate when the handed-off order is `Submitted`, but it differs from the harness assertion
  in the issue. If the list should carry the validation-specific reason instead, the change is
  local to `deny_submission`.
- Mechanism choice: engine-local state rather than a cache or order-level signal, because the
  cache client claim is set at `add_order` time (before any handoff) and existing tests rely on
  denying claimed `Initialized` orders.
- The handoff record is process-local and not persisted, and an adapter that returns `Ok` from
  `submit_order` but never emits any order event keeps its record, so a later re-submission of that
  same client order ID is skipped rather than routed. Intended (the client owns the order after
  handoff), but a semantic change for such adapters.


cjdsellers (commented):
Hi @s1amese2003,

Thanks for the PR. This closes the lifecycle gap between dispatching an order and applying its first status event, so a later command can't deny or resubmit an order already owned by the client. The implementation looks good from my side :ok_hand:.

**A couple of changes before this can merge**:

I'd prefer `orders_dispatched`, `is_dispatched()`, and `SubmissionValidationResult::Dispatched` for the new names. “Dispatched” describes the execution-engine boundary more directly; could you use that terminology consistently in the accompanying comments and logs?

Could you also add a regression test that submits an order, resets the engine before its first status event, and submits the same order ID again? The reset already clears the set correctly; the test should confirm the second submission reaches the client.

s1amese2003:
Thanks for the review. I've renamed the tracking to `orders_dispatched`, `is_dispatched()`, and `SubmissionValidationResult::Dispatched`, with consistent terminology in the comments, logs, tests, and docs.

I also added a regression test that resets the engine before the first status event and confirms the same order ID reaches the client again. The test fails if the reset cleanup is removed.

The execution suite passes all 1,616 tests on Windows, and `make format` and the full `make pre-commit` pass on Linux.

cjdsellers (approved):
@s1amese2003, thanks for the quick turnaround on both changes. The `dispatched` naming is consistent across the engine, logs, and tests, and the reset regression test covers exactly the case I was after, so this looks good to merge 👌.

When you get a chance could you refresh the PR description to match the revised code? It still uses the earlier `handed_off` wording and doesn't mention the reset regression test.

I'll pick up the small docs addition after merge, noting that the dispatch record is process-local, cleared on reset and absent after a restart.

s1amese2003:
@cjdsellers Thanks for the review and merge! I've updated the PR description to use the `dispatched` terminology consistently, include the reset regression test, and reflect the final validation results. Thanks also for taking care of the docs addition.

PR #5022, opened 2026-09-18
Fix AroonOscillator MAX_PERIOD window capacity

### LINKED ISSUE #4995: AroonOscillator accepts MAX_PERIOD but cannot retain its required period + 1 window
# Bug Report

## Confirmation

- [x] I've re-read the relevant implementation and tests.
- [x] I've searched existing issues, PRs, and discussions to avoid duplicating active work.
- [x] I've reviewed the current `develop` implementation and this does not appear to be intentional.
- [x] I've reproduced the issue against the published `v2.0.0rc5` pre-release wheel.
- [x] I've confirmed the same capacity invariant is present in current `develop` at `608200232f17f6944bc247515c9ee8352bbbdaa1`.

I found this while looking through the current Rust `AroonOscillator` implementation.

## Expected behavior

`AroonOscillator::new(MAX_PERIOD)` should be a valid configuration if the constructor explicitly accepts it.

The current maximum is:

```rust
pub const MAX_PERIOD: usize = 1_024;
```

and the constructor accepts every period up to and including that value:

```rust
assert!(
    period <= MAX_PERIOD,
    "AroonOscillator: period must be ≤ {MAX_PERIOD} (received {period})"
);
```

Aroon uses a `period + 1` observation window, so a period of `1024` requires the oscillator to retain `1025` highs and `1025` lows.

The implementation itself relies on this invariant:

```rust
let required = self.period + 1;

if !self.initialized && self.total_count >= required {
    self.initialized = true;
}
```

and `calculate_aroon()` asserts the same thing:

```rust
debug_assert_eq!(self.high_inputs.len(), self.period + 1);
```

So if `period = MAX_PERIOD`, I would expect both internal buffers to be able to retain `MAX_PERIOD + 1` values.

## Actual behavior

The buffers are currently defined with capacity `MAX_PERIOD`, not `MAX_PERIOD + 1`:

```rust
high_inputs: ArrayDeque<f64, MAX_PERIOD, Wrapping>,
low_inputs: ArrayDeque<f64, MAX_PERIOD, Wrapping>,
```

This means the public maximum period and the internal storage invariant disagree.

For:

```rust
let period = MAX_PERIOD;
```

we get:

```text
accepted period        = 1024
required window        = 1025
ArrayDeque capacity    = 1024
```

`update_raw()` also only explicitly removes an old value when:

```rust
if self.count == self.period + 1 {
    let _ = self.high_inputs.pop_front();
    let _ = self.low_inputs.pop_front();
} else {
    self.count += 1;
}
```

At the boundary, the deque reaches its fixed capacity before `count` can reach `period + 1`.

The repository already has a unit test documenting how `Wrapping` behaves when a deque is full:

```rust
#[rstest]
fn arraydeque_wraps_when_full() {
    const CAP: usize = 3;
    let mut buf: ArrayDeque<usize, CAP, Wrapping> = ArrayDeque::new();

    for i in 0..=CAP {
        let _ = buf.push_back(i);
    }

    assert_eq!(buf.len(), CAP);
    assert_eq!(buf.front().copied(), Some(1));
    assert_eq!(buf.back().copied(), Some(3));
}
```

So pushing the `1025`th Aroon observation into a `Wrapping` deque with capacity `1024` evicts the oldest observation and leaves the length at `1024`.

Because these are `ArrayDeque<_, MAX_PERIOD, Wrapping>` instances, the storage can never satisfy the invariant checked by:

```rust
debug_assert_eq!(self.high_inputs.len(), self.period + 1);
```

for a period of `1024`.

In a debug build, `calculate_aroon()` reaches that assertion expecting `1025` while the deque can only contain `1024`.

In a release build, the debug assertion is removed, so calculation can proceed against a window smaller than the period the user requested.

This exact boundary problem was also called out, but intentionally left unfixed, in merged PR #4914:

> `AroonOscillator::new(MAX_PERIOD)` is accepted, but the window needs `period + 1` elements while the backing `ArrayDeque` capacity is `MAX_PERIOD`, so at `period = 1024` the window silently caps at 1024 elements (and the `debug_assert_eq!` fires in debug builds).

## Runtime reproduction on the published pre-release

I reproduced the issue directly against the published `nautilus_trader==2.0.0rc5` wheel.

Environment:

```text
nautilus_trader=2.0.0rc5
python=3.14.3
platform=Linux-7.0.0-1009-aws-x86_64-with-glibc2.36
period=1024
```

Reproduction:

```python
from nautilus_trader.indicators import AroonOscillator

period = 1024
aroon = AroonOscillator(period)

# Oldest observation is the unique highest high.
aroon.update_raw(1000.0, 5.0)

# Fill the remainder of the required 1025-observation window.
for _ in range(period):
    aroon.update_raw(10.0, 1.0)

print(f"initialized={aroon.initialized}")
print(f"count={aroon.count}")
print(f"aroon_up={aroon.aroon_up}")
print(f"aroon_down={aroon.aroon_down}")
print(f"value={aroon.value}")
```

Actual output from the published `v2.0.0rc5` wheel:

```text
initialized=True
count=1025
aroon_up=100.0
aroon_down=100.0
value=0.0
```

With a correct `1025` element window, the unique high of `1000.0` is still exactly 1024 periods back when the indicator first initializes.

The expected Aroon Up is therefore:

```text
100 * (1024 - 1024) / 1024 = 0
```

The lowest low is present in the newer observations and, because ties resolve to the most recent occurrence, Aroon Down should be `100`.

So the expected state is:

```text
aroon_up=0.0
aroon_down=100.0
value=-100.0
```

The published pre-release instead returns:

```text
aroon_up=100.0
aroon_down=100.0
value=0.0
```

This confirms that the boundary mismatch produces an actual indicator signal error in the published pre-release, not only an internal capacity mismatch.

## Why this affects the indicator value

This is not only an internal length mismatch.

For a period of `1024`, the first observation in the reproduction contains the unique highest high:

```text
high = 1000.0
```

With the required `1025` observation window, that value should still be present exactly `1024` periods back when the oscillator initializes.

That makes the correct Aroon Up:

```text
0.0
```

With the current `Wrapping` capacity, the first observation has already been displaced before `calculate_aroon()` runs.

The remaining highs are all:

```text
10.0
```

Because equal highs resolve to the most recent occurrence, the calculation treats the newest `10.0` high as the current highest high and returns:

```text
aroon_up=100.0
```

instead of:

```text
aroon_up=0.0
```

This changes the final oscillator value from:

```text
-100.0
```

to:

```text
0.0
```

at the exact maximum period accepted by the constructor.

### Published pre-release

1. Install `nautilus_trader==2.0.0rc5`.
2. Construct `AroonOscillator(1024)`.
3. Feed one observation with a unique highest high.
4. Feed another `1024` observations to reach the required `period + 1` window.
5. Inspect `aroon_up`, `aroon_down`, and `value`.

Minimal reproduction:

```python
from nautilus_trader.indicators import AroonOscillator

period = 1024
aroon = AroonOscillator(period)

aroon.update_raw(1000.0, 5.0)

for _ in range(period):
    aroon.update_raw(10.0, 1.0)

print(aroon.aroon_up)
print(aroon.aroon_down)
print(aroon.value)
```

Current `v2.0.0rc5` output:

```text
100.0
100.0
0.0
```

Expected:

```text
0.0
100.0
-100.0
```

### Rust boundary regression

A minimal Rust regression case for the underlying window invariant would be:

```rust
#[test]
fn max_period_retains_period_plus_one_window() {
    let mut aroon = AroonOscillator::new(MAX_PERIOD);

    for _ in 0..=MAX_PERIOD {
        aroon.update_raw(1.0, 0.0);
    }

    assert_eq!(aroon.high_inputs.len(), MAX_PERIOD + 1);
    assert_eq!(aroon.low_inputs.len(), MAX_PERIOD + 1);
}
```

With the current implementation, the backing `ArrayDeque` cannot retain `MAX_PERIOD + 1` elements.

In a debug build, the existing `debug_assert_eq!(self.high_inputs.len(), self.period + 1)` is reached while the deque is still limited to `MAX_PERIOD` elements.

## Relevant existing tests

There is already a test which documents the intended Aroon window invariant:

```rust
#[rstest]
fn test_window_size_period_plus_one() {
    let period = 7;
    let mut aroon = AroonOscillator::new(period);

    for _ in 0..=period {
        aroon.update_raw(1.0, 0.0);
    }

    assert_eq!(aroon.high_inputs.len(), period + 1);
    assert_eq!(aroon.low_inputs.len(), period + 1);
}
```

The test is correct for `period = 7`, but it does not exercise the same invariant at `MAX_PERIOD`.

The current implementation therefore has an invariant which is tested at a normal period but cannot hold at the constructor's accepted upper bound.

A regression test at `MAX_PERIOD` should ideally cover both the internal window size and an extrema pattern which makes the lost oldest observation visible in the resulting Aroon value.

## Possible direction

I see two possible fixes:

1. Keep `MAX_PERIOD = 1024` as the supported user-facing maximum and give the internal Aroon buffers capacity for `MAX_PERIOD + 1`.
2. Change the accepted maximum period to `MAX_PERIOD - 1` so the existing `1024` element buffers remain sufficient.

I think the first option is less surprising because the current public constructor explicitly says that `1024` is valid, and the implementation already defines the required window as `period + 1`.

If keeping `1024` valid is the preferred direction, I can send a focused PR with regression coverage for:

- initialization at exactly `MAX_PERIOD + 1` observations
- correct extrema retention at `MAX_PERIOD`
- the resulting Aroon values at the boundary
- rollover after the next observation

## Specifications

- `nautilus_trader`: `v2.0.0rc5`
- Runtime reproduction: confirmed against the published `2.0.0rc5` wheel
- Python: `3.14.3`
- Platform: `Linux-7.0.0-1009-aws-x86_64-with-glibc2.36`
- Current `develop`: same capacity invariant present at `608200232f17f6944bc247515c9ee8352bbbdaa1`
- Component: Rust `AroonOscillator` / Python binding
- Adapter/venue: n/a

Tyagiquamar:
### Description

In `AroonOscillator` (`crates/indicators/src/momentum/aroon.rs`), initialization accepts any `period <= MAX_PERIOD` (1024). However, calculating the Aroon oscillator requires examining the highest high and lowest low across a window of `period + 1` inputs (`self.period + 1`, or up to 1025 observations).

Because `high_inputs` and `low_inputs` were declared with `ArrayDeque<[f64; MAX_PERIOD], Wrapping>` (capacity 1024), pushing the 1025th observation when `period == 1024` triggered the `Wrapping` eviction strategy on the 1024-element deque, retaining only 1024 values instead of the required 1025. This caused:
- In debug builds: `debug_assert_eq!(self.high_inputs.len(), self.period + 1)` panic on the 1025th tick.
- In release builds: calculation over a truncated window of length 1024 rather than 1025, producing incorrect oscillator outputs.

### Changes
- Defined `const MAX_CAPACITY: usize = MAX_PERIOD + 1;` in `crates/indicators/src/momentum/aroon.rs`.
- Backed `high_inputs` and `low_inputs` with `ArrayDeque<[f64; MAX_CAPACITY], Wrapping>`.
- Added unit tests in Rust (`aroon.rs`) and Python (`python/tests/unit/indicators/test_aroon.py`) verifying window retention and expected oscillator values at `period = MAX_PERIOD`.

Closes #4995


wbizmo:
One regression case still looks worth adding here.

The `MAX_PERIOD` test covers the case identified in #4995 where the oldest high is the unique maximum and must survive the full `period + 1` window. However, the current fixture does not exercise the symmetric boundary for `low_inputs`: the first low is `5.0`, while subsequent lows are `1.0`, so the oldest low is not the minimum.

Since the capacity mismatch affected both deques, I think the regression coverage should also include a mirrored case where the oldest observation contains the unique lowest low. At initialization, that should produce:

    aroon_up   = 100.0
    aroon_down = 0.0
    value      = 100.0

followed by the corresponding rollover on the next observation.

That would cover both sides of the boundary behavior originally described in #4995.

cjdsellers:
Hi @Tyagiquamar,

Thanks for the PR, and thanks @wbizmo for the follow-up on top of the original report in #4995.

The fix looks right to me: giving both deques capacity for `MAX_PERIOD + 1` restores the `period + 1` window at the boundary. We'd like to include the mirrored low-side case @wbizmo suggested, since the current fixture only proves high-side retention through values and the mismatch affected both deques. If you're able, add the unique-lowest-low case with the rollover assertion so both sides stay covered.

The PR template hasn't been observed here, which means the `CONTRIBUTING.md` and `AI_POLICY.md` confirmations it contains haven't been agreed to. Could you confirm you've read both guides and run `make format`, `make pre-commit`, and the indicator tests locally? Future PRs need to keep the template intact.

Tyagiquamar:
Hi @cjdsellers, thanks for the guidance! I have added the mirrored unique-lowest-low boundary regression test in \	est_aroon.py\ and \roon.rs\ with the rollover assertion as suggested. I also confirm that I have read and agree to \CONTRIBUTING.md\ and \AI_POLICY.md\, and have verified formatting, linter, and indicator tests locally. Thanks!

wbizmo:
@Tyagiquamar I need to be much more explicit here because the concern I raised has still not actually been addressed.

This is not about two contributors independently arriving at the same solution.

I had already done the investigation, identified the failure mode, worked through the boundary condition, and documented the solution on the issue. You then took that work forward into your own contribution without coordinating with me or acknowledging where the substantive investigation came from.

Even after that, I continued engaging constructively. I reviewed the implementation, identified what was still missing, and pointed out the additional regression coverage required so that the fix would actually be correct.

And yet, even now, there has still been no acknowledgement of any of that.

At this point, the way this has been handled gives the appearance that obtaining another merged PR on your GitHub profile is being prioritized over the collaborative process and contribution etiquette expected in this project. I cannot know your personal motivation, but that is how the conduct reads from the outside: the investigation was already done by another contributor, the solution was already laid out, concerns about coordination were explicitly raised, and instead of addressing them, the work continued toward getting the PR merged.

That is exactly the kind of behavior open-source projects should discourage.

A merged PR is not supposed to be a statistic collected at any cost. Open source depends on contributors being willing to share investigations, partial solutions, reviews, and technical findings openly. If someone can take another contributor's detailed investigation, race it into a PR, decline to coordinate, ignore requests for acknowledgement, and still receive the merged contribution as though nothing happened, the message to everyone else is obvious:

**Do not share your investigation until you already have a PR ready, because somebody else can simply package your work first.**

That would be an extremely unhealthy precedent for this repository.

I would also ask you, genuinely, to consider how you would regard this situation if the roles were reversed. If you had spent your time debugging an issue, identified the exact problem, documented the solution publicly for the maintainers, and another contributor then turned that work into their own PR without coordinating with you or even acknowledging you after you raised the concern, would you consider that good open-source etiquette?

I don't think contributors should have to compete against the very people they are trying to collaborate with.

What makes this more concerning is that you have explicitly confirmed that you have read and agreed to `CONTRIBUTING.md` and `AI_POLICY.md`. Contribution guidelines are meaningful only if the standards they describe are actually enforced when contributor conduct is tested.

@cJDsellers, I would therefore strongly ask for maintainer guidance before this is treated as an ordinary merge.

The relevant sequence is straightforward:

1. I investigated the issue and documented the technical problem and proposed solution.
2. That work was subsequently incorporated into this contribution without coordination or attribution.
3. I raised the concern directly.
4. I nevertheless continued helping and identified an additional problem in the implementation.
5. That correction was then incorporated as well.
6. The underlying attribution and coordination concern has still not been acknowledged.

I am not claiming that commenting on an issue gives somebody permanent ownership of it. It does not.

I am saying that there is a substantial difference between independently working on the same problem and taking a contributor's already-published investigation and solution, converting it into a PR, and refusing to meaningfully acknowledge or coordinate with that contributor even after they call attention to it.

Those should not be treated as the same thing.

I think simply merging this and moving on without addressing what happened would set precisely the wrong precedent. If this repository has contribution rules around collaboration, attribution, and responsible participation, then situations like this are where those rules actually need to mean something.

I have already gone considerably further than necessary in trying to be constructive about this, including helping correct the implementation itself.

What I am asking for now is very simple: **acknowledge what happened, properly acknowledge the work that was used, and establish clearly that this is not how contributors should obtain merged PRs in this project.**

Otherwise, the precedent being created is that contributors can ignore coordination and attribution, collect the merge, and face no consequence as long as the resulting code eventually passes review.

That would be deeply discouraging for anyone contributing serious investigative work in good faith.

Tyagiquamar:
@wbizmo, you're right to call this out. I went back through #4995 and the sequence is clear: you had already done the substantive investigation, identified the `MAX_PERIOD` / `period + 1` capacity mismatch, proposed expanding the buffers to `MAX_PERIOD + 1`, and outlined the boundary/value/rollover regression coverage before I opened #5022.

I should have coordinated with you before implementing that direction, and when you raised the concern on the issue I should have addressed it directly instead of only continuing with the technical changes. I also incorporated the mirrored low-side regression you suggested without properly acknowledging your contribution to the fix.

I'm sorry for handling that badly.

@cjdsellers, I'm going to close #5022 so @wbizmo can submit the implementation they were already preparing. Thank you both for the feedback, and apologies for the unnecessary friction here.

wbizmo:
@Tyagiquamar thanks for acknowledging this properly. I appreciate that.

Just to clarify one thing: I was not asking you to close #5022 or give up the PR.

My concern was never about taking the implementation away from you or competing over who gets the merge. It was about acknowledgement, coordination, and attribution, especially where another contributor's investigation and proposed solution materially informed the work.

I actually think the healthier outcome is for the contribution to continue, with the history and contributions properly acknowledged, rather than treating this as though one of us has to "own" the fix.

So if #5022 can still be reopened, I have no objection to that at all.

What mattered to me was that what happened was acknowledged and that we establish the right precedent around collaborating with and crediting other contributors. You've now done that, and I appreciate the apology.

No need to throw away the work you've already done.

wbizmo:
@Tyagiquamar just a quick update: I’ve now opened a pull request with my implementation of the fix.

Thanks again for acknowledging the earlier coordination and attribution concerns. I appreciate you addressing that directly.

I’ve gone ahead with the implementation I had been preparing, including the relevant regression coverage, and the PR is now up for review.

cjdsellers:
@wbizmo, thanks for laying this out clearly and bringing the coordination concern to our attention. We'd missed that you had already offered to submit the fix when you raised the issue. My earlier response acknowledged your report and test feedback but overlooked that offer and your concern about the work being taken forward without coordination.

Our contributing guidelines already ask contributors to coordinate and avoid competing implementations, but the wording needs to cover this situation more clearly: when an issue author has offered a fix and is preparing it, another contributor should coordinate with them before opening a PR. We'll clarify that shortly and improve how we check linked issue discussions during PR triage so these offers and ongoing work aren't overlooked. Thank you for continuing to help with the regression coverage while raising this.

wbizmo:
@cjdsellers, thanks for taking the time to revisit this and for the clarification. I really appreciate it.

My main concern was the coordination rather than ownership of the implementation, so I’m glad the distinction is clearer now. Improving the contributor guidance around cases where an issue author has already offered to implement a fix should help avoid similar situations going forward.

I also want to acknowledge @Tyagiquamar’s work on #5022. He had already taken the same fix direction forward and incorporated the mirrored low-side regression case after the discussion there. I’ve kept #5022 explicitly referenced in #5037 as prior related work so that contribution history is preserved rather than lost.

I’ve now submitted the focused fix in #5037, including the boundary, value, and rollover regression coverage discussed in this issue.

Thanks again for looking into this and for being receptive to the feedback.

PR #5025, opened 2026-09-19
Create ParquetDataCatalog base directory when missing

### LINKED ISSUE #4950: ParquetDataCatalog reports an opaque canonicalization error when the local base path does not exist
# Bug Report

## Confirmation

Tested on `2.0.0rc4`. The relevant code is unchanged on `develop` HEAD.

## Expected behavior

`ParquetDataCatalog("some/path")` on a local path either creates the directory, or fails with a message that says the directory must exist.

Creating it would be consistent with the rest of the catalog's behavior: once the base directory exists, the catalog creates its per data type and per instrument subdirectories on write without being asked.

## Actual behavior

Constructing a catalog on a local path that does not yet exist raises an `OSError` whose message describes an internal object store operation and does not indicate what the user should do.

## Steps to reproduce

```python
import tempfile
from pathlib import Path
from nautilus_trader.persistence import ParquetDataCatalog

missing = Path(tempfile.gettempdir()) / "nautilus-does-not-exist" / "catalog"
print(f"path exists: {missing.exists()}")   # False
ParquetDataCatalog(str(missing))
```

Creating the directory first makes it work.

## Code snippets or logs

```
OSError: Failed to create ParquetDataCatalog: Generic LocalFileSystem error:
Unable to canonicalize filesystem root: /var/folders/.../nautilus-does-not-exist/catalog
```

`crates/persistence/src/backend/catalog.rs`: `ParquetDataCatalog::new` forwards to `from_uri`, which calls `create_object_store_location_from_path`. For a local path that builds an `object_store` `LocalFileSystem` rooted at the prefix, which requires an existing directory and reports the failure as a canonicalization error.

## Impact

Mild, but worth reporting because of where it surfaces. A live trading node aborted at trader startup because a recording actor constructed its catalog on a fresh checkout where `catalog/` had not been created yet. "Unable to canonicalize filesystem root" does not point at "this directory does not exist", which made it slower to diagnose than it should have been.

The docstrings throughout `catalog.rs` use `/tmp/nautilus_data` as the example path, which reads as though a not-yet-existing path is fine.

## Suggested fix

For local paths, `create_dir_all` before constructing the store. Failing that, map the error to something actionable, for example:

```
catalog base path does not exist: <path>
```

Happy to open a small PR for whichever behavior you prefer.

## Specifications

- Python version: 3.13.15
- Installed from: PyPI wheel
- Adapter/venue: n/a

AmitKumarDeoghoria:

## Summary

Constructing `ParquetDataCatalog` on a local base path that does not yet exist raised an internal `object_store` canonicalization error ("Unable to canonicalize filesystem root: <path>") rather than either creating the directory or reporting an actionable message. `create_local_store` now calls `std::fs::create_dir_all` on the resolved local path before `LocalFileSystem::new_with_prefix`, matching the catalog's existing behavior of creating per-type and per-instrument subdirectories on write, and matching the `/tmp/nautilus_data` usage in the docstring examples.

## Related issues/PRs

Resolves #4950

## Type of change

- [x] Bug fix (non-breaking)

## Testing

Added \`test_create_object_store_from_path_local_creates_missing_directory\` covering construction against a nested not-yet-existing path and asserting the base directory is created. All 271 tests in \`nautilus-persistence\` pass; \`cargo fmt\` and \`prek run\` on the modified file both pass.


cjdsellers:
Hi @AmitKumarDeoghoria,

Thanks for the PR. Creating the directory is fine, but `create_local_store` is the wrong layer for it.

It's also the fallback for unmatched schemes, and `normalize_path_to_uri` passes anything with `://` through untouched. So `ParquetDataCatalog::from_uri("S3://my-bucket/data")` now creates `./S3:/my-bucket/data` and silently writes to local disk instead of erroring. That also flips `test_register_object_store_from_uri_invalid_uri` to a failure.

`write_batches_to_parquet`, `combine_parquet_files`, and `min_max_from_parquet_metadata` pass a Parquet *file* path in, so they now get a directory created over the file: writes land at `/tmp/out.parquet/tmp/out.parquet`, and an existing file fails with `AlreadyExists`.

Scoping this to the catalog base path avoids both: `create_storage_backend_from_path` in `common/storage.rs` already does it for `file://` URIs. `test_register_object_store_from_uri_nonexistent_path` needs updating either way, since it asserts the error this removes.

Two conventions that fail `make pre-commit`: `|e|` not `|err|`, and `, got` should be `, was` / `, received` / `, found`.

AmitKumarDeoghoria:
Thanks for the review :

- Reverted the `create_local_store` change (both the scheme-fallthrough on `S3://` and the Parquet-file-path regression in `write_batches_to_parquet`/`combine_parquet_files`/`min_max_from_parquet_metadata` would have surfaced there).
- Added the `create_dir_all` to `ParquetDataCatalog::from_uri` for `file://` URIs, mirroring `create_storage_backend_from_path`.
- Moved the regression test to `tests/integration/test_catalog.rs` as `test_from_uri_local_creates_missing_directory`.
- Switched to `|e|` and `received` per the review conventions.

On `test_register_object_store_from_uri_nonexistent_path`: with the scoped fix, the session's `register_object_store_from_uri` goes through `create_object_store_location_from_path` directly (unchanged), so the assertion still holds.

cjdsellers:
@AmitKumarDeoghoria, thanks for the update. Scoping creation to the catalog fixes the earlier store-factory issues. Two cases still need addressing:

- Windows UNC paths such as `\\server\share\catalog` become a relative native path through `file_uri_to_native_path`. Creating that path now lets the catalog silently open under the working directory. Network-share paths need to retain their location or fail before creating a local directory.
- Migration callers also use `from_uri` to open their source. A missing source now gets created and produces a successful zero-file migration, including during a dry run. Those callers need to preserve the missing-source error and leave the source unchanged.

Regression coverage for both cases would protect these boundaries.

AmitKumarDeoghoria:
Thanks — pushed `65e48127cb`:

- Reverted the auto-create in `ParquetDataCatalog::from_uri` so all Rust callers (including migration sources) get the strict behaviour again.
- Moved the auto-create into `PyParquetDataCatalog::py_new`, using the same `Url::to_file_path()` pattern the migration destination handling already uses. UNC (`file://server/…`) and cloud URIs return `Err(())` from `to_file_path()`, so they skip the create and fail loudly in `from_uri` instead of landing under the CWD.
- Added `migration_rejects_missing_source` (parameterised over dry-run and apply) asserting the migration errors and leaves both the missing source and the target untouched.
- Added `test_parquet_data_catalog_creates_missing_local_directory` and `test_parquet_data_catalog_does_not_create_local_directory_for_unc_uri` on the Python side.
- Existing `test_from_uri_returns_error_on_missing_local_path` was updated to assert the reverted strict behaviour of `from_uri`.

All 596 Rust persistence tests pass; the 6 Python tests in `test_parquet_catalog.py` pass; `make pre-commit` on all changed files is clean.

AmitKumarDeoghoria:
Just noticed #4950 is closed and the error-message fix landed in `519ed08` — thanks for taking that on directly.

Do you still want the auto-create-on-missing behaviour from this PR as a follow-up on top of your error message, or should I close #5025?

cjdsellers:
Thanks for working through the revisions. I think the clearer error message resolves #4950, and I'd prefer to keep directory creation explicit. That also avoids silently opening an empty catalog when a path is mistyped, so let's close this PR.

If we pursued auto-creation, the path-conversion mismatch, Windows UNC handling, and conflicting Python tests would still need addressing and validation. There's no need to spend another round on those now. Thanks again for the contribution.

PR #5027, opened 2026-09-19
Use typed selectors for catalog operations

faysou:
### Summary

Parquet catalog operations identified their target with a directory-name string, so a typo or an
unsupported family surfaced as a runtime error and Python callers had no type to check against.
This change introduces `CatalogDataType`, the persistence sum of `NautilusDataType`,
`NautilusRecordType`, and `NautilusInstrumentType`, and takes it in every family-scoped catalog
operation. `NautilusDataType.Instrument` addresses every instrument class the catalog stores, so
callers never name a class to reach instrument coverage, consolidation, listing, or metadata.
`BacktestDataConfig` takes a `NautilusDataType` value rather than a name.

```mermaid
flowchart LR
    DT["NautilusDataType
    incl. Instrument"] --> CT[CatalogDataType]
    RT[NautilusRecordType] --> CT
    IT[NautilusInstrumentType] --> CT
    CT --> Res["parquet_catalog_data_type_path_prefixes
    Data(Instrument) -> every class dir
    otherwise -> one dir"]
    Res --> Ops["coverage, consolidation,
    reset/extend, listing, metadata"]
    DT --> Del["delete_data_range
    data families only"]
    IT --> Filter["optional instrument_type
    filter on instrument reads"]
    DT --> Cfg["BacktestDataConfig
    data_type"]
```

### Design decisions

- `CatalogDataType` lives in `crates/persistence/src/catalog/types.rs`, beside the query types and path
  prefix functions it selects over, because only catalogs address stored families. It is a sum of
  the three existing selectors rather than a new data kind, and `From` impls let call sites pass any
  of them. `From` and `Into` are the only construction paths, so every value of the three families
  is a valid selector and no accepted value silently produces empty coverage or a no-op.
- Parquet directory names are unchanged. `parquet_catalog_data_type_path_prefixes` maps a selector onto
  the existing layout and is the single place that knows the aggregate instrument family spans every
  class directory. Coverage intervals, first and last timestamps, missing intervals, consolidation
  (whole and by period), file-name reset, file listing and filtering, identifier listing, metadata,
  and instrument listing all iterate its result, so the fan-out cannot drift between operations.
  `parquet_catalog_data_type_table_stem` names registered query tables and never addresses storage.
- `extend_file_name` needs a directory to write into, so for the aggregate family it resolves the
  instrument class that already stores the identifier and reports an error naming
  `NautilusInstrumentType` when none does.
- Instrument reads keep an optional class filter. `CatalogInstrumentQuery::with_instrument_type`
  targets one class, and omitting it reads every class.
- `delete_data_range` takes `NautilusDataType` rather than `CatalogDataType`. Ranged deletion applies to
  data files only, so a record or instrument family is a compile error rather than a runtime
  rejection. Families without ranged deletion are named explicitly, so a new data type fails to
  compile rather than falling into a wildcard.
- `consolidate_data_by_period` keeps the `CatalogDataType` signature shared by the other file
  operations but rejects record and instrument selectors up front with a message naming the
  selector and pointing to `consolidate_data`. Period consolidation is a typed rewrite that only
  data families have, and the previous failure named a directory instead of the selector.
- `list_instruments` deduplicates across instrument classes, because one identifier can be stored
  under more than one class, and aggregate `extend_file_name` extends every class directory that
  stores the identifier rather than stopping at the first.
- The trait-level `query_metadata` honors the instrument class filter carried by `CatalogQuery`,
  so an equity-only query reads the equity directory alone, as the data, identifier, and display
  reads already did.
- The selector enum is `CatalogDataType`; the trait that previously held that name, mapping a Rust
  data type to its `NautilusDataType`, is `HasCatalogDataType`, following the `HasTsInit` naming.
- Every family-scoped operation names its selector `data_type`, the keyword the typed query methods
  already use, so one keyword covers the catalog surface. The Rust parameter type stays
  `CatalogDataType`; only the Python keyword changed.
- `ParquetDataCatalog.query` takes a `NautilusDataType` and dispatches one explicit arm per variant,
  so a new data family is a compile error rather than a fall-through to the custom-data path.
  `Instrument` returns instrument objects; records stay on the record methods because `query`
  returns `Data` values and a record is not one.
- The optional class filter on instrument reads accepts a `NautilusInstrumentType` only; the string
  spelling is gone from the changed surface.
- Instrument requests carry no class selector. `RequestInstrument`, `RequestInstruments`, the actor
  and strategy request methods, and the data engine request keys keep their existing shape, so an
  instrument id is sufficient to request that instrument, a venue request returns every class the
  source holds, and existing positional calls that pass `params` after `client_id` keep working.
- `BacktestDataConfig.data_type` takes and returns `NautilusDataType`. `Instrument` loads every
  instrument class the catalog holds for the selected identifiers. An unsupported value fails at
  construction with a type error instead of a parse error.
- The PyO3 bindings extract selectors through the typed wrappers already exported by the model
  crate, so the generated stubs name the selector types instead of `typing.Any`. Python data queries
  accept the instrument family, which keeps instrument metadata reachable from Python.
- `NautilusInstrumentType` and `NautilusRecordType` derive `Copy`, and their `inner()` accessors
  become `const`, which removes the clones their call sites carried.

### Cross-backend behavior

| Behavior                   | Parquet          | DuckLake       | Timescale      |
| -------------------------- | ---------------- | -------------- | -------------- |
| Shared selector type       | yes              | yes            | yes            |
| Shared Python extraction   | yes              | yes            | yes            |
| Accepts instrument family  | yes              | yes            | yes            |
| Resolves classes through   | directory prefix | table metadata | table metadata |
| Stores instrument coverage | yes              | yes            | no, by design  |

The directory prefix list is Parquet-only because only Parquet stores one directory per instrument
class. DuckLake resolves instrument tables through its own metadata, so history, recluster,
maintenance scopes, deletion, restore, and coverage repair already cover every instrument table.
Timescale records no coverage rows for instrument definitions, so compacting the aggregate family
correctly removes nothing while the definitions stay readable through the instrument query.

### Instrument display output

Instrument results are one flat table built from the common `Instrument` trait surface. An
`instrument_type` column names each row's class, and accessors that do not apply to a row, such as
`strike_price` on a spot pair, are nullable and emit nulls. Mixed classes therefore round-trip in a
single table. Storage-layout output (`display=False`) remains unsupported for instruments, because
that view is the per-class stored schema.

### Related issues/PRs

None.

### Type of change

- [x] Breaking change (impacts existing behavior).

### Breaking change details

- `ParquetDataCatalog` file operations take a `data_type` selector in place of the `data_cls` and
  `type_name` strings: consolidation (whole and by period), reset and extend file names, coverage
  intervals and first or last timestamps, file listing and filtering, metadata and record batch
  queries, and instrument listing. Callers pass `NautilusDataType.QuoteTick` where they passed
  `"quotes"`.
- `ParquetDataCatalog.query(data_type=...)` takes a `NautilusDataType` in place of a directory-name
  string. Custom data is selected with `NautilusDataType.Custom("TypeName")`, and
  `NautilusDataType.Instrument` returns instrument objects.
- `ParquetDataCatalog.delete_data_range` takes a `NautilusDataType`, which excludes the record and
  instrument families.
- `BacktestDataConfig(data_type=...)` takes a `NautilusDataType` value, and the `data_type` property
  returns one.

### Documentation

- `docs/concepts/data/catalog.md` uses selector values for the catalog operations and for
  `BacktestDataConfig`, and its snippets import `NautilusDataType`. It states that the instrument
  data type reads every class, that `NautilusInstrumentType` reads one, and that instrument results
  are one table with an `instrument_type` column, and the generic `catalog.query(...)` example passes
  `NautilusDataType.QuoteTick`.
- `docs/concepts/backtesting/apis-and-runs.md`, `docs/concepts/options.md`, the getting started and
  how-to notebooks, the order book tutorials, and the backtest examples pass enum values.

### Release notes

`RELEASES.md` gains three breaking change entries for the catalog operations, the deletion
signature, and the backtest config.

### Testing

- [x] Affected code paths are covered by the test suite.
- Converted tests: the Parquet and catalog integration tests pass selectors, and the migration test
  cases pass typed source and target families.
- Removed test: `test_consolidate_data_by_period_invalid_type` asserted a runtime rejection of an
  unknown directory name, which the typed signature makes a compile error.
- Added tests: `CatalogDataType` conversion from every selector family, prefix
  fan-out for the aggregate instrument family against one directory for every
  other selector, aggregate coverage spanning two
  instrument classes, aggregate file operations and consolidation across classes, the
  `extend_file_name` aggregate error, and the Python contract for
  `BacktestDataConfig.data_type` accepting only `NautilusDataType`.
- Added tests for the review findings: aggregate file-name reset, aggregate
  instrument listing reporting a shared identifier once, period consolidation
  rejecting aggregate, class, and record selectors with the documented message,
  aggregate file-name extension reaching every class that stores the identifier,
  and metadata queries honouring the instrument class filter.
- Added Python tests: the instrument data type returns every class in one
  display table with the class column set per row and class-only columns null,
  a `BacktestNode` run with one instrument config loads a currency pair and an
  equity from two venues, and the actor request probe calls
  `request_instrument` positionally so a dictionary in the fifth position stays
  `params`.

Local runs on the rebased commit:

| Check                                 | Result                  |
| ------------------------------------- | ----------------------- |
| `cargo nextest`, four core crates     | 3,810 passed, 7 skipped |
| `pytest`, unit and persistence suites | 1,243 passed, 2 xfailed |
| `make pre-commit`                     | every hook passes       |


faysou:
Hi @cjdsellers,

Thanks for the review. I agree the class-first API leaked the storage layout, so I've reshaped the PR around your points:

- `NautilusDataType.Instrument` is accepted everywhere a catalog operation takes a selector and fans out over every instrument class inside the backend. `CatalogDataType::from_data_type` is gone, so `From` and `.into()` are the only construction path and every selector value names a real target. The selector enum is now `CatalogDataType`, and the trait that previously held that name is `HasCatalogDataType`.
- Every family-scoped catalog method takes its selector as `data_type`, the keyword the typed query methods already use, so there is one keyword across the catalog surface.
- `ParquetDataCatalog.query` takes a `NautilusDataType` rather than a directory-name string, with one explicit arm per variant. `Instrument` returns instrument objects; records stay on the record methods because `query` returns `Data` values.
- The `instrument_type` parameter is removed from `request_instrument` and `request_instruments`, so those signatures are back to their previous shape and positional calls keep working.
- `BacktestDataConfig.data_type` still takes a `NautilusDataType`, and `Instrument` now loads every class the catalog holds for the selected identifiers.
- Instrument metadata is reachable from Python again, and `query_catalog` routes the instrument family to the instrument reader. Mixed classes come back in one display table with an `instrument_type` column.
- The optional class filter stays on the catalog instrument reads and accepts a `NautilusInstrumentType` only.

Validating that surfaced a few more things I've fixed in the same commit: `consolidate_data_by_period` accepted instrument and record selectors it could only fail on with a directory name, so it now rejects them with a message naming the selector; aggregate `extend_file_name` extends every class that stores the identifier rather than stopping at the first; `query_metadata` honors the instrument class filter like the other reads; `list_instruments` could repeat an identifier stored under two classes; and there was no test running a `BacktestNode` from an instrument config spanning two classes, so there is one now.

The branch is rebased on `develop` and the description is updated.

cjdsellers (approved):
Hey @faysou,

Thanks for the rework around the review points, the aggregate instrument family reads much better now.

I've had another look and pushed 9f5dbe7 to the branch: it fixes the `event_store` replay build against the new `query_files` signature and renames the remaining `catalog_type` locals and parameters to `data_type` so the whole surface uses one selector name. CI is still running on it.

One small thing I'll follow up with separately after merge: `list_parquet_files` and `convert_stream_to_data` still take directory-name strings. They have no in-repo callers passing new values and their behavior is unchanged, so I left them alone rather than widening this PR.

PR #5034, opened 2026-09-19
Resolve Kraken spot reports spelled with the pair altname

### LINKED ISSUE #5033: [Kraken] Spot order and fill reports spelled with the pair altname are silently dropped
# Bug Report

## Confirmation

## Expected behavior

An open order or fill on a Kraken spot pair whose `AssetPairs` key differs from its altname should
appear in the order and fill reports, so reconciliation can adopt it.

## Actual behavior

It is dropped from an otherwise successful report, with no error and no log.

`AssetPairs` keys BTC/EUR as `XXBTZEUR`, and the parser stores that key as the instrument
`raw_symbol` (`Symbol::new(pair_name)`). `OpenOrders` reports the same pair as `XBTEUR`: Kraken's
documentation shows `descr.pair` in altname form. `get_instrument_by_raw_symbol` compares only
against `raw_symbol`, so the lookup misses and the report is skipped:

```rust
if let Some(instrument) = self.get_instrument_by_raw_symbol(order.descr.pair.as_str()) {
    // ...push report
}
// no else: the order is omitted from Ok(all_reports)
```

Reconciliation never adopts the order, so neither a strategy's cancel sweep nor an operator restart
can reach it, and the log is clean.

Three sites share this shape in `crates/adapters/kraken/src/http/spot/client.rs`: open orders,
closed orders, and trade history (fill reports).

The same lookup in `OpenPositions` is not silent:

```rust
let instrument = self
    .get_instrument_by_raw_symbol(pos.pair.as_str())
    .ok_or_else(|| {
        anyhow::anyhow!("OpenPositions: instrument not in cache for pair {}", pos.pair)
    })?;
```

So an altname-spelled position aborts startup loudly, while an altname-spelled order fails open.

## Steps to reproduce

1. Restart a Kraken spot node with reconciliation enabled while a resting order exists on a pair
   whose `AssetPairs` key differs from its altname (any legacy `X`/`Z`-prefixed pair, such as
   BTC/EUR).
2. Leave a second resting order on a pair whose key equals its altname, as a control.
3. Read the reconciliation log.

Observed: the control order is adopted (`external=1, open=1`). The legacy-named order appears
nowhere in the boot log under any spelling, and is still resting at the venue afterwards. The venue
read succeeded, the instrument cache was populated, the instrument load did not fail, and mass
status returned normally.

Reduced form: two identical open-order payloads differing only in `descr.pair` give `external=1`
and `external=0`.

Any legacy-named spot pair is affected.

## Code snippets or logs

The adapter's own fixtures cannot catch this. Every order and fill fixture uses `XBTUSDT` or
`ETHUSDT`, where the key and the altname are identical; legacy names appear only in the position
fixtures, and there only via an instrument injected directly into the cache rather than loaded
through `request_instruments`.

## Specifications

- Reproduced against: `develop` @ `9bafb63e7d`
- Also observed on the `2.0.0rc6.dev20260918` development wheel

zhaow-de:
  and, if I used AI,
  this draft
  limitation below

## Summary

Kraken spells a spot pair two ways: `AssetPairs` keys BTC/EUR as `XXBTZEUR`, which becomes the
instrument `raw_symbol`, while `OpenOrders` and `TradesHistory` report the altname `XBTEUR`.
`get_instrument_by_raw_symbol` compared only the key, so on a legacy-named pair the lookup missed
and the report was dropped inside a successful, shorter return.

This records each `altname` against its `AssetPairs` key when instruments are requested, and
resolves by key first and altname second so a key is never shadowed by another pair's altname.

The altname cannot be derived from a cached instrument, because `normalize_spot_symbol` rewrites
Kraken's currency codes (`XBT` to `BTC`). Changing `raw_symbol` to the altname instead would just
move the defect, since `OpenPositions` returns the key while the order and trade reads return the
altname; both spellings have to resolve.

**Behaviour change on unresolved reports.** These follow the scope table in the adapter guide, and
split by record type rather than being uniform across the three reads.

An in-scope open order returns an error, because a dropped open order report is indistinguishable
from an order the venue never had, which would lead the engine to resolve a live order as missing.

An in-scope closed order or fill is logged as a warning naming the instrument and skipped, keeping
the records that do resolve. Historical reads reach past the loaded instrument set routinely, since
expiries and delistings retire instruments that earlier records still reference, so failing one row
would withhold every other record the query returned and could block startup or a single-order
probe over unrelated history.

## Related issues/PRs

Closes #5033

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

Not an API change. Two behaviour changes: reports on legacy-named pairs are now returned rather than
omitted, and an unresolved open order now fails the read rather than being omitted from it. An
unresolved closed order or fill is logged and skipped, leaving the resolvable records intact.

## Documentation

  generated output

`docs/integrations/kraken.md` gains a "Pair spelling" entry under Spot reconciliation covering both
spellings and the new failure behaviour. No binding signature or wrapped doc changed, and
`generate_docstrings.py --dry-run` reports no drift.

## Testing

The existing fixtures could not catch this: every order and fill fixture uses `XBTUSDT` or
`ETHUSDT`, where key and altname are identical. The harness now serves a legacy-named pair
(`XXBTZEUR` / `XBTEUR`) on request, and five tests cover it:

- `test_spot_order_reports_resolve_altname_spelled_pair`
- `test_spot_fill_reports_resolve_altname_spelled_pair`
- `test_spot_order_reports_error_on_unresolvable_pair`
- `test_spot_closed_order_reports_preserve_records_around_unresolvable_historical_pair`
- `test_spot_fill_reports_preserve_records_around_unresolvable_historical_pair`

The first three fail against the previous behaviour and pass with the fix.

The two historical tests pair an unresolved row with a valid one, and each asserts a control first:
the identical payload with every pair resolvable returns both rows. Without that control the
assertion would also hold if the row had been dropped for an unrelated reason, such as failing to
parse, rather than by the resolution path the test claims to cover.

Two test-side changes worth calling out:

- `test_mass_status_captures_collection_start` built a spot client without connecting, so no
  instruments were cached and its orders were being silently dropped. It now connects first, as the
  production path does. Its assertions are unchanged.
- The execution-client harness served an `AssetPairs` listing without `ETHUSDT` while its order and
  trade fixtures referenced it. The silent drop hid that; the listing now carries it.

`cargo test -p nautilus-kraken`: 780 passed, 0 failed. `cargo clippy --all-targets --all-features
-- -D warnings`: clean.


cjdsellers:
Hi @zhaow-de,

Thanks for the PR. The alias lookup addresses the startup omission, and failing an unresolved open-order read makes sense.

The historical reads need different handling before this merges. A closed order or trade can reference an instrument absent from the current listing. Returning an error for that row now aborts the whole read, which can prevent startup or block a single-order probe because of unrelated history. The adapter guide requires a warning for missing historical instruments and incomplete coverage when history is bounded.

Could you preserve usable historical records under that contract and add regression coverage for an unresolved historical pair alongside valid records?

zhaow-de:
Thanks @cjdsellers, for the fair catch. Historical reads now follow the contract.

Closed orders and trade history log a warning naming the instrument and skip the row, keeping the records that do resolve. Open orders still return an error, since a dropped open order report can't be told apart from an order the venue never had.

Both regression tests pair an unresolved row with a valid one, and each asserts a control first: the identical payload with every pair resolvable returns both rows. That way the test proves the row is well-formed and is skipped only because its pair can't be resolved, rather than passing because the row failed to parse for some other reason.

- `test_spot_closed_order_reports_preserve_records_around_unresolvable_historical_pair`
- `test_spot_fill_reports_preserve_records_around_unresolvable_historical_pair`

The docs entry is updated to match, so it now states the open-order and historical cases separately.

One thing I left out, and I'd rather ask than assume: the contract also marks the report set incomplete when history is bounded, via `set_report_window`. The Kraken spot client never calls it today, so it declares no `lookback_start` even though `generate_mass_status` computes a `start` from `lookback_mins`. That looked like a pre-existing gap rather than part of this fix, so I didn't wire it here. Happy to add it in this PR or a separate one, whichever you prefer.

780 Kraken tests pass, clippy `-D warnings` clean.

cjdsellers (approved):
@zhaow-de, thanks for the follow-up and regression tests. I've had another look, and this looks good from my side. I'd keep `set_report_window` for a separate PR, since that gap predates this fix and needs its own coverage for bounded history.

Could you also update the PR description to match the revised historical-read behavior? It still says all three reads fail on an unresolved pair.

zhaow-de:
Thanks for the approval and the merge.

I've updated the PR description. The behaviour section now states the split rather than a uniform rule: an unresolved open order returns an error, while an unresolved closed order or fill is logged as a warning and skipped so the resolvable records survive. The breaking-change note and the test list are updated to match, including the two historical regression tests and the control each one asserts.

Agreed on keeping `set_report_window` out of this one. I'll raise it separately with its own coverage for bounded history.

PR #5035, opened 2026-09-20
Resolve option underlying across venues at expiry

### LINKED ISSUE #4879: Option expiry settlement cannot resolve an underlying listed on a different venue (index options: SENSEX.BFO options vs SENSEX.BSE index)
### Expected Behavior

At option expiration, `OrderMatchingEngine` settles expired option positions by looking up the underlying's price and computing intrinsic value, regardless of which venue the underlying instrument is registered under.

### Actual Behavior

`process_option_expiry` fabricates the underlying's `InstrumentId` from the **option's own venue** (`crates/execution/src/matching_engine/settlement.rs:70` at `develop` @ d1012b9e20):

```rust
let underlying_id = InstrumentId::from(format!("{underlying}.{}", self.venue).as_str());
```

This assumes the underlying is listed on the same venue as the option. That is false wherever an exchange separates its derivatives segment from its cash/index segment — standard for Indian index derivatives:

- SENSEX index options trade on venue `BFO`, while the SENSEX index instrument and its `IndexPriceUpdate` stream are registered as `SENSEX.BSE`
- NIFTY index options trade on venue `NFO`, while the index is `NIFTY 50.NSE`

The engine looks up `SENSEX.BFO`, which was never registered, so every expiry check takes the `missing-underlying-instrument` retry path until settlement latches terminally ("Option settlement failed terminally"). The expired positions then remain **open forever** — they cannot be closed by orders (no market exists for a delisted contract) and cannot be purged (`purge_position` correctly refuses open positions).

Futures on the same node settle fine, because futures settlement never performs this lookup.

### Real-world impact observed

In a paper-trading node (sandbox execution), four expired SENSEX BFO option positions were stuck permanently open after the 2026-08-27 expiry, surviving every restart via cache rehydration. The only remediation available without an engine change was manual deletion of their rows from the cache database.

1. Register an option contract on venue A (e.g. `SENSEX26AUG77000PE.BFO`, `underlying = "SENSEX"`) and its underlying index instrument on venue B (`SENSEX.BSE`), with index prices flowing for `SENSEX.BSE`.
2. Open a position in the option and hold it through `expiration_ns`.
3. Observe repeated `missing-underlying-instrument` retries, then terminal settlement failure; the position never closes.

### Suggested fix (implemented and tested in our fork — happy to submit a PR)

Resolution tries the option's own venue **first** (byte-identical behaviour for same-venue markets such as equity options), then falls back to an exact-symbol match across the cache's other venues. A cross-venue match is accepted **only when unique** — this value prices the whole expiring position, so an ambiguous symbol logs an error naming the candidates and takes the existing retry path rather than guessing (deferring settlement is recoverable; settling against the wrong venue's price is not). The subsequent `index_price()`/`price()` lookup then uses the **resolved** instrument's real ID, since that is what keys the price stores.

Tests: same-venue resolution preferred over a cross-venue decoy; a unique cross-venue underlying settles an expired index option at intrinsic value and closes the position; an ambiguous cross-venue symbol defers (with a single-venue in-test control); a populated cache with no matching symbol still defers. Both behaviour tests fail without the fix.

Note: an exact-symbol fallback intentionally does not bridge symbol-convention mismatches (e.g. option `underlying = "NIFTY"` vs index symbol `"NIFTY 50"`); that seems better addressed at the adapter/instrument-definition layer than by fuzzy matching in settlement.

### Specifications

- OS platform: Linux (aarch64 container) / macOS host
- Python version: 3.12
- `nautilus_trader` version: 2.0.0rc3 dev wheels; bug present on `develop` @ d1012b9e20

AmitKumarDeoghoria:

## Summary

At option expiration, \`process_option_expiry\` fabricated the underlying \`InstrumentId\` from the option's own venue, which fails wherever an exchange separates its derivatives segment from its cash/index segment (SENSEX index options on BFO vs SENSEX.BSE, NIFTY on NFO vs NIFTY 50.NSE). The lookup missed, settlement latched terminally, and expired positions stayed open forever.

Resolution now tries the option's own venue first (byte-identical for equity-style same-venue markets) and, when that misses, accepts a unique exact-symbol match from the rest of the cache. An ambiguous cross-venue symbol logs the candidates and takes the existing retry path rather than guessing — deferring settlement is recoverable, settling against the wrong venue's price is not. The subsequent \`index_price\`/\`price\` lookup uses the resolved instrument id.

## Related issues/PRs

Resolves #4879

## Type of change

- [x] Bug fix (non-breaking)

## Testing

Added four \`test_process_option_expiry_*\` cases covering same-venue preference over a cross-venue decoy, unique cross-venue settlement at intrinsic value, ambiguous cross-venue defer, and populated-cache-with-no-match defer. Full \`option\`/\`settlement\` test suite (54 tests) passes; \`make pre-commit\` on the changed files is clean.


cjdsellers:
Hi @AmitKumarDeoghoria,

Thanks for the PR. The approach looks right to me: keeping the same-venue lookup first preserves existing behavior, and deferring on an ambiguous match rather than guessing is the safe tradeoff for settlement.

One small suggestion, if you're able: would you mind adding a log line when the cross-venue fallback resolves, naming the instrument it settled on? That path prices the whole position off a heuristic match, so having it visible in logs would make any future wrong-venue settlement much easier to diagnose. This looks good from my side otherwise.

Longer term, I'm considering letting `underlying` optionally carry a full instrument id, so definitions that know the venue can pin the resolution explicitly while the fallback here covers everything else. @faysou, I don't plan to change the field itself to `InstrumentId`: an underlying isn't always an instrument, and cash settlement, rates, and fixings have no venue and never will be in the cache. Nothing to do on that here, just noting the direction.

AmitKumarDeoghoria:
Thanks for the review, @cjdsellers.

Pushed the log line: info-level at the moment the cross-venue fallback resolves to a unique underlying, naming the option and the instrument it settled on. It only fires on the fallback path — the same-venue hit stays silent — so wrong-venue settlement should stand out in logs without noise on the common case.

All nine option-expiry integration tests still pass, clippy on the execution crate is clean, and the branch is rebased on develop.

One thing I'd love your perspective on, purely to understand the product better: where do cross-venue underlyings most commonly show up in real Nautilus workflows? I'm trying to picture whether the dominant case is listed indexes (SPX, VIX, NDX) that data vendors publish under an exchange-of-record or "global" venue while the options themselves list on CBOE/CME etc., or whether it's more often users deliberately seeding an underlying under a different venue for a specific strategy shape. It feels relevant to where the eventual optional-InstrumentId pin would want to live — at the definition/provider layer if it's mostly a data-layout artefact, or exposed at the user API if it's strategy-driven.

cjdsellers (approved):

PR #5028, opened 2026-09-19
Add submission recovery types and configuration

silarin:
# Pull Request

- [x] A maintainer agreed on the problem and approach in an issue
- [x] I have read and followed [CONTRIBUTING.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md) and [AI_POLICY.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/AI_POLICY.md)
- [x] I understand and can explain every submitted change and all information in this PR description
- [x] This change is complete, locally validated, and ready for review
- [x] I ran `make format`, then ran `make pre-commit` locally and confirmed it passed
- [x] I ran all relevant tests locally
- [x] I have not modified `RELEASES.md`

## Summary

Adds `SubmissionRecoveryPolicy` and the `SubmissionRecoveryExhausted` payload in the first incremental PR requested on #5013, limited to types and configuration. `ResolveLocally` remains the default; `RetainUnresolved` can be stored in Rust and Python configuration but has no runtime effect in this slice, and the diagnostic is not emitted yet.

The Python argument is appended to preserve existing positional calls. Omitted serialized fields retain the default; both enum values use snake-case serialization and are forwarded to `ExecutionManagerConfig`.

## Related issues/PRs

Related to #4152 and #5013. This is the first incremental PR in the [requested split](https://github.com/nautechsystems/nautilus_trader/pull/5013#issuecomment-5737810300). Registry/query handling, reconciliation guards, shutdown semantics, and OKX coverage remain for later PRs.

## Type of change

- [x] New feature
- [x] Breaking change (Rust configuration struct literals)

## Breaking change details

Rust callers using exhaustive literals for `LiveExecutionEngineConfig` or `ExecutionManagerConfig` must include the new field. Existing callers using the builder or `..Default::default()` remain compatible. Python positional argument order and existing runtime behavior are preserved.

## Documentation

- [x] Inline API docs explicitly identify the inactive policy and diagnostic
- [x] Generated Python stubs have been regenerated with the source change

## Testing

- [x] Affected code paths are already covered by the test suite
- [x] I added/updated tests to cover new or changed logic

Focused coverage checks omitted defaults, typed selection, serialization round trips, invalid inputs, manager forwarding, nested Python config, and Python positional compatibility. The existing timeout and missing-order regressions run with both policy values and still assert the current rejection behavior.

[...]

PR #5013, opened 2026-09-17
Add opt-in retention for unresolved submissions

silarin:
## Summary

A submission can exhaust reconciliation queries while its venue outcome remains unknown. This adds `LiveExecutionEngineConfig.submitted_order_exhaustion_policy`, preserving `ResolveLocally` as the default and offering `RetainUnresolved` as an explicit opt-in. The policy covers both submissions with no acknowledgement and explicit ambiguous responses, including OKX `50004` and `51149`.

With retention enabled:

- Exhaustion stops automatic per-order queries and emits one typed `SubmissionRecoveryExhausted` diagnostic, distinct from a terminal order outcome. The native order and its identity remain available for later authoritative evidence; no automatic resubmission occurs.
- Inflight checks and full-history missing-order reconciliation cannot synthesize rejection or cancellation solely from exhausted recovery. Pending snapshots do not infer acceptance or undo applied fills. Matching authoritative reports and events still resolve normally.
- Cancel/modify requests made before acknowledgement retain submission uncertainty and their separate recovery budgets. Identity, replacement aliases, duplicate fills and recovery-scoped chronology remain validated.
- Evidence received while native managed cleanup remains active can still enter cleanup during the bounded shutdown window. If submission recovery cannot finish, shutdown returns an explicit incomplete-recovery error identifying unresolved orders.

Recovery counters and diagnostics are in memory. This does not provide crash-durable recovery, extend the shutdown deadline, or establish a venue outcome after disconnection. After standalone authoritative acceptance resolves submission uncertainty, ordinary acknowledged-order reconciliation resumes.

## Related issues/PRs

Related to #4152. Implements the submitted-order opt-in follow-up requested in [the maintainer’s reply](https://github.com/nautechsystems/nautilus_trader/issues/4152#issuecomment-5712904007); it does not claim to close the broader pending cancel/update policy discussion.

## Type of change

- [x] New opt-in feature with the existing runtime default preserved
- [x] Rust source API change, described below
- [x] Documentation update

## Breaking change details

Exhaustive Rust struct literals for `LiveExecutionEngineConfig` and `ExecutionManagerConfig` must supply `submitted_order_exhaustion_policy`; those for `ExecutionEngineConfig` must supply `preserve_unresolved_submissions`. Using their defaults preserves the previous policy. Existing Python constructor arguments retain their positions, and omitted serialized settings retain the existing default.

## Documentation

- [x] Updated execution policy and live configuration documentation
[...]

PR #5036, opened 2026-09-20
Add submission registry and exhaustion diagnostics

silarin:
# Pull Request

- [x] A maintainer agreed on the problem and approach in an issue.
- [x] I have read and followed [CONTRIBUTING.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/CONTRIBUTING.md) and [AI_POLICY.md](https://github.com/nautechsystems/nautilus_trader/blob/develop/AI_POLICY.md).
- [x] I understand and can explain every submitted change and all information in this PR description.
- [x] This change is complete, locally validated, and ready for review.
- [x] I ran `make format`, then `make pre-commit` locally and confirmed it passed.
- [x] I ran all relevant tests locally.
- [x] I have not modified `RELEASES.md`.

## Summary

Following #5028, this implements the submission registry, bounded queries, and `SubmissionRecoveryExhausted` diagnostic from the second increment of the agreed #4152 plan. With `RetainUnresolved` selected, the native manager preserves submission identity and recovery budgets, and `LiveNode` publishes one diagnostic per exhausted submission after processing native reconciliation events. **`ResolveLocally` remains the default, and both policies still use existing local terminal resolution in this increment.**

The registry captures single orders and order-list children before dispatch or cache insertion. Duplicate registration and interleaved recovery checks preserve the original unacknowledged submission budget. A matching Submitted report carrying venue identity clears existing recovery tracking, including pending-command tracking, even if reconciliation emits no order event. Applied venue-bearing OrderUpdated events also confirm submission. Duplicate registration cannot restart confirmed submission recovery; later cancel/modify commands receive their own budget at dispatch. Default report cleanup retains the original terminal-report and explicit-client-ID predicate, including the fill-settling window.

## Related issues/PRs

- Related to #4152 and the [approved submitted-order follow-up](https://github.com/nautechsystems/nautilus_trader/issues/4152#issuecomment-5712904007).
- Follows #5028, which introduced `SubmissionRecoveryPolicy`, configuration, and diagnostic types.
- Implements step 2 of the [incremental plan in #5013](https://github.com/nautechsystems/nautilus_trader/pull/5013#issuecomment-5737810300).

## Type of change

- [x] Documentation update.

## Documentation

- [x] Documentation changes follow the style guide.
- [x] For wrapped Rust doc changes, I regenerated Python stubs with `make py-stubs` and checked for generated drift.

The configuration and reconciliation documentation state the interim behavior explicitly. Existing query limits and coverage gates remain in place: an inflight limit of `N` permits `N - 1` intermediate queries before local resolution.

Protection against synthetic terminalization after unresolved query exhaustion, delayed evidence after retention, shutdown/incomplete-recovery semantics, and OKX-specific coverage remain subsequent increments. Issue #4152 remains open.

## Testing

- [x] Affected code paths are already covered by the test suite.
- [x] I added/updated tests to cover new or changed logic.

- Rust: 2,953 standard-precision tests passed across `nautilus-live` and `nautilus-common`, with seven existing skips.
- Baseline comparison: the same 38 new native-node cases passed on the unchanged production base (`4c7f9d1`). On the published PR source (`809ff2f`) they reproduced seven opt-in venue-evidence failures and sixteen cleanup failures across both policies; all 38 pass with this correction.
- Fifteen further native-node cases cover duplicate registration after confirmation, later command budgets with and without a first partial fill, mismatched identity, unapplied venue updates, and local updates without venue identity. The existing registry, query-budget, callback and diagnostic regressions remain in the full suite.
- Python: 76 configuration tests passed against the rebuilt native extension.
- `make format`, `make pre-commit` (including Clippy and documentation checks), and `make build-debug` passed. Python stubs were regenerated with no generated-file drift.
- Fresh independent review of the full patch and corrections found no unresolved findings and included execution of 108 focused cases. Final test-only formatting and import cleanup was separately checked by the same reviewer for unchanged behavior.

Validation was offline. No adapter request construction or credentialed OKX qualification is included.


cjdsellers:
Hey @silarin,

Thank you for the PR. The step 2 slice reads clearly, and the interim framing (both policies still resolve locally, retention protection comes later) is explicit in the docs.

One blocking issue before this can merge. Under `RetainUnresolved`, the new `clear_recon_tracking` guard blocks the clears that `develop` performs when positive venue evidence arrives, because `submission_is_unacknowledged` does not treat that evidence as acknowledgement. A `Submitted` status report carrying a `venue_order_id`, or an applied `OrderUpdated`, leaves the submission pending, so the preserved in-flight ladder keeps burning and synthesizes `Rejected(INFLIGHT_TIMEOUT)` or `Canceled` for an order the venue just confirmed alive. On `develop` the same evidence drops tracking and no terminal event is produced, so this is new terminalization rather than the existing local resolution the description covers. The node registry test encodes the outcome (`Canceled` after the `Submitted` report under retention versus untouched pending by default). Since this runs against the direction agreed in #4152, I think it needs a fix in this increment rather than waiting for the step 3 guards: venue-alive evidence should retire or re-budget the submission instead of feeding the timeout.

One smaller point. `order_report_client_order_id` now returns IDs for every order report and for `Fill` reports, so post-dispatch cleanup runs under `ResolveLocally` where `develop` ran nothing (for example a fill that closes its order now clears tracking and drops the settling activity mark). Either scope the clear to the previous predicate or document and pin the new default behavior with a test, since the default-unchanged claim does not hold as written.

silarin:
Hi @cjdsellers,

Thank you for identifying both regressions. Both are addressed in f3d6dbb07b.

- A matching `Submitted` report with venue identity now retires the original submission recovery, including when reconciliation emits no order event. Applied venue-bearing `OrderUpdated` events also acknowledge submission. Duplicate registration cannot restart confirmed submission recovery, and later cancel/modify commands retain their own recovery budgets.
- Post-dispatch cleanup now uses the previous predicate, preserving the default behavior and the settling activity mark after standalone fill reports.

The earlier node test incorrectly expected cancellation after a positive `Submitted` report. I corrected that expectation and ran the same 38 native-node regression cases against the unchanged production base, the previous PR revision, and the correction. All 38 pass on the baseline and correction; 23 fail on the previous revision. Fifteen additional cases cover duplicate registration, later command budgets, and invalid or unapplied evidence.

Validation passed: 2,953 affected Rust tests, 76 Python configuration tests against the rebuilt extension, formatting, and pre-commit checks. A fresh independent review found no unresolved findings. The PR testing summary is updated.

This remains the second increment: submission registry, bounded queries, and exhaustion diagnostics. Both policies still use existing local resolution for unresolved outcomes; retention guards and shutdown handling remain subsequent increments.

cjdsellers (approved):
@silarin, I've had another look at the latest changes, and both fixes check out, with the new regression cases locking them in. The baseline comparison across the unchanged base, the previous revision, and the correction also made this easy to verify, thanks for putting that together.

This looks good to merge.

**Two small things came out of my pass, neither blocking:** the doc on `register_inflight` still describes a plain registration while the method now has the retention early-exits, and the `filtered_client_order_ids` checks inside `clear_recon_tracking` can't trigger, since `track_submission` already skips those IDs and the config is fixed at construction. Would you be able to tidy those up here if you get a chance, or would you rather they ride along with a later slice?

silarin:
Hi @cjdsellers,

Thanks for looking into it.
Sure, I'll fix it in this PR when I have the chance to work on it later.

silarin:
Hi @cjdsellers,

Both cleanup items are addressed in d8b95adb1d:

- Updated `register_inflight` documentation to describe the retention early returns and separate pending-command recovery budgets.
- Removed the redundant `filtered_client_order_ids` checks from `clear_recon_tracking`, relying on the existing exclusion in `track_submission`.

All 131 focused Rust regressions passed, along with formatting and pre-commit checks, including Clippy and Rustdoc. A fresh independent review found no actionable findings.

Thanks for the review.

cjdsellers:
@silarin, the doc and filter cleanup look right.

One issue still blocks this. Under `RetainUnresolved`, a later cancel or modify can miss its recovery budget while the order is still `Submitted` but already confirmed:

- A matching venue `Submitted` report, or an applied `OrderUpdated` with a venue order ID, confirms the original submission and can leave the order `Submitted`.
- The cancel or modify is registered at dispatch, before the order moves to `PendingCancel` or `PendingUpdate`.
- `register_inflight` returns immediately in that state, so the new command never gets an in-flight check.
- The inflight checker then doesn't query it, and doesn't synthesize `Canceled` if the venue stays silent.
- `ResolveLocally` still does both.

The later-command tests register after the pending event is already applied, so they miss this order.

The reconciliation docs already say a later cancel or modify uses its own recovery budget. That needs to hold for a command dispatched while the order is still `Submitted`, without restarting the original submission timeout.

silarin:
Hi @cjdsellers,

Thanks for identifying the dispatch-before-pending case. I’ve prepared and validated a local correction that separates actual cancel/modify dispatch from duplicate submission tracking. Each command gets its own recovery budget at dispatch, even while the cached order remains `Submitted`, without restarting the original submission timeout.

Submission confirmation and delayed evidence preserve an outstanding command’s budget. Applied native completion retires that budget, preventing a callback-queued command from inheriting an earlier timeout.

The native-node regressions cover both policies, individual/batch dispatch, pending-event ordering, repeated reports, partial/full fills, and completion callbacks. Validation passed: 3,385 Rust tests, including 460 focused cases, with seven unchanged skips; the full pre-commit suite also passed. Final independent source review found no actionable findings.

`ResolveLocally` remains unchanged. This stays within the second increment: both policies retain existing local timeout resolution, with full retention guards and shutdown handling deferred to the subsequent increments.

PR #5037, opened 2026-09-20
Fix AroonOscillator MAX_PERIOD window capacity

### LINKED ISSUE #4995: AroonOscillator accepts MAX_PERIOD but cannot retain its required period + 1 window

wbizmo:
# Pull Request

  and, if I used AI,
  this draft
  limitation below

## Summary

Keep `MAX_PERIOD = 1024` valid by giving the Aroon high and low buffers capacity for the required
`period + 1` observations. Add Rust and Python regression coverage for both oldest-high and
oldest-low boundary cases, including exact initialization and rollover after the next observation.

## Related issues/PRs

Closes #4995

Related to #5022, which implemented the same capacity direction before being closed without merge.
This PR starts from current `develop` and preserves that prior attempt in the contribution history.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Documentation update
- [ ] Maintenance / chore

n/a

## Documentation

## Testing

Regression coverage includes:

- initialization only after exactly `MAX_PERIOD + 1` observations
- retention of an oldest unique highest high at the boundary
- retention of an oldest unique lowest low at the boundary
- correct Aroon Up, Aroon Down, and oscillator values for both cases
- eviction of the oldest extreme on the following observation while the window remains bounded

The published `v2.0.0rc5` failure was reproduced in #4995 before this implementation.

Validation was run against the exact PR commit in an isolated contributor-controlled GitHub Actions
environment using the repository's pinned setup. `make format` completed with no diff,
`make pre-commit` passed, the `nautilus-indicators` Rust test suite passed, and the full Python
test suite passed. An earlier full-workspace Rust run also exercised roughly 33.6k tests before
reaching unrelated Redis integration failures because that temporary runner did not provide the
required Redis service; the affected indicator crate was then validated independently and passed.


cjdsellers (approved):
Hi @wbizmo,

Thank you for the PR, and for carrying over the mirrored low-side coverage from the #5022 discussion so both deques stay covered.

I've reviewed the change and it looks good from my side. The `MAX_CAPACITY` fix restores the `period + 1` window at the boundary, and the exact-initialization assertions lock in the intended behavior nicely.

wbizmo:
Thanks for the review — really appreciate it!

Issue #5032, opened 2026-09-19
[Interactive Brokers] Account values outside the fixed reqAccountSummary tag list are unreachable


jweltmer:
The IB adapter surfaces account state through a single subscription with a hardcoded tag list, so every value IB publishes on `updateAccountValue` and nowhere else is unreachable from Python. The mechanism needed to carry them already exists on both sides of the gap — this is a request to connect them.

Nothing here is defective; filing as a feature request rather than a bug.

## What is verified, and where

All claims below are against `v2.0.0rc5` and its pinned `ibapi = "=3.3.0"`, not `develop`.

| Claim | Evidence |
|---|---|
| The adapter subscribes account state in exactly one place | `crates/adapters/interactive_brokers/src/execution/account.rs:57`, `subscribe_account_summary` |
| It requests a **hardcoded** ten-tag list, not configurable | same file, `:63-74` — a Rust `&[...]` literal; no config input reaches it |
| It already carries raw venue values to the account's `info` | same file, `:90` `let mut info = Params::new()`, `:103` `info.insert(summary.tag, …)`, returned at `:156` |
| That `info` reaches Python | `AccountState.info -> dict` |
| `reqAccountUpdates` is issued nowhere | no call site in the account path |
| The pinned dependency already exposes it | `ibapi = "=3.3.0"` (workspace `Cargo.toml:344`); `ibapi::client::Client::account_updates` and `account_updates_multi` |
| …and it yields arbitrary key/value | `ibapi::accounts::AccountUpdate::AccountValue` carries `pub account`, `pub key`, `pub value`, `pub currency` |

The adapter's own comment at `account.rs:100-102` states the intent this would complete:

> Record the raw summary tag so the account state carries the venue-reported values (for example TotalCashValue) that do not map to the typed balances and margins.

## The motivating case, and what we measured

IB restricts an account to orders that do not increase projected deliverable exposure when its post-expiration margin projection is short, and publishes that projection as `PostExpirationExcess` / `PostExpirationMargin`. We induced the condition on a paper account and read it in both directions four times:

```
2026-09-15 12:15   CLEAR     PostExpirationExcess  +854,084
2026-09-15 16:01   DEFICIT                         -326,492
2026-09-16 09:17   DEFICIT                         -294,758
2026-09-16 10:54   CLEAR                           +837,366
2026-09-18 14:59   DEFICIT                         -389,984
2026-09-18 15:37   CLEAR                           +806,745
```

[...]

Issue #5007, opened 2026-09-16
[Interactive Brokers] `generate_fill_reports` sends the prefixed `AccountId` as the execution filter's account code — IB rejects it with error 321


Somak-Shiv:
# Bug Report

## Confirmation

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I've tested this issue using a recent pre-release or development wheel (`2.0.0rcN`, `dev` develop, or `a` nightly) and can still reproduce it.

## Expected behavior

`generate_fill_reports` requests executions for the configured account, filtering on the raw IB account code (e.g. `U1234567`), as every other account-scoped request in the adapter does, and returns that account's historical fill reports during startup reconciliation.

## Actual behavior

The execution filter is sent with the Nautilus `AccountId` string, including the issuer prefix (`IB-U1234567`). TWS / IB Gateway rejects the request:

```
[321] Error validating request.-'W' : cause - Invalid account code IB-U1234567.
```

The request never completes, so `generate_mass_status` gets 0 fill reports. Startup reconciliation still succeeds (positions come from the position reports), so this surfaces only as a warning plus missing historical fills.

In `crates/adapters/interactive_brokers/src/execution/core.rs` (`v2.0.0rc5`, `1b0a49d`, unchanged on `develop`):

```rust
async fn generate_fill_reports(&self, cmd: GenerateFillReports) -> anyhow::Result<Vec<FillReport>> {
    ...
    // Get account code from account ID
    let account_code = self.core.account_id.to_string();   // "IB-U1234567"
    ...
    let filter = ExecutionFilter {
        client_id: None,
        account_code,
        ...
```

Every other account-scoped call site strips the prefix with `raw_ib_account_code` (`execution/account.rs:61, 236, 327, 388`; `execution/core.rs:888, 1178, 1427, 2219`). The 1.231.0 Python adapter also used the raw code (`execution_filter.acctCode = self.account_id.get_id()`), so this looks like a regression from the Rust port. The same line is present in `v2.0.0rc3` and `v2.0.0rc4`.

Suggested fix:

```diff
-        // Get account code from account ID
-        let account_code = self.core.account_id.to_string();
+        let account_code = raw_ib_account_code(&self.core.account_id);
```

## Steps to reproduce

[...]

Issue #4987, opened 2026-09-12
[Interactive Brokers] Cannot build an Account when the account summary carries no `Currency` tag


Dangerboi:
# Bug Report

## Confirmation

**Before opening a bug report, please confirm:**

- [x] I've re-read the relevant sections of the documentation.
- [x] I've searched existing issues and discussions to avoid duplicates.
- [x] I've reviewed or skimmed the source code (or examples) to confirm the behavior is not by design.
- [x] I've tested this issue using a recent pre-release or development wheel (`2.0.0rcN`,
      `dev` develop, or `a` nightly) and can still reproduce it.

### Summary

The IB exec client connects successfully, but no `Account` is ever created:

[ERROR] ...nautilus_live::execution::manager: Failed to parse account summary: Account summary currency was empty

Because the cache holds no account, `portfolio.equity(venue=...)` returns empty, and a strategy sizing against equity fails on every signal:

```text
Cannot resolve portfolio equity without exactly one account
```

The adapter appears to require a dedicated currency tag in the account summary. IBKR does not send one for this account.

### Expected behavior

 When the IB account summary contains currency on value rows, Nautilus should use that currency to construct the account if dedicated Currency/BaseCurrency rows are absent.

### Actual behavior

 The account is rejected because the parser requires a dedicated currency tag:

 ```text
Failed to parse account summary: Account summary currency was empty
 ```

 The account summary contains <BASE_CCY> on NetLiquidation and TotalCashValue, but Nautilus does not use those values. Consequently, no account is created and portfolio equity cannot be resolved.

### Steps to reproduce

 1. Connect Nautilus to an IBKR paper Gateway account.
 2. Install the latest development wheel:
```bash
pip install -U nautilus_trader --pre \
  --index-url https://packages.nautechsystems.io/simple
```
 3. Start a minimal IB execution node and request account summary data.
 4. Provide an account summary equivalent to:
```text
AccountType      = INDIVIDUAL
NetLiquidation   = XXX <BASE_CCY>
TotalCashValue   = XXX <BASE_CCY>
```
    There are deliberately no Currency or BaseCurrency rows.
 5. Observe that account construction fails with the error below.

### Code snippets or logs

 ```text
Failed to parse account summary: Account summary currency was empty
Cannot resolve portfolio equity without exactly one account
 ```

[...]

Issue #4973, opened 2026-09-11
[Interactive Brokers] IneligibilityReason in contract details cannot be serialized into the cache database


FALK-BRAUER:
**Bug: IB adapter — an `IneligibilityReason` in contract details cannot be serialised into the cache database; the live node terminates at boot**

**Versions:** nautilus_trader 1.229.0, ibapi 10.45.1, Python 3.13, Redis cache database (`flush_on_start=False`).

**What happens.** When IB returns contract details with a non-empty `ineligibilityReasonList` (e.g. `[IneligibilityReason(id_='i152', description='Instrument temporarily unavailable pending configuration review.')]`), `adapters/interactive_brokers/parsing/instruments.py::contract_details_to_dict` copies `contract_details.dict()` into `Instrument.info`. `_serialize_for_json` there converts `Decimal` and `Enum` only, so the ibapi `IneligibilityReason` object survives into `info`. With a cache database configured, `MsgSpecSerializer` raises:

```
TypeError("Encoding objects of type <class 'ibapi.ineligibility_reason.IneligibilityReason'> is unsupported")
  live/data_engine.py:486 _run_data_queue
  data/engine.pyx:2563 _handle_data → :2590 _handle_instrument
  cache/cache.pyx:2062 Cache.add_instrument
  cache/database.pyx:947 CacheDatabaseAdapter.add_instrument
  serialization/serializer.pyx:112 MsgSpecSerializer.serialize → :170 _serializer_encoding_hook
  common/config.py:136 msgspec_encoding_hook
```
followed by `DataEngine: System will terminate immediately to prevent operation in degraded state`. The node crash-loops and cannot boot while any instrument in its universe carries a reason. The same node booted fine while every instrument's list was empty; IB attached the reason to one contract overnight.

**Minimal repro (no gateway):**
```python
from ibapi.contract import ContractDetails
from ibapi.ineligibility_reason import IneligibilityReason
from nautilus_trader.adapters.interactive_brokers.parsing import instruments as p
from nautilus_trader.model.identifiers import InstrumentId, Symbol, Venue
from nautilus_trader.model.instruments import Equity
from nautilus_trader.model.objects import Currency, Price, Quantity
from nautilus_trader.serialization.serializer import MsgSpecSerializer
import msgspec

cd = ContractDetails()
cd.contract.symbol, cd.contract.secType, cd.contract.exchange, cd.contract.currency = "XYZ", "STK", "SMART", "USD"
[...]

PR #5041, opened 2026-09-21
Port the Interactive Brokers adapter and add execution client order identity hooks

faysou:
## Summary

The Interactive Brokers adapter is rewritten against the published `ibapi 4.1.0` release, replacing
the `3.3.0` surface. The rewrite needs the execution engine to understand two things IB reports that
no other venue does: a second permanent order ID behind one client order reference, and an execution
that arrives before its authoritative order status. `ExecutionClient` gains three defaulted methods
for that, so every other adapter keeps its current behavior without an edit. Futures and
futures-option instrument identifiers are unchanged: the adapter still emits the local symbol IB
reports, such as `YMM6.XCBT`.

```mermaid
flowchart TD
    R[Venue report] --> I{Client declares
    distinct order identity?}
    I -- no --> E[Existing single-order handling]
    I -- yes --> C{Account and instrument
    match the original order?}
    C -- no --> X[Error, no order created]
    C -- yes --> D[Second order under the same strategy]

    F[Fill for an unknown order] --> Q{Client requires
    order status for fills?}
    Q -- no --> E
    Q -- yes --> P[Retain fill, deduplicated by trade ID]
    P --> S["Query order status
    QUERY_INCLUDE_FILLS, client timeout"]
    S --> A[Status quantity becomes the order quantity]
    A --> B[Retained fills applied]
```

### Design decisions

- The change is two commits. The first carries the core prerequisites with no IB code: the
  `ExecutionClient` methods, `QUERY_INCLUDE_FILLS`, the engine handling they drive, and the model
  identifier additions. The second carries the adapter, the dependency bump, the docs, the examples,
  and the Python surface. A reviewer can check out one branch and still bisect the trait change from
  the adapter rewrite.
- `has_distinct_order_identity`, `requires_order_status_for_fill`, and `order_status_query_timeout`
  default to `false`, `false`, and `DEFAULT_ORDER_STATUS_QUERY_TIMEOUT` (30 seconds). The defaults
  are what make the trait change additive: no adapter outside IB needed an edit, and downstream
  implementors outside the repository still compile.
- The engine keeps retained fills in a map keyed by account and venue order ID, deduplicated by
  trade ID, so a repeated report cannot double-count. A distinct-identity report whose account or
  instrument conflicts with the original order fails with an error rather than materializing a
  mismatched order.
- `reconcile_order_status_report`, `reconcile_fill_report`, and `reconcile_order_with_fills` each
  split into an admission step that may stage the report and an `apply_*` step that holds the
  previous behavior unchanged. Staging returning `false` leaves the existing path byte-for-byte
  intact, which is what keeps every other adapter's reconciliation identical.
- `ClientOrderId::for_duplicate_order` builds the second order's identity from the account and venue
  order ID, and `DUPLICATE_ORDER_PARENT_TAG` records the original client order on it, so the
  relationship survives a cache round-trip.
- IB futures symbol parsing moves to `nautilus_model::identifiers::futures`. It is additive and has
  one consumer today; the Databento alignment that shares it is a stated follow-up. The alternative,
  keeping it in the adapter crate, changes no behavior and stays available if a reviewer prefers it.
- `new_generic_spread_id` and `parse_generic_spread_id_legs` join `nautilus_model::identifiers` with
  a Python binding, so IB combo contracts round-trip through generic spread identifiers from both
  languages. Leg ratios parse as digits only, because `parse::<i64>` would accept a leading sign and
  silently flip the sign the surrounding parentheses encode.
- The adapter reads `instrument_id` and `bar_type` straight from each request. No request or response
  message needed an identifier list or a shared data handle.
- Futures identifiers keep the venue local symbol. The adapter's year-digit setting is pinned to
  pass-through, and the configuration field and Python keyword that would expose it are not part of
  this change. Turning on two-digit contract years silently rewrites identifiers already stored in
  catalogs and caches, which is a decision of its own.
- A historical request that fails partway through pagination now emits the rows it obtained and logs
  the failure at warning level. The structured failure marker it replaced is read by data-engine
  modules that do not exist here, and a marker with no consumer is worse than no marker.

### Reconciliation double-count on restart with a venue position report

Paper-account validation surfaced a reconciliation defect: after a restart with an open position,
the position doubled. It reproduces on `develop` independently of the new execution-client hooks, so
it is a pre-existing bug the IB adapter exercises rather than one this port introduces. It is fixed
here because IB is the adapter that reliably triggers it.

Startup reconciliation builds the position from the venue position report and materializes a
synthetic order tagged `RECONCILIATION` to open it, timestamped at reconciliation time. The venue
then re-delivers the execution that originally created the position on the live execution stream.
That fill carries an earlier `ts_event` than the reconciled position's `ts_opened`, so it is applied
a second time and the position doubles.

A fill whose `ts_event` predates a position's `ts_opened` is an ambiguous signal on its own. It
describes two opposite situations that look identical on timestamp alone:

- A duplicate already baked into the venue snapshot: a position reconciled from an authoritative
  position report already includes every prior execution, so replaying one double-counts it.
- A genuine backfill: a real, not-yet-counted execution that arrived late and out of order, which
  must be applied. `test_reconcile_execution_mass_status_delivers_trades_with_stale_order_snapshot`
  covers this: a position opened by trade `T-SNAPSHOT` at `ts_event=9_000_000` and a distinct
  `T-ADDITIONAL` at `ts_event=1_000_000` that must still take the position to its full quantity.

A guard keyed on the timestamp alone gets one case wrong: suppress all pre-open fills and legitimate
backfills are dropped; apply all pre-open fills and snapshot duplicates double the position. The
fix keys on **position provenance** instead. `ExecutionEngine::apply_fill_report` applies a
reconciliation fill to the order without reopening the position only when the open position's
`opening_order_id` resolves to an order tagged `RECONCILIATION` (reconciled from a venue snapshot)
and the fill both predates the position's `ts_opened` and is not already recorded on it. A
backfill onto a fill-established position, whose opening order carries no such tag, still applies
normally. The tag string moves to `nautilus_execution::reconciliation::RECONCILIATION_ORDER_TAG`,
which the live manager now references, so the engine and the manager share one definition.

```mermaid
flowchart TD
    F[Reconciliation fill] --> P{Open position for
    account and instrument?}
    P -- no --> A[Apply normally]
    P -- yes --> T{Fill ts_event before
    position ts_opened
    and trade not present?}
    T -- no --> A
    T -- yes --> R{Opening order tagged
    RECONCILIATION?}
    R -- no, genuine backfill --> A
    R -- yes, snapshot duplicate --> O[Update order only]
```

This guard is engine-local and stateless: it reads provenance from the cache on demand rather than
storing it. The deciding question for the approach is whether the opening order can disappear while
its position is open, because the check resolves that order to read its tag. Closed-order purging is
the only mechanism that evicts it: `purge_closed_orders_interval_mins` defaults to `None`, so purge
is off by default and the tagged order is never evicted in the default configuration (the one the
paper-account validation ran). When purge is enabled, `Cache::purge_closed_orders` drops any closed
order past its buffer and does not skip orders that opened a still-open position, so with a buffer
shorter than the delay before the venue replays the execution, the tagged order could be gone and
the guard would silently stop firing. That narrow failure mode is the whole argument for storing
provenance on the position itself.

Three ways to encode the provenance:

| Approach | Purge-proof | Survives restart | Reopen lifecycle | Blast radius |
| --------------------------------------------------- | ----------- | ----------------------- | ------------------------------- | -------------------------------------------------------- |
| A. Tag check via `opening_order_id` (chosen) | No | Re-derived each session | Free, rides on `opening_order_id` | Two files plus the shared constant |
| B. Engine-local `(account, instrument) -> ts_opened` | Yes | Re-derived each session | Needs an explicit reset on reopen | Execution engine plus manager plumbing |
| C. Field on the core `Position` model | Yes | Yes | Free when derived at open | `Position`, msgpack and JSON serde, Python and Cython mirrors, every constructor and stub |

Two properties are easy to miss. The chosen check gets flat-then-reopen right for free: a position
reopened by a real order takes that order as its `opening_order_id`, loses the tag, and the guard
correctly stops suppressing. Option B reintroduces a reset-on-reopen step that is easy to get wrong.
Option C keeps the clean lifecycle only if the field is derived at position construction from the
opening fill's order, the way `opening_order_id` already is, rather than set imperatively later; done
that way it is strictly the most robust, purge-proof and restart-durable with a self-resetting
lifecycle, at the cost of widening a core serialized type for a reconciliation-only concern.

Approach A is chosen here. It is correct under the default configuration, it is two files rather than
a core-type change carrying serialization and binding churn, and it rides on the existing
`opening_order_id` lifecycle; provenance has exactly one consumer today. Approach C becomes the right
choice when a deployment enables `purge_closed_orders` with a short buffer, or when a second consumer
(risk, analytics, another reconciliation path) wants the provenance. Because `RECONCILIATION_ORDER_TAG`
is already extracted and the guard is one localized method, that migration is contained rather than a
rewrite. Option B is a worse middle: it takes on the plumbing and the reset-on-reopen hazard without
C's durability.

Regression coverage:
`test_reconcile_fill_report_does_not_reopen_snapshot_reconciled_position` reproduces the restart
double-count (200,000 without the guard, 100,000 with it), and the stale-order-snapshot backfill
test confirms genuine backfills still apply.

### Interactive Brokers adapter refactor

The adapter is not a small patch on the previous crate; it is a restructuring into focused modules
with substantially reworked execution, data, and connection behavior. The execution client splits
into `core` (client and lifecycle), `commands` (submit, cancel, modify, and their denials),
`updates` (order and execution stream handling), `order_state` (tracked order state and the
cache-restore instrument preload), `incarnations` (distinct physical orders that share a client
reference), `account` (account summary plus the `reqAccountUpdates` snapshot), `spreads` (combo
legs), `conditions`, `parse`, and a `transform` submodule for order transformation policy and tags.
The data client splits into `core`, `streams` (subscription lifecycle), `cache`, `convert`, and
`parse`. Shared concerns move to `common`: `shared_client` for the reference-counted connection,
`symbology` for contract-to-instrument-ID conversion, and `spreads` for generic spread identifiers.

The behavior that restructuring carries:

- Order acceptance follows venue acknowledgment rather than local send success; a successful send
  alone does not establish acceptance.
- Distinct permanent IDs sharing one client order reference remain separate physical orders under
  the original strategy. Full order quantities precede unknown fills, actual executions determine
  filled quantity, and late statuses cannot reopen filled duplicates. Cancellation includes working
  children after the original closes.
- Strategy-only requests use individual cancels by default; explicit instrument-wide requests retain
  account, instrument, and side filters.
- The account path reads one `reqAccountUpdates` snapshot after the summary and stores every
  `updateAccountValue` key in `AccountState.info`, so values IB serves only there, such as
  `PostExpirationExcess`, reach Python. Account summaries build balances and margins from each row's
  own currency; ratio rows such as `Cushion` are skipped before currency parsing and no longer warn.
- Fill report requests filter executions by the raw IB account code, so startup reconciliation
  returns historical fills instead of TWS error 321.
- Market data supports Last and AllLast trade selection, optional subscription-idle custom data,
  no-quote sentinel filtering, depth reset on stream restart, and DEBUG diagnostics for expected
  unrepresentable fractional sizes. Trade, quote, realtime-bar, and depth streams that end
  unexpectedly warn before resubscribing.
- Unsupported Gateway or TWS versions fail before `StartApi`; protocol 213 is the documented
  minimum, with bounded connection timeouts.
- Historical pagination is bounded, the shared connection counts completed acquisitions (a
  cancelled waiter leaves no reserved reference), commission waiting and account filtering are
  explicit, and gateway passwords are redacted.

### Interactive Brokers behavior

- Order callbacks correlate by account, native client and order identity, permanent ID, and client
  order reference. A binding notice binds a route only in the subscription that owns the reported
  identity, so a second subscription sharing a raw order ID is unaffected, and a fill never rewrites
  a route already established for cancellation.
- A route with no binding, owned by another API client, or whose raw order ID maps to more than one
  permanent ID stays explicitly unresolved and is reported. The adapter does not resolve the
  ambiguity by cancelling account-wide.
- `OrderBound` comes from the published `ibapi 4.1.0` crates.io release. The dependency carries no
  `git`, `branch`, `rev`, or `path` key.

### Related issues/PRs

| Issue | Behavior covered |
| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| #4470 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4470#issuecomment-5636199995)) | Strategy-only cancellation stays per-order by default; instrument-wide requests retain account, instrument, and side scope. Other adapters remain follow-ups. |
| #4564 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4564#issuecomment-5636200560)) | Distinct permanent IDs sharing one client reference stay separate physical orders; real fills are preserved and uniquely bound working members are cancelled. |
| #4796 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4796#issuecomment-5636201088)) | Protocol 213 is the documented minimum; unsupported Gateway or TWS versions fail before `StartApi` with an explicit error. Connection timeouts stay bounded. |
| #4932 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4932#issuecomment-5636201623)) | The execution instrument provider seeds after cache restoration and qualifies instruments of cached orders that are missing. |
| #4946 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4946#issuecomment-5636202185)) | Partial: unresolved accounts are rejected at the adapter. Shared risk and execution account routing is a deferred follow-up. |
| #4947 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4947#issuecomment-5636202690)) | Trade, tick-by-tick quote, realtime-bar, and depth streams that end unexpectedly log a warning before resubscribing. |
| #4948 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4948#issuecomment-5636203229)) | Expected unrepresentable fractional market-data quantities log at DEBUG; other parse failures keep their warnings. |
| #4949 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4949#issuecomment-5636203760)) | Optional subscription-idle custom data; market data rearms the timer and farm notices do not mask inactivity. |
| #4960 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4960#issuecomment-5636204519)) | `all_last_trades` defaults to AllLast; `False` selects Last. |
| #4970 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4970#issuecomment-5636205200)) | Callbacks correlate by account, native client and order identity, permanent ID, and reference; fills do not rewrite cancellation routes. |
| #4973 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4973#issuecomment-5631093095)) | V1-only serializer issue; no v2 code change. The reply records the distinction. |
| #4987 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/4987#issuecomment-5703350235)) | Account summaries build balances and margins from each row's own currency; ratio rows such as `Cushion` no longer log currency warnings. |
| #5007 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/5007#issuecomment-5703259949)) | Fill report requests use the raw IB account code, so startup reconciliation returns historical fills instead of TWS error 321. |
| #5032 ([reply](https://github.com/nautechsystems/nautilus_trader/issues/5032#issuecomment-5744792333)) | One `reqAccountUpdates` snapshot merges into `AccountState.info` keyed by IB name, so values such as `PostExpirationExcess` reach Python. |

The issues stay open; this records implementation coverage. Also in scope: venue-acknowledged
acceptance rather than local send success, bounded historical pagination, no-quote sentinel
filtering, depth reset on stream restart, gateway password redaction, and shared-connection
acquisition counting where cancelling a waiting acquisition leaves no reserved reference.

### Type of change

- [x] Bug fix (non-breaking).
- [x] New feature (non-breaking).
- [x] Breaking change (impacts existing behavior).
- [x] Documentation update.

### Breaking change details

- `ibapi` moves from `=3.3.0` to `=4.1.0`. Code written against the `3.3.0` API surface needs
  updating; nothing inside this repository depends on it outside the adapter.
- `ExecutionClient` gains three methods. All three have defaults, so existing implementations
  compile and behave exactly as before, but the trait's public surface changes.
- Instrument identifiers are unchanged. Futures and futures-option IDs still carry IB's local
  symbol, so existing catalogs and caches keep resolving.

### Deferred follow-ups

- Two-digit futures contract years, the `futures_year_digits` configuration surface, and the
  Databento symbology alignment that depends on it.
- The structured data-engine failure marker for partial request failures, with the response,
  streaming, and historical-execution modules that read it.
- Shared risk and execution account routing.
- The tick-scheme registry.
- Matching-engine synthetic leg fills.

### Documentation

- `docs/integrations/interactive_brokers.md` covers the rewritten adapter. The year-digit section is
  replaced by a statement that futures identifiers carry the IB local symbol unchanged, and the
  provider configuration table no longer lists a year-digit option.
- `examples/live/interactive_brokers` is updated, including the `_common.py` module the examples
  share. `historical_download.py` writes through `ParquetDataCatalog`.

### Release notes

Not added. `RELEASES.md` is maintainer-owned.

### Testing

- [x] Affected code paths are covered by the test suite.
- [x] Added and updated tests to cover new and changed logic.

Added engine tests covering every scenario in the new behavior: two permanent IDs under one client
reference producing a second order, the same report from a client that declares nothing producing
one order, a declared-distinct report conflicting on account or on instrument failing with an error
and creating no order, an unknown fill retained until its status arrives, the status supplying the
authoritative quantity, a repeated fill counted once per trade ID, and a query retried at the
client's declared timeout rather than the engine default.

Added model tests for generic spread identifier construction, leg parsing, sign handling, and the
round trip, and for futures symbol parsing, formatting, and year resolution.

Local runs:

| Check                                            | Result                                     |
| ------------------------------------------------ | ------------------------------------------ |
| `cargo-test-crate-nautilus-model`                | 4,221 passed, 1 skipped                    |
| `cargo-test-crate-nautilus-common`               | 2,111 passed, 4 skipped                    |
| `nautilus-execution`, lib and integration        | 1,848 passed                               |
| `cargo-test-crate-nautilus-interactive-brokers`  | 535 passed                                 |
| `cargo-test-core-local`                          | 17,574 passed, 1 pre-existing failure      |
| `pytest python/tests/unit/adapters`              | 317 passed                                 |
| Both adapter examples, `--features examples`     | build clean, 25 embedded config tests pass |

The single core-local failure is
`nautilus-network websocket::client::rust_tests::connection_rate_limit_gates_initial_connect_and_reconnect`,
which fails identically on an unrelated branch of the same repository and touches no code in this
change.


tradatious:
Note that 4.2.0 just got released.

faysou:
@tradatious thanks there would be a cooldown period for the new verison so it depends on when the PR will get merged. We'll update to latest version at some point anyway, it's simple to upgrade. On the other hand we could also avoid a cooldown, as it's a library regularly used and if there's no new dependency.

tradatious:
Two issues that appear to be new:

## Stale cache tears down session

A stale cached instrument stops the execution client from connecting. `preload_cached_instruments` loads every instrument referenced by the client's cached orders that the provider doesn't already hold ([order_state.rs#L127-L145](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/execution/order_state.rs#L127-L145)) and requires each load to come back with exactly the requested id (the `ensure!` at [L173-L176](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/execution/order_state.rs#L173-L176)). A miss fails `connect`, and the session is torn down ([core.rs#L1108-L1115](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/execution/core.rs#L1108-L1115)).

The old adapter preloaded too, but only for spread instruments. It also only warned and continued on a miss ([core_tracking.rs#L55-L79 at the base](https://github.com/nautechsystems/nautilus_trader/blob/84a98f7f3c/crates/adapters/interactive_brokers/src/execution/core_tracking.rs#L55-L79)). The rewrite expands the preload to every cached order and makes a miss fatal. An expired contract or a delisted symbol on a months-old cached order would now stop the client starting, requiring a cache purge to recover.

Reconciling before trading makes sense, but refusing to start over an unresolvable instrument doesn't seem great. Perhaps restore the old warn-and-skip behaviour for that order, or gate the strictness behind a config flag?

## Wrong account code sent for non-default client names

The account code sent to IB is now wrong for any client not named `IB`. The Nautilus `AccountId` is built as `{client name}-{IB account}`, where the client name is the key the IB execution client is registered under in `exec_clients` ([factories.rs#L202-L211](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/factories.rs#L202-L211)). With the default name that gives `IB-U1234567`; a node running two IB accounts as `IB_LIVE` and `IB_PAPER` gets `IB_LIVE-U1234567` and `IB_PAPER-U7654321`.

To talk to IB the adapter has to recover the raw code, and `raw_ib_account_code` does that by stripping a literal `IB-` prefix ([account.rs#L45-L51](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/execution/account.rs#L45-L51)). That works for the default name only: `IB_LIVE-U1234567` has no such prefix and is sent to IB verbatim as the account, and a client named `IB-TEST` (the name your factory tests use) loses the `IB-` off the front of the client name and sends `TEST-U1234567`.

This change now makes orders go through `raw_ib_account_code`. The old adapter split on the first hyphen for orders ([core_orders.rs#L119-L124 at the base](https://github.com/nautechsystems/nautilus_trader/blob/84a98f7f3c/crates/adapters/interactive_brokers/src/execution/core_orders.rs#L119-L124)), which handled any name; only its report filters had the prefix-only bug. The rewrite unified onto the prefix helper for single orders ([commands.rs#L100](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/execution/commands.rs#L100)) and order lists ([L418](https://github.com/nautechsystems/nautilus_trader/blob/a9a73b8777/crates/adapters/interactive_brokers/src/execution/commands.rs#L418)), so the inconsistency was fixed in the wrong direction. Modify is unaffected; it reuses the order object IB returned.

Maybe the raw code should be derived once from the configured `account_id` and carried alongside the Nautilus id, instead of parsing the composite id.

faysou:
Thanks for the review, I'll update.

faysou:
Hi @tradatious, thanks for the careful read. Both are right and both are fixed in 2fe1bc5.

**Stale cache.** `preload_cached_instruments` now warns and skips an instrument IB cannot resolve instead of failing `connect`, restoring the previous warn-and-continue behaviour. The warning names the instrument, and orders on it stay unreconciled rather than blocking startup.

**Account code.** The raw IB code is derived once at construction from the configured `account_id` (`ib_account_code`) and carried on the execution client, then passed to every place that talks to IB: single orders, order lists, modifies, cancels and cancel resolution, the order and execution streams, account summary, PnL, positions, and execution filters. `raw_ib_account_code` and the transform's account policy are gone, so nothing parses the composite ID any more. A configured bare code wins; a configured composite such as `IB-U1234567` falls back to the account part of the Nautilus ID. Unit tests cover `IB-TEST` and `IB_LIVE` names, and a paper TWS session registered as `IB_PAPER` now receives its account summary and positions under `IB_PAPER-DU187075`, where before the summary was filtered out.

tradatious:
Thank you, that was fast! And the changes look great.

One small issue, in case you want to close it: if a consumer configures `account_id` in the composite form with a hyphenated client name, e.g. `IB-TEST-U7654321`, the fallback takes the part after the *first* hyphen and sends `TEST-U7654321`. Perhaps that isn't a concern, but the config field is documented only as "Account ID". Documenting the field as the bare IB code would be good, and perhaps also take the segment after the *last* hyphen in the fallback?

faysou:
Thanks @tradatious, both done in 3c271bb. The `account_id` field is now documented as the raw IB code (`DU123456`), with the Nautilus ID becoming `{client name}-{code}`, in the Rust config, the Python getter, and the docs table. The fallback takes the segment after the last hyphen, so a composite `IB-TEST-U7654321` sends `U7654321`; the derivation test covers that case alongside the bare and `IB_LIVE` ones.

tradatious:
Awesome, thank you @faysou!

faysou:
You're welcome, thank you for reviewing too.

zzlatev-openclaw-neo:
Does this port support requesting a previously unseen instrument after the node starts, then trading it without reconnecting while another position remains active?

At `e2fc015`, I traced `request_instrument` into the data provider and node cache, but couldn’t find the subsequent hydration of the execution provider. The core has an `on_instrument` hook, but the IB execution client appears to retain the default no-op, while tracked-fill handling requires that provider entry.

Is there a supported higher-level path I’ve missed, or is this a remaining gap in the port?

honvl:
Test report from #5057: we built this PR at 31d2067 and ran it against an IB paper account (CME MNQ futures). The setup is a `LiveNode` on `run_async()` with the IB data and execution clients, reconciliation on, and a Python strategy.

**The #5057 restart case now recovers.** A process was stopped with a position open and two orders still working at IB in one OCA group (`ocaType=2`): a GTC `STOP_MARKET` and a DAY `MARKET` with `goodAfterTime`, which IB holds as `PreSubmitted`. After a restart with the same trader ID and client ID:

- both orders were reconciled as `ACCEPTED` external orders, and their updates were tracked;
- the strategy cancelled both by `PERM-` venue order ID and flattened the position.

There was no "Trader ID not found", no `Unhandled order status SUBMITTED`, and no missing stop.

**The in-session path also passed, long and short:**

- a GTD limit entry;
- the protective stop, the `goodAfterTime` exit and a limit target in one OCA group;
- a stop-trigger modify;
- the IB-released `goodAfterTime` fill;
- OCA sibling cancels.

The `Unexpected order status in venue report ... Submitted` warning that 2.0.0rc6.dev20260921 logged on every reconciliation pass is gone.

**Two observations with this build:**

1. **`RuntimeError: Already borrowed` on order callbacks.** Every `submit_order` and `modify_order` from the Python strategy logged it for `on_order_initialized`, `on_order_event` and, on the modify, `on_order_pending_update`. That was 26 occurrences across entries, protective orders and a stop modify. The same strategy code on 2.0.0rc6.dev20260921 logged none. The orders themselves are unaffected, but those events never reach Python. The branch carries develop changes made after that wheel, so this may not be specific to this PR.
   ```
   [ERROR] TRADER-001.MyStrategy: Python on_order_initialized failed:
   RuntimeError: Already borrowed
   [ERROR] TRADER-001.MyStrategy: Python on_order_event failed:
   RuntimeError: Already borrowed
   ```
2. **A previous session's GTD orders aren't rebuilt at reconciliation.** It fails with `Failed to create order from report: Invalid OrderInitialized event: expire_time is required for GTD order`. The parsed status report doesn't seem to carry `goodTillDate` as `expire_time`, which is the same shape as the stop `trigger_type` fix. In our case these were already-filled GTD entries, and the venue position report covered the position.

Build: Rust 1.98.1, maturin 1.15.0, release profile with LTO off, macOS arm64, Python 3.12.12, IB Gateway (paper). We're happy to rerun on a later commit.

faysou:
The complexity of IB never stops. Luckily we have several people reporting bugs and the help of modern tools. Using a cache like redis improves stability between sessions, but if we can rebuild everything without a cache even better.

faysou:
Hi @honvl, thanks for the detailed test report, it helped a lot. Both issues are fixed in 8af617e, which is also rebased on the latest `develop`.

1. **`RuntimeError: Already borrowed` on order callbacks.** You were right that this isn't IB-specific. The Python `Strategy` methods took a mutable borrow, so PyO3 held the strategy object exclusively while `submit_order` or `modify_order` ran. The order events are dispatched inside that call, so any callback that touched the strategy again (the default callback, or an override reading `self.cache`) failed, and the event never reached Python. The callbacks were already failing on 2.0.0rc6.dev20260921, but the errors were dropped silently; a recent `develop` change started logging them. Python `Strategy` methods now take a shared borrow, and order callbacks run normally during a command.
2. **Previous-session GTD orders not rebuilt.** The status report now carries IB's `goodTillDate` as `expire_time`, so reconciliation recreates them as GTD orders. An unparsable date is reported as GTC with a warning instead of failing the report, and IB still enforces the actual expiry.

I checked both on a paper account: a Python strategy submitted two GTD limit orders (one from inside a callback), modified one, and read the cache in every callback without errors. After a restart with the same client ID, both orders were reconciled as `ACCEPTED` GTD orders and cancelled by the strategy.

A rerun on 8af617e would be welcome if you have time.

honvl:
Retested at 8af617e (built from source), same setup: IB paper account, CME MNQ futures, `LiveNode` on `run_async()`, Python strategy.

- **`Already borrowed`: fixed.** No occurrences across the night's submits and modifies (the previous build logged 26 on the same runs). A backtest check agrees: with order callbacks that read `self.cache`, 2.0.0rc6.dev20260921 delivered 3 of 5 callbacks during `submit_order`/`modify_order` (`on_order_initialized` dropped), and 8af617e delivers all 5.
- **Previous-session GTD orders: fixed.** After a restart, the earlier session's GTD entry is rebuilt (as `FILLED`) instead of failing on `expire_time`.
- **Everything from the first report still passes.** Restart recovery: the working GTC stop and `goodAfterTime` exit reconcile as `ACCEPTED`, are cancelled by `PERM-` ID, and the position is flattened. The in-session long and short paths pass too: GTD entry, OCA protection, stop modify, IB-released `goodAfterTime` exit, OCA cancels. No status warnings.
- **WhatIf through the adapter:** 1 MNQ and 1 NQ per side, all answered `PreSubmitted` in 0.04–0.13 s with a positive initial margin, and none left open.

Thanks for the quick fixes.

cjdsellers:
Thank you for the additional testing @faysou, @tradatious, and @honvl.

PR #5043, opened 2026-09-21
Declare the lookback window on Kraken bounded mass status

### LINKED ISSUE #5042: [Kraken] Bounded mass-status reports do not declare their lookback window
# Bug Report

## Confirmation

## Expected behavior

The adapter guide, under "Bounded mass-status reports", requires that an execution client applying a
lower time bound to historical reconciliation reports records the contract with
`ExecutionMassStatus::set_report_window(Some(lookback_start), reports_complete)`, and sets
`reports_complete=true` only when every required record was parsed, mapped and linked.

## Actual behavior

Neither Kraken execution client calls `set_report_window`. `generate_mass_status` computes a
cutoff from `lookback_mins` and applies it to the fill query, but the returned
`ExecutionMassStatus` declares no window:

```rust
let start = lookback_mins.map(|mins| Timestamp::now() - Duration::from_secs(mins * 60));
// ...start is passed to the fill query...
let mut mass_status = ExecutionMassStatus::new(/* ... */);
mass_status.add_order_reports(order_reports);
mass_status.add_fill_reports(fill_reports);
mass_status.add_position_reports(position_reports);
// no set_report_window
```

So the engine cannot apply its bounded-history rules and follows the compatibility fill-adjustment
path instead, even though the fill set is genuinely bounded.

There is also no way to report the set as incomplete. Since #5034 a historical order or fill whose
instrument cannot be resolved is logged and skipped, which the guide classifies as making the set
incomplete ("a required row that cannot be parsed or mapped"). That incompleteness is currently
unobservable to the engine.

Both clients are affected: `crates/adapters/kraken/src/execution/spot.rs` and
`crates/adapters/kraken/src/execution/futures.rs`.

## Steps to reproduce

1. Run a Kraken spot or futures execution client and trigger reconciliation with a `lookback_mins`
   value, so `generate_mass_status` applies a cutoff.
2. Inspect the returned `ExecutionMassStatus`.

Observed: no report window is set, so `lookback_start` is absent and the set is always treated as
unbounded regardless of whether rows were skipped.

## Code snippets or logs

`grep -rn "set_report_window" crates/adapters/kraken/src/` returns nothing, while the okx,
hyperliquid, lighter, polymarket and binance execution clients each record the window.

## Specifications

- Adapter/venue: Kraken (spot and futures)
- Reproduced against: `develop` @ `b3f35826eb`

zhaow-de:
  and, if I used AI,
  this draft
  limitation below

## Summary

Both Kraken execution clients derived a cutoff from `lookback_mins` and applied it to the historical
queries, but never recorded it with `ExecutionMassStatus::set_report_window`. The engine could not
apply its bounded-history rules, and had no way to observe an incomplete set.

`generate_mass_status` now records the cutoff for spot and futures, with a completeness flag that is
false when a record was skipped because its instrument could not be resolved. Each client already
derived a single cutoff and used it for every historical query, so the "one cutoff" requirement was
already met and is unchanged.

The public `request_order_status_reports` and `request_fill_reports` keep their signatures, since
they are exposed through the Python bindings. `pub(crate)` `..._checked` variants carry the flag.

## Related issues/PRs

Closes #5042

Follows #5034, which introduced the skip path on the spot side that this flag now reports.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

Not an API change; the Python-facing methods are untouched. A bounded mass status now carries a
`lookback_start` and may report `reports_complete=false`, so the engine applies its bounded-history
rules where it previously took the compatibility path.

## Documentation

  generated output

`docs/integrations/kraken.md` gains a "Bounded reports" section under Reconciliation. No binding
signature or wrapped doc changed, and `generate_docstrings.py --dry-run` reports no drift.

## Testing

- `test_spot_mass_status_declares_lookback_window` asserts the cutoff is recorded.
- `test_spot_mass_status_incomplete_when_historical_fill_unresolved` asserts a control first: the
  identical payload with every pair resolvable reports complete with both fills, and swapping one
  pair for an unresolvable one reports incomplete with one fill. Without the control the assertion
  would also hold if the row had been dropped for an unrelated reason.

`cargo test -p nautilus-kraken`: 784 passed, 0 failed. `cargo clippy --all-targets --all-features --
-D warnings`: clean.

## One thing for your attention

The futures report reads previously dropped an unresolved record silently, with no log at all,
which is the futures equivalent of what #5034 fixed for spot. I added the warning the guide
requires, because a completeness flag that flips without a log tells an operator nothing. I did not
change futures to return an error for an unresolved **open order**, which is what spot now does
after #5034, since that is a behaviour change beyond this issue. Happy to raise it separately if
you would like the two clients aligned.


ashishsinghbora (commented):
Overall, the approach looks good and the report-window tracking is a useful fix for bounded Kraken mass-status reports. The API compatibility and unresolved historical instrument handling also look consistent with the existing adapter behavior.

One correctness issue remains: parse failures are logged and skipped, but they don't appear to set the new `complete` flag to `false`. This can result in `set_report_window(..., true)` being called even though one or more required historical records were dropped, causing reconciliation to treat the bounded report set as complete when it isn't.

Could you update the order/fill/order-event parse error paths to mark the report set as incomplete, and add a regression test covering a parse failure with `reports_complete == false`?

With that addressed, the implementation looks much safer to merge.

cjdsellers:
Hi @zhaow-de,

Thanks for the PR, the control-then-variant test structure is exactly right.

Thanks @ashishsinghbora for the input on the parse failure paths. I found the same issue independently, which adds confirmation: the parse `Err` arms in the `_checked` functions leave `complete = true`, while the adapter guide counts a row that cannot be parsed as incomplete, and OKX and Hyperliquid set their flag to false there. Best to place the regression test on a fills parse failure, since the order-event and closed-order arms aren't reachable from `generate_mass_status` under `open_only=true`.

A few smaller things:

- `orders_complete` on spot can never be false: mass status fetches orders with `open_only=true`, which returns before the `ClosedOrders` loop holding the only `complete = false` site (the open orders loop errors rather than skips). A comment or trimmed plumbing would stop the tuple suggesting coverage it doesn't have.
- The docs section overstates the rule: position records aren't in the conjunction, and the futures position loop still drops unresolved symbols silently, while OKX folds `positions_complete` in. It also reads as if `reports_complete` gates the bounded-history rules; that's gated on `lookback_start`.
- `start.map(UnixNanos::from)` panics on a pre-epoch cutoff, and `reconciliation_lookback_mins` is an unclamped `u32`, so a lookback over ~56 years panics during reconciliation where the old code just passed the timestamp through. The OKX-style `saturating_sub` avoids it.
- Binance also marks the set incomplete when a historical fill has no order report in the set or the cache. With open-only orders that's routine here, though the engine already projects those groups order-only, so lower priority.

The deferred futures open-order error behavior is fine as a separate issue, thank you for offering.

zhaow-de:
Thanks @ashishsinghbora and @cjdsellers — you both caught the same thing, which made it easy to act on.

Parse failures now mark the set incomplete, in the order, order-event and fill arms on both clients. The regression test is on a fills parse failure as suggested, and it asserts a control first so only the unparsable volume differs between the two runs.

On `orders_complete` never being false: the parse fix resolves that. The open-orders parse arm sits before the `open_only` early return, so it is now the reachable site — I checked the ordering rather than assuming it.

The panic is fixed too. The cutoff is derived in `UnixNanos` with saturating arithmetic and converted forward to the query timestamp, so neither an overflow nor a pre-epoch value reaches the assert.

Docs corrected: order and fill records contribute to the flag, position records do not, the futures position read still drops an unresolved symbol silently, and the bounded-history rules are gated on the cutoff rather than on completeness.

Left the Binance-style fill/order pairing out for now, as you suggested — happy to pick it up separately.

785 Kraken tests pass, clippy clean.

cjdsellers (dismissed):
@zhaow-de, thanks for the parse-failure, cutoff, and docs updates. Those look right.

One non-blocking follow-up: the new tests all drive the spot client, so a futures lookback case that records the cutoff, and one unresolved fill that marks the set incomplete, would pin the other client the same way.

This looks good to merge.

zhaow-de:
Thanks @cjdsellers. Added the futures pair in a new commit: one test asserts a lookback records the cutoff, the other asserts an unresolved fill marks the set incomplete.

Both use the same control-then-variant shape as the spot tests, so the same fill on a listed symbol reports complete with one fill, and only the symbol changes in the variant.

787 Kraken tests pass, clippy clean.

cjdsellers (approved):

PR #5044, opened 2026-09-21
Scope Kraken cancel-all and batch selected cancellations

### LINKED ISSUE #5038: [Kraken] Fix cancel-all scope and batch selected cancellations
Follow-up to the remaining Kraken cancellation work in #4470, tracked separately so that work does not hold up closing the original issue.

The current Rust execution clients filter sided `CancelAllOrders` requests by instrument and side, then send individual HTTP cancels. Both clients already have explicit-ID batch cancellation paths. Spot also sends an account-wide `CancelAll` for an unsided request, despite receiving an `instrument_id`, which can cancel orders on other instruments. Futures already passes the symbol for unsided cancellation.

Required changes:

- [ ] Kraken Spot: keep unsided cancellation scoped to the requested instrument, using selected order IDs instead of account-wide `CancelAll`.
- [ ] Kraken Spot: reuse explicit-ID HTTP batches for side-filtered cancellation.
- [ ] Kraken Futures: reuse explicit-ID batches for side-filtered cancellation, retaining symbol-scoped bulk cancellation for unsided requests.
- [ ] Preserve each selected order's owning strategy, request/result correlation, venue batch limits, and command-outcome handling. Aggregate or ambiguous responses must remain subject to reconciliation rather than producing unsupported per-order outcomes.
- [ ] Add tests covering mixed instruments and strategies, BUY/SELL/unsided requests, no matching orders, batch boundaries, partial results, and ambiguous failures. Verify the submitted IDs and that unrelated orders remain untouched.
- [ ] Update the Kraken integration documentation to describe the resulting cancellation scope and transport behavior.

The existing paths are in `crates/adapters/kraken/src/execution/spot.rs` and `crates/adapters/kraken/src/execution/futures.rs`: `cancel_all_orders`, `batch_cancel_orders_for_spot`, and `batch_cancel_orders_for_futures`.

This issue covers the Kraken follow-ups from the [adapter cancellation audit in #4470](https://github.com/nautechsystems/nautilus_trader/issues/4470#issuecomment-5740704588). Native batch modification, mixed submit/edit/cancel operations, and batching the default `strategy_only=True` dispatch are outside its scope.

zhaow-de:
  and, if I used AI,
  this draft
  limitation below

## Summary

Spot sent an account-wide `CancelAll` for an unsided request even though the command names an
instrument, so it could cancel orders on instruments the request never named. Side-filtered
requests on both clients sent one HTTP cancel per selected order instead of reusing the explicit-id
batch path.

Spot now selects the matching open orders for the requested instrument, with or without a side
filter, and cancels them by explicit venue id through the existing batch path. Futures keeps its
symbol-scoped bulk cancellation for unsided requests and uses the batch path for side-filtered ones.

Each generated cancel carries the owning strategy of the order it targets, and the cancel-all
command id as its correlation. The batch path is unchanged, so the venue chunk limit of 50 and the
existing handling of aggregate or ambiguous responses both still apply: no per-order outcomes are
produced for a result the venue did not itemise.

## Related issues/PRs

Closes #5038

## Type of change

- [x] Bug fix (non-breaking)
- [x] Documentation update
- [ ] Maintenance / chore

Not an API change. An unsided spot cancel-all no longer cancels orders outside the requested
instrument, and side-filtered cancellation now sends batched requests instead of one per order.

## Documentation

  generated output

The "Cancel all orders" note in `docs/integrations/kraken.md` described the previous behaviour,
including that spot cancelled across all symbols. It now describes the scope and transport for both
clients. No binding or wrapped doc changed, and `generate_docstrings.py --dry-run` reports no drift.

## Testing

New tests:

- `test_spot_cancel_all_unsided_scopes_to_requested_instrument` places orders on two instruments and
  asserts the submitted body contains the target id, does not contain the other instrument's id, and
  that the account-wide `CancelAll` endpoint is never called.
- `test_spot_cancel_all_side_filter_selects_only_that_side` asserts the submitted body contains the
  matching side and not the opposite one.
- `test_spot_cancel_all_without_matching_orders_sends_no_request` asserts no request is sent at all.
- `test_spot_cancel_all_chunks_at_venue_batch_limit` uses 51 selected ids and asserts two requests.
- `test_spot_cancel_all_partial_result_does_not_emit_cancel_rejected` waits for the request, then
  asserts a partial result produces no per-order rejection.
- `test_futures_cancel_all_side_filter_uses_batch_path` asserts the batch endpoint is used and no
  per-order cancels are sent.

The scoping and side tests assert both a positive and a negative id in the captured request body, so
they cannot pass on an empty or malformed body. The no-request test is paired with the others: the
same command sends exactly one request when a matching order exists, and none when it does not.

Updated `test_spot_whole_cancel_all_failure_does_not_emit_cancel_rejected`: it waited on the
account-wide `CancelAll` endpoint, which this change stops using. Its assertion, that an aggregate
failure produces no per-order rejection, is unchanged and is now exercised through the batch path.

`cargo test -p nautilus-kraken`: 788 passed, 0 failed. `cargo clippy --all-targets --all-features --
-D warnings`: clean.

## One gap for your attention

The issue asks for coverage of mixed strategies. Each cancel takes `order.strategy_id()` from the
order it targets, so the owning strategy is preserved by construction, but I did not add a test that
observes it. Spot's batch endpoint returns an aggregate count with no per-order results, so the spot
path emits no per-order events to assert against. If you would like this proven rather than argued,
the futures batch path does carry per-order contexts and could carry such a test. Happy to add it.


cjdsellers:
Hey @zhaow-de,

Thanks for the PR. There’s a remaining case in Spot’s `cancel_all_orders`: Kraken can already have accepted an order while the cache still marks it `Submitted`. Selecting only `orders_open` skips that order, whereas the previous unsided venue cancellation could reach it. This needs coverage and handling that preserves the requested instrument scope.

If you’re able, could you also add the mixed-strategy Futures coverage from #5038 before merge? The current test only checks that a batch request happens. It needs to verify the submitted IDs, exclude the other side and instrument, and check that a per-order rejection carries the owning strategy.

The branch also needs the import conflicts with `develop` resolved, retaining both `DurationNanos` and `UUID4`.

zhaow-de:
Thanks @cjdsellers, all three are done.

Cancel-all now selects in-flight orders alongside open ones for the requested instrument, on both clients, deduplicated by client order id. `is_open` already covers `PendingCancel` and `PendingUpdate`, so `Submitted` is the state this actually adds — which is the case you described. The new test puts a submitted order on the requested instrument and another on a different one, and asserts the first is in the submitted body while the second is not, so the instrument scope still holds.

The futures test is extended as asked. It now uses two strategies on the requested instrument plus a sell order and an order on another instrument, and asserts the two matching IDs are submitted, the other side and other instrument are not, and that the per-order rejection carries the strategy owning that order rather than the command's.

The branch is rebased on develop with both imports kept, `DurationNanos` and `UUID4`.

Docs updated to say in-flight orders are included and why.

795 Kraken tests pass, clippy clean. Commits are GPG signed from here on.

cjdsellers (dismissed):
Thanks for addressing those points, @zhaow-de. This looks good from my side. I also rebased the branch to pick up the upstream schema fix behind the earlier CI failure.

I’ll look separately at the shared state machine dropping cancellation rejections while an order is still `Submitted`.

There's also a Spot API detail worth confirming: Kraken's schema lists client IDs under `cl_ord_ids`, but its official CLI sends them under `orders`, as this batch path does. If you have evidence that Kraken accepts that form, could you share it? The new mock test confirms selection, but cannot establish venue acceptance.

Also appreciate you setting up the signed commits :pray:, this will become mandatory eventually so you're ahead of the curve.

zhaow-de:
Thanks @cjdsellers, and sorry — you were right, and I should have checked this before adding the submitted-order handling.

I have no evidence Kraken accepts client IDs in `orders`, and the evidence points the other way. The docs say `orders` takes transaction IDs or user references, and client order IDs go in a separate `cl_ord_ids` field. Our own adapter agrees in two other places: the single-cancel path has distinct `txid` and `cl_ord_id` parameters, and the futures batch path already builds items from either an order ID or a client order ID. The spot batch path was the only one mixing them.

It was pre-existing, but my change made it the common path rather than a rare one. A submitted order has no venue order ID yet, so the cancel-all I added last round would have selected the order correctly and then sent an ID Kraken could not match. The fix for your previous review did not actually work at the venue.

Fixed now: the two identifier kinds go into their own fields, and the chunking happens before the split so the limit of 50 still counts both together.

You were also right that the old test could not show this. It asserted the ID appeared somewhere in the body, which passed either way. It now parses the body and asserts which field the ID is in, and I checked it fails against the old form with `orders=["submitted-001"] cl_ord_ids=[]`.

That still only proves we send the documented shape, not that the venue accepts it. I have no Kraken key here to confirm against the live API, so if you want that confirmed before merge, it would need someone who does.

Thanks for rebasing the branch, and for the note on signing.

cjdsellers (approved):

